PR #53, opened 2020-07-22
k8s: null coalesce clientset on pod delete

gumpt:
Our internal resolver deduces the clientset based on the user and clusters
passed. Coalescing this to `"undefined"` matches the HPA frontend as well.


gumpt:
Screenshot as suggested for frontend changes:

Screen Shot 2020-07-22 at 2 02 11 PM

PR #55, opened 2020-07-22
backend: feature flag api

danielhochman:
### Description
This is a backend that will allow us to eventually flip frontend features on and off.

It's very simple for now, no user or group targeting. Figured we could play around with it on the frontend before committing to this design (versus something like embedding flags in assets).

### Testing Performed
Unit and manual.

### GitHub Issue
See #4 

### TODOs
- [ ] Frontend integration


danielhochman:
@dschaller threw this simple API together for us to experiment with on the frontend. left some more thoughts in the parent issue as well. we can discuss frontend design there in #4  for the most part.

mikecutalo (approved):

PR #70, opened 2020-07-23
ec2: display tags during termination flow

mcutalo88:
### Description

Appends all available tags for the instance, as they are secondary information I opted to display them below our original data set.

Before
Screen Shot 2020-07-23 at 10 12 39 AM

After
Screen Shot 2020-07-23 at 10 10 12 AM

### Testing Performed

Local testing.

### GitHub Issue

n/a

### DIFF
--- a/frontend/workflows/ec2/src/terminate-instance.tsx
+++ b/frontend/workflows/ec2/src/terminate-instance.tsx
@@ -31,19 +31,25 @@ const InstanceDetails: React.FC<WizardChild> = () => {
   const resourceData = useDataLayout("resourceData");
   const instance = resourceData.displayValue();
 
+  const data = [
+    { name: "Instance ID", value: instance.instanceId },
+    { name: "Region", value: instance.region },
+    { name: "State", value: instance.state },
+    { name: "Instance Type", value: instance.instanceType },
+    { name: "Public IP Address", value: instance.publicIpAddress },
+    { name: "Private IP Address", value: instance.privateIpAddress },
+    { name: "Availability Zone", value: instance.availabilityZone },
+  ];
+
+  if (instance.tags) {
+    Object.keys(instance.tags).forEach(key => {
+      data.push({ name: key, value: instance.tags[key] });
+    });
+  }
+
   return (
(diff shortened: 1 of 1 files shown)


frontend/workflows/ec2/src/terminate-instance.tsx
```diff
@@ -31,19 +31,25 @@ const InstanceDetails: React.FC<WizardChild> = () => {
...
+    { name: "State", value: instance.state },
+    { name: "Instance Type", value: instance.instanceType },
+    { name: "Public IP Address", value: instance.publicIpAddress },
+    { name: "Private IP Address", value: instance.privateIpAddress },
+    { name: "Availability Zone", value: instance.availabilityZone },
+  ];
+
+  if (instance.tags) {
+    Object.keys(instance.tags).forEach(key => {
+      data.push({ name: key, value: instance.tags[key] });
```
dschaller, line 46: Should we ensure title casing of the name?

danielhochman: i don't think so. tags are a case-sensitive field on the server and should be presented directly to the user.

danielhochman: however we probably want to differentiate between tags and regular informational fields. i could have a tag which is Region: iad, then you would get two rows in the table with the same header and diff value.

mcutalo88: @danielhochman what if we prepended the name with `tag: {key}` to indicate it is a tag?

danielhochman: i like that or
Screenshot from 2020-07-23 14-53-37


a separate table with its own header could work too.

let's do the easy thing now (text label or icon prepended) and we can ask our UX person soon what the ideal solution would be and follow up.

dschaller: Sorry all. I merged this too early. @mcutalo88 feel free to open a follow up :)

PR #79, opened 2020-07-23
core: kubernetes deployment update api

mcutalo88:
### Description

This adds an API to update kubernetes deployment objects, currently the support fields to mutate are `labels` and `annotations`.

### Testing Performed

Unit test and local integration testing with a local k8s cluster (v1.16.6).

before
```
kubectl describe deployment envoy
Name:                   envoy
Namespace:              choice
Labels:                 app=envoy
Annotations:            deployment.kubernetes.io/revision: 1
Selector:               app=envoy
...
Pod Template:
  Labels:  app=envoy
````

```
curl -X POST \
  http://localhost:8080/v1/k8s/updateDeployment \
  -H 'Content-Type: application/json' \
  -d '{
    "clientset": "docker-for-desktop",
    "cluster": "docker-desktop",
    "namespace": "choice",
    "name": "envoy",
    "fields": {
	"labels": {
		"newlabel": "labelvalue"
	},
	"annotations": {
		"newannotation": "annotationvalue"
	}
    }
}'
```

after
```
Name:                   envoy
Namespace:              choice
Labels:                 app=envoy
                        newlabel=labelvalue
Annotations:            deployment.kubernetes.io/revision: 2
                        newannotation: annotationvalue
Selector:               app=envoy
...
Pod Template:
  Labels:       app=envoy
                newlabel=labelvalue
  Annotations:  newannotation: annotationvalue
```

### GitHub Issue

https://github.com/lyft/clutch/issues/78

### CODE REVIEW

File: backend/service/k8s/k8s.go [hunk 1]
```diff
@@ -60,6 +60,10 @@ type Service interface {
 	// HPA management functions.
 	DescribeHPA(ctx context.Context, clientset, cluster, namespace, name string) (*k8sapiv1.HPA, error)
 	ResizeHPA(ctx context.Context, clientset, cluster, namespace, name string, sizing *k8sapiv1.ResizeHPARequest_Sizing) error
+
+	// Deployment management functions.
+	DescribeDeployment(ctx context.Context, clientset, cluster, namespace, name string) (*k8sapiv1.Deployment, error)
+	UpdateDeployment(ctx context.Context, clientset, cluster, namespace, name string, labels map[string]string, annotations map[string]string) error
```
mcutalo88, line 66: I dont really like how I'm passing through `labels map[string]string, annotations map[string]string` here, will probably refactor this.

mcutalo88:
I need to add tests and want to redesign a few things, mostly the func signatures but I'm keeping the API limited to what we need internally we can add more capabilities if the need arises such as update the image of a deployment for example.

@danielhochman mind if you give me a first pass to make sure I'm on the right track? Also I have this fully working locally.

api/k8s/v1/k8s.proto
```diff
@@ -166,3 +174,35 @@ message ResizeHPARequest {
...
+    pattern : "{cluster}/{namespace}/{name}"
+  };
+
+  string clientset = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  string cluster = 2 [ (validate.rules).string = {min_bytes : 1} ];
+  string namespace = 3 [ (validate.rules).string = {min_bytes : 1} ];
+  string name = 4 [ (validate.rules).string = {min_bytes : 1} ];
+
+  map<string, string> labels = 5 [ (validate.rules).map.keys.string = {min_bytes : 1} ];
+  map<string, string> annotations = 6 [ (validate.rules).map.keys.string = {min_bytes : 1} ];
```
danielhochman, line 204: what are the update semantics?

let's say existing labels are 
```
foo: bar
baz: yes
```

and i post 
```
bop: bop
```

is it 
```
```
bop: bop
```

or 

```
foo: bar
baz: yes
bop: bop
```

**Comment by danielhochman** [comment]
overall looks good, i think you touched on the API needing a bit more thought, left a question of my own

File: api/k8s/v1/k8s.proto (hunk 2 above)
Comment (mcutalo88) [reply api/k8s/v1/k8s.proto:204]: Its a merge of all labels, so in your example it would be 
```
foo: bar
baz: yes
bop: bop
```

File: api/k8s/v1/k8s.proto [hunk 3]
```diff
@@ -166,3 +174,39 @@ message ResizeHPARequest {
...
+
+  string clientset = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  string cluster = 2 [ (validate.rules).string = {min_bytes : 1} ];
+  string namespace = 3 [ (validate.rules).string = {min_bytes : 1} ];
+  string name = 4 [ (validate.rules).string = {min_bytes : 1} ];
+
+  message Fields {
+    map<string, string> labels = 1 [ (validate.rules).map.keys.string = {min_bytes : 1} ];
+    map<string, string> annotations = 2 [ (validate.rules).map.keys.string = {min_bytes : 1} ];
+  }
```
danielhochman, line 206: can you document the update semantics we discussed earlier?

also guessing it's not possible to delete a label in this API?

mcutalo88: Yeah sure let me add some documentation to the proto. Currently no there is not a way to delete labels or annotations in this implementation, if we needed that functionality at some point we could add a new message say `message DeleteFields | RemoveFields`.

PR #84, opened 2020-07-23
Pass 3rd argument

Augustyniak:
Pass 3rd argument that's missing after the recent changes to gateway's API.


dschaller:
Thank you!

PR #101, opened 2020-07-24
scaffolding: remove dependency on local checkout

### LINKED ISSUE #88: scaffolding: removing replace in go.mod breaks scaffolded app
**Description**

We had intended to remove the replace directive after open source since the code would now be publicly downloadable: https://github.com/lyft/clutch/blob/124f38193eb44cd29aad4d236fa32ef48ba30257/tools/scaffolding/templates/gateway/backend/go.mod#L9

However, this breaks all of the frontend build scripts (well anything in `tools/` really) because go modules only downloads the backend folder and not the rest of the repository. We will need to figure out a different way of vendoring code outside of backend/ tooling. Or move our go.mod to the top of the project.

**Expected Behavior**
A scaffolded version of Clutch builds successfully without the code being checked out in the GOPATH.

**Actual Behavior**
```
bash: /home/dhochman/go/pkg/mod/github.com/lyft/clutch/backend@v0.0.0-20200723221653-124f38193eb4/../tools/install-yarn.sh: No such file or directory
```

```
~/go/pkg/mod/github.com/lyft/clutch/backend@v0.0.0-20200723221653-124f38193eb4 $ ls
api  clutch-config.yaml  cmd  gateway  go.mod  go.sum  id  LICENSE  main.go  middleware  mock  module  resolver  service  tools.go
```
(no frontend code available from module)

**Version**

124f38193eb4

**Other Context**
N/A

danielhochman:
### Description
Remove dependency on a full local checkout of Clutch for a custom gateway.

### Testing Performed
Manual + CI.

### GitHub Issue
Fixes #88 

### TODOs
- [x] remove frontend build dependency on linker
- [x] vendor API definitions
- [x] use tools version from PR in tests


.github/workflows/scaffolding.yml
```diff
@@ -28,8 +28,8 @@ jobs:
         with:
           node-version: '12.x'
       - name: scaffold new app
-        run: yes | make scaffold-gateway
-        working-directory: ${{ env.PACKAGEPATH }}
+        run: yes | go run scaffolder.go -m gateway -p ${{ github.event.pull_request.head.sha }}
```
dschaller, line 31:
```suggestion
        run: yes | make scaffold-gateway
```

.github/workflows/scaffolding.yml
```diff
@@ -28,8 +28,8 @@ jobs:
...
+        working-directory: ${{ env.PACKAGEPATH }}/tools/scaffolding
```
dschaller, line 32:
```suggestion
```

dschaller:
We might want to move clutch config back up to the project root since that's probably where we will invoke the binary. You just won't be able to run the backend from the backend directory.

PR #102, opened 2020-07-24
ci: publish beta versions of frontend packages

dschaller:
### Description
Publish and consume beta versions of clutch packages/workflows.

This allows us to still maintain some correlation between the frontend and backend versioning while also allowing us to use semantic version resolution during installations.


danielhochman:
looks like we might want to separate the publish and consume into separate PRs

dschaller:
I think we just need to update the package names to include the beta. This is yarn not being able to match the package name for workspaces.

dschaller:
I think we can leave old packages around and just publish beta going forward

PR #107, opened 2020-07-24
k8s: RetryOnConflict for resource mutations

mcutalo88:
### Description

This ensure that the k8s logic when mutating a resource has basic conflict retries as provided by upstream. Mutations of k8s objects happen very frequently this is important for the services reliability. We will use the default [retry policy](https://github.com/kubernetes/client-go/blob/3473491eeb2e22384b846a27f3dceb9aaa73dd76/util/retry/util.go#L26-L33).

### Testing Performed

Manual testing of both k8s endpoints on a local k8s cluster (v1.16.6).

### GitHub Issue

n/a

### DIFF
--- a/backend/service/k8s/deployment.go
+++ b/backend/service/k8s/deployment.go
@@ -7,6 +7,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/client-go/util/retry"
 
 	appsv1 "k8s.io/api/apps/v1"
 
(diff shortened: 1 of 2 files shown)


backend/service/k8s/deployment.go
```diff
@@ -58,12 +59,11 @@ func (s *svc) UpdateDeployment(ctx context.Context, clientset, cluster, namespac
 		return err
 	}
 
-	_, err = cs.AppsV1().Deployments(cs.Namespace()).Patch(oldDeployment.Name, types.StrategicMergePatchType, patchBytes)
-	if err != nil {
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		_, err = cs.AppsV1().Deployments(cs.Namespace()).Patch(oldDeployment.Name, types.StrategicMergePatchType, patchBytes)
```
danielhochman, line 63: err should be scoped to the function

```suggestion
		_, err := cs.AppsV1().Deployments(cs.Namespace()).Patch(oldDeployment.Name, types.StrategicMergePatchType, patchBytes)
```

backend/service/k8s/hpa.go
```diff
@@ -53,8 +54,12 @@ func (s *svc) ResizeHPA(ctx context.Context, clientset, cluster, namespace, name
 	}
 
 	normalizeHPAChanges(hpa, sizing)
-	_, err = cs.AutoscalingV1().HorizontalPodAutoscalers(cs.Namespace()).Update(hpa)
-	return err
+
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		_, err = cs.AutoscalingV1().HorizontalPodAutoscalers(cs.Namespace()).Update(hpa)
```
danielhochman, line 59:
```suggestion
		_, err := cs.AutoscalingV1().HorizontalPodAutoscalers(cs.Namespace()).Update(hpa)
```

PR #149, opened 2020-07-29
resolver: display error for empty required option fields

dschaller:
### Description
If the resolver has an option field has errors we will display that instead of any schema fields.

Screen Shot 2020-08-05 at 1 40 55 PM
Screen Shot 2020-08-05 at 1 40 59 PM


frontend/packages/core/src/Resolver/hydrator.tsx
```diff
@@ -108,6 +108,11 @@ const OptionField = (
     });
   }, []);
 
+  const hasError = field.metadata.required && options.length === 0;
+  if (hasError) {
+    options = ["Error: Missing options for required field"];
```
pcn, line 113: Is there a breadcrumb that could be left here to describe what the missing option/options might be?

danielhochman: interesting idea. that information would need to come from server side.

would need to add some kind of error state to the schema for rendering. i'll think on it and file a followup issue.

frontend/packages/core/src/Resolver/hydrator.tsx
```diff
@@ -108,6 +108,11 @@ const OptionField = (
     });
   }, []);
 
+  const hasError = field.metadata.required && options.length === 0;
```
danielhochman, line 111: if len === 0 and not required we should just omit the display of the dropdown altogether.

frontend/packages/core/package.json
```diff
@@ -38,12 +40,12 @@
...
     "react": "^16.8.0",
     "react-dom": "^16.8.0",
-    "react-hook-form": "^5.7.2",
-    "react-hook-form-devtools": "^1.2.0",
+    "react-hook-form": "^6.3.0",
     "react-is": "^16.8.0",
     "react-router": "^6.0.0-alpha",
     "react-router-dom": "^6.0.0-alpha",
     "styled-components": "^5.1.1",
+    "superstruct": "~0.8.4",
```
dschaller, line 48: we need to include this until https://github.com/react-hook-form/resolvers/issues/20 is resolved

PR #157, opened 2020-07-30
k8s: add method to update pod annotations

gkleiman:
### Description
Added a function to the k8s service to update pod annotations.

### Testing Performed
Unit tests with fake k8s client.


backend/service/k8s/pods.go
```diff
@@ -62,6 +62,79 @@ func (s *svc) ListPods(ctx context.Context, clientset, cluster, namespace string
 	return pods, nil
 }
 
+func (s *svc) UpdatePodAnnotations(ctx context.Context, clientset, cluster, namespace, name string, expectedAnnotations, newAnnotations Annotations) error {
+	cs, err := s.manager.GetK8sClientset(clientset, cluster, namespace)
+	if err != nil {
+		return err
+	}
+
+	pod, err := cs.CoreV1().Pods(namespace).Get(name, metav1.GetOptions{})
```
danielhochman, line 71:
```suggestion
	pod, err := cs.CoreV1().Pods(cs.Namespace()).Get(name, metav1.GetOptions{})
```

backend/service/k8s/pods.go
```diff
@@ -62,6 +62,79 @@ func (s *svc) ListPods(ctx context.Context, clientset, cluster, namespace string
...
+
+	for newAnnotations, newValue := range newAnnotations {
+		if newValue != nil {
+			podAnnotations[newAnnotations] = newValue.GetValue()
+		} else {
+			delete(podAnnotations, newAnnotations)
+		}
+	}
+
+	_, err = cs.CoreV1().Pods(namespace).Update(pod)
```
danielhochman, line 91:
```suggestion
	_, err = cs.CoreV1().Pods(cs.Namespace()).Update(pod)
```

backend/service/k8s/pods.go
```diff
@@ -62,6 +62,79 @@ func (s *svc) ListPods(ctx context.Context, clientset, cluster, namespace string
 	return pods, nil
 }
 
+func (s *svc) UpdatePodAnnotations(ctx context.Context, clientset, cluster, namespace, name string, expectedAnnotations, newAnnotations Annotations) error {
```
danielhochman, line 65: can you add a few comments throughout this func

api/k8s/v1/k8s.proto
```diff
@@ -132,6 +132,14 @@ message DeletePodRequest {
 message DeletePodResponse {
 }
 
+// Annotation value is wrapped in the message field to make it nullable.
+// Setting AnnotationValue to nil means the annotation shouldn't exist or
+// should be deleted.
+message AnnotationValue {
```
danielhochman, line 138: google provides the StringValue wrapper in core proto for this purpose but maybe it's still good to have it's own object for documentation purposes. just thinking out loud.

backend/service/k8s/k8s.go
```diff
@@ -56,6 +58,7 @@ type Service interface {
 	DescribePod(ctx context.Context, clientset, cluster, namespace, name string) (*k8sapiv1.Pod, error)
 	DeletePod(ctx context.Context, clientset, cluster, namespace, name string) error
 	ListPods(ctx context.Context, clientset, cluster, namespace string, listPodsOptions *k8sapiv1.ListPodsOptions) ([]*k8sapiv1.Pod, error)
+	UpdatePodAnnotations(ctx context.Context, clientset, cluster, namespace, name string, expectedAnnotations, newAnnotations Annotations) error
```
danielhochman, line 61: it would be good if we could come up with a common and extensible way of handling metadata updates. this API works much different than UpdateDeployment for example.

i think the precondition functionality is great so wondering if we can somehow unify the way all of these work.

cc @mcutalo88 for thoughts

backend/service/k8s/pods.go
```diff
@@ -62,6 +62,79 @@ func (s *svc) ListPods(ctx context.Context, clientset, cluster, namespace string
...
+		return err
+	}
+
+	for newAnnotations, newValue := range newAnnotations {
+		if newValue != nil {
+			podAnnotations[newAnnotations] = newValue.GetValue()
+		} else {
+			delete(podAnnotations, newAnnotations)
+		}
+	}
```
mcutalo88, line 89: This is very destructive and should avoid deleting annotations that we are not controlling via clutch. For example at Lyft there are several k8s systems that rely on pod annotations such as secret injection (confidant).

backend/service/k8s/k8s.go
mcutalo88: I was just going to comment on the precondition and the label / annotation update process in general. I dont really see the value in doing a precondition check given that updating a map is easily unit testable at least in this case Im not seeing the value and it looks too complex for the given task.

Ideally I would prefer us to use a `Patch()` here since we should be merging the new and old objects together, I realized that the original intention was to delete all annotations that were not supplied by the caller I left a comment below explaining why we should not do that.

Let me digest this PR a bit more.

backend/service/k8s/pods.go
mcutalo88: ah nvm this is fine

backend/service/k8s/pods.go
gkleiman: `DescribePod` and `ListPods` use `namespace`.  Also `ContextClientset` has no field or method `namespace` =).

backend/service/k8s/pods.go
gkleiman: Ditto `ContextClientset has no field or method namespace`

backend/service/k8s/k8s.go
gkleiman: The goal of the preconditions is to prevent some race conditions or updates using stale views.

The updates will always be applied if no preconditions are specified, but if preconditions are specified the client will first get the pod and check whether the preconditions match before applying the updates.

In the case of the Squeeze workflow it prevents clutch from overwriting load balancing weight updates in scenarios like the following:

a. User opens multiple edit tabs, which show the current load balancing weight and then tries to change the weight with both of the tabs.
b. User A opens the edit tab and sees weight X. User B opens the edit tab, sees weight X and sets weight to Y. User A still seeing weight X tries to update weight to Z, which would overwrite the change that user B did.

gkleiman:
@dschaller The `Run make frontend-verify` CI step failed, but I haven't touched the frontend here.

backend/service/k8s/pods.go
danielhochman: they also need to be updated.

https://github.com/lyft/clutch/blob/30dc4f3db3cf06b9bb49867c2a8848c7a3ee6f0f/backend/service/k8s/clientset.go#L18

confused where `as no field or method namespace` is coming from based on ^, can you share additional output?

danielhochman:
@gkleiman merge main for fix on the frontend failure, #159 fixed it

backend/service/k8s/pods.go
gkleiman: Nevermind, I made a typo which was causing the compilation failure =).

I added a commit fixing the other occurrences.

backend/service/k8s/k8s.go
mcutalo88: Thanks for the clarification @gkleiman I see what you'r trying to accomplish now.

gkleiman:
@mcutalo88 note that in this PR passing an empty string as value will set an empty annotation, but not remove it. In order to remove an annotation the value has to be `nil`.

Having said that I do like the more explicit protobuf messages you propose. The only nit is that k8s allows empty annotations/labels, so we should remove the `min_bytes` validation.

mcutalo88:
@gkleiman true good call on the empty strings the validation, if we decide to pursue this design I can refactor the deployment code in a separate PR. @danielhochman what do you think of this proposal?

danielhochman:
sounds good to me

backend/service/k8s/pods.go
```diff
@@ -62,6 +64,104 @@ func (s *svc) ListPods(ctx context.Context, clientset, cluster, namespace string
...
+	// Delete annotations to be removed
+	for _, annotation := range removeObjectMetaFields.GetAnnotations() {
+		delete(podAnnotations, annotation)
+	}
+
+	_, err = cs.CoreV1().Pods(cs.Namespace()).Update(pod)
+	return err
+}
+
+func checkExpectedObjectMetaFields(expectedObjectMetaFields *k8sapiv1.ExpectedObjectMetaFields, pod *corev1.Pod) error {
```
mcutalo88, line 117: This looks highly reusable why not just pass the resources `ObjectMeta` instead of a pod?

PR #164, opened 2020-07-31
metadata table: remove validation on load

dschaller:
### Description
Validating on load can be a bit jarring for fields without default values. Since validation will happen on submit it's ok to delay this.


dschaller:
CLA Trigger

PR #172, opened 2020-07-31
tools/scaffolding: fix walkfunc error

alrs:
This fixes a `filepath.WalkFunc` that wasn't handling its incoming error.

https://golang.org/pkg/path/filepath/#WalkFunc


dschaller:
Thank you!

PR #193, opened 2020-07-31
k8s: use ClusterName if available

mcutalo88:
### Description

We should defer to the k8s object `ClusterName` before falling back to what the clientset is configured for. We changed this behavior slightly in https://github.com/lyft/clutch/commit/ad118d0265aa4505f912da779f6edc12d74fce4d and this preserves the original intent but falling back to `cs.Cluster()` if `ClusterName` is not set.

### DIFF
--- a/backend/service/k8s/deployment.go
+++ b/backend/service/k8s/deployment.go
@@ -30,11 +30,12 @@ func (s *svc) DescribeDeployment(ctx context.Context, clientset, cluster, namesp
 }
 
 func ProtoForDeployment(cluster string, deployment *appsv1.Deployment) *k8sapiv1.Deployment {
-	if deployment.ClusterName == "" {
-		deployment.ClusterName = cluster
+	clusterName := deployment.ClusterName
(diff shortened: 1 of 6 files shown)


dschaller:
Testing something

backend/service/k8s/deployment.go
```diff
@@ -30,8 +30,11 @@ func (s *svc) DescribeDeployment(ctx context.Context, clientset, cluster, namesp
 }
 
 func ProtoForDeployment(cluster string, deployment *appsv1.Deployment) *k8sapiv1.Deployment {
+	if deployment.ClusterName == "" {
+		deployment.ClusterName = cluster
```
danielhochman, line 34: this will modify the incoming object, which may not be desirable in some cases.

i think we should stick with the local var, override it with deployment.ClusterName if deployment.ClusterName not empty

danielhochman:
oh, can you add tests if not already there and simple to do

PR #239, opened 2020-08-04
experimentation: refactor the package to make it more generic

Augustyniak:
### Description
Split the existing implementation of the `experimentation` package into `experimentation` and `serverexperimentation`. This will allow us to ensure that `serverexperimentation` is treated exactly the same way any other package depending on the `experimentation` package is (i.e. a hypothetic `mobile experimentation` package). 

The supported structure of packages once the PR is merged:

Code Organization 2 (5)

### The organization of components once all of the changes are completed

Copy of  Clutch  register

### Config

Add the following code to the `clutch.config.js` file to enable experimentation package frontend:
```javascript
  "@clutch-sh/experimentation": {
    listExperiments: {
      trending: true,
      componentProps: {
        columns: ["identifier", "targets", "type", "description"],
        experimentTypes: {
          "type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig": {
            mapping: e => {
              const clusters = e.clusterPair;
              return {
                targets: `"${clusters.downstreamCluster}" ➡️ "${clusters.upstreamCluster}"`,
                type: "server",
                description: e.abort
                  ? `${e.abort.percent}% abort with ${e.abort.httpStatus} status code`
                  : `${e.latency.percent}% latency with ${e.latency.durationMs}ms delay`,
              };
            },
            links: [
              {
                displayName: "Start Abort Experiment",
                path: "/server-experimentation/startabort",
              },
              {
                displayName: "Start Latency Experiment",
                path: "/server-experimentation/startlatency",
              },
            ],
          },
        },
      },
    },
  },
  "@clutch-sh/server-experimentation": {
    startAbortExperiment: {},
    startLatencyExperiment: {},
  },
```

### Screenshots

| List View  | Start New Test  |
| ------------- | ------------- |
| `experimentation` package, editable list of columns that can specified in `clutch.config.js` file |Screen Shot 2020-08-17 at 1 19 32 PM|
| `serverexperimentation` package | Screen Shot 2020-08-14 at 11 48 56 AM|

### Change log
- move server experimentation specific modules to `serverexperimentation` module. Keep existing generic protos inside of an `experimentation` module.
- change the type of test config to `google.protobuf.Any` to make it generic.
- Allow the users of experimentation modules to configure the columns that are a part of "experiments table view".
- Allow the users of `experimentation` package to register a "converter" that's used to transform raw experiments into view models that are are displayed in experiments list view.

### TODOs
- [x] Disable Postgres and experimentation packages
- [x] Add missing doc strings
- [x] Make it possible for the users of `experimentation` package to register a screen responsible for the addition of new type of tests (i.e. "a mobile fault injection test")


api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -0,0 +1,50 @@
...
+import "google/api/annotations.proto";
+import "validate/validate.proto";
+
+import "api/v1/annotations.proto";
+
+option go_package = "serverexperimentationv1";
+
+message TestSpecification {
+  oneof config {
+    AbortFault abort = 1;
```
Augustyniak, line 14: @dschaller as discussed offline yesterday, eventually we should be able to reorganize models for server experimentation but I decided against it (when it comes to making the change as part of this PR) to limit the amount of changes in my PR.

frontend/packages/app/src/clutch.config.js
```diff
@@ -46,4 +46,41 @@ module.exports = {
...
     },
   },
+  "@clutch-sh/experimentation": {
+    listExperiments: {
+      trending: true,
+      componentProps: {
+        columns: ["identifier", "targets", "type", "description"],
+        experimentTypes: {
+          "type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestSpecification": {
+            mapping: e => {
```
Augustyniak, line 56: This is still a function -> not ideal for a config file.

Augustyniak: I wanted to move the implementation of this function to another file and import it here but I ran into a few issues. 

First, nothing from outside of `/src` directory can be imported in here which means that I could move the implementation of the function to `serverexperimentation` package:

Screen Shot 2020-08-11 at 4 29 47 PM

Secondly, even If I move the implementation of the function to the file that lives inside of the `/src` directory (again, this limitation makes this far from being perfect) I see the following error once I add an import statement at the top of the file:

Screen Shot 2020-08-11 at 4 37 53 PM

frontend/packages/app/src/clutch.config.js
```diff
@@ -46,4 +46,41 @@ module.exports = {
       },
     },
   },
+  "@clutch-sh/experimentation": {
+    listExperiments: {
+      trending: true,
+      componentProps: {
+        columns: ["identifier", "targets", "type", "description"],
+        experimentTypes: {
+          "type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestSpecification": {
```
Augustyniak, line 55: ideally, this should be a constant that lives inside of a server experimentation module and that could be imported here but I couldn't make importing work per my comment in [here](https://github.com/lyft/clutch/pull/239/files#r469641669).

Augustyniak:
@gkleiman Could you take a look at the parts that (slightly) modify the implementation of rtds to make sure that it still looks fine? Obviously, any other feedback is welcome too. :)

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -3,55 +3,16 @@ syntax = "proto3";
...
-  oneof target {
-    ClusterPairTarget cluster_pair = 1;
-  }
-
-  // The percentage of requests that will be slowed down.
-  float percent = 2 [ (validate.rules).float = {gt : 0.0, lte : 100.0} ];
-
-  // The latency duration in milliseconds.
-  int32 duration_ms = 3 [ (validate.rules).int32.gt = 0 ];
+  google.protobuf.Any test_config = 2;
```
gkleiman, line 15: I'd remove `test` from the name of the field and call it either `spec` or `config`.

api/chaos/serverexperimentation/v1/serverexperimentation.proto
gkleiman: I agree this should be reorganized in another PR.

I think we can probably:

1. Try to get rid of this wrapper message, although I think @wgallagher needed to add it to handle storing the specs in an `Any` field.
2. Rename the `AbortFault` and `LatencyFault` to something like `AbortExperimentSpec` and `LatencyExperimentSpec`.

`oneof` can't be extended in other proto files, so using a `oneof` would make it hard to add new kinds of experiments in lyft/clutch-private, e.g., squeeze experiments.

backend/module/chaos/serverexperimentation/rtds/rtds_cache_test.go
```diff
@@ -8,74 +8,92 @@ import (
...
-			},
+				},
+				Percent:    faultPercent,
+				HttpStatus: httpStatus},
 		},
 	}
+
+	anyConfig, err := ptypes.MarshalAny(testConfig)
+	if err != nil {
+		t.Errorf("setSnapshot failed %v", err)
```
gkleiman, line 37: This looks like copy & pasta =)

```suggestion
		t.Errorf("couldn't marshal experiment config: %v", err)
```

backend/module/chaos/serverexperimentation/rtds/rtds_cache_test.go
```diff
@@ -8,74 +8,92 @@ import (
...
-			},
+				},
+				Percent:    latencyPercent,
+				DurationMs: duration},
 		},
 	}
+
+	anyConfig, err := ptypes.MarshalAny(testConfig)
+	if err != nil {
+		t.Errorf("setSnapshot failed %v", err)
```
gkleiman, line 60: Ditto updating the message here and in all other occurrences below

backend/module/chaos/serverexperimentation/rtds/rtds_cache_test.go
```diff
@@ -8,74 +8,92 @@ import (
...
+func mockGenerateFaultData(t *testing.T) []*experimentation.Experiment {
 	return []*experimentation.Experiment{
-		createAbortExperiment("serviceA", "serviceB", 10, 404),
-		createAbortExperiment("serviceC", "", 20, 504),
-		createLatencyExperiment("serviceA", "serviceD", 30, 100),
-		createLatencyExperiment("serviceE", "", 40, 200),
+		createAbortExperiment(t, "serviceA", "serviceB", 10, 404),
+		createAbortExperiment(t, "serviceC", "", 20, 504),
+		createLatencyExperiment(t, "serviceA", "serviceD", 30, 100),
+		createLatencyExperiment(t, "serviceE", "", 40, 200),
```
gkleiman, line 71: Do we need to pass `t`? It doesn't look like it's used by `createAbortExperiment()`/`createLatencyExperiment()`

frontend/packages/app/src/clutch.config.js
```diff
@@ -46,4 +46,41 @@ module.exports = {
...
+            mapping: e => {
+              const config = e.latency || e.abort;
+              const clusters = config.clusterPair;
+              return {
+                targets: `from "${clusters.downstreamCluster}" to "${clusters.upstreamCluster}"`,
```
gkleiman, line 60: Nit: I'd make this less verbose, e.g.,: `"${clusters.downstreamCluster}" <right arrow icon> "${clusters.upstreamCluster}"`

frontend/workflows/experimentation/src/list-experiments.tsx
```diff
@@ -1,49 +1,73 @@
...
-    defaultData.push((data as IClutch.chaos.experimentation.v1.AbortFault).httpStatus);
+  const types = experimentTypes || {};
+  const registeredExperimentType = types[experiment.testConfig["@type"]];
+  const isExperimentTypeRegistered = typeof registeredExperimentType !== "undefined";
+  if (!isExperimentTypeRegistered) {
+    const data = columns.map(() => {
+      return experiment.testConfig["@type"];
+    });
+    return <Row data={data} />;
   }
```
gkleiman, line 27: I am no JS/typescript expert, but I think this may be cleaner:

```suggestion
  if (!types.hasOwnProperty(experiment.testConfig["@type"])) {
    const data = columns.map(() => {
      return experiment.testConfig["@type"];
    });
    return <Row data={data} />;
  }

  const registeredExperimentType = types[experiment.testConfig["@type"]];
```

frontend/workflows/experimentation/src/list-experiments.tsx
```diff
@@ -1,49 +1,73 @@
...
 
-  return <Row data={defaultData} />;
+  const mapper = registeredExperimentType.mapping;
+  const mapperExists = typeof mapper !== "undefined";
```
gkleiman, line 30: Ditto using `registeredExperimentType.hasOwnProperty(registeredExperimentType.mapping)`, but check with @dschaller  before taking my word on it.

frontend/workflows/experimentation/src/list-experiments.tsx
```diff
@@ -1,49 +1,73 @@
...
+  const model = mapperExists ? mapper(experiment.testConfig) : experiment;
+
+  const data = columns.map(column => {
+    let value: string;
+    if (column === "identifier") {
+      value = experiment.id.toString();
+    } else if (column in model) {
```
gkleiman, line 37: I _think_ we may want to also use `hasOwnProperty` instead of `in`

frontend/packages/app/src/clutch.config.js
```diff
@@ -46,4 +46,41 @@ module.exports = {
       },
     },
   },
+  "@clutch-sh/experimentation": {
+    listExperiments: {
+      trending: true,
```
dschaller, line 51: We should probably remove this from trending until we finish making changes

frontend/workflows/serverexperimentation/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@clutch-sh/serverexperimentation",
```
dschaller, line 2: Could we possibly look at keeping it within the experimentation package?

frontend/workflows/experimentation/src/list-experiments.tsx
dschaller: 👍

frontend/workflows/experimentation/src/list-experiments.tsx
dschaller: 👍

frontend/workflows/experimentation/src/list-experiments.tsx
```diff
@@ -56,26 +80,37 @@ const ListExperiments: React.FC<BaseWorkflowProps> = () => {
       });
   }
 
+  const types = experimentTypes || [];
+  let links: ExperimentTypeLinkProps[] = [];
+  Object.keys(types).forEach(experimentType => {
+    const specification = types[experimentType];
+    if (typeof specification.links !== "undefined") {
```
dschaller, line 87: same comment @gkleiman left above

api/chaos/experimentation/v1/experimentation.proto
Augustyniak: yea, the name should be clear enough even without a `test` prefix - will update

backend/module/chaos/serverexperimentation/rtds/rtds_cache_test.go
Augustyniak: it is used by both of these methods for logging errors that can come out of `MarshalAny` method

frontend/packages/app/src/clutch.config.js
Augustyniak: 👍

frontend/packages/app/src/clutch.config.js
Augustyniak: I forgot to clean it up 👍 updating

api/chaos/serverexperimentation/v1/serverexperimentation.proto
wgallagher: I think that since this is being changed to an any, we can remove the wrapper message now. The oneof was the problem, not the any

frontend/workflows/serverexperimentation/package.json
Augustyniak: Not sure whether we want to keep this as part of `experimentation` package - let me explain why. So the idea here is to make `experimentation` package a generic package for people to use when they want to implement their own type of experimentation package on top of the provided package. 

`serverexperimentation` is supposed to be one of packages that benefits from the abstractions that are provided by `experimentation` - if we combine `serverexperimentation` and `experimentation` packages it will become harder for us to ensure that experimentation stays a generic type of module. Since they are separated it's kind of a given that `experimentation` packages provides all of the hooks that are needed in order for any other package to benefit from functionality provided by `experimentation`.

I would like to know whether @wgallagher has any thoughts about the current organization of code/packages too.

api/chaos/serverexperimentation/v1/serverexperimentation.proto
wgallagher: regarding the naming, I'd probably stick with 'config': AbortFaultConfig, LatencyFaultConfig, etc.

Augustyniak: @wgallagher I used this as an opportunity to modify the model layer. I kept the top level `TestConfig` object and made `ClusterPair` a property on the root object (as opposed to a nested property on the fault type - abort/latency). Let me know what you think.

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -0,0 +1,46 @@
...
+import "api/v1/annotations.proto";
+
+option go_package = "serverexperimentationv1";
+
+message TestConfig {
+  oneof target {
+    ClusterPairTarget cluster_pair = 1;
+  }
+
+  oneof fault {
```
wgallagher, line 17: ideally, we should get rid of this oneof (now that the config is described as an any in Experiment). We can do this in a separate change, though

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -0,0 +1,46 @@
...
+import "google/api/annotations.proto";
+import "validate/validate.proto";
+
+import "api/v1/annotations.proto";
+
+option go_package = "serverexperimentationv1";
+
+message TestConfig {
+  oneof target {
+    ClusterPairTarget cluster_pair = 1;
```
wgallagher, line 14: this won't apply for all server-side tests. redline tests won't have a clusterpair, for example.

wgallagher: I can take care of this in a subsequent PR.

frontend/workflows/serverexperimentation/package.json
wgallagher: I agree that it would be better to keep them separated. The goal of 'experimentation' is to provide a generic experimentation framework. 'serverexperimentation' is a specific instance of this.

frontend/workflows/experimentation/src/list-experiments.tsx
```diff
@@ -1,49 +1,72 @@
...
-
-  const defaultData = [
-    data.clusterPair.downstreamCluster,
-    data.clusterPair.upstreamCluster,
-    data.percent,
-  ];
-  if (specification?.abort) {
-    defaultData.push((data as IClutch.chaos.experimentation.v1.AbortFault).httpStatus);
+  const types = experimentTypes || {};
+  if (!Object.prototype.hasOwnProperty.call(types, experiment.config["@type"])) {
```
wgallagher, line 20: what is this checking for? could you please add a comment.

api/chaos/serverexperimentation/v1/serverexperimentation.proto
Augustyniak: >I can take care of this in a subsequent PR.

Great 👍 

I think that for squeeze tests we will introduce a separate root object (other than `TestConfig`) so that the targeting in squeeze tests won't have to be the same as it is here.

frontend/workflows/experimentation/src/list-experiments.tsx
Augustyniak: added comment 👍

Augustyniak:
@dschaller @mcutalo88 this PR should be ready to be re-reviewed now 🙂

frontend/workflows/serverexperimentation/package.json
dschaller: Sounds good. One nit would be to use a hyphen here so it's a bit more readable.
```suggestion
  "name": "@clutch-sh/server-experimentation",
```

Augustyniak: >Ideally the single entry point is the experimentation package which lists all experiments and gives users the ability to create experiments (maybe with a select for the type)

That's the goal but I think that dependency wise we will want to "inject" screens responsible for the creation of new tests into the `experimentation` package via `clutch.config.js`. In fact, this is how `server-experimentation` package does it now - see `links` part below (other packages will follow this pattern):

```javascript
"@clutch-sh/experimentation": {
    listExperiments: {
      trending: true,
      componentProps: {
        columns: ["identifier", "targets", "type", "description"],
        experimentTypes: {
          "type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig": {
            mapping: e => { return e },
            links: [
              {
                displayName: "Start Abort Experiment",
                path: "/serverexperimentation/startabort",
              },
              {
                displayName: "Start Latency Experiment",
                path: "/serverexperimentation/startlatency",
              },
            ],
          },
        },
      },
    },
  },
  "@clutch-sh/serverexperimentation": {
    startAbortExperiment: {},
    startLatencyExperiment: {},
  },
```

Augustyniak: Renamed the module @dschaller (per your suggestion)

dschaller: Yeah, I guess for now this works but going forward it might be better to embed these with react components instead of relying on a URL. The main thing I'm worried about is we are still flushing out what the end product will be and by publishing an external package we won't be able to remove it if we decide to take an approach that doesn't mesh with having this code be external.

Augustyniak: Yea, we will need to figure this out - I know that it's not a great answer to your concerns (that I share too) but both of these packages are marked as beta ones so breaking changes - although not something that we want to introduce - are acceptable until these packages stabilize more.

Happy to chat more about how we could organize this to avoid using URL going forward *and* having the level of flexibility we need to have to support external packages (from outside of clutch repo) depending on the `experimentation` package at the same time.

frontend/workflows/serverexperimentation/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@clutch-sh/server-experimentation",
+  "version": "0.0.0-beta",
```
dschaller, line 3:
```suggestion
  "version": "0.0.0-beta",
  "private": true,
```

Augustyniak: Change commited and the PR is still green 🙂

PR #249, opened 2020-08-05
kubernetes: rest client configuration

mcutalo88:
### Description

Expose common configuration options for the Kuberentes rest client.

### Testing Performed

unit test and tested locally with this configuration
```
  - name: clutch.service.k8s
    typed_config:
      "@type": types.google.com/clutch.config.service.k8s.v1.Config
      rest_client_config:
        timeout: 10s
        QPS: 1
        burst: 1
```


api/config/service/k8s/v1/k8s.proto
```diff
@@ -8,4 +8,23 @@ import "validate/validate.proto";
 
 message Config {
   repeated string kubeconfigs = 1 [ (validate.rules).repeated = {unique : true} ];
+  RestClientConfig restclientconfig = 2;
```
danielhochman, line 11: nit: name should be snake case. `kubeconfigs` is one word because `kubeconfig` is a thing but this can be snake case.

```suggestion
  RestClientConfig rest_client_config = 2;
```

api/config/service/k8s/v1/k8s.proto
```diff
@@ -8,4 +8,23 @@ import "validate/validate.proto";
...
+}
+
+// These configuration values are passed directly through to the rest config object.
+message RestClientConfig {
+  // The maximum length of time to wait before giving up on a server request.
+  // By default a value of zero means no timeout.
+  // https://github.com/kubernetes/client-go/blob/00dbcca6ee44c678754d3f5fda1bd0e704b26fe2/rest/config.go#L130-L131
+  uint32 timeout = 1 [(validate.rules).uint32.gte = 1];
```
danielhochman, line 19: we'll almost always use duration to represent time (the underlying config uses duration as well)

```suggestion
  google.protobuf.Duration timeout = 1 [(validate.rules).duration.gt.seconds = 1];
```
(also needs import)

api/config/service/k8s/v1/k8s.proto
```diff
@@ -8,4 +8,23 @@ import "validate/validate.proto";
...
+
+  // QPS indicates the maximum QPS to the master from this client.
+  // If it's zero, the created RESTClient will use DefaultQPS: 5
+  // https://github.com/kubernetes/client-go/blob/00dbcca6ee44c678754d3f5fda1bd0e704b26fe2/rest/config.go#L115-L117
+  float QPS = 2 [(validate.rules).float.gte = 1];
+
+  // Maximum burst for throttle.
+  // If it's zero, the created RESTClient will use DefaultBurst: 10.
+  // https://github.com/kubernetes/client-go/blob/00dbcca6ee44c678754d3f5fda1bd0e704b26fe2/rest/config.go#L119-L121
+  uint32 burst = 3 [(validate.rules).uint32.gte = 1];
```
danielhochman, line 29: gte 1 doesn't make sense if we want to allow 0, same for QPS

api/config/service/k8s/v1/k8s.proto
```diff
@@ -5,7 +5,27 @@ package clutch.config.service.k8s.v1;
...
   repeated string kubeconfigs = 1 [ (validate.rules).repeated = {unique : true} ];
+  RestClientConfig rest_client_config = 2;
+}
+
+// These configuration values are passed directly through to the rest config object.
+message RestClientConfig {
+  // The maximum length of time to wait before giving up on a server request.
+  // By default a value of zero means no timeout.
+  // https://github.com/kubernetes/client-go/blob/00dbcca6ee44c678754d3f5fda1bd0e704b26fe2/rest/config.go#L130-L131
+  google.protobuf.Duration timeout = 1 [(validate.rules).duration.gt.seconds = 1];
```
danielhochman, line 20: err whoops
```suggestion
  google.protobuf.Duration timeout = 1 [(validate.rules).duration.gte.seconds = 1];
```

mcutalo88: no worries i caught it before

mcutalo88:
@danielhochman sorry missed this comment, I added impl to this PR. I initially just wanted to get your thoughts on the proto and the idea of exposing this configuration before I spent time on the Impl.

backend/service/k8s/clientset.go
```diff
@@ -95,6 +103,20 @@ func newClientsetManager(rules *clientcmd.ClientConfigLoadingRules, logger *zap.
...
+	if restClientConfig.Burst != 0 {
+		restConfig.Burst = int(restClientConfig.Burst)
+	}
+
+	if restClientConfig.Qps >= 0 {
+		restConfig.QPS = restClientConfig.Qps
+	}
+
+	if restClientConfig.Timeout != nil {
+		restConfig.Timeout = time.Duration(restClientConfig.Timeout.Seconds) * time.Second
```
danielhochman, line 116: use `ptypes.Duration` to convert from restClientConfig.Timeout

mcutalo88: ah much nicer

backend/service/k8s/k8s.go
```diff
@@ -38,9 +39,13 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
...
 			}
 		}
+
+		if c.RestClientConfig != nil {
+			restClientConfig = *c.RestClientConfig
+		}
 	}
 
-	c, err := newClientsetManager(loadingRules, logger)
+	c, err := newClientsetManager(loadingRules, restClientConfig, logger)
```
danielhochman, line 48: pass in `c.RestClientConfig` and add a nil check to the apply func instead.

then we don't have to do the nil check, assignment dance in every caller if we want to reuse func.

backend/service/k8s/clientset.go
```diff
@@ -95,6 +107,25 @@ func newClientsetManager(rules *clientcmd.ClientConfigLoadingRules, logger *zap.
 	return &managerImpl{clientsets: lookup}, nil
 }
 
+func applyRestClientConfig(restConfig *restclient.Config, restClientConfig k8sconfigv1.RestClientConfig) error {
```
danielhochman, line 110: take ptr to RestClientConfig and add nil check

also probably want to export this func so we can use in private

PR #250, opened 2020-08-05
ci: restructure e2e tests

dschaller:
### Description
Restructured E2E tests to consolidate the assertion of drawer links being displayed and hidden into one. Before this was causing random failures since there was a dependency between the two tests executing in the correct order.

Additionally, this scoped the menu expansion and collapse click to the icon to prevent drawer resizing from causing issues.

### Testing Performed
CI


dschaller:
I've run this ~5 times and it's passed consistently

PR #254, opened 2020-08-06
core: minor css tweaks

danielhochman:
### Description

In advance of the UX designer outcomes, wanted to make some minor tweaks that I think improve the experience a bit.

<hr />

:arrow_forward: Header spacing constant so it aligns with stepper, also strong.

**Before**
Screenshot from 2020-08-06 11-25-41

**After**
Screenshot from 2020-08-06 11-25-46

<hr />

:arrow_forward: Increased margin on form elements so shrunken labels don't nest immediately under previous form element.

**Before**
Screenshot from 2020-08-06 11-45-15
**After**
Screenshot from 2020-08-06 11-44-41

<hr />

:arrow_forward: Added an overrideable max width constraint to the wizard.

**Before**
image

**After**
Screenshot from 2020-08-06 11-42-37

### Testing Performed
Manual and unit/e2e.


frontend/packages/wizard/src/wizard.tsx
```diff
@@ -112,6 +113,7 @@ const Spacer = styled.div<SpacerProps>`
 interface WizardProps {
   heading?: string;
   dataLayout: ManagerLayout;
+  maxWidth?: "xs" | "sm" | "md" | "lg" | "xl" | false;
```
dschaller, line 116: can you use the Container type for this field so it stays up to date automatically?

PR #261, opened 2020-08-07
scaffolding: improve workflow completion message

dschaller:
### Description
The old message was confusing as it directed users to run a make target in the directory of the new workflow.

This improves the scaffolder by telling the user where the generated workflow is and points them to the configuration documentation since we aren't able to predict the destination's contents.


dschaller:
We don't. Feel free to open one. :)

PR #271, opened 2020-08-10
kubernetes: omni search fix for all resources

### LINKED ISSUE #257: resolver: Kubernetes lookup by ID doesn't work
**Description**

The Kubernetes resolver's `Search()` fails when querying by a resource ID.

For example, when trying to use the "K8s: Delete Pod" workflow, I can't obtain a pod by ID using the input labeled "Search by pod ID", as it always returns an error. The same also happens for the HPA search.

**Expected Behavior**

I would expect searching for the pod on all namespaces (which might be costly) or just allowing prefixing the pod ID with the namespace (e.g. `default/podname-58448d6d95-znr75`) and passing the namespace along.

**Actual Behavior**

There is always a "Failed Precondition" error, for example: `Failed Precondition: Did not find a 'type.googleapis.com/clutch.k8s.v1.Pod', looked by: pod ID`.

**Version**

Docker image: 
`docker.io/lyft/clutch@sha256:6da4e0c07c93eb744ec213a86951730e4a22496663a2826eace66f17f87b67ef`

**Other Context**

When looking at the source code, I noticed that the k8s backend invokes the service's `DescribePod()` without a namespace:
https://github.com/lyft/clutch/blob/94145f2f7db6dc3abcfe0ec2874de3ee57a3f2e0/backend/resolver/k8s/k8s.go#L158

That will result in a call to `clientset.CoreV1().Pods("").Get(...)` - which is invalid - and will always return the following error due to the empty namespace: `an empty namespace may not be set when a resource name is provided`.

This is also the case when one tries to run `kubectl get pod -A pod-name`

mcutalo88:
### Description

Fixing omni search functionality for all Kubernetes resource types supported by clutch workflows. We do this by using `fieldSelectors` to search cluster wide for a resources `metadata.name`. A note about performance, the [kubernetes/apiserver](https://github.com/kubernetes/apiserver) does implement a [cache](https://github.com/kubernetes/apiserver/blob/fe7e3b0680a96aa6a3e30185c1d2644f9bc6b3ed/pkg/storage/cacher/watch_cache.go) for `List()` lookups with that being said Im not worried about the performance impact on kubernetes apiservers.

The way that this is implemented, `Describe[Pod | HPA | Deployment]` which are mostly invoked by the kubernetes resolver now use a `List()` operation to search for a resource, rather than using `Get()` as `List()` allows us to do a cluster wide search we lack sufficient information for a targeted lookup namely the `namespace`. Additionally I modified the `GetK8sClientset` to return `clientset`s with `""` or more accurately [metav1.NamespaceAll](https://github.com/kubernetes/apimachinery/blob/da970ee67a0683f95ab94dece488e7adb4eaacdf/pkg/apis/meta/v1/types.go#L288) as this is a valid input. This also makes it easier for callers as they dont need override the `default` namespace of the clientset, in addition this makes `cs.Namespace()` the source of truth.

This is functionality that can be improved with clutch's topology cache work that is coming up soon.

Also in this PR:
* fixed resolver regex matching

### Testing Performed

* unit / CI
* manual against a local kubernetes cluster

omnisearchdeletepod
omnisearchhparesize

### GitHub Issue

Fixes #257 

### TODOs

- [x] Fix search regex
- [x] Refactor and reuse list pods functionality 
- [x] GetK8sClientset honor `""` as `metav1.NamespaceAll` to search all namespaces
- [x] Better errors around describe operations that yield > 1 resource
- [x] tests
- [ ] maybe add `ListDeployments` & `ListHPAs` functionality although currently we dont need this.


danielhochman:
is there a scenario in which this behavior would be undesirable? not sure if it should be configurable.

the cost of maintaining the two codepaths also seems like an issue. why not always use `ListHPAs` instead of using List in one context and Get in another?

mcutalo88:
Nothing comes to mind in term of undesirable effects, but let me think on it a bit more.

The two codepaths do have a code smell but when is a `Get()` better than a `List()`?  Probably with high RPS it might make a difference, would have to look at the apiserver implementation more to understand the implications of always using `List()` I assume its more intensive as it must traverse all of its cache and or make actual API calls to retrieve all matching resources while a `Get()` is a targeted lookup for a single resource.

Let me see if there is compelling reason to use `Get()` at all, if there is not I will refactor this to only use `List()` for all resources.

mcutalo88:
As for `Get` vs `List` they both use the same cache ([Get](https://github.com/kubernetes/apiserver/blob/fe7e3b0680a96aa6a3e30185c1d2644f9bc6b3ed/pkg/registry/generic/registry/store.go#L615) [List](https://github.com/kubernetes/apiserver/blob/fe7e3b0680a96aa6a3e30185c1d2644f9bc6b3ed/pkg/registry/generic/registry/store.go#L334)). `List` is always going to do more work as thats the nature of the API but for our use case it seems ok.

If this API was going to receive a high amount of throughput I would have branching code paths to invoke either of these API's depending on use case.

With that being said i'm going to remove all of the `Get` calls and replace them with `List` to simplify our implementation.

As for undesirable effects, I'm still thinking on this one.

danielhochman:
Cool. To be clear, I think we keep the Describe/Get methods just have them proxy to List instead of either doing Get or List. As opposed to current PR impl which can do List or Get in Get.

This is good compromise that leaves us flexibility in the future to optimize if needed.

mcutalo88:
Yes were on the same page there, I was thinking of doing just that.

api/k8s/v1/k8s.proto
```diff
@@ -113,8 +113,10 @@ message Pod {
   map<string, string> annotations = 10;
 }
 
-message ListPodsOptions {
-  map<string, string> labels = 9;
+// List Options are suitable for all resources types eg: pods, HPA, and deployment.
+message ListOptions {
+  map<string, string> labels = 1;
+  string field_selectors = 2 [ (validate.rules).string = {min_bytes : 1} ];
```
danielhochman, line 119: can you add individual docstrings for these fields? particularly field_selectors.

it's also a bit weird to me that we have labels instead of label_selectors. or should we eventually have both?

finally wondering field_selectors should be a repeated string where the comma separated join for multiple selectors is done on backend

mcutalo88: The `labels` field existed before this pr, but yes I can refactor them to be `label_selectors` as thats what it translates into.

I had mix feelings on `field_selectors` being a string vs repeated, as they support a verity of [operators](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) such as `!= | ==` I could support this of course, but in the moment it felt fine to just let the caller construct their own query string based on k8s docs and will just pass this through to the `ListOptions` untouched.

If you feel strongly about the more structured approach let me know and I can do so.

mcutalo88: Talked offline, going to revert these proto changes for now as its not longer being used in this PR. Going to make a follow up PR to enable field selector functionality with additional List functionality for all k8s resources.

backend/service/k8s/pods.go
```diff
@@ -18,12 +17,20 @@ func (s *svc) DescribePod(ctx context.Context, clientset, cluster, namespace, na
...
+	})
 	if err != nil {
 		return nil, err
 	}
-	return podDescription(pod, cs.Cluster()), nil
+
+	if len(pods.Items) == 1 {
+		return podDescription(&pods.Items[0], cs.Cluster()), nil
+	} else if len(pods.Items) > 1 {
+		return nil, fmt.Errorf("Located multipule Pods")
```
danielhochman, line 31: spelling
```suggestion
		return nil, fmt.Errorf("Located multiple Pods")
```

PR #313, opened 2020-08-15
github: adding GetCommit method to service

smercatoris:
Description
Adds the GitCommit method into the v3repositories interface and a service method to get commit details for a given repository.  

### Testing Performed
`make backend-test` :+1:
`make backend-verify` :+1:
`make backend` :+1:

## Open Questions
In github.go, I've added a [Commit object](https://github.com/lyft/clutch/pull/313/files#diff-1b9746ee0afeded06f610fd4fb017a3fR296) exposing only the data that I'll need to access (files and their filenames), but the GetCommit method could just as easily return a [githubv3.RepositoryCommit](https://github.com/google/go-github/blob/master/github/repos_commits.go#L18) object. Not sure which is the preferable option in this case, but went with this for the time being. Let me know if you have a preference.

### DIFF
--- a/backend/mock/service/githubmock/githubmock.go
+++ b/backend/mock/service/githubmock/githubmock.go
@@ -43,6 +43,10 @@ func (s svc) CompareCommits(ctx context.Context, ref *github.RemoteRef, compareS
 	panic("implement me")
 }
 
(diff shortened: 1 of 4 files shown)


smercatoris:
@danielhochman I've rebased on most recent main branch and should be ready for review. Let me know if you have any questions/concerns. Thanks for the review in advance.

PR #317, opened 2020-08-17
scaffolding: pin material table to fix dep issue

danielhochman:
### Description
Fixes
```
./node_modules/jspdf/dist/jspdf.umd.min.js
Critical dependency: the request of a dependency is an expression
```

Root cause: https://github.com/mbrn/material-table/pull/2311

### Testing Performed
Local


dschaller:
Might be good to pin this in the core package as well

PR #319, opened 2020-08-17
core: fix dependency pin of material-table

dschaller:
### Description
Pin material-table for https://github.com/mbrn/material-table/issues/2326

### DIFF
--- a/frontend/packages/core/package.json
+++ b/frontend/packages/core/package.json
@@ -37,7 +37,7 @@
     "js-cookie": "^2.2.1",
     "jwt-decode": "^2.2.0",
     "lodash": "^4.17.15",
-    "material-table": "1.60.0",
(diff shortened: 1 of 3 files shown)


frontend/packages/core/package.json
```diff
@@ -37,7 +37,7 @@
     "js-cookie": "^2.2.1",
     "jwt-decode": "^2.2.0",
     "lodash": "^4.17.15",
-    "material-table": "^1.60.0",
+    "material-table": "1.60.0",
```
danielhochman, line 40: you can pin up to 1.68.0 i think
```suggestion
    "material-table": "1.68.0",
```

dschaller: Done in the commit below to upgrade the lockfile

PR #330, opened 2020-08-21
frontend: allow workflow routes to be private

dschaller:
### Description
Allow workflows to define private routes. This will prevent the route from being discoverable via the drawer and search bar while still allowing them to be registered on the application.

**note**: this supersedes the trending property and will hide trending workflows.

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/AppProvider/index.tsx
+++ b/frontend/packages/core/src/AppProvider/index.tsx
@@ -29,9 +29,17 @@ interface ClutchAppProps {
 
 const ClutchApp: React.FC<ClutchAppProps> = ({ availableWorkflows, configuration }) => {
   const workflows = registeredWorkflows(availableWorkflows, configuration);
+
+  /** Filter out all of the workflows that are configured to be `hideNav: true`.
+   * This prevents the workflows from being discoverable by the user from the UI,
+   * both search and drawer navigation.
+   *
+   * The routes for all configured workflows will still be reachable
+   * by manually providing the full path in the URI.
+   */
   const publicWorkflows = _.cloneDeep(workflows).filter(workflow => {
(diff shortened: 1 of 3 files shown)


mcutalo88:
This looks good, just cleaned up the test configuration for terminate instance.

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -28,13 +29,21 @@ interface ClutchAppProps {
 
 const ClutchApp: React.FC<ClutchAppProps> = ({ availableWorkflows, configuration }) => {
   const workflows = registeredWorkflows(availableWorkflows, configuration);
+  const publicWorkflows = _.cloneDeep(workflows).filter(workflow => {
+    const publicRoutes = workflow.routes.filter(route => {
+      return !(route?.private !== undefined ? route.private : false);
+    });
+    workflow.routes = publicRoutes; /* eslint-disable-line no-param-reassign */
+    return publicRoutes.length !== 0;
+  });
```
danielhochman, line 38: comment this block

frontend/packages/core/src/AppProvider/workflow.tsx
```diff
@@ -47,6 +47,8 @@ interface Route {
   path: string;
   /** Properties required by the Component that are set only via the config. */
   requiredConfigProps?: string[];
+  /** Is the workflow discoverable via search and drawer navigation. This defaults to false. */
+  private?: boolean;
```
danielhochman, line 51: would `hideNav` be a better var? it's not really private.

mcutalo88: Sure. `discoverable`, `hidden`, `hideNav`?

PR #336, opened 2020-08-25
chaos: split experiments table into config and run

wgallagher:
split experiments table into experiment_config and experiment_run
remove squirrel library


backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,66 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
 }
 
 // GetExperiments gets all experiments
 func (fs *experimentStore) GetExperiments(ctx context.Context) ([]*experimentation.Experiment, error) {
-	psql := sq.StatementBuilder.PlaceholderFormat(sq.Dollar)
-	builder := psql.Select(experimentColumns...).From(experimentsTableName)
+	sql := `
+        SELECT experiment_run.id, details FROM experiment_config, experiment_run
+        WHERE experiment_config.id = experiment_run.experiment_config_id
+        	AND experiment_run.start_time < now()`
```
Augustyniak, line 110: Can we default to returning all of the experiments? We can add more conditions later on but I think that returning all of them here (the removal of `experiment_run.start_time < now()`) condition could help us with debugging/development.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,66 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
-		builder = builder.Where(sq.Eq{"id": ids})
-	}
-
-	sql, args, err := builder.ToSql()
-	if err != nil {
-		return err
-	}
-
-	_, err = fs.db.ExecContext(ctx, sql, args...)
+	sql := `DELETE FROM experiment_run WHERE id IN $1`
```
Augustyniak, line 100: not a part of this PR: we will need to probably rename this method to something like `StopExperiment` instead and avoid removing the records from SQL.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,66 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
+			return err
+		}
+
+		// Step 2) start a new experiment with the config
+		runSql := `
+			INSERT INTO experiment_run (
+                id, 
+		        experiment_config_id,
+		        start_time,
+			    end_time,
```
Augustyniak, line 84: Per our offline discussion that we had the other day - `scheduled_end_time` should be set to the same value `end_time` is set when a new experimentation_record is created.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,66 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
+		configSql := `INSERT INTO experiment_config (id, details) VALUES ($1, $2)`
+		_, err = fs.db.ExecContext(ctx, configSql, configID, configJson)
+		if err != nil {
+			return err
+		}
+
+		// Step 2) start a new experiment with the config
+		runSql := `
+			INSERT INTO experiment_run (
+                id,
```
Augustyniak, line 81: Is this just github preview or are column names misaligned in here?

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,66 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
+		runSql := `
+			INSERT INTO experiment_run (
+                id, 
+		        experiment_config_id,
+		        start_time,
+			    end_time,                        
+		        creation_time)
+            VALUES ($1, $2, NOW(), NULL, NOW())`
+
+		_, err = fs.db.ExecContext(ctx, runSql, id.NewID(), configID)
```
Augustyniak, line 88: for `start_time` and end_time (`scheduled_end_time`) - can we add an option to pass it as part of the `CreateExperiments` method call - both of them would be optional so we could have the existing fallbacks in place (setting `start_time` to current time if it's not provided from the outside)? I know that they are not used by `experimentation` as of now but adding them now would allow us to avoid a follow up change in Clutch to unlock other projects.

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
```diff
@@ -0,0 +1,14 @@
+DROP TABLE IF EXISTS experiments;
+CREATE TABLE IF NOT EXISTS experiment_config (
+    id BIGINT,
+    details JSONB
```
Augustyniak, line 4: Does not have to be a part of this PR.

One thing that we haven't touched on yet - do you think that there would be any value in us adding some kind of a `string type` field in here to help packages depending on `experimentation` package to query the tests they are interested in? 

As of now, the only way to learn about the type of a given experiment config is to look at the value stored for `@type` in `details` JSON (i.e. if the value is equal to `type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig` then we are dealing with a server side fault injection test). //cc @donyu since he will need to query these tables when he started pushing postgres changes to other parts of Lyft infra.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: Yeah, I can do that. I'll have to add additional filtering to the RTDS code but that's not a big deal.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: Yeah, I'd like to have a separate PR that's dedicated mostly to API changes.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: right but currently FI experiments aren't started with an end time. I'd like to get parity with the way things currently work and then add end_time to FI experiments in a followup change.

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
wgallagher: I think adding a type might make sense. You can currently determine the type during queries by using the postgres json syntax but it's cumbersome.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: Yeah, we should do that in the followup API change.

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
Augustyniak: Do you think we could at the column itself as part of this PR to avoid SQL migration going forward or would it be better to add the column in the follow up PR?

backend/service/chaos/experimentation/experimentstore/experiment_store.go
Augustyniak: ping

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
```diff
@@ -0,0 +1,14 @@
...
+);
+CREATE TABLE IF NOT EXISTS experiment_run (
+    id BIGINT,
+    experiment_config_id BIGINT,
+    start_time TIMESTAMP WITH TIME ZONE,
```
donyu, line 9: How were you planning to calculate the start_time? This would involve an update to this row whenever the experiment is actually started vs. created. Do you expect this to be that different enough to provide value?

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
donyu: Yes we'll need some easy way to filter out only mobile fault injection scenarios.

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    end_time TIMESTAMP WITH TIME ZONE,
+    scheduled_end_time TIMESTAMP WITH TIME ZONE,
```
danielhochman, line 11: thoughts on incorporating the range type?

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: https://jira.lyft.net/browse/RES-1550

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: https://jira.lyft.net/browse/RES-1551

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: https://jira.lyft.net/browse/RES-1552

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: i fixed lint issues

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: https://jira.lyft.net/browse/RES-1369

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
wgallagher: I tried to use it but it doesn't support 'TIMESTAMP WITH TIME ZONE' apparently

wgallagher: it supports TIMESTAMP

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
wgallagher: https://jira.lyft.net/browse/RES-1369

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
wgallagher: right now start and create happen in the same query, at least for FI. When we support scheduling, create will be NOW() and start will be some time in the future.

backend/cmd/migrate/migrations/000004_experiments_schema_changes.up.sql
danielhochman: https://www.postgresql.org/docs/current/rangetypes.html
```
tsrange — Range of timestamp without time zone
tstzrange — Range of timestamp with time zone
```

why do we need a time zone anyway? isn't everything UTC?

wgallagher: i think we can use tstzrange

wgallagher: the only reason why i wanted to use timezone was b/c the other tables that are in clutch (auditing) use it. Eventually when we have scheduling it may matter but right now it probably doesn't.

Augustyniak: scheduling (using start_time and end_time) will be a part of the initial implementation of `mobile_experimentation` package so lets make sure that we are not blocking ourselves. That being said, both timestamp and timestamp with time zone should work here - from the perspective of the end user a frontend will always keep using a local time zone and we can convert to UTC as needed (if we decided against timestamp with timezone).

wgallagher: ok, i switched to tstzrange

wgallagher:
PTAL @danielhochman

wgallagher:
:dash:

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,69 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
+	}
 
 	for _, experiment := range experiments {
-		marshaler := jsonpb.Marshaler{}
-		buf := &bytes.Buffer{}
-		err := marshaler.Marshal(buf, experiment.GetConfig())
+		// Step 1) create the config
+		configID := id.NewID()
+
+		configJson, err := marshalConfig(experiment)
```
danielhochman, line 67: i think this should be experiment.Config, no? in fact it seems like this method should take an array of configs, not an array of experiments, as experiments already have an id.

if i'm reading correctly experiments created here would look like this in the db:
```yaml
id: 123
details:
  id: null
  config:
    @type: foo
    bar: baz
```

wgallagher:
PTAL @danielhochman

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -58,61 +50,75 @@ func New(cfg *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
 }
 
 // GetExperiments gets all experiments
 func (fs *experimentStore) GetExperiments(ctx context.Context) ([]*experimentation.Experiment, error) {
-	psql := sq.StatementBuilder.PlaceholderFormat(sq.Dollar)
-	builder := psql.Select(experimentColumns...).From(experimentsTableName)
+	sql := `
+        SELECT experiment_run.id, details FROM experiment_config, experiment_run
+        WHERE experiment_config.id = experiment_run.experiment_config_id
+        	AND now() <@ experiment_run.execution_time`
```
Augustyniak, line 119: is this condition needed in this test?

backend/service/chaos/experimentation/experimentstore/experiment_store_test.go
```diff
@@ -85,13 +103,16 @@ func TestCreateExperiments(t *testing.T) {
...
-			expected := mock.ExpectExec(regexp.QuoteMeta(test.sql))
-			if test.err != nil {
-				expected.WillReturnError(test.err)
-			} else {
-				expected.WithArgs(sqlmock.AnyArg(), test.args[1]).WillReturnResult(sqlmock.NewResult(1, 1))
+			mock.ExpectBegin()
+			for _, query := range test.queries {
+				expected := mock.ExpectExec(regexp.QuoteMeta(query.sql))
+				if test.err != nil {
+					expected.WillReturnError(test.err)
```
Augustyniak, line 110: is this ever executed? We don't have an "erroring" test case now, no? Can we get rid of it? Or maybe we need to modify our test case so that some of them are expected to fail (return an error) again?

examples/experimentation-framework/clutch-config.yaml
```diff
@@ -25,5 +25,15 @@ modules:
       rtds_layer_name: foo
       cache_refresh_interval: "60s"
 services:
+  - name: clutch.service.db.postgres
```
Augustyniak, line 28: we should probably remove this

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: It's not needed now but will be in the next change. I suppose I could remove it.

examples/experimentation-framework/clutch-config.yaml
wgallagher: Are you using this file? If so, are you using a mock server or the real server?

If the purpose of this file to serve as an example, I think it makes sense for it to use a real postgres DB instead of the mock store.

Augustyniak: Ah, I see - I think that it makes sense to keep it then - leaving the decision up to you. I haven't used this file but I agree - it's good to have some kind of an example for how it should look like.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
danielhochman: better with change, but can you address question of whether or not to change interface to array of configs vs array of experiments?

wgallagher: It needs to be an array of experiments. I set the experiment.Id for each experiment on line 96.

danielhochman: still doesn't quite track for me because the user input will allow an id if we gen docs for the API or someone looks at the protos but i think it's trivial for now

wgallagher:
:dash:

wgallagher:
:dash:

PR #339, opened 2020-08-25
migrate: connect to database with config

danielhochman:
### Description
Changes migrate to read from the clutch config file for the database connection instead of requiring a handcrafted connection string.

### Testing Performed
Local
### GitHub Issue

Fixes #

### TODOs
- [x] test local


backend/gateway/gateway.go
```diff
@@ -32,28 +31,31 @@ type ComponentFactory struct {
...
+				panic(err)
+			}
+		}()
+		tmpLogger.Fatal("could not instantiate logger", zap.Error(err))
 	}
+	defer func() {
+		if err := logger.Sync(); err != nil {
+			panic(err)
+		}
+	}()
```
mcutalo88, line 51: Why was the previous `tmpLogger` so much simpler? Looks like a lot of code for one log statement.

backend/cmd/migrate/migrate.go
```diff
@@ -2,43 +2,44 @@ package main
...
-func run(args []string) error {
-	flags := flag.NewFlagSet(args[0], flag.ExitOnError)
-	var (
-		migrationDir = flags.String("migrations", "./migrations", "Directory with sql migrations.")
-		// TODO: Read from the gateway config to create the connection.
-		//config       = flags.String("config", "../backend/clutch-config.yaml", "config yaml to parse for db connection")
-		connection = flags.String("connection", "", "Connection string for the database.")
-	)
+func run() error {
+	f := gateway.ParseFlags()
```
mcutalo88, line 21: A dangerous side effect of running this without args is the usage of ones `clutch-config.yaml` which could be their production configuration. We should probably output what datastore we are modifying and ask the user to confirm before proceeding.

backend/gateway/gateway.go
danielhochman: the tmp logger is only intended for use when the logger can't be configured. before this change the user will always get log output in mixed formats for the config filename log line.

i can simplify this though if i change the interface to RunWithConfig to accept flags so it has the filename for logging purposes.

backend/cmd/migrate/migrate.go
```diff
@@ -2,43 +2,85 @@ package main
...
+		logger.Fatal("no database found in config", zap.String("file", f.BaseFlags.ConfigPath))
+	}
+	fmt.Printf("connecting to '%s' for migration\n", hostInfo)
+	if !f.Force {
+		fmt.Printf("this could cause irrevocable data loss, is this okay? [Y/n] ")
+		var answer string
+		if _, err := fmt.Scanln(&answer); err != nil && err.Error() != "unexpected newline" {
+			logger.Fatal("could not read user input", zap.Error(err))
+		}
+		if answer != "" || strings.ToLower(answer) == "y" {
```
mcutalo88, line 77:
```suggestion
		if answer == "" || strings.ToLower(answer) == "y" {
```

Right now this always aborts.

backend/module/authn/authn.go
```diff
@@ -92,9 +92,9 @@ func (a *api) Callback(ctx context.Context, request *authnv1.CallbackRequest) (*
 	// set the cookie header
 	// redirect back to original location
 	md := metadata.New(map[string]string{
-		"Location":         redirectURL,
-		"Location-Status":  "303",
-		"Set-Cookie-Token": token,
+		"Location":          redirectURL,
+		"Location-Status":   "303",
+		"Link-Cookie-Token": token,
```
mcutalo88, line 97: This looks like an unrelated change?

backend/gateway/mux/mux.go
```diff
@@ -76,7 +76,7 @@ func customResponseForwarder(ctx context.Context, w http.ResponseWriter, resp pr
 		return nil
 	}
 
-	if cookies := md.HeaderMD.Get("Set-Cookie-Token"); len(cookies) > 0 {
+	if cookies := md.HeaderMD.Get("Link-Cookie-Token"); len(cookies) > 0 {
```
mcutalo88, line 79: Same comment as below

backend/module/authn/authn.go
danielhochman: whoops, was overzealous goland refactor. good catch.

PR #366, opened 2020-09-02
experimentation: add experiment details view

Augustyniak:
This PR adds a detail view for experiments managed by `experimentation` package. It's going to be merged after the following PRs are merged:
- https://github.com/lyft/clutch/pull/377
- https://github.com/lyft/clutch/pull/378
- ~https://github.com/lyft/clutch/pull/380~
- https://github.com/lyft/clutch/pull/381

Screen Shot 2020-09-03 at 11 39 25 PM

When it comes to the implemented approach, it's the server that's responsible for deciding what types of fields should be displayed as part of the "experiment details screen". The decision to move the logic responsible for making a decision for what fields to display in frontend to backend was made mainly because of the increased flexibility of such solution. 

At Lyft, we have multiple types of experiments. Server-side experiments are going to be handled by Clutch's `serverexperimentation` package while mobile experiments will be handled by a package that's internal to Lyft. Both `mobileexperimentation` and `serverexperimentation` packages depend on `experimentation` package.

Each package may want to display the details of their experiments in a different way (addition fields displayed for  mobile experiments as compared to server experiments). For this reason, moving rendering logic to `experimentation_store` gives us an option to implement a `registry` that would allow each module that depends on `experimentation` package to register a function that's used to transform their experiment into the list of fields that are then rendered in the frontend. The registry is going to be implemented as part of the separate PR. For now, all types of experiments are displayed in the same way.


api/chaos/experimentation/v1/experiment_run_config_pair_details.proto
```diff
@@ -0,0 +1,24 @@
...
+package clutch.chaos.experimentation.v1;
+
+import "google/protobuf/any.proto";
+import "google/protobuf/timestamp.proto";
+
+import "chaos/experimentation/v1/form.proto";
+
+option go_package = "experimentationv1";
+
+message ExperimentRunConfigPairDetails {
```
wgallagher, line 12: any reason this is in its own proto?

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -31,6 +33,14 @@ message GetExperimentsResponse {
   repeated Experiment experiments = 3;
 }
 
+message GetExperimentRunConfigPairDetailsRequest {
```
wgallagher, line 36: this is a pretty long name. maybe GetExperimentRunRequest? or something similar?

api/chaos/experimentation/v1/experiment_run_config_pair_details.proto
wgallagher, line 12: I think it would be fine to call this 'ExperimentRunDetails'

wgallagher: or just ExperimentDetails?

api/chaos/experimentation/v1/form.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+package clutch.chaos.experimentation.v1;
+
+option go_package = "experimentationv1";
+
+message Form {
```
wgallagher, line 7: I think it would make more sense to name this 'Properties'/'Property'. Form/TextField seem like UI elements. I don't think that's what these represent?

frontend/workflows/experimentation/src/view-experiment-run.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { useState } from "react";
```
wgallagher, line 1: would be better to have Derek review the UI parts

api/chaos/experimentation/v1/experiment_run_config_pair_details.proto
Augustyniak: I like ExperimentRunDetails (experiment itself is a an extremely overloaded word)

Augustyniak: just to avoid polluting the 'main' proto - do you think it should be a  part of the main experimentation file?

api/chaos/experimentation/v1/form.proto
Augustyniak: `Properties` sounds good 👍

api/chaos/experimentation/v1/experiment_run_config_pair_details.proto
Augustyniak: I would prefer to keep it where it is for now since experimentation.proto is getting big anyway. Let me know if you have as strong opinion here.

frontend/workflows/experimentation/src/view-experiment-run.tsx
Augustyniak: Could you take a look too - you have more context as when it comes to how this view should/could work. I can ask @dschaller  for a separate review when it comes to the correctness of React etc. but would prefer this PR not to be blocked on this.

api/chaos/experimentation/v1/experiment_run_config_pair_details.proto
wgallagher: no strong opinion

api/chaos/experimentation/v1/properties.proto
```diff
@@ -0,0 +1,14 @@
...
+
+package clutch.chaos.experimentation.v1;
+
+option go_package = "experimentationv1";
+
+message Properties {
+  repeated Text items = 1;
+}
+
+message Text {
```
wgallagher, line 11: rename to Property?

wgallagher (approved):

api/chaos/experimentation/v1/properties.proto
Augustyniak: done

frontend/workflows/experimentation/src/index.tsx
```diff
@@ -18,6 +19,13 @@ const register = (): WorkflowConfiguration => {
         description: "List Experiments.",
         component: ListExperiments,
       },
+      viewExperimentRun: {
```
Augustyniak, line 22: @dschaller @mcutalo88 Is it possible for one route (`listExperiments` from line 16.) to require the existance of another route (`viewExperimentRun`)?

PR #368, opened 2020-09-02
experimentation framework: rename DeleteExperiments -> StopExperiments

wgallagher:


wgallagher:
simple rename, followup change will stop deleting records from the experiment_run table on experiment completion

PTAL @danielhochman

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -54,9 +54,9 @@ service ExperimentsAPI {
     };
     option (clutch.api.v1.action).type = READ;
   }
-  rpc DeleteExperiments(DeleteExperimentsRequest) returns (DeleteExperimentsResponse) {
+  rpc StopExperiments(StopExperimentsRequest) returns (StopExperimentsResponse) {
     option (google.api.http) = {
-      post : "/v1/experiments/delete"
+      post : "/v1/experiments/stop"
```
wgallagher, line 59: @Augustyniak your pending change will need to be adjusted to point to this URI

Augustyniak: makes sense 👍

PR #369, opened 2020-09-02
don't delete experiment_run row when stopping experiment

wgallagher:


backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -82,7 +82,7 @@ func (fs *experimentStore) CreateExperiments(ctx context.Context, experiments []
 		        experiment_config_id,
 		        execution_time,
 		        creation_time)
-            VALUES ($1, $2, tstzrange(NOW(), NULL), NOW())`
+            VALUES ($1, $2, tstzrange(NOW(), NULL, '[]'), NOW())`
```
Augustyniak, line 85: noob question - what is this `[]` for?

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -114,7 +118,8 @@ func (fs *experimentStore) StopExperiments(ctx context.Context, ids []uint64) er
 func (fs *experimentStore) GetExperiments(ctx context.Context) ([]*experimentation.Experiment, error) {
 	sql := `
         SELECT experiment_run.id, details FROM experiment_config, experiment_run
-        WHERE experiment_config.id = experiment_run.experiment_config_id`
+        WHERE experiment_config.id = experiment_run.experiment_config_id
+        AND NOW() <@ experiment_run.execution_time`
```
Augustyniak, line 122: Can we keep this 'generic' GetExperiments method return all experiments (unless the caller specifies otherwise)?

backend/service/chaos/experimentation/experimentstore/experiment_store.go
wgallagher: this means that both sides of the range bound are inclusive

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -138,10 +138,14 @@ func toProto(t *time.Time) (*timestamp.Timestamp, error) {
...
 func (fs *experimentStore) StopExperiments(ctx context.Context, ids []uint64) error {
 	if len(ids) != 1 {
-		// TODO: This API will be renamed to StopExperiment and will take a single ID as a parameter
+		// TODO: This API will be changed to take a single ID as a parameter
 		return errors.New("A single ID must be provided")
 	}
-	sql := `DELETE FROM experiment_run WHERE id = $1`
+	sql :=
+		`UPDATE experiment_run 
+         SET execution_time = tstzrange(lower(execution_time), NOW(), '[]')
```
Augustyniak, line 146: does this update execution time only if:
1. end_time is nil
2. Or `end_time` > `Now()`

Asking mainly because of 2) which is needed if we don't want to allow people to stop already stopped experiment.

wgallagher: I'll have to add an additional predicate to the where clause to support this: 'AND (end_time == nil OR end_time > Now())

Augustyniak: sounds good

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -138,10 +138,14 @@ func toProto(t *time.Time) (*timestamp.Timestamp, error) {
...
 	if len(ids) != 1 {
-		// TODO: This API will be renamed to StopExperiment and will take a single ID as a parameter
+		// TODO: This API will be changed to take a single ID as a parameter
 		return errors.New("A single ID must be provided")
 	}
-	sql := `DELETE FROM experiment_run WHERE id = $1`
+	sql :=
+		`UPDATE experiment_run 
+         SET execution_time = tstzrange(lower(execution_time), NOW(), '[]') 
+         WHERE id = $1 AND (lower(execution_time) IS NOT NULL OR clock_timestamp() <@ execution_time)`
```
Augustyniak, line 147: 1. shouldn't this be `upper` (as opposed to `lower`)?
1. Can we use now() instead of `clock_timestamp()`. Per https://www.cybertec-postgresql.com/en/postgresql-now-vs-nowtimestamp-vs-clock_timestamp/#:~:text=Keep%20in%20mind%3A%20now(),()%20changes%20all%20the%20time I think that `NOW()` is what we want since the value returned from this function does NOT change within a given SQL transaction (which matters in our case since we call `NOW()` twice)

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -138,10 +138,14 @@ func toProto(t *time.Time) (*timestamp.Timestamp, error) {
...
 	if len(ids) != 1 {
-		// TODO: This API will be renamed to StopExperiment and will take a single ID as a parameter
+		// TODO: This API will be changed to take a single ID as a parameter
 		return errors.New("A single ID must be provided")
 	}
-	sql := `DELETE FROM experiment_run WHERE id = $1`
+	sql :=
+		`UPDATE experiment_run 
+         SET execution_time = tstzrange(lower(execution_time), NOW(), '[]') 
+         WHERE id = $1 AND (upper(execution_time) IS NOT NULL OR NOW() < upper(execution_time))`
```
Augustyniak, line 147:
```suggestion
         WHERE id = $1 AND (upper(execution_time) IS NULL OR NOW() < upper(execution_time))`
```

Augustyniak (approved):
Pls commit my suggestion. Other than that it looks fine - thanks!

wgallagher:
:dash:

PR #373, opened 2020-09-03
add CreateExperimentRequest start/end time

wgallagher:


api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -2,21 +2,34 @@ syntax = "proto3";
...
 import "validate/validate.proto";
 
-import "api/v1/annotations.proto";
-
 option go_package = "experimentationv1";
 
+message ExperimentParams {
+
+  // The experiment configuration specific to the type of experiment
+  google.protobuf.Any config = 2;
```
Augustyniak, line 16: this should start with `1`

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -2,21 +2,34 @@ syntax = "proto3";
...
+
+  // The time when the experiment should end, If unspecified, the experiment runs indefinitely.
+  google.protobuf.Timestamp end_time = 4;
+}
+
 message Experiment {
   fixed64 id = 1;
-  google.protobuf.Any config = 2;
+
+  ExperimentParams params = 2;
```
Augustyniak, line 28: If we change `Experiment` type we will also change the look of the Experiment that are returned from this method (which will break existing usages of `GetExperiments`):

For example:
before change:
```
Experiment:
    id: 123
    config: {}
```

After change
```
Experiment:
   id: 123
   experiment_params: 
       config: {}
``

From the perspective of the user of this API, the second proto looks harder to use - it uses nested `experiment_params` type that users of this api should not really need to care about.

Comment (Augustyniak) [reply api/chaos/experimentation/v1/experimentation.proto:28]: I think that your initial idea to just introduce a separate proto for experiment creation would result in a "nicer to use API" here:

```proto
message ExperimentCreationParams {
    int64 id = 1
    google.protobuf.Timestamp start_time = 2;
    google.protobuf.Timestamp end_time = 3;
    // ...
}

message CreateExperimentsRequest {
   repeated ExperimentCreationParams experiments = 1
}

message CreateExperimentsResponse {
  repeated Experiment experiments = 1;
}
```

With this API, a user quering the list of experiments receives the following "JSON" (notice the flat structure of the returned object, `config` is not nested):
```
Experiment:
   id: 123
   config: {}
```

wgallagher: but that gets us back to the problem that @danielhochman was trying to avoid: passing an uninitialized id to the create function.

wgallagher: this is the comment I'm referring to: https://github.com/lyft/clutch/pull/336#discussion_r478695477

Augustyniak: The more I think about the more I like the proposed approach ^. Lets avoid coupling of request and response data. Pretty soon, our `GetExperimentsResponse` will need to start returning fields such as `termination_reason` or `scheduled_end_time` - both of them should *not* be a part of `CreateExperimentRequest`

wgallagher: another option is to only support creating a single experiment at a time. then we don't need a struct to carry the params, they can be defined in CreateExperimentRequest.

Augustyniak: @wgallagher my bad - `ExperimentCreationParams` doesn't have to have `id` field.

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -2,21 +2,37 @@ syntax = "proto3";
...
+  google.protobuf.Timestamp start_time = 3;
+
+  // The time when the experiment should end, If unspecified, the experiment runs indefinitely unless it is manually
+  // stopped by a user.
+  google.protobuf.Timestamp end_time = 4;
 }
 
 message CreateExperimentsRequest {
-  repeated Experiment experiments = 1 [ (validate.rules).repeated = {min_items : 1} ];
+  // The experiment configuration specific to the type of experiment
```
Augustyniak, line 28: we need to rename `CreateExperimentsRequest` to `CreateExperimentRequest`

backend/module/chaos/experimentation/api/experimentation.go
```diff
@@ -59,15 +60,29 @@ func (s *Service) Register(r module.Registrar) error {
...
+	// If start time is not provided, default to starting now
+	now := time.Now()
+	startTime := &now
+	if req.StartTime != nil {
+		s := req.StartTime.AsTime()
+		startTime = &s
+	}
+
+	var endTime *time.Time = nil
+	if req.EndTime != nil {
```
Augustyniak, line 75: If passed time is not null we assign the current time as the end time?

backend/module/chaos/experimentation/api/experimentation.go
```diff
@@ -59,15 +60,29 @@ func (s *Service) Register(r module.Registrar) error {
...
+func (s *Service) CreateExperiment(ctx context.Context, req *experimentation.CreateExperimentRequest) (*experimentation.CreateExperimentResponse, error) {
 	s.createExperimentsStat.Inc(1)
 
-	err := s.experimentStore.CreateExperiments(ctx, req.Experiments)
+	// If start time is not provided, default to starting now
+	now := time.Now()
+	startTime := &now
+	if req.StartTime != nil {
+		s := req.StartTime.AsTime()
+		startTime = &s
```
Augustyniak, line 71: nit: combine these two lines into one line?

```suggestion
		startTime = &req.StartTime.AsTime()
```

wgallagher: go doesn't permit taking the address of a function result

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -2,25 +2,41 @@ syntax = "proto3";
...
+
+  // The experiment configuration specific to the type of experiment
   google.protobuf.Any config = 2;
+
+  // The time when the experiment should start. If unspecified, defaults to 'now'
+  google.protobuf.Timestamp start_time = 3;
+
+  // The time when the experiment should end, If unspecified, the experiment runs indefinitely unless it is manually
+  // stopped by a user.
+  google.protobuf.Timestamp end_time = 4;
```
Augustyniak, line 24: what about `scheduled_end_time` and `creation_time` fields? Should their addition be a part of this PR?

wgallagher: i can add creation_time if you want to put it in the UI. Do you currently have a use for scheduled_end_time? should it be the same as end_time initially?

Augustyniak: We have a use case for `scheduled_end_time` - when a user taps a "Stop experiment" button we will want to update the end_time and leave `scheduled_end_time` unchanged - both end_time and scheduled_end_time will be displayed in experiment details view (which is going to be merged into the main 'soon').

Augustyniak: `creation_time` will be displayed in experiment details view too.

Augustyniak:
@wgallagher there is a change in `backend/backend` whose size is equal to 83MB. Should we get rid of it? (not sure what this file is)

api/chaos/experimentation/v1/experimentation.proto
Augustyniak: nvm, we can address this in a separate PR too - we use `ExperimentRunDetails` when we display creation_time and scheduled_end_time in the UI so the changes that I requested are not a must here.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -49,55 +50,62 @@ func New(_ *any.Any, _ *zap.Logger, _ tally.Scope) (service.Service, error) {
...
 		// TODO(bgallagher) temporarily returning the experiment run ID. Eventually, the CreateExperiments function
 		// will be split into CreateExperimentConfig and CreateExperimentRun in which case they will each return
 		// their respective IDs
-		experiment.Id = uint64(runId)
-	}
-
-	err = tx.Commit()
-	return err
+		Id:     uint64(runId),
+		Config: config,
```
Augustyniak, line 107: we should be setting `start_time` and `end_time` too.

Augustyniak (approved):
Approving but pls address the comment regarding returning of start_time and end_time fields when experiment is created

Augustyniak (approved):

wgallagher:
:dash:

PR #381, opened 2020-09-08
core: call function only if it's available

Augustyniak:
Call `onChange` function only when it's passed by the caller (user of a `TextField`).


frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -34,7 +34,9 @@ const TextField: React.FC<TextFieldProps & MuiTextFieldProps> = ({
   const onKeyDown = (
     e: React.KeyboardEvent<HTMLDivElement | HTMLTextAreaElement | HTMLInputElement>
   ) => {
-    onChange(e as React.ChangeEvent<any>);
+    if (onChange) {
```
dschaller, line 37:
```suggestion
    if (onChange !== undefined) {
```

Augustyniak: updated

dschaller (approved):

PR #385, opened 2020-09-08
k8s: fix ListPods validation

gkleiman:
### Description
This change allows specifying an empty cluster in a ListPodsRequest.
This can be used in a multi cluster environment to indicate that we want
to list pods from all clusters.

### Testing Performed
Manual test using Lyft's multi-cluster k8s client.


danielhochman:
can we pass a sentinel value instead e.g. `*`? using default values with no validation means that not passing a cluster accidentally could result in a fanout.

gkleiman:
Closing since our internal implementation already fans out when an unknown/invalid cluster name (e.g., `*`) is passed.

Issue #30, opened 2020-07-18
core: implement topology API


danielhochman:
**Description**

We have a topology API definition from the original private implementation of Clutch.

https://github.com/lyft/clutch/blob/master/api/topology/v1/topology_api.proto

It is not implemented in the public repo. This API will be used in the future for auto-complete, fast resource lookup (compared to vendor API), and relationship discovery (e.g. which service or user owns a resource such as a DB).

Complexity [S/M/L]: L

Milestones
- [x] Database schema and configuration
- [x] Topology cache scaffolding
- [x] Topology cache protos
- [x] Topology cache leader election
- [x] Topology cache cashing functionality
- [x] Topology cache actually populating cache (k8s)
- [x] Topology cache actually populating cache (aws)
- [x] Cache expiry
- [x] Topology Search API Protos
- [x] Topology Search API

Will add more milestones as we make progress.

**Wireframes, Mockups, or Screenshots (if applicable)**

PR #391, opened 2020-09-09
kubernetes: getter for clientsets

mcutalo88:
### Description
While prototyping for the topology feature set https://github.com/lyft/clutch/issues/30, I needed to access all of the clientsets that were currently configured in the k8s service. This was so I could [setup informers](https://github.com/lyft/clutch/pull/362/files#diff-30185c438f5bbaf098b6a77eab58cf02R25-R28 ) for all configured clusters to populate the cache.

### Testing Performed

n/a

### GitHub Issue

To unblock on https://github.com/lyft/clutch/issues/30.

### DIFF
--- a/backend/mock/service/k8smock/k8smock.go
+++ b/backend/mock/service/k8smock/k8smock.go
@@ -7,6 +7,7 @@ import (
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/uber-go/tally"
 	"go.uber.org/zap"
+	"k8s.io/client-go/kubernetes"
 
 	k8sv1 "github.com/lyft/clutch/backend/api/k8s/v1"
 	"github.com/lyft/clutch/backend/service"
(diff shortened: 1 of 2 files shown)


danielhochman:
This is where hanging the object discovery off of the service in some kind of `GetCacheableObjects() ([]*any.Any, error) or `GetTopologyObjects() ([]*any.Any, error)` would make more sense I think.

Exposing the clientsets in this way is a bit odd because they are fungible (can change namespace dynamically), which is why calls to get the clientset normally always go through the manager. For this case it's fine but I worry that future users of that method may not realize that.

mcutalo88:
I like the idea of `GetCacheableTopologyObjects` to deter users from using this method, however I dont like returning `any.Any` if I can avoid. Might as well just return the appropriate object based on the service in question, in this case it will still always be `clientsets`.

danielhochman:
To clarify, my suggestion for `GetTopologyObjects()` was to return the actual pods, HPAs, etc to be cached. Keeping all the crawling, enumeration logic in the service. Maybe we should sketch out this code structure in the RFC?

mcutalo88:
I see. k8s has [event handlers (informers)](https://github.com/lyft/clutch/pull/362/files#diff-10ffe5b68096f01c415e1e354687e5afR15-R37) which makes it unique in this case as no other providers that we are support have this type of functionality. Maybe in this case `GetTopologyObjects()` would just return informer objects.

Even for aws `GetTopologyObjects` would work best if it returned a channel I guess of `any` type? and we process entries from each paginated call as we get them.

I wanted to centralize all caching to the topology package and simply have service expose whatever metadata was necessary I felt it was easier to control.

I'll add some pseudo code structure to the RFC.

PR #396, opened 2020-09-10
gateway: move to protojson marshaler

danielhochman:
### Description
`jsonpb` has a bug where deserializing a map with `null` messages results in an empty message with empty values. This is fixed in `protojson`. The signature changed for the ContentType() method of the marshaler interface from v1 to v2 of grpc-gateway so a simple 1:1 shim is needed to use the new marshaler.


danielhochman:
Closing in favor of #575

PR #399, opened 2020-09-10
header: persist theme through global class overrides

dschaller:
### Description

Persist theme using CSS instead of relying on inferred CSS via the global theme. This fixes an issue where Material UI would change the global styles and any styles that were inferred (e.g. `color="secondary"` when a select menu was opened) would update.

Screen Shot 2020-09-10 at 1 15 12 PM

### Testing Performed
Manual


danielhochman (approved):
:balloon:

Augustyniak:
Thanks for fixing this 💯

PR #400, opened 2020-09-10
core: rename topology db migration steps

mcutalo88:
### Description
Rename topology migration steps, `000004` was already taken by experiments.


danielhochman (approved):
can we add a lint or test or something in the future?

mcutalo88:
Yes that would be most ideal, i'll make an issue.

PR #402, opened 2020-09-11
core: add new select component

dschaller:
### Description

Add a select component to @clutch/core.

Screen Shot 2020-09-10 at 5 01 31 PM

Screen Shot 2020-09-10 at 5 01 27 PM

This is just a first pass at adding a basic select. In follow up PRs we can add validation and error messaging.

### DIFF
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -36,7 +36,7 @@ const StyledSelect = styled(MuiSelect)`
 
 interface SelectOption {
   label: string;
-  value: string;
+  value?: string;
 }
 
 interface SelectProps {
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/select.tsx
```diff
@@ -0,0 +1,80 @@
...
+  max-width: ${props["data-max-width"] || "500px"};
+  `}
+`;
+
+interface SelectProps {
+  defaultOption?: string;
+  label: string;
+  maxWidth?: string;
+  name: string;
+  options: string[];
```
Augustyniak, line 42: would it make sense to be able to specify options as as list of `labels/values`? Pretty often, we want to have a dropdown that displays one string but is backed-up by some internal value that's not visible to a user. Something like:


```typescript
interface SelectProps {
    //...
    options: Option[]
    //...
}

interface Option {
    label: string
    value: string
}
```

Augustyniak: Obviously, this is not a blocker - just a proposal for a small modification of API.

dschaller: 💯

Augustyniak: Happy to help with the change - let me know and I can commit changes and push them.

dschaller: Should have it finished this morning :)

Augustyniak (approved):

danielhochman (approved):

PR #412, opened 2020-09-15
frontend: add ClipboardButton component

gkleiman:
### Description
This change adds a clipboard button component.

When clicked a clipboard button will copy text to the clipboard and briefly change its icon to a checkbox to provide feedback to the user.

### Testing Performed
Manual testing on a private workflow.


danielhochman (approved):
nice, looks familiar :)

PR #413, opened 2020-09-15
frontend: add children param to NotePanel

gkleiman:
### Description
This change adds a `children` parameter to the `NotePanel` frontend component.

This makes it possible to create a note panel that contains complex notes.

### Testing Performed
Manual testing using a private workflow.

### DIFF
--- a/frontend/packages/core/src/Feedback/note.tsx
+++ b/frontend/packages/core/src/Feedback/note.tsx
@@ -24,10 +24,10 @@ export interface NoteConfig extends NoteProps {
 
 interface NotePanelProps {
   direction?: "row" | "column";
-  notes: NoteConfig[];
+  notes?: NoteConfig[];
 }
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Feedback/note.tsx
```diff
@@ -24,10 +24,10 @@ export interface NoteConfig extends NoteProps {
 
 interface NotePanelProps {
   direction?: "row" | "column";
-  notes: NoteConfig[];
+  notes?: NoteConfig[];
```
danielhochman, line 27: should something like `children?: React.ReactElement` be part of the props here?

dschaller: We get this for free with the functional component type interface

dschaller (approved):

PR #424, opened 2020-09-17
experimentation: set scheduled_end_time of created experiments

Augustyniak:
### Description
What `scheduled_end_time` is supposed to be used for if we have start_time and end_time fields?

When a new experiment is created its `end_time` and `scheduled_end_time` fields are set to exactly the same value. If an experiment is manually stopped by a user - with the use of a stop button in the experiment details view - we stop the experiment by setting its `end_time` to the current time. For these cases, we can use scheduled_end_time field to see whether there is a difference between originally `scheduled_end_time` and the actually `end_time` to find experiments that were stopped manually.

### Testing Performed
Unit Tests

### DIFF
--- a/backend/service/chaos/experimentation/experimentstore/experiment_store.go
+++ b/backend/service/chaos/experimentation/experimentstore/experiment_store.go
@@ -85,7 +85,7 @@ func (fs *experimentStore) CreateExperiment(ctx context.Context, config *any.Any
 	// Step 2) start a new experiment with the config
 	runSql := `
 			INSERT INTO experiment_run (
-                id, 
+            id,
 		        experiment_config_id,
 						execution_time,
 						scheduled_end_time,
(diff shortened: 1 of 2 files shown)


kathan24 (approved):

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -87,9 +87,10 @@ func (fs *experimentStore) CreateExperiment(ctx context.Context, config *any.Any
 			INSERT INTO experiment_run (
                 id, 
 		        experiment_config_id,
-		        execution_time,
+						execution_time,
+						scheduled_end_time,
```
danielhochman, line 91: nit: indentation doesn't match

(unfortunately i don't know of a way to automatically lint these sql statements with how we're using them currently)

PR #425, opened 2020-09-17
resolver: automatically trim whitespace from strings

danielhochman:
### Description
Automatically trim whitespace from any string values for resolver input.

From internal chat:

>UX suggestion - at least for the resize ASG workflow, any leading/trailing whitespace in the “search by name” form input leads to a “not found” error, which is something one can run into when copy/pasting ASG names around and fat-fingering selection. What about sanitizing the input somehow by trimming leading/trailing whitespace?

### Testing Performed
Manual

### DIFF
--- a/frontend/packages/core/src/Resolver/index.tsx
+++ b/frontend/packages/core/src/Resolver/index.tsx
@@ -69,13 +69,16 @@ const Resolver: React.FC<ResolverProps> = ({ type, searchLimit, onResolve, varia
   React.useEffect(() => loadSchemas(type, dispatch), []);
 
   const submitHandler = () => {
+    // Move to loading state.
     dispatch({ type: ResolverAction.RESOLVING });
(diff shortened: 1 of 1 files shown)


danielhochman:
@dschaller opened as draft, let me know what you think of this approach.

dschaller (approved):
Generally looks good to me. It would be awesome if we had the text field component configurable to trim leading and trailing whitespace such that the user can see it and it'll be safe for any values (e.g. this won't handle nested dicts at the moment) but that can happen at a later point

danielhochman:
@dschaller i thought about that but i think for it to work as expected it would need to be done on focus change and on submit which is pretty complicated. i will open a follow-up issue for that type of field prop enhancement.

danielhochman:
filed https://github.com/lyft/clutch/issues/450

PR #426, opened 2020-09-18
experimentation: rename ExperimentStore to Storer

Augustyniak:
I read a little bit about Go and apparently `-er` is the default choice when it comes to the names of interfaces in Go.

* https://stackoverflow.com/questions/38842457/interface-naming-convention-golang
* https://golang.org/doc/effective_go.html#interface-names

### DIFF
--- a/backend/module/chaos/experimentation/api/experimentation.go
+++ b/backend/module/chaos/experimentation/api/experimentation.go
@@ -25,7 +25,7 @@ const (
 
 // Service contains all dependencies for the API service.
 type Service struct {
-	experimentStore             experimentstore.ExperimentStore
+	storer                      experimentstore.Storer
 	logger                      *zap.SugaredLogger
 	createExperimentStat        tally.Counter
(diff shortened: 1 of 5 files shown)


backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -24,7 +24,7 @@ import (
 const Name = "clutch.service.chaos.experimentation.store"
 
 // ExperimentStore stores experiment data
-type ExperimentStore interface {
+type ExperimentStorer interface {
```
danielhochman, line 27: While we're at it, might as well also avoid stutter/repetition per convention.

See [Go Blog: Package Names -- Naming package contents](https://blog.golang.org/package-names#TOC_3.)

```suggestion
type Storer interface {
```

Augustyniak: @danielhochman could you take another look at this one, please?

danielhochman (approved):

PR #427, opened 2020-09-18
experimentation: query experiments with a given config type

Augustyniak:
### Description

`experiment_config` table that's managed by `experimentStore` can contain multiple types of experiments (i.e., "server", "mobile") at the same time. `rtds` needs to be able to get the list of server experiments only. This PRs adds an option for the callers of `experimentation` API to specify the type of the experiments they want to fetch, if a caller omits the type experiments of all types are returned.

### Testing Performed
Unit Tests


api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -42,7 +42,7 @@ message CreateExperimentResponse {
 }
 
 message GetExperimentsRequest {
-  repeated uint64 ids = 1;
+  string config_type = 1;
 }
```
danielhochman, line 46: how do we plan to expand usage of this function in the future? isn't fetching specific IDs also a valid use case? or was this already replaced by `GetExperimentRunDetailsRequest`?

Augustyniak: >how do we plan to expand usage of this function in the future?

We are planning to add an option to ask `GetExperimentsRequest ` to return only running/scheduled/completed tests. I'm currently working on the PR for this.

>isn't fetching specific IDs also a valid use case? or was this already replaced by GetExperimentRunDetailsRequest?

For now, `GetExperimentRunDetailsRequest`  replaced `GetExperimentsRequests` when it comes to getting experiments by their ID. It's possible that we will need to re-add `ids` as part of the `GetExperimentsRequest` but that's not something that we are planning on doing in the near future.

danielhochman (approved):

PR #432, opened 2020-09-18
docs: API convention and guidelines

danielhochman:
### Description
Document some of the API conventions that are used in Clutch to make code review easier.

### Testing Performed
N/A


api/README.md
```diff
@@ -21,4 +39,144 @@ enum Foo {
...
+
+The rules are meant to take as much guesswork out of naming things as possible by making the filename, package name, and services match. HTTP annotations are literal RPC, instead of trying to map RESTful concepts back onto the API.
+
+### All HTTP annotations should use the POST method
+
+**Examples of incorrect definition for this rule:**
+```
+rpc GetThing(GetThingRequest) returns (GetThingResponse) {
+  option (google.api.http) = {
+    get : "/v1/foo/thing"
```
scarlettperry, line 56: should this path match the path in the correct example `/v1/foo/getThing`?

api/README.md
```diff
@@ -21,4 +39,144 @@ enum Foo {
...
+
+**Examples of correct definition for this rule:**
+```proto
+package clutch.healthcheck.v1;
+```
+
+```proto
+package clutch.aws.ec2.v1;
+```
+
```
scarlettperry, line 95: nit: perhaps swap the order of the correct definitions so that it matches the order of the incorrect definitions?

api/README.md
```diff
@@ -21,4 +39,144 @@ enum Foo {
...
+- Filename `clutch/api/aws/ec2/v1/ec2.proto`
+```proto
+package clutch.aws.ec2.v1;
+```
+
+### Service name should match last non-version component of package name
+
+**Examples of incorrect definition for this rule:**
+
```
scarlettperry, line 117: filenames are mentioned in the correct definitions but not in the incorrect definition, so initially was a lil confused on what to compare

api/README.md
```diff
@@ -21,4 +39,144 @@ enum Foo {
...
+    option (google.api.http) = {
+      post : "/v1/bar/create"
+      body : "*"
+    };
+  }
+}
+```
+
+- :x: Leading path is `/v1/experiments`, not `/v1/foo` in accordance with package name.
+- :x: RPC method is `CreateExperiment` but HTTP mapping is `/experiments/create`.
```
scarlettperry, line 166: this block doesn't match the incorrect definition

scarlettperry (approved):

PR #435, opened 2020-09-18
experimentation: update API so that it follows guidelines

Augustyniak:
### Description
Update experimentation API according to guidelines introduced in https://github.com/lyft/clutch/pull/432

### Testing Performed
N/A


api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -68,7 +68,7 @@ message StopExperimentsResponse {
 service ExperimentsAPI {
   rpc CreateExperiment(CreateExperimentRequest) returns (CreateExperimentResponse) {
     option (google.api.http) = {
-      post : "/v1/experiments/create"
+      post : "/v1/experiment/create"
```
danielhochman, line 71: still doesn't quite match, probably the guidelines need to be improved a bit more so they are clearer.

this would ideally be `/v1/chaos/experimentation/createExperiment` to match the package `clutch.chaos.experimentation.v1` and method name `CreateExperiment`

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -80,18 +80,18 @@ service ExperimentsAPI {
     };
     option (clutch.api.v1.action).type = READ;
   }
-  rpc GetExperimentRunDetails(GetExperimentRunDetailsRequest) returns (GetExperimentRunDetailsResponse) {
+  rpc StopExperiments(StopExperimentsRequest) returns (StopExperimentsResponse) {
```
Augustyniak, line 83: I reordered methods here and the diff is bigger than it had to be

api/chaos/experimentation/v1/experimentation.proto
Augustyniak: I see, thanks for the explanation. Is `createExperiment` really preferred over `create-experiment`?

danielhochman: yep, it's arbitrary i guess but we chose to go with camelcase throughout since it matches javascript naming conventions where these names are mostly used. google also uses camelcase throughout their APIs, and while we differ from their mappings in some ways (e.g. only using POST), that's one area we didn't feel was beneficial to diverge.

Augustyniak: Updated per your comments. Thanks for the detailed explanation.

danielhochman (approved):

PR #438, opened 2020-09-18
tools: link types for gateways

dschaller:
### Description

This will allow us to use a unified type system when consuming Clutch as a submodule.

### Testing Performed

Manual


danielhochman (approved):
can you add a comment on what belongs in the linked packages list?

PR #443, opened 2020-09-19
experimentation: log error when malformed configuration is detected

Augustyniak:
### Description
Current implementation returns an empty list of experiments every time there is an issue with any of the experiments that are stored in the database. Keep this logic but add additional logging to make acting on these kind of issues easier.

### Testing Performed
Unit Tests.


backend/service/chaos/experimentation/experimentstore/experiment_store_test.go
```diff
@@ -169,74 +169,84 @@ func TestStopExperiments(t *testing.T) {
 	}
 }
 
-var getExperimentsTests = []struct {
```
Augustyniak, line 172: make this test less generic - current generic implementation adds a lot of code that's never evaluated since  the list of experiment tests has only 1 element. We can make it more complex later on if we decided that we need more complexity here.

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -179,7 +182,8 @@ func (fs *experimentStore) GetExperiments(ctx context.Context, configType string
 
 		anyConfig := &any.Any{}
 		if nil != jsonpb.Unmarshal(strings.NewReader(details), anyConfig) {
-			return nil, err
+			fs.getExperimentsLoadConfigurationError.Inc(1)
```
gkleiman, line 185: I'd log the error here to make it easier to find out why experiments can't be unmarshalled.

I am not a big fan of skipping errors, because they can hide real problems. But in this case skipping some experiments could provide a better UX than failing if there are multiple experimentation modules and one is missing.

So I think skipping here would be acceptable if we recommend users to set an alarm on this metric.

Augustyniak: addressed

backend/service/chaos/experimentation/experimentstore/experiment_store_test.go
```diff
@@ -169,74 +169,72 @@ func TestCancelExperimentRun(t *testing.T) {
...
-			if test.err != nil {
-				a.Equal(test.err, err)
-			} else {
-				a.NoError(err)
-			}
+	config := &serverexperimentation.TestConfig{}
+	err2 := ptypes.UnmarshalAny(experiment.GetConfig(), config)
+	if err2 != nil {
+		t.Errorf("setSnapshot failed %v", err2)
+	}
```
gkleiman, line 202: Is this different from `assert.NoError(err2)`?

gkleiman (approved):

backend/service/chaos/experimentation/experimentstore/experiment_store_test.go
Augustyniak: it should be similar - let me check

Augustyniak: updated, thanks for the comment!

PR #444, opened 2020-09-19
experimentation: fix get experiments being called infinite number of times

Augustyniak:
### Description
Do not attempt to fetch experiments if:
* getExperiments call failed
* getExperiments called returned 0 experiments

### Testing Performed
Verified locally that the list of experiments works if GetExperiments returns no experiments or more than 0 experiments.

### DIFF
--- a/frontend/workflows/experimentation/src/list-experiments.tsx
+++ b/frontend/workflows/experimentation/src/list-experiments.tsx
@@ -81,7 +81,7 @@ const ListExperiments: React.FC<ListExperimentsProps> = ({ columns, experimentTy
 
   const navigate = useNavigate();
 
-  if (experiments === undefined) {
+  React.useEffect(() => {
     client
       .post("/v1/chaos/experimentation/getExperiments")
       .then(response => {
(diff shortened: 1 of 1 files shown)


frontend/workflows/experimentation/src/list-experiments.tsx
```diff
@@ -74,12 +74,14 @@ interface ListExperimentsProps {
...
-  const [experiments, setExperiments] = useState([]);
+  const [experiments, setExperiments] = useState<
+    IClutch.chaos.experimentation.v1.Experiment[] | undefined
+  >(undefined);
   const [error, setError] = useState("");
 
   const navigate = useNavigate();
 
-  if (experiments.length === 0) {
+  if (experiments === undefined) {
```
dschaller, line 84: If you only want to call this once on page load you could use a react hook

```jsx
React.useEffect(() => {
  client....
}, []);
```

Augustyniak: nice! fixed - thanks!

donyu (approved):

PR #453, opened 2020-09-21
topology types

mcutalo88:
### GitHub Issue

Fixes #


mcutalo88:
@danielhochman this is ready for review

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,29 @@ message Edge {
...
+  string id = 1;
+
+  // Pb is the clutch proto object.
+  google.protobuf.Any pb = 2;
+
+  // Metadata is set by the service which produces TopologyObjects, for example k8s would extract
+  // relevent metadata that gives the TopologyAPI the ability to query against it.
+  map<string, string> metadata = 3;
+
+  // TopologyCacheAction signifies to the Topology service what to do with an incomeing TopologyObject
```
dschaller, line 174:
```suggestion
  // TopologyCacheAction signifies to the Topology service what to do with an incoming TopologyObject
```

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,29 @@ message Edge {
   // edge.
   map<string, google.protobuf.Value> metadata = 4;
 }
+
+message TopologyObject {
```
dschaller, line 163: Discussing IRL but TopologyObject makes me think this is an instance of Topology. This is pretty minor and I could just be alone here since I can't really think of something better at the moment. Any thoughts on this @danielhochman ?

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,29 @@ message Edge {
...
+  map<string, string> metadata = 3;
+
+  // TopologyCacheAction signifies to the Topology service what to do with an incomeing TopologyObject
+  //
+  // The topology service gets TopologyObjects off of the `GetTopologyObjectChannel` which is processed
+  // and stored in the topology_cache table.
+  enum TopologyCacheAction {
+    // UPSERT creates or updates items in the topology_cache table
+    UNSPECIFIED = 0;
+    UPSERT = 1;
```
dschaller, line 181:
```suggestion
    UNSPECIFIED = 0;
    // UPSERT creates or updates items in the topology_cache table
    UPSERT = 1;
```

dschaller: Is there no use case for explicit update or insert actions?

mcutalo88: Yeah good question, some background here as to why I came to this initially while prototyping, for kubernetes its easy to know if there the object such as a pod is being created or updated as there are [specific event handlers](https://github.com/lyft/clutch/pull/362/files#diff-926c4afc5e2909828dc818d6579cbf24R39-R43) just for this. Even with knowing if there is an insert or update its easier to just have `SetCache()`[perform an upsert]
(https://github.com/lyft/clutch/pull/362/files#diff-30185c438f5bbaf098b6a77eab58cf02R46-R54) instead of having an associated `UpdateCache()` function. Initially this is when I deiced to just use `UPSERT` and drop the explicit insert / update.

However now taking the use case of AWS and the implementation changes a bit, AWS client libs dont have this notion of event handlers for listening to AWS objects. Instead will have to craft this ourselves by polling, when we do poll there isn't enough information to say is this a new object or should I update an existing one? To my knowledge there isn't a way to easily see what AWS objects like EC2 instances have been remove / delete. So when we implement the AWS topology cache it seems will only ever use 1 action and thats `UPSERT`. The way will remove AWS objects is with cache expiry, unless there is a better way once I start work on the code. 

And that is how I ended up with `UPSERT` & `DELETE`. I do like being explicit though and I'm happy to change if up to `INSERT`, `UPDATE`, `DELETE` to be more explicit.

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,29 @@ message Edge {
...
+  // and stored in the topology_cache table.
+  enum TopologyCacheAction {
+    UNSPECIFIED = 0;
+    // UPSERT creates or updates items in the topology_cache table
+    UPSERT = 1;
+    DELETE = 2;
+  }
+
+  // Action denotes what the topology service should do with this object.
+  TopologyCacheAction action = 4;
```
danielhochman, line 186: does this belong on the object itself? seems like it would be in a parent message used on channel with a nested topo object.

mcutalo88: Yeah I see what you'r saying, that makes sense let me change it up a bit.

api/topology/v1/topology_api.proto
danielhochman: yeah AWS we'll have to manage with TTL + UPSERT.

i think UPDATE or CREATE_OR_UPDATE might be the correct term here, as UPSERT would be a merging of old and new. for AWS it's definitely CREATE_OR_UPDATE.

does k8s informer send a diff or the entire object?

mcutalo88: It does not send a diff, it does send the old and the new object if you wanted to perform your own diff on whatever fields your interested in. However in practice and in this use case it makes more sense to just take the entire new object and upsert it.

I like the `CREATE_OR_UPDATE` its much more clear for those who dont immediately know what an upsert is.

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,32 @@ message Edge {
...
+
+  // Pb is the clutch proto object.
+  google.protobuf.Any pb = 2;
+
+  // Metadata is set by the service which produces TopologyObjects, for example k8s would extract
+  // relevent metadata that gives the TopologyAPI the ability to query against it.
+  map<string, string> metadata = 3;
+}
+
+message TopologyCacheObject {
```
mcutalo88, line 175: Im not sure how I feel about this name and I'm open to suggestions.

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,34 @@ message Edge {
...
+  // The topology service gets TopologyCacheObject off of the `GetTopologyObjectChannel` which is processed
+  // and stored in the topology_cache table.
+  enum TopologyCacheAction {
+    UNSPECIFIED = 0;
+    CREATE_OR_UPDATE = 1;
+    DELETE = 2;
+  }
+
+  // Action denotes what the topology service should do with this object.
+  TopologyCacheAction action = 2;
```
danielhochman, line 191: avoid stutter / repetition of package name
```suggestion
  enum Action {
    UNSPECIFIED = 0;
    CREATE_OR_UPDATE = 1;
    DELETE = 2;
  }

  // Action denotes what the topology service should do with this object.
  Action action = 2;
```

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,34 @@ message Edge {
...
+
+message Resource {
+  // Id is the unique identifer of the Resource.
+  string id = 1;
+
+  // Pb is the clutch proto object.
+  google.protobuf.Any pb = 2;
+
+  // Metadata is set by the service which produces the topology Resource, for example k8s would extract
+  // relevent metadata that gives the Topology API the ability to query against it.
```
danielhochman, line 171:
```suggestion
  // relevant metadata that gives the Topology API the ability to query against it.
```

api/topology/v1/topology_api.proto
```diff
@@ -158,3 +159,34 @@ message Edge {
...
+  map<string, string> metadata = 3;
+}
+
+// A UpdateCacheRequest is used when a service such as kuberentes or aws produces objects for
```
danielhochman, line 175:
```suggestion
// A UpdateCacheRequest is used when a service such as kubernetes or aws produces objects for
```

danielhochman (approved):

PR #455, opened 2020-09-22
experimentation: fix canceling of scheduled experiments

Augustyniak:
### Description
Currently, each experiment has up to 3 timestamps associated with them:

- `execution_time` which consist of `start_time` and `end_time` (optional)
- `scheduled_end_time` which is a single timestamp
 
Every time a new experiment is started it's `start` time is set to either the current time or the time provided by a user. If `end_time` value is provided by a user both `end_time` and `scheduled_end_time` fields are set to exactly the same value.

The idea for cancelation was to set the `end_time` to the `current_time` (for when a user taps a cancel button). This does NOT work for experiments with `start_time` in the future. For experiments like this, if we try to cancel them we end up with `end_time` < `start_time` which is not accepted by postgres database since `execution_time` is of `timerange` type and this type requires the `end_time` to be greater than or equal to `start_time`.

The new structure of `experiment_run` model uses `cancellation_time` field to mark experiments that have been canceled. To get the list of running experiments one needs to use the following query:
```
select 
   *
from experiment_run 
where
  lower(execution_time) < NOW() and NOW() < upper(execution_time) and cancellation_time is NULL
```

### Testing Performed
Unit tests + verified that manual creation and cancelation of server experiments worked as expected.


danielhochman:
you may want to consider keeping a log of events on each test rather than using individual fields for this purpose. then use individual fields only used for indexing purposes not recordkeeping.

```yaml
id: 1234
status: CANCELED
schedule: [2013-12-12T09:00:00Z,inf]
events:
  - status_new: SCHEDULED
    updated_at: 2013-12-12T09:00:00Z
    user: rafal@lyft.com
  - status_new: STARTED
    updated_at: 2013-12-12T09:00:00Z
    user: system
  - status_old: STARTED
    status_new: CANCELED
    updated_at: 2013-12-12T10:00:00Z
    user: dhochman@lyft.com
```

Augustyniak:
> you may want to consider keeping a log of events on each test rather than using individual fields for this purpose. then use individual fields only used for indexing purposes not recordkeeping.

Interesting. I like the idea - I think that we will want to have a list of events sooner than later since we want to support experimentation steps (a slow rollout of a given test) and events could be helpful there. That being said, for status field since we don't have a long running job that would be updating this field automatically as the time flies we want to derive the status from experiment's `start_time`, `end_time` and `cancelation_time` fields. Your point with regard the the list of events still makes sense - the log would just need to look a little bit differently here.

Augustyniak:
@danielhochman do we have an example of a log like this in Clutch already?

donyu (approved):

danielhochman:
@rafal there is no current example of that. we aren't using the DB much as of yet outside of this flow.

danielhochman (approved):

PR #460, opened 2020-09-23
experimentation: make rtds query running experiments

Augustyniak:
### Description
To be merged after https://github.com/lyft/clutch/pull/455 is merged.

Change RTDS implementation so that it fetches only experiments that are currently running (instead of all experiments of a given type stored that are stored in a database).

### Testing Performed
- User Tests
- Manual creation and cancelation of experiments.


api/chaos/experimentation/v1/experiment_run_details.proto
```diff
@@ -10,17 +10,17 @@ import "chaos/experimentation/v1/properties.proto";
 option go_package = "experimentationv1";
 
 message ExperimentRunDetails {
+  enum Status {
+    UNSPECIFIED = 0;
+    SCHEDULED = 1;
+    RUNNING = 2;
+    COMPLETED = 3;
+    CANCELED = 4;
+    STOPPED = 5;
```
kathan24, line 19: nit: Most of these are self-explanatory but would be nice to add comments

api/chaos/experimentation/v1/experimentation.proto
```diff
@@ -41,8 +41,14 @@ message CreateExperimentResponse {
   Experiment experiment = 1;
 }
 
+enum GetExperimentsStatus {
```
kathan24, line 44: Not sure why do we have Status here and in Experiment Run details? Also, why do we only have `running` and not other statuses? Can we not have a generic status and inherit it in both places if we need?

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -148,14 +148,18 @@ func (fs *experimentStore) CancelExperimentRun(ctx context.Context, id uint64) e
...
+func (fs *experimentStore) GetExperiments(ctx context.Context, configType string, status experimentation.GetExperimentsStatus) ([]*experimentation.Experiment, error) {
+	query := `
+        SELECT 
+			experiment_run.id, 
+			details
+		FROM experiment_config, experiment_run
+		WHERE
+			experiment_config.id = experiment_run.experiment_config_id
+			AND ($1 = '' OR $1 = experiment_config.details ->> '@type')
+			AND ($2 != 'RUNNING' OR (experiment_run.cancellation_time is NULL AND NOW() > lower(experiment_run.execution_time) AND (upper(experiment_run.execution_time) IS NULL OR NOW() < upper(experiment_run.execution_time))))`
```
kathan24, line 160: Could you please add comments for each WHERE clause and explain what is being done here?

backend/service/chaos/experimentation/experimentstore/expriment_run_details.go
```diff
@@ -53,53 +59,53 @@ func NewRunDetails(fetchedID uint64, creationTime time.Time, startTime sql.NullT
...
+		return experimentation.ExperimentRunDetails_COMPLETED, nil
+	}
+}
+
+func timeToString(t sql.NullTime) string {
+	layout := "01/02/06 15:04:05"
+	if t.Valid {
+		return t.Time.Format(layout)
+	} else {
+		return "Undefined"
```
kathan24, line 92: Have we checked if Lyft has a go library for DateTime like python? If so, maybe the library has some functionality

api/chaos/experimentation/v1/experimentation.proto
Augustyniak: Because for quering we support only 2 statuses - `Running` and `Unspecfied` - and for returning the details of an experiment we do support 6 of them. If I wanted to use the same enum in both places I would need to implement more querying logic that we currently need + I do not think that we should be reusing the same `Status` types for quering and returning the results of quering since it's not clear for me whether these 2 will ever become the same.

backend/service/chaos/experimentation/experimentstore/expriment_run_details.go
Augustyniak: What functionality? I'm just providing a date format I want to use here (that's the only custom logic here?) but I think that this is fine since I want to be able to specify the format that users see in the UI

kathan24 (approved):

backend/service/chaos/experimentation/experimentstore/expriment_run_details.go
Augustyniak: Discussed offline - we can improve this by making it possible for frontend to send `Date` directly to frontend instead of being forced to convert to string before sending a given value to frontend

Augustyniak:
A lot of changes here after https://github.com/lyft/clutch/pull/455 was merged into master but these are mostly compilations fixes. A few small organizational tweaks:
- moved `experiment` proto to a separate proto file.
- moved `Status` enum from global scope to `experimentation.proto`
- made `GetExperimentsStatus` to be a part of `GetExperimentsRequest` proto.

PR #461, opened 2020-09-23
tools/scaffolding: Add Dockerfile Template for Custom Gateway

daviddyball:
### Description
This adds a `Dockerfile` to the templates used for `make scaffold-gateway` so that users can have a reference `Dockerfile` ready to go after scaffolding a custom gateway

### Testing Performed

Local testing with:
* `make scaffold-gateway` 
* `docker build -t clutch-custom-gateway:latest .`

### GitHub Issue

N/A

Fixes #

### TODOs

N/A

### DIFF
--- a/tools/scaffolding/templates/gateway/Dockerfile
+++ b/tools/scaffolding/templates/gateway/Dockerfile
@@ -1,0 +1,20 @@
+FROM node:14-buster as nodebuild
+COPY ./frontend ./frontend
+COPY ./tools/install-yarn.sh ./tools/install-yarn.sh
+COPY Makefile .
+RUN tools/install-yarn.sh && yarn --cwd frontend install
+
+FROM golang:1.15-buster as gobuild
+WORKDIR /go/src/{{ .RepoProvider}}/{{ .RepoOwner }}/{{ .RepoName }}
+COPY ./backend ./backend
+COPY Makefile .
+COPY --from=nodebuild ./frontend/build ./frontend/build/
+RUN cd backend \
+    && go mod download \
+    && go run $(go list -f "{{ .Dir }}" -m "github.com/lyft/clutch/backend")/cmd/assets/generate.go ../frontend/build \
+    && go build -tags withAssets -o ../build/clutch -ldflags="-X main.version=${VERSION}"
+
+FROM gcr.io/distroless/base-debian10
+COPY --from=gobuild /go/src/{{ .RepoProvider}}/{{ .RepoOwner }}/{{ .RepoName }}/build/clutch /
+COPY backend/clutch-config.yaml /
+CMD ["/clutch"]
--- a/tools/scaffolding/templates/gateway/Dockerfile
+++ b/tools/scaffolding/templates/gateway/Dockerfile
@@ -11,7 +11,7 @@ COPY Makefile .
 COPY --from=nodebuild ./frontend/build ./frontend/build/
 RUN cd backend \
     && go mod download \
-    && go run $(go list -f "{{ .Dir }}" -m "github.com/lyft/clutch/backend")/cmd/assets/generate.go ../frontend/build \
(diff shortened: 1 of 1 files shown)


tools/scaffolding/templates/gateway/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM node:14-buster as nodebuild
```
mcutalo88, line 1: Does `node:14-buster` already have `yarn` installed? I see in our main `dockerfile` we run our install tool for it. We could be doing extra work in our main `dockerfile` if this is the case.
https://github.com/lyft/clutch/blob/fb1e24c8bf1cedc6dcc1ce066141f99d4ead3c29/Dockerfile#L4

mcutalo88:
@daviddyball could you also do a `docker run` with this image to ensure it's working as expected?

Also thanks for the contribution!

tools/scaffolding/templates/gateway/Dockerfile
danielhochman: Our yarn policy uses yarn from a specific path to avoid version floats. This is required.

tools/scaffolding/templates/gateway/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+COPY ./frontend ./frontend
+COPY Makefile .
+RUN yarn --cwd frontend install
+
```
danielhochman, line 5: `RUN make frontend` appears to be missing here
```suggestion

RUN make frontend

```

tools/scaffolding/templates/gateway/Dockerfile
daviddyball: This gets a bit fiddly. At this stage of the build it is running in a `yarn` container that only includes `frontend/` and we can't copy `tools/install-yarn.sh` from `./` unless we include it in the gateway template.

Thoughts? Should we add `tools/install-yarn.sh` to `templates/gateway/`? Would you rather I just extract the logic from `install-yarn.sh` and run it in the `Dockerfile` directly to avoid having to include `install-yarn.sh` in the template?

tools/scaffolding/templates/gateway/Dockerfile
daviddyball: Yes indeed it was missing in my local build as well. I'll add it :+1:

tools/scaffolding/templates/gateway/Dockerfile
danielhochman: actually at the moment we don't lock the yarn version in the scaffolded gateway (i.e. no .yarnrc in frontend/). i think for now let's avoid making the change, it's a separate enhancement from just getting the build up and running in the Dockerfile. in reality, we haven't run into any issues with yarn floating yet.

normally in the gateway we access parent scripts from a go module which complicates things in this dockerfile where it's not immediately in context.

easiest way to solve it would likely be to thread a variable through with the proper version and in the dockerfile just make sure it matches.

filed https://github.com/lyft/clutch/issues/472 to track

tools/scaffolding/templates/gateway/Dockerfile
tkohrumel: @daviddyball did you manage to get a working Dockerfile for the scaffolded `clutch-gateway`? I've just started working on this myself using your preliminary template as a starting point. Thanks for sharing your work on it by the way!

daviddyball: I've actually not delved into this since I first raised the PR :flushed: If you've got something working then it should be fine to close this PR.

tkohrumel: Ah ok, thanks for the update! I'm running into some issues executing `RUN make frontend`, so will continue to debug on my end. I'll report back once I've got it sorted.

PR #467, opened 2020-09-24
serverexperimentation: support xDS V3

snowp:
### Description

V2 is being sunset EOY, so we should expect all users to want to
migrate to V3. This sets up the RTDS cache so that we support handling
both V2 and V3 streams: we maintain two separate go-control-plane
servers/caches and check both of them for pending watches that we
populate with the requested runtime layer.

### Testing Performed

Updated the existing unit tests to cover the V3 RTDS cache

### TODOs

- [] Maybe: Add integration tests for the actual server
- [] Maybe: Reduce some of the duplication between versions
- [] Maybe: Per version stats (right now we just join it all)


backend/module/chaos/serverexperimentation/rtds/rtds_cache.go
```diff
@@ -86,7 +155,7 @@ func refreshCache(ctx context.Context, store experimentstore.ExperimentStore, sn
 	}
 }
 
-func setSnapshot(snapshotCache gcpCache.SnapshotCache, rtdsLayerName string, upstreamCluster string,
+func setSnapshot(snapshotCache cacheWrapper, rtdsLayerName string, upstreamCluster string,
```
kathan24, line 158: I really like the enhancements in this function (along with the necessary change). Thank you

backend/module/chaos/serverexperimentation/rtds/rtds.go
```diff
@@ -11,10 +11,15 @@ import (
...
 	gcpV2 "github.com/envoyproxy/go-control-plane/envoy/api/v2"
-	gcpCore "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
-	gcpDiscovery "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v2"
-	gcpCache "github.com/envoyproxy/go-control-plane/pkg/cache/v2"
-	gcpServer "github.com/envoyproxy/go-control-plane/pkg/server/v2"
+	gcpCoreV2 "github.com/envoyproxy/go-control-plane/envoy/api/v2/core"
+	gcpCoreV3 "github.com/envoyproxy/go-control-plane/envoy/config/core/v3"
+	gcpDiscoveryV2 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v2"
+	gcpDiscoveryV3 "github.com/envoyproxy/go-control-plane/envoy/service/discovery/v3"
+	gcpDiscoveryServiceV3 "github.com/envoyproxy/go-control-plane/envoy/service/runtime/v3"
```
kathan24, line 18: nit : can we rename this to reflect `runtime` in it? Maybe gcpRuntimeDiscoveryServer as gcpRTDS

kathan24 (approved):
Thank you for making this change :)

backend/module/chaos/serverexperimentation/rtds/rtds_cache.go
```diff
@@ -120,33 +189,26 @@ func setSnapshot(snapshotCache gcpCache.SnapshotCache, rtdsLayerName string, ups
...
 		logger.Errorw("Error computing version", "error", err)
 		return err
 	}
 
-	currentSnapshot, _ := snapshotCache.GetSnapshot(upstreamCluster)
-	if !reflect.DeepEqual(currentSnapshot, gcpCache.Snapshot{}) {
-		currentVersion := currentSnapshot.GetVersion(gcpResource.RuntimeType)
-		if currentVersion == computedVersion {
+	currentSnapshotVersion, err := snapshotCache.GetSnapshotVersion(upstreamCluster)
+	if err == nil {
```
Augustyniak, line 203: are we fine with proceeding for cases when error is equal to nil here?

snowp: The underlying API from GCP for snapshot lookups is a bit odd, it uses error to denote whether the snapshot was found in the cache. This is how we end up with this off-looking code. 

I think ideally it would have an API similar to map lookups, e.g. `func GetSnapshot(key string) (Snapshot, bool)`, but right now the only "error" case is that the key doesn't exist in the cache, so it's keep going. This is basically doing the exact same thing as the previous deepEquals check but without having to use reflection.

PR #468, opened 2020-09-24
core / frontend: passthrough host support

mcutalo88:
### Description
Add passthrough host support for frontend static assets. If the deployed version of clutch does not have the requested frontend assets on disk the passthrough host will be used to fetch the correct assets assuming they exist for the configured provider.

In a different PR we would have a more pretty failure mode, currently you simply get a white page.

### Testing Performed

Testing was performed internally at Lyft, forcing all static asset traffic to go through the passthrough host.
```
gateway:
  assets:
    s3:
      region: us-east-1
      bucket: clutch
      key: static
```
### GitHub Issue 
Passthrough host support #471


api/config/gateway/v1/gateway.proto
```diff
@@ -85,6 +85,12 @@ message GatewayOptions {
   Timeouts timeouts = 5;
 
   repeated Middleware middleware = 6;
+
+  Frontend frontend = 7;
+}
+
+message Frontend {
```
danielhochman, line 92:
```suggestion
  Assets assets = 7;
}

message Assets {
```

api/config/gateway/v1/gateway.proto
```diff
@@ -85,6 +85,12 @@ message GatewayOptions {
...
+  string cdn_url = 1;
```
danielhochman, line 93: it's not really a CDN, it's a passthrough. a CDN would go in front of the server whereas this is behind it.

also in our case we use s3, not a publicly accessible URL, so the config will need to a be bit different.

danielhochman (commented):
thanks for picking this up. some initial comments on naming clarity.

api/config/gateway/v1/gateway.proto
mcutalo88: Yeah sure sounds good

api/config/gateway/v1/gateway.proto
mcutalo88: Yup makes sense I can change it up. 

Why will the config be different for us? Our IAM role will have access to these assets?

danielhochman: we need to use an s3 client in order to sign the requests, a regular HTTP request won't do

mcutalo88: Yeah i see what your saying now after reading a bit, let me compensate.

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,51 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Serve!
 	if f, err := a.fileSystem.Open(r.URL.Path); err != nil {
+		// If not a known static asset and a asset provider is configured, try streaming from the configured provider.
+		if a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static") {
+			// We attatch this header simply for observibility purposes.
+			// Otherwise its difficult to know if the assets are being served from the configured provider.
+			w.Header().Set("x-clutch-asset-passthrough", "true")
+
+			// TODO: handle errors
```
mcutalo88, line 78: I still need to handle errors some how, I haven't decided on what way would be best yet.

api/config/gateway/v1/gateway.proto
```diff
@@ -85,6 +85,23 @@ message GatewayOptions {
   Timeouts timeouts = 5;
 
   repeated Middleware middleware = 6;
+
+  Assets assets = 7;
+}
+
+// Assets configuration provide a passthrough cdn for frontend static assets.
```
dschaller, line 92:
```suggestion
// Assets configuration provide a passthrough host for frontend static assets.
```

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,51 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Serve!
 	if f, err := a.fileSystem.Open(r.URL.Path); err != nil {
+		// If not a known static asset and a asset provider is configured, try streaming from the configured provider.
```
dschaller, line 72:
```suggestion
		// If not a known static asset and an asset provider is configured, try streaming from the configured provider.
```

backend/gateway/mux/mux.go
dschaller: We can open another PR this week to handle the error state when either both the host and the pass through fail or only the host if the pass through isn't configured.

backend/clutch-config.yaml
```diff
@@ -15,6 +15,12 @@ gateway:
       address: 0.0.0.0
       port: 8080
       secure: false
+  assets:
+    s3:
```
dschaller, line 19: Should we have this in the default config? We aren't hosting this bucket so it might cause confusion.

mcutalo88: Ah no sorry let me delete this, this was leftover from testing.

dschaller (approved):

Issue #471, opened 2020-09-25
frontend: passthrough host support


mcutalo88:
**Description**
Currently when deploying out a new version of clutch to production, depending on your ingress configuration a user could be requesting frontend resource asynchronously from two or more clutch instances. All of which may have different frontend assets locally on disk.

This feature will give deployers of clutch a passthrough host option, when a client requests frontend resources if the version they requested is not available on disk we will fall back to the configured passthrough host.

Complexity [S/M/L]: S

PR #482, opened 2020-09-28
core / frontend: passthrough host support

mcutalo88:
### Description
Add passthrough host support for frontend static assets. If the deployed version of clutch does not have the requested frontend assets on disk the passthrough host will be used to fetch the correct assets assuming they exist for the configured provider.

In a different PR we would have a more pretty failure mode, currently you simply get a white page.

### Testing Performed

Testing was performed internally at Lyft, forcing all static asset traffic to go through the passthrough host.
```
gateway:
  assets:
    s3:
      region: us-east-1
      bucket: clutch
      key: static
```
### GitHub Issue 
Passthrough host support #471

### DIFF
--- a/api/config/gateway/v1/gateway.proto
+++ b/api/config/gateway/v1/gateway.proto
@@ -85,6 +85,23 @@ message GatewayOptions {
   Timeouts timeouts = 5;
 
   repeated Middleware middleware = 6;
+
+  Assets assets = 7;
+}
+
+// Assets configuration provide a passthrough host for frontend static assets.
+// This is useful if you dont have the ability to enable sticky sessions or a blue/green deploy system in place.
+message Assets {
+  // To use the S3Provider you must have the AWS service configured
+  message S3Provider {
+    string region = 1;
+    string bucket = 2;
+    string key = 3;
+  }
+
+  oneof provider {
+    S3Provider s3 = 1;
+  }
 }
 
 message Logger {
--- a/api/config/gateway/v1/gateway.proto
+++ b/api/config/gateway/v1/gateway.proto
@@ -96,6 +96,7 @@ message Assets {
   message S3Provider {
     string region = 1;
     string bucket = 2;
+    // key is the path to clutchs frontend static assets in the configured bucket
     string key = 3;
   }
 
--- a/backend/gateway/mux/mux.go
+++ b/backend/gateway/mux/mux.go
@@ -70,17 +70,16 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Serve!
 	if f, err := a.fileSystem.Open(r.URL.Path); err != nil {
 		// If not a known static asset and an asset provider is configured, try streaming from the configured provider.
-		if a.assetCfg != nil {
-			if a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static") {
-				// We attatch this header simply for observibility purposes.
-				// Otherwise its difficult to know if the assets are being served from the configured provider.
-				w.Header().Set("x-clutch-asset-passthrough", "true")
+		if a.assetCfg != nil && a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static/") {
+			// We attatch this header simply for observibility purposes.
+			// Otherwise its difficult to know if the assets are being served from the configured provider.
+			w.Header().Set("x-clutch-asset-passthrough", "true")
 
-				// TODO: handle errors
-				asset, _ := a.assetProviderHandler(r.URL.Path)
-				_, _ = io.Copy(w, asset)
-				return
-			}
+			asset, _ := a.assetProviderHandler(r.Context(), r.URL.Path)
+			defer asset.Close()
+
+			_, _ = io.Copy(w, asset)
+			return
 		}
 
 		// If not a known static asset serve the SPA.
@@ -92,7 +91,7 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	a.fileServer.ServeHTTP(w, r)
 }
 
-func (a *assetHandler) assetProviderHandler(urlPath string) (io.Reader, error) {
+func (a *assetHandler) assetProviderHandler(ctx context.Context, urlPath string) (io.ReadCloser, error) {
 	switch a.assetCfg.Provider.(type) {
 	case *gatewayv1.Assets_S3:
 		aws, ok := service.Registry[awsservice.Name]
@@ -106,13 +105,13 @@ func (a *assetHandler) assetProviderHandler(urlPath string) (io.Reader, error) {
 		}
 
 		return awsClient.S3StreamingGet(
-			context.Background(),
+			ctx,
 			a.assetCfg.GetS3().Region,
 			a.assetCfg.GetS3().Bucket,
 			path.Join(a.assetCfg.GetS3().Key, strings.TrimPrefix(urlPath, "/static")),
 		)
 	default:
-		return nil, fmt.Errorf("No asset provider is configured")
+		return nil, fmt.Errorf("configured asset provider has not been implemented")
 	}
 }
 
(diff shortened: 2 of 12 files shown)


backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,53 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			return nil, fmt.Errorf("The AWS service must be configured to use the asset s3 provider.")
+		}
+
+		awsClient, ok := aws.(awsservice.Client)
+		if !ok {
+			return nil, fmt.Errorf("Unable to aquire the aws client")
+		}
+
+		return awsClient.S3StreamingGet(
+			context.Background(),
```
danielhochman, line 109: thread through request context `r.Context()` in parent func

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,53 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		// If not a known static asset and an asset provider is configured, try streaming from the configured provider.
+		if a.assetCfg != nil {
+			if a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static") {
+				// We attatch this header simply for observibility purposes.
+				// Otherwise its difficult to know if the assets are being served from the configured provider.
+				w.Header().Set("x-clutch-asset-passthrough", "true")
+
+				// TODO: handle errors
+				asset, _ := a.assetProviderHandler(r.URL.Path)
+				_, _ = io.Copy(w, asset)
```
danielhochman, line 81: need to defer a Close() somewhere

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,53 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	// Serve!
 	if f, err := a.fileSystem.Open(r.URL.Path); err != nil {
+		// If not a known static asset and an asset provider is configured, try streaming from the configured provider.
+		if a.assetCfg != nil {
+			if a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static") {
+				// We attatch this header simply for observibility purposes.
+				// Otherwise its difficult to know if the assets are being served from the configured provider.
+				w.Header().Set("x-clutch-asset-passthrough", "true")
+
+				// TODO: handle errors
```
danielhochman, line 79: why is this a TODO? only a few lines of code to handle err unless i'm missing something

api/config/gateway/v1/gateway.proto
```diff
@@ -85,6 +85,23 @@ message GatewayOptions {
...
+}
+
+// Assets configuration provide a passthrough host for frontend static assets.
+// This is useful if you dont have the ability to enable sticky sessions or a blue/green deploy system in place.
+message Assets {
+  // To use the S3Provider you must have the AWS service configured
+  message S3Provider {
+    string region = 1;
+    string bucket = 2;
+    string key = 3;
```
danielhochman, line 99: what is key in this context? is it actually prefix? add comment.

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,53 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// Serve!
 	if f, err := a.fileSystem.Open(r.URL.Path); err != nil {
+		// If not a known static asset and an asset provider is configured, try streaming from the configured provider.
+		if a.assetCfg != nil {
+			if a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static") {
```
danielhochman, line 74: though collision is unlikely this is slightly more correct
```suggestion
			if a.assetCfg.Provider != nil && strings.HasPrefix(r.URL.Path, "/static/") {
```

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,53 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	case *gatewayv1.Assets_S3:
+		aws, ok := service.Registry[awsservice.Name]
+		if !ok {
+			return nil, fmt.Errorf("The AWS service must be configured to use the asset s3 provider.")
+		}
+
+		awsClient, ok := aws.(awsservice.Client)
+		if !ok {
+			return nil, fmt.Errorf("Unable to aquire the aws client")
+		}
```
danielhochman, line 106: would prefer we don't do this at runtime per-request, can we do it when the asset handler is instantiated so user knows if they have completely misconfigured?

backend/mock/service/awsmock/awsmock.go
```diff
@@ -92,6 +93,10 @@ func (s *svc) TerminateInstances(ctx context.Context, region string, ids []strin
 	return nil
 }
 
+func (s *svc) S3StreamingGet(ctx context.Context, region string, bucket string, key string) (io.ReadCloser, error) {
+	return nil, nil
```
danielhochman, line 97: either panic, return nil, err or return a fake file. nil nil is not a valid value could cause mock server to behave funny or crash unexpectedly.

backend/gateway/mux/mux.go
```diff
@@ -62,14 +69,53 @@ func (a *assetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		return awsClient.S3StreamingGet(
+			context.Background(),
+			a.assetCfg.GetS3().Region,
+			a.assetCfg.GetS3().Bucket,
+			path.Join(a.assetCfg.GetS3().Key, strings.TrimPrefix(urlPath, "/static")),
+		)
+	default:
+		return nil, fmt.Errorf("No asset provider is configured")
```
danielhochman, line 115: i think this error is actually 
```suggestion
		return nil, fmt.Errorf("configured asset provider has not been implemented")
```

backend/gateway/mux/mux.go
danielhochman, line 74: this can be one if

api/config/gateway/v1/gateway.proto
mcutalo88: This is just S3 terminology, I could rename to `path` if that makes more sense although that does not alight with S3 terms. Otherwise ill just comment.

backend/gateway/mux/mux.go
mcutalo88: Here is the original context. https://github.com/lyft/clutch/pull/468#discussion_r496060061

backend/gateway/mux/mux.go
mcutalo88: What needs a close?

backend/gateway/mux/mux.go
mcutalo88: Can we? What is the load order here? Gateway then Services?

backend/gateway/mux/mux.go
mcutalo88: How would you suggest handling these errors? Panic? throw 5xx back to the user? Or have a better 5xx error page?

backend/gateway/mux/mux.go
danielhochman: s3 streamingget returns a readcloser that is currently not closed

mcutalo88: I see my issue `assetProviderHandler` was returning a `io.Reader`.

api/config/gateway/v1/gateway.proto
danielhochman: i think of key as the entire path following bucket, whereas afaik this represents everything before the requested filename, and then the filename is appended to it. comment is enough.

backend/gateway/mux/mux.go
danielhochman: throw 5xx, it's at least debuggable in browser tools. we can fix up with nicer error page later if it makes sense.

danielhochman (approved):
lgtm, a few more minor comments

PR #487, opened 2020-09-29
authn: request refresh token from issuer

danielhochman:
### Description
1. Start requesting a refresh token from the OIDC issuer via the `offline_access` scope.
1. Issue a rich token object instead of a string from `Exchange` so the refresh token can be appended later.
1. Rename the `token` field in the callback response to prepare for `refresh_token` field to be added.
1. Fix `12345` group bug typo leftover from testing.


dschaller (approved):

backend/service/authn/authn.go
```diff
@@ -121,31 +127,43 @@ type stateClaims struct {
...
+	// Sign and issue token.
+	accessToken, err := jwt.NewWithClaims(jwt.SigningMethodHS256, claims).SignedString([]byte(p.sessionSecret))
+	if err != nil {
+		return nil, err
+	}
+	t := &oauth2.Token{
+		AccessToken:  accessToken,
+		Expiry:       time.Unix(claims.ExpiresAt, 0),
+		RefreshToken: "", // TODO: implement refresh_token flow with stateful sessions.
+		TokenType:    "Bearer",
```
dschaller, line 164: super nit: should we leave this off since it's the default?

backend/service/authn/authn.go
```diff
@@ -26,7 +26,13 @@ import (
 	"github.com/lyft/clutch/backend/service"
 )
 
-var scopes = []string{oidc.ScopeOpenID, "email"} // TODO(maybe): make scopes part of config?
+// TODO(maybe): make scopes part of config?
+// For more documentation on scopes see: https://developer.okta.com/docs/reference/api/oidc/#scopes
+var scopes = []string{
+	oidc.ScopeOpenID,
+	oidc.ScopeOfflineAccess, // offline_access is used to request issuance of a refresh_token
+	"email",
```
rithujohn191, line 34: Do we want to add in profile here as well so we request access to the end user's default profile claims like name?

rithujohn191: that someone could be me :)

PR #489, opened 2020-09-29
frontend: add storybook

### LINKED ISSUE #24: ui: add Storybook
**Description**

Add [Storybook](https://github.com/storybookjs/storybook) to Clutch to provide a design system for components and workflows.

Complexity [S/M/L]: M

dschaller:
### Description

Initial storybook implementation.

**Note**: Due to the babel changes you MUST remove your root node_modules directory before starting storybook for the first time.

In subsequent PRs we can:
- build out the component library
- add a friendlier landing page
- add netlify support to host and publish storybook

Screen Shot 2020-09-29 at 1 16 03 PM

### Testing Performed

manual

### GitHub Issue

Fixes #24


frontend/.storybook/theme.js
```diff
@@ -0,0 +1,30 @@
+import { create } from '@storybook/theming/create';
```
mcutalo88, line 1: I assume this theme is to theme storybook its self, would be cool if we could use the clutch theme one day.

frontend/packages/core/src/AppLayout/stories/logo.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
```
mcutalo88, line 1: Just for my understanding, it seems like you need to have a `stories` dir in each component and thats automatically pulled in by storybook at build time?

frontend/.storybook/theme.js
dschaller: Yeah that would be ideal. I think the best we can get is reusing the same palette sine the keys don't map.

frontend/packages/core/src/AppLayout/stories/logo.stories.tsx
dschaller: yupp, we can update this to be any pattern: https://github.com/lyft/clutch/pull/489/files#diff-c0daeb9d52f8778948a4e8a2835a953dR3

frontend/.storybook/preview.js
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+import { Theme } from "./../packages/core/src/AppProvider/themes";
+
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
```
danielhochman, line 5: comment what the intent is here

danielhochman (commented):
awesome!

mcutalo88 (commented):
Just ran this locally and it looks great!

frontend/.storybook/preview.js
dschaller: I'll actually remove this for now since we aren't really using it

danielhochman (approved):

PR #490, opened 2020-09-29
topology: leader election

mcutalo88:
### Description
Perform simple leader election utilizing postgres advisory locks.

Each clutch instances will infinitely attempt to acquire this lock, when one is successful that instance will be in charge of managing the topology cache. In follow up PRs will actually implement the logic to populate the cache for some of our current service providers (k8s & aws).

If the connection that is used to create the advisory lock is severed for any reason an unlock occurs, so in the event of an unexpected shutdown we should not deadlock.

### Testing Performed
Manually testing against a local postgres, locking / unlocking advisory locks to test application behavior. 

### GitHub Issue

This PR makes progress on the Topology API story https://github.com/lyft/clutch/issues/30


backend/service/topology/cache.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"context"
+	"time"
+
+	"go.uber.org/zap"
+)
+
+// Perfroms leader election by acquiring a postgres advisory lock.
+// Once the lock is acquired topology caching is started.
+func (c *client) acquireTopologyCacheLock() {
```
mcutalo88, line 12: Still trying to craft some test for this, might refactor a bit but overall its ready for initial review.

backend/service/topology/cache_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTryAdvisoryLock(t *testing.T) {
```
mcutalo88, line 11: Im still not entirely sure how to test advisory locks with sqlmock, but still looking at it.

backend/service/topology/cache.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Infinitely try to acquire the advisory lock
+	// Once the lock is acquired we start caching, this is a blocking operation
+	for {
+		c.log.Debug("trying to acquire advisory lock")
+
+		// Advisory locks only take a arbitrary bigint value
+		// In this case 100 is the lock for topology caching
+		if tryAdvisoryLock(conn, 100) {
+			c.log.Debug("acquired the advisory lock, starting to cache topology now...")
```
danielhochman, line 29: another thing, it would be better if each individual topology updater (e.g. AWS, K8s, etc) had its own lock to distribute the work among different instances. good TODO.

danielhochman (commented):
cool stuff, mostly forward looking comments, some of which i think are worth incorporating now

backend/service/topology/cache.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"go.uber.org/zap"
+)
+
+// Perfroms leader election by acquiring a postgres advisory lock.
+// Once the lock is acquired topology caching is started.
+func (c *client) acquireTopologyCacheLock() {
+	// We create our own connection to use for acquiring the advisory lock
+	// If the connection is severed for any reason the advisory lock will automatically unlock
+	conn, err := c.db.Conn(context.Background())
```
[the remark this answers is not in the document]
danielhochman: in this case context could hang off of client as well, but still probably better to thread so subcontext with individual deadlines, timeouts, cancellation, etc could be introduced later by the parent.

backend/service/topology/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// Infinitely try to acquire the advisory lock
+	// Once the lock is acquired we start caching, this is a blocking operation
+	for {
+		c.log.Debug("trying to acquire advisory lock")
+
+		// Advisory locks only take a arbitrary bigint value
+		// In this case 100 is the lock for topology caching
+		if tryAdvisoryLock(conn, 100) {
```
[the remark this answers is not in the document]
mcutalo88: global locking service would be pretty choice actually, will leave a TODO.

backend/service/topology/cache.go
mcutalo88: Interesting thought, that would make the locking service more compelling also. Ill leave some comments.

backend/service/topology/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+func tryAdvisoryLock(conn *sql.Conn, lockId int) bool {
+	var lock bool
+
+	// Notably we do not use `pg_advisory_lock` as the behavior of this function stack locks.
+	// For each lock invocation requires the same number of unlocks to release the advisory lock.
+	// https://www.postgresql.org/docs/12/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS-TABL
+	_ = conn.QueryRowContext(context.Background(), "SELECT pg_try_advisory_lock($1);", lockId).Scan(&lock)
```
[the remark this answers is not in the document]
mcutalo88: Oops forgot scan returns an error.

backend/service/topology/cache.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func convertLockIdToAdvisoryLockId(lockId string) (uint32, error) {
+	h := fnv.New32a()
+	_, err := h.Write([]byte(lockId))
+	if err != nil {
+		return 0, err
+	}
+	return h.Sum32(), nil
+}
```
danielhochman, line 75: can avoid err

```suggestion
func convertLockIdToAdvisoryLockId(lockID string) uint32 {
	x := sha256.New().Sum([]byte(lockID))
	return binary.BigEndian.Uint32(x)
}
```

backend/service/topology/cache.go
```diff
@@ -0,0 +1,75 @@
...
+	// Infinitely try to acquire the advisory lock
+	// Once the lock is acquired we start caching, this is a blocking operation
+	for ; true; <-ticker.C {
+		c.log.Debug("trying to acquire advisory lock")
+
+		advisoryLockId, err := convertLockIdToAdvisoryLockId(topologyCacheLockId)
+		if err != nil {
+			c.log.Error("Unable to convert the advisory lock id", zap.Error(err))
+			return
+		}
```
danielhochman, line 35: nit: can move this outside of the ticker loop

danielhochman (approved):
nice, two more nits

backend/service/topology/cache.go
mcutalo88: Ah nice sounds good

PR #495, opened 2020-10-01
frontend: preserve terminal output when app starts

dschaller:
### Description

Currently react-scripts clears the contents of the terminal window when it builds and starts. This is particularly frustrating when the background process we spawn to run babel results in an error that gets thrown away. This change prevents react-scripts from clearing any information upon start.

### Testing Performed

manual


danielhochman (approved):
wow i have wanted this for so long

PR #499, opened 2020-10-01
experimentation: add more columns to experiments table

Augustyniak:
### Description

This PR will be merged into master after https://github.com/lyft/clutch/pull/523 is in master.

Add more columns to experiments table view. Make it possible for packages that depend on the `experimentation` package - `serverexperimentation` in our example - to define how the value of a given column should be populated. For example, `serverexperimentation` package can tell `experimentation` package how to populate the value of `description` column for all experiment config of `type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig` type.

ezgif com-gif-maker (6)


backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, errors.New("service was not the correct type")
+	}
+
+	return &Service{
+		experimentStore: experimentStore,
+	}, nil
+}
+
+func (s *Service) Register(r module.Registrar) error {
+	s.experimentStore.RegisterTransformation("type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig", transform)
```
Augustyniak, line 46: register a handler that tells `experimentation` package how to transform configs of `type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig` into the list of `Properties` that can be displayed in experiments table and/or experiment details view.

backend/service/chaos/experimentation/experimentstore/experiment_config.go
```diff
@@ -0,0 +1,29 @@
...
+	experimentation "github.com/lyft/clutch/backend/api/chaos/experimentation/v1"
+)
+
+type ExperimentConfig struct {
+	id     uint64
+	Config *any.Any
+}
+
+func (ec *ExperimentConfig) CreateProperties(transformer *Transformer) ([]*experimentation.Property, error) {
+	properties, err := transformer.CreateProperties(ec)
```
Augustyniak, line 15: create properties with transformers registered by external packages - `serverexperimentation` in our case.

backend/service/chaos/experimentation/experimentstore/experiment_config.go
```diff
@@ -0,0 +1,29 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	idProperty := []*experimentation.Property{
```
Augustyniak, line 20: create standard properties that can be created by the `experimentation` package itself without any help from external packages.

backend/service/chaos/experimentation/experimentstore/transformer.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	experimentation "github.com/lyft/clutch/backend/api/chaos/experimentation/v1"
+)
+
+type Transformer struct {
+	nameToConfigTransformMap map[string]func(*ExperimentConfig) ([]*experimentation.Property, error)
+}
+
+func (c *Transformer) CreateProperties(config *ExperimentConfig) ([]*experimentation.Property, error) {
```
Augustyniak, line 11: External modules - including serverexperimentation - register their own handlers with Transformer which is then used to create properties for a given experiment configuration

backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -260,29 +241,56 @@ func (fs *experimentStore) GetListView(ctx context.Context) ([]*experimentation.
...
 
 	row := fs.db.QueryRowContext(ctx, sqlQuery, id)
 
-	var fetchedID uint64
-	var startTime, endTime, cancellationTime sql.NullTime
-	var creationTime time.Time
 	var details string
+	run := ExperimentRun{}
+	config := ExperimentConfig{Config: &any.Any{}}
+	err := row.Scan(&run.id, &run.startTime, &run.endTime, &run.cancellationTime, &run.creationTime, &config.id, &details)
```
kathan24, line 259: Is there a better way to query DB? Afaik, `scan` affects the performance.

backend/service/chaos/experimentation/experimentstore/experiment_run.go
```diff
@@ -0,0 +1,93 @@
...
+			Value: &experimentation.Property_DateValue{DateValue: startTimeTimestamp},
+		},
+	}
+
+	var time sql.NullTime
+	if er.endTime.Valid {
+		time = er.endTime
+	} else if er.cancellationTime.Valid {
+		time = er.cancellationTime
+	}
```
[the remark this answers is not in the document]
Augustyniak: both of them can be `invalid` (invalid just means that they are not set and it's a valid case to have a test that has only a start time - all server experiments have only start time when they are created)

backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+func transform(config *experimentstore.ExperimentConfig) ([]*experimentation.Property, error) {
+	var testConfig = serverexperimentation.TestConfig{}
+	err := ptypes.UnmarshalAny(config.Config, &testConfig)
+	if err != nil {
+		return []*experimentation.Property{}, nil
```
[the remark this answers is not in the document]
Augustyniak: we should, good catch

backend/service/chaos/experimentation/experimentstore/experiment_store.go
Augustyniak: frankly speaking I'm not sure - can we leave it the way it's as part of this PR (since it's been like this previously too) and address this if performance is an issue? Reading through the docs for Scan method I do not see any mentions of performance issues (https://golang.org/pkg/database/sql/#Row.Scan and https://golang.org/pkg/database/sql/#Rows.Scan)

kathan24 (approved):

backend/module/chaos/serverexperimentation/rtds/mock_experiment_store.go
```diff
@@ -38,5 +39,8 @@ func (fs *mockExperimentStore) GetExperimentRunDetails(ctx context.Context, id u
 	return nil, nil
 }
 
+func (fs *mockExperimentStore) RegisterTransformation(typeUrl string, transformation func(config *experimentstore.ExperimentConfig) ([]*experimentation.Property, error)) {
+}
```
mcutalo88, line 43: return a not implemented or something for now

backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return &Service{
+		storer: storer,
+	}, nil
+}
+
+func (s *Service) Register(r module.Registrar) error {
+	s.storer.RegisterTransformation("type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig", transform)
+	return nil
+}
```
mcutalo88, line 47: Handle error for `RegisterTransformation`

backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -0,0 +1,84 @@
...
+	err := ptypes.UnmarshalAny(config.Config, &testConfig)
+	if err != nil {
+		return []*experimentation.Property{}, err
+	}
+
+	upstream := testConfig.GetClusterPair().GetUpstreamCluster()
+	downstream := testConfig.GetClusterPair().GetDownstreamCluster()
+
+	var description string
+	switch testConfig.GetFault().(type) {
```
mcutalo88, line 60: add a default case here, log error and or return error.

backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -0,0 +1,84 @@
...
+	s.storer.RegisterTransformation("type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig", transform)
+	return nil
+}
+
+func transform(config *experimentstore.ExperimentConfig) ([]*experimentation.Property, error) {
+	var testConfig = serverexperimentation.TestConfig{}
+	err := ptypes.UnmarshalAny(config.Config, &testConfig)
+	if err != nil {
+		return []*experimentation.Property{}, err
+	}
```
mcutalo88, line 54:
```suggestion
	if err := ptypes.UnmarshalAny(config.Config, &testConfig); err != nil {
		return []*experimentation.Property{}, err
	}
```

nit: up to you.

backend/service/chaos/experimentation/experimentstore/experiment_run.go
```diff
@@ -0,0 +1,93 @@
...
+		Label: "End Time",
+		Value: endTimeTimestamp,
+	})
+
+	cancelationTimeTimestamp, err := TimeToPropertyDateValue(er.cancellationTime)
+	if err != nil {
+		return nil, err
+	}
+
+	if status == experimentation.Experiment_STOPPED {
```
mcutalo88, line 78: nit: use switch and default case to handle unknown status.

backend/service/chaos/experimentation/experimentstore/property_date.go
```diff
@@ -0,0 +1,20 @@
...
+	"database/sql"
+
+	"github.com/golang/protobuf/ptypes"
+
+	experimentation "github.com/lyft/clutch/backend/api/chaos/experimentation/v1"
+)
+
+func TimeToPropertyDateValue(t sql.NullTime) (*experimentation.Property_DateValue, error) {
+	if t.Valid {
+		timestamp, err := ptypes.TimestampProto(t.Time)
```
mcutalo88, line 13: Handle this error case

backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -219,35 +210,25 @@ func (s *storer) GetListView(ctx context.Context) ([]*experimentation.ListViewIt
...
 		if err != nil {
 			return nil, err
 		}
 
-		anyConfig := &any.Any{}
-		err = jsonpb.Unmarshal(strings.NewReader(details), anyConfig)
+		err = jsonpb.Unmarshal(strings.NewReader(details), config.Config)
 		if err != nil {
 			return nil, err
 		}
```
mcutalo88, line 224:
```suggestion
		if err = jsonpb.Unmarshal(strings.NewReader(details), config.Config); err != nil {
			return nil, err
		}
```
nit: up to you.

backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -260,29 +241,56 @@ func (s *storer) GetListView(ctx context.Context) ([]*experimentation.ListViewIt
...
 }
 
+func (s *storer) RegisterTransformation(typeUrl string, transformation func(*ExperimentConfig) ([]*experimentation.Property, error)) {
+	s.transformer.nameToConfigTransformMap[typeUrl] = transformation
+}
+
+func toProto(t *time.Time) (*timestamp.Timestamp, error) {
+	if t == nil {
+		return nil, nil
+	}
```
mcutalo88, line 284: If the caller is expecting a `timestamp` why not throw an error here?

backend/module/chaos/serverexperimentation/server_experimentation.go
Augustyniak: RegisterTransformation does not return an error

backend/module/chaos/serverexperimentation/server_experimentation.go
Augustyniak: nice!

backend/service/chaos/experimentation/experimentstore/experiment_run.go
Augustyniak: for unknown cases we just want to do nothing - not sure whether we need to have a switch's default for this. What do you think?

backend/service/chaos/experimentation/experimentstore/storer.go
Augustyniak: I just copied this method over from https://github.com/lyft/clutch/pull/499/files#diff-c4a96668461e1e041b98305db571af49L127 (a few lines above). This method handles the nil time case so that the code on the caller side can be simpler I guess. A nil time is a valid input. What would be your recommendation here (I can implement it as a part of a separate PR)?

backend/module/chaos/serverexperimentation/rtds/mock_experiment_store.go
Augustyniak: this method does not return anything

mcutalo88: ah i mis read the signature

backend/module/chaos/serverexperimentation/server_experimentation.go
mcutalo88: Yeah I misread the signature for that function. Will `Register` ever return an error then?

backend/service/chaos/experimentation/experimentstore/experiment_run.go
mcutalo88: Most of the time Ill just log in the event of an unknown type just so its observable.

Augustyniak: I agree in general but in this specific case other types are not unknown per se - we just do not want to create one extra property for them but logging an error/warning for this case would result in a non actionable noise.

backend/module/chaos/serverexperimentation/server_experimentation.go
Augustyniak: Made this function to return an `error` in case we need this in the future (it's possible). It's good to have this method to be future proof since it will be used by external packages.

backend/service/chaos/experimentation/experimentstore/storer.go
mcutalo88: The caller should check nil before invoking this function, if they dont need a proto for it. Feels weird that we would would `return nil, nil`. You can have a follow up PR to change this.

mcutalo88 (approved):

PR #500, opened 2020-10-01
topology: cache Set/Delete operations

mcutalo88:
### Description
Implements some of the basic caching functionality needed to make progress on the Topology API #30. 

I did think about making this functionality public, however for consistency across services they should implement the standardized interface to populate cache by implementing the [CacheableTopology](https://github.com/lyft/clutch/pull/362/files#diff-a352069dcb53d7f4a25245319d355ea8R39-R48) interface.

### Testing Performed
Manual with https://github.com/lyft/clutch/pull/362 test implementation. 

### GitHub Issue
This PR makes progress on the Topology API story #30

### DIFF
--- a/backend/service/topology/cache.go
+++ b/backend/service/topology/cache.go
@@ -70,7 +70,7 @@ func convertLockIdToAdvisoryLockId(lockID string) uint32 {
 	return binary.BigEndian.Uint32(x)
 }
 
-func (c *client) SetCache(obj *topologyv1.Resource) error {
+func (c *client) SetCache(ctx context.Context, obj *topologyv1.Resource) error {
 	const upsertQuery = `
 		INSERT INTO topology_cache (id, resolver_type_url, data, metadata)
 		VALUES ($1, $2, $3, $4)
@@ -95,7 +95,7 @@ func (c *client) SetCache(obj *topologyv1.Resource) error {
 	}
 
 	_, err = c.db.ExecContext(
-		context.Background(),
+		ctx,
 		upsertQuery,
 		obj.Id,
 		obj.Pb.GetTypeUrl(),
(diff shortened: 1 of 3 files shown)


backend/service/topology/cache.go
```diff
@@ -66,3 +69,61 @@ func convertLockIdToAdvisoryLockId(lockID string) uint32 {
 	x := sha256.New().Sum([]byte(lockID))
 	return binary.BigEndian.Uint32(x)
 }
+
+// TODO (mcutalo): make this private once its in use
+func (c *client) SetCache(ctx context.Context, obj *topologyv1.Resource) error {
```
mcutalo88, line 74: Left some todo's but I couldn't keep around private unused func's. In my next PR the will be changed once I implement the CacheableTopology interface.
```
> make backend-lint-fix
tools/golangci-lint.sh run --fix
service/topology/cache.go:114:18: func `(*client).deleteCache` is unused (unused)
service/topology/cache.go:73:18: func `(*client).setCache` is unused (unused)
make: *** [backend-lint-fix] Error 1
```

danielhochman, line 74: can use lint directives instead

```suggestion
// nolint:unused
// TODO (mcutalo): remove lint directive once in use
func (c *client) setCache(ctx context.Context, obj *topologyv1.Resource) error {
```

backend/service/topology/cache.go
```diff
@@ -66,3 +69,61 @@ func convertLockIdToAdvisoryLockId(lockID string) uint32 {
...
+	const upsertQuery = `
+		INSERT INTO topology_cache (id, resolver_type_url, data, metadata)
+		VALUES ($1, $2, $3, $4)
+		ON CONFLICT (id) DO UPDATE SET
+			id = EXCLUDED.id,
```
danielhochman, line 79: nit: this line is unnecessary since the id will not change. could be minor implication in perf, i really don't know much about postgres perf, probably would realize that it's not changed thus no impact.

danielhochman (approved):

PR #504, opened 2020-10-02
authn: refresh token encryption

danielhochman:
### Description
As part of preparation to store the authority's refresh token, create a path to use a passphrase for encrypting and decrypting credentials that will be stored in the database.

### Testing Performed
Unit


backend/service/authn/crypto.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+)
+
+// AES-256 encryption.
+type cryptographer struct {
+	gcm cipher.AEAD
+}
+
+func newCryptographer(passphrase string) (*cryptographer, error) {
+	if len(passphrase) == 0 {
```
dschaller, line 17: should we use something here to enforce password complexity? Maybe [zxcvbn](https://github.com/nbutton23/zxcvbn-go)?

backend/service/authn/crypto.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, fmt.Errorf("passphrase was empty")
+	}
+
+	key := sha256.Sum256([]byte(passphrase))
```
dschaller, line 21: Should we use bcrypt instead of sha256 hashing here? From my understanding it's slower but more secure

backend/service/authn/authn.go
```diff
@@ -262,12 +271,22 @@ func NewProvider(config *authnv1.Config) (Provider, error) {
...
 		verifier:            verifier,
 		oauth2:              oc,
 		httpClient:          httpClient,
 		sessionSecret:       config.SessionSecret,
 		claimsFromOIDCToken: DefaultClaimsFromOIDCToken,
-	}, nil
+	}
+
+	if config.Storage != nil {
+		c, err := newCryptographer(config.Storage.EncryptionPassphrase)
```
mcutalo88, line 284: Would it be also beneficial to also support environment variable parsing for`config.Storage.EncryptionPassphrase`?

mcutalo88 (approved):
Left some questions but looks good

backend/service/authn/authn.go
danielhochman: all configuration is already possible via environment variable, no additional support needed

backend/service/authn/crypto.go
danielhochman: we're just ensuring entropy here from the passphrase i.e. we're not storing a hash so bcrypt is not useful here.

backend/service/authn/crypto.go
danielhochman: we could but maybe there's some confusion? this is not a user password. it is an encryption passphrase configured on the server. seems a little handholdy to enforce some arbitrary strength constraints on the clutch administrator. however if we want to ensure a stronger key we could maybe try to do so with protoc-gen-validate annotations rather than introducing a separate mechanism.

api/config/service/authn/v1/authn.proto
```diff
@@ -14,11 +14,18 @@ message OIDC {
...
+
 message Config {
   // Used to sign the nonce or any other JWT secrets.
   string session_secret = 1 [ (validate.rules).string = {min_bytes : 1} ];
 
   oneof type {
     OIDC oidc = 2;
   }
+
+  Storage storage = 3;
```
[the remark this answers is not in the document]
danielhochman: :+1: added

backend/service/authn/authn.go
mcutalo88: Ah yes I dont know what I was thinking.

backend/service/authn/crypto.go
danielhochman: added comment explaining use of sha256

backend/service/authn/crypto.go
dschaller: sounds good to me. I follow the source of the passphrase but since this will be potentially user generated I was wondering if it would make sense to ensure that they don't dump in something simple such as a single digit or character

Issue #12, opened 2020-07-18
gateway: graceful shutdown


danielhochman:
**Description**

The gateway should cancel contexts and wait for them to complete before shutting down on `SIGTERM`, etc. 

We have code in our private impl that we can provide for reference.

In the future, we will also want to send some kind of cleanup signal to services to allow them to gracefully shutdown. Probably filed as a separate issue once we fix this one, as it way require changes to the `service` constructor.

Complexity [S/M/L]: S

gumpt:
I'd like to work on this next week. I think I know what private code you're referring to, and either wrote or modified it at some point. 😄

PR #517, opened 2020-10-05
topology / k8s: topology caching

mcutalo88:
### Description
This introduces topology caching for the Kubernetes service. It is the first service to implement the `CacheableTopology` interface which is also included in this PR.

If the topology service is enabled it will look through all configured serves to see who implements the `CacheableTopology` interface, If one is found then caching is automatically started.

### Testing Performed
Unit & local testing against a local kube cluster and postgres db.
```
clutch=# select id, resolver_type_url, metadata, updated_at from topology_cache;
                      id                      |              resolver_type_url               |                                                              metadata                                                              |         updated_at
----------------------------------------------+----------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+----------------------------
 envoy-main                                   | type.googleapis.com/clutch.k8s.v1.Deployment | {"app": "envoy-main"}                                                                                                              | 2020-10-08 18:31:07.140647
 envoy-main-596fb7fbfd-4sptx                  | type.googleapis.com/clutch.k8s.v1.Pod        | {"app": "envoy-main", "pod-template-hash": "596fb7fbfd"}                                                                           | 2020-10-08 18:31:07.144498
 coredns                                      | type.googleapis.com/clutch.k8s.v1.Deployment | {"k8s-app": "kube-dns"}                                                                                                            | 2020-10-08 18:31:07.128286
 envoy-main-596fb7fbfd-8jqzz                  | type.googleapis.com/clutch.k8s.v1.Pod        | {"app": "envoy-main", "pod-template-hash": "596fb7fbfd"}                                                                           | 2020-10-08 18:31:07.169238
 kube-apiserver-clutch-control-plane          | type.googleapis.com/clutch.k8s.v1.Pod        | {"tier": "control-plane", "component": "kube-apiserver"}                                                                           | 2020-10-08 18:31:07.162023
 coredns-66bff467f8-fm5r7                     | type.googleapis.com/clutch.k8s.v1.Pod        | {"k8s-app": "kube-dns", "pod-template-hash": "66bff467f8"}                                                                         | 2020-10-08 18:31:07.165616
```

### GitHub Issue

This PR makes progress on the Topology API #30 issue.

### DIFF
--- a/backend/clutch-config.yaml
+++ b/backend/clutch-config.yaml
@@ -24,16 +24,6 @@ modules:
   - name: clutch.module.k8s
   - name: clutch.module.kinesis
 services:
-  - name: clutch.service.db.postgres
-    typed_config:
-      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
-      connection:
-        host: 0.0.0.0
-        port: 5432
-        user: clutch
-        ssl_mode: DISABLE
-        dbname: clutch
-        password: clutch
   - name: clutch.service.aws
     typed_config:
       "@type": types.google.com/clutch.config.service.aws.v1.Config
@@ -48,9 +38,6 @@ services:
   - name: clutch.service.k8s
     typed_config:
       "@type": types.google.com/clutch.config.service.k8s.v1.Config
-  - name: clutch.service.topology
-    typed_config:
-      "@type": types.google.com/clutch.config.service.topology.v1.Config
 resolvers:
   - name: clutch.resolver.aws
   - name: clutch.resolver.k8s
--- a/backend/clutch-config.yaml
+++ b/backend/clutch-config.yaml
@@ -24,6 +24,16 @@ modules:
   - name: clutch.module.k8s
   - name: clutch.module.kinesis
 services:
+  - name: clutch.service.db.postgres
+    typed_config:
+      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
+      connection:
+        host: 0.0.0.0
+        port: 5432
+        user: clutch
+        ssl_mode: DISABLE
+        dbname: clutch
+        password: clutch
   - name: clutch.service.aws
     typed_config:
       "@type": types.google.com/clutch.config.service.aws.v1.Config
@@ -38,6 +48,9 @@ services:
   - name: clutch.service.k8s
     typed_config:
       "@type": types.google.com/clutch.config.service.k8s.v1.Config
+  - name: clutch.service.topology
+    typed_config:
+      "@type": types.google.com/clutch.config.service.topology.v1.Config
 resolvers:
   - name: clutch.resolver.aws
   - name: clutch.resolver.k8s
--- a/backend/service/k8s/cache.go
+++ b/backend/service/k8s/cache.go
@@ -17,8 +17,17 @@ import (
 	topologyv1 "github.com/lyft/clutch/backend/api/topology/v1"
 )
 
-// TODO (mcutalo): Make this configurable or keep this at a high value
+// TODO (mcutalo): Make this configurable or keep this at a high value.
+// The implications of resync running often is both the pressure it puts on our datastore,
+// and the additional requests that we place on all configured kubernetes cluster.
+// We have to be mindful of our clients Burst & QPS configuration,
+// which is overrideable by the user.
 const informerResyncTime = time.Hour * 1
+
+// Setting a large channel buffer mostly for first boot and the  resync timer,
+// this really should be sized according to the size of your k8s deployment.
+// However this should be a large enough buffer for the datastore to keep up with.
+const topologyObjectChanBufferSize = 5000
 
 func (s *svc) CacheEnabled() bool {
 	return true
@@ -40,7 +49,6 @@ func (s *svc) StartTopologyCaching(ctx context.Context) <-chan *topologyv1.Updat
 func (s *svc) startInformers(ctx context.Context, cs ContextClientset) {
 	factory := informers.NewSharedInformerFactoryWithOptions(cs, informerResyncTime)
 	stop := make(chan struct{})
-	defer close(stop)
 
 	podInformer := factory.Core().V1().Pods().Informer()
 	deploymentInformer := factory.Apps().V1().Deployments().Informer()
@@ -62,6 +70,7 @@ func (s *svc) startInformers(ctx context.Context, cs ContextClientset) {
 
 	<-ctx.Done()
 	s.log.Info("Shutting down the kubernetes cache informers")
+	close(stop)
 	close(s.topologyObjectChan)
 	s.topologyInformerLock.Unlock()
 }
@@ -89,9 +98,8 @@ func (s *svc) processInformerEvent(obj interface{}, action topologyv1.UpdateCach
 		}
 		s.topologyObjectChan <- &topologyv1.UpdateCacheRequest{
 			Resource: &topologyv1.Resource{
-				Id:       pod.Name,
-				Pb:       protoPod,
-				Metadata: pod.GetLabels(),
+				Id: pod.Name,
+				Pb: protoPod,
 			},
 			Action: action,
 		}
@@ -104,9 +112,8 @@ func (s *svc) processInformerEvent(obj interface{}, action topologyv1.UpdateCach
 		}
 		s.topologyObjectChan <- &topologyv1.UpdateCacheRequest{
 			Resource: &topologyv1.Resource{
-				Id:       deployment.Name,
-				Pb:       protoDeployment,
-				Metadata: deployment.GetLabels(),
+				Id: deployment.Name,
+				Pb: protoDeployment,
 			},
 			Action: action,
 		}
@@ -119,9 +126,8 @@ func (s *svc) processInformerEvent(obj interface{}, action topologyv1.UpdateCach
 		}
 		s.topologyObjectChan <- &topologyv1.UpdateCacheRequest{
 			Resource: &topologyv1.Resource{
-				Id:       hpa.Name,
-				Pb:       protoHpa,
-				Metadata: hpa.GetLabels(),
+				Id: hpa.Name,
+				Pb: protoHpa,
 			},
 			Action: action,
 		}
--- a/backend/service/k8s/cache.go
+++ b/backend/service/k8s/cache.go
@@ -1,0 +1,107 @@
+package k8s
+
+import (
+	"context"
+	"log"
+	"time"
+
+	"github.com/golang/protobuf/ptypes"
+	"go.uber.org/zap"
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+
+	topologyv1 "github.com/lyft/clutch/backend/api/topology/v1"
+)
+
+func (s *svc) CacheEnabled() bool {
+	return true
+}
+
+func (s *svc) GetTopologyObjectChannel(ctx context.Context) chan topologyv1.UpdateCacheRequest {
+	stop := make(chan struct{})
+
+	for name, cs := range s.manager.Clientsets() {
+		log.Printf("starting informer for cluster: %s", name)
+		s.startInformers(cs, stop)
+	}
+
+	return s.topologyObjectChan
+}
+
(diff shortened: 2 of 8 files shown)


backend/service/topology/cache.go
```diff
@@ -12,13 +12,16 @@ import (
...
+	"github.com/lyft/clutch/backend/service"
 )
 
 const topologyCacheLockId = "topology:cache"
 
 // Performs leader election by acquiring a postgres advisory lock.
 // Once the lock is acquired topology caching is started.
-func (c *client) acquireTopologyCacheLock(ctx context.Context) {
+func (c *client) acquireTopologyCacheLock() {
+	ctx, ctxCancelFunc := context.WithCancel(context.Background())
```
mcutalo88, line 23: I need to find a better way of handling / canceling the context here, open to any suggestions.

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,126 @@
...
+	"k8s.io/client-go/tools/cache"
+
+	topologyv1 "github.com/lyft/clutch/backend/api/topology/v1"
+)
+
+func (s *svc) CacheEnabled() bool {
+	return true
+}
+
+func (s *svc) GetTopologyObjectChannel(ctx context.Context) chan *topologyv1.UpdateCacheRequest {
```
danielhochman, line 23: what happens if this function is called twice? do we need to protect against that?

danielhochman (commented):
some initial comments, will need to dig in a bit more on context, shutdown, etc

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,126 @@
...
+		DeleteFunc: s.informerDeleteHandler,
+	}
+
+	podInformer.AddEventHandler(informerHandlers)
+	deploymentInformer.AddEventHandler(informerHandlers)
+	hpaInformer.AddEventHandler(informerHandlers)
+
+	go func() { podInformer.Run(stop) }()
+	go func() { deploymentInformer.Run(stop) }()
+	go func() { hpaInformer.Run(stop) }()
```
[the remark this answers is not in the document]
mcutalo88: yeah idk what i was thinking here let me clean this up.

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,126 @@
...
+		s.topologyObjectChan <- &topologyv1.UpdateCacheRequest{
+			Resource: &topologyv1.Resource{
+				Id:       hpa.Name,
+				Pb:       protoHpa,
+				Metadata: hpa.GetLabels(),
+			},
+			Action: action,
+		}
+	default:
+		s.log.Warn("unable to determin topology object type", zap.Any("object type", objType))
```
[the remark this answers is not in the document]
mcutalo88: Ah yes good idea thanks

backend/service/topology/topology.go
```diff
@@ -27,6 +28,17 @@ type client struct {
...
 
+// CacheableTopology is implemented by a service that wishes to enable the topology API feature set
+//
+// By implmenting this interface the topology service will automatically setup all services which implement it.
+// Automatically ingesting Resource objects via the `GetTopologyObjectChannel()` function.
+// This enables users to make use of the Topology APIs with these new Topology Resources.
+//
+type CacheableTopology interface {
+	CacheEnabled() bool
+	GetTopologyObjectChannel(ctx context.Context) chan *topologyv1.UpdateCacheRequest
```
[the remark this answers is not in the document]
mcutalo88: Yeah I agree, let me change this up.

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,126 @@
...
+	go func() { podInformer.Run(stop) }()
+	go func() { deploymentInformer.Run(stop) }()
+	go func() { hpaInformer.Run(stop) }()
+
+	select {
+	case <-ctx.Done():
+		s.log.Info("Shutting down the kubernetes cache informers")
+		close(stop)
+		close(s.topologyObjectChan)
+	default:
```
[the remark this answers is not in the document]
mcutalo88: I wanted to not block the return of this function as its invoked in `GetTopologyObjectChannel` which needs to return the `topologyObjectChan`, let me just just spawn a go routine instead.

backend/service/k8s/cache.go
mcutalo88: Yeah we need to protect against this in some way

backend/service/topology/cache.go
```diff
@@ -27,22 +36,27 @@ func (c *client) acquireTopologyCacheLock(ctx context.Context) {
...
+	// catch signals, shutdown caching by canceling the context, and release the advisory lock.
+	sigc := make(chan os.Signal, 1)
+	signal.Notify(sigc, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
+	go func() {
+		<-sigc
+		c.log.Info("Caught shutdown signal, shutting down topology caching and releasing advisory lock")
+		ticker.Stop()
+		ctxCancelFunc()
+		c.unlockAdvisoryLock(context.TODO(), conn, advisoryLockId)
+	}()
```
danielhochman, line 48: i would push the signal catching up another level, and use the incoming context for cancellation.

eventually the gateway itself will catch the signal and propagate the event, see https://github.com/lyft/clutch/issues/12 graceful shutdown

backend/service/topology/cache.go
```diff
@@ -62,17 +76,53 @@ func (c *client) tryAdvisoryLock(ctx context.Context, conn *sql.Conn, lockId uin
 	return lock
 }
 
-func (c *client) startTopologyCache() {
-	c.log.Debug("implement me")
+func (c *client) unlockAdvisoryLock(ctx context.Context, conn *sql.Conn, lockId uint32) bool {
+	var unlock bool
+	if err := conn.QueryRowContext(ctx, "SELECT pg_advisory_unlock($1)", lockId).Scan(&unlock); err != nil {
+		c.log.Error("Unable to perfrom an advisory unlock", zap.Error(err))
```
danielhochman, line 82:
```suggestion
		c.log.Error("Unable to perform an advisory unlock", zap.Error(err))
```

backend/clutch-config.yaml
```diff
@@ -24,6 +24,16 @@ modules:
...
+  - name: clutch.service.db.postgres
+    typed_config:
+      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
+      connection:
+        host: 0.0.0.0
+        port: 5432
+        user: clutch
+        ssl_mode: DISABLE
+        dbname: clutch
+        password: clutch
```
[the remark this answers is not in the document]
mcutalo88: Yeah will remove when this PR is closer to being ready.

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,131 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+
+	topologyv1 "github.com/lyft/clutch/backend/api/topology/v1"
+)
+
+// TODO (mcutalo): Make this configurable or keep this at a high value
```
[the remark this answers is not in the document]
mcutalo88: Sure ill expand my comment.

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,131 @@
...
+		protoPod, err := ptypes.MarshalAny(pod)
+		if err != nil {
+			s.log.Error("unable to marshal pod", zap.Error(err))
+			return
+		}
+		s.topologyObjectChan <- &topologyv1.UpdateCacheRequest{
+			Resource: &topologyv1.Resource{
+				Id:       pod.Name,
+				Pb:       protoPod,
+				Metadata: pod.GetLabels(),
```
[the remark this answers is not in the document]
mcutalo88: Lets chat over VC and I can echo it back on this PR.

backend/service/topology/cache.go
```diff
@@ -6,19 +6,28 @@ import (
...
+	"github.com/lyft/clutch/backend/service"
 )
 
 const topologyCacheLockId = "topology:cache"
 
 // Performs leader election by acquiring a postgres advisory lock.
 // Once the lock is acquired topology caching is started.
-func (c *client) acquireTopologyCacheLock(ctx context.Context) {
+func (c *client) acquireTopologyCacheLock() {
+	ctx, ctxCancelFunc := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
mcutalo88: Hum ok, I mean thats still going to be from background unless we change the way this looks in the future during service construction.

backend/service/topology/cache.go
mcutalo88: Doesn't that mean i need to pass the cancel func to `acquireTopologyCacheLock` in addition to the context? or is their an easier way of doing this?

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,137 @@
...
+// The implications of resync running often is both the pressure it puts on our datastore,
+// and the additional requests that we place on all configured kubernetes cluster.
+// We have to be mindful of our clients Burst & QPS configuration,
+// which is overrideable by the user.
+const informerResyncTime = time.Hour * 1
+
+// Setting a large channel buffer mostly for first boot and the  resync timer,
+// this really should be sized according to the size of your k8s deployment.
+// However this should be a large enough buffer for the datastore to keep up with.
+const topologyObjectChanBufferSize = 5000
```
danielhochman, line 30: random q, do we have a sense for what kind of memory consumption a value of 5000 will require? how big are k8s objects in the buffer?

backend/service/topology/cache.go
```diff
@@ -27,22 +31,22 @@ func (c *client) acquireTopologyCacheLock(ctx context.Context) {
 	}
 	defer conn.Close()
 
-	advisoryLockId := convertLockIdToAdvisoryLockId(topologyCacheLockId)
+	go func() {
+		<-ctx.Done()
+		ticker.Stop()
+		c.unlockAdvisoryLock(context.TODO(), conn, advisoryLockId)
```
danielhochman, line 37:
```suggestion
		c.unlockAdvisoryLock(context.Background(), conn, advisoryLockId)
```

TODO should only be used in the case that we plan to later plumb a different context or something.

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (s *svc) CacheEnabled() bool {
+	return true
+}
+
+func (s *svc) StartTopologyCaching(ctx context.Context) <-chan *topologyv1.UpdateCacheRequest {
+	// Their should only ever be one instances of all the informers for topology caching
```
danielhochman, line 37:
```suggestion
	// There should only ever be one instances of all the informers for topology caching
```

danielhochman (commented):
last round, minor

backend/service/k8s/cache.go
```diff
@@ -0,0 +1,137 @@
...
+	// We lock here until the context is closed
+	s.topologyInformerLock.Lock()
```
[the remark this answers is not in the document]
mcutalo88: Ah yeah sure we can do that

backend/service/k8s/cache.go
mcutalo88: Good question, I would have to measure the size of our protos. Just to be clear this isn't buffering the entire k8s object just Clutch's v1 k8s protos (pod, hpa, deployment), so there is substantially less data.

mcutalo88: I've taken the following to calculate rough memory consumption, this example is for a pod.
```
resource := topologyv1.Resource{
	Id: pod.Name,
	Pb: protoPod,
}
update := topologyv1.UpdateCacheRequest{
	Resource: &resource,
	Action:   action,
}
log.Printf("update %d", reflect.TypeOf(update).Size())
log.Printf("resource %d", reflect.TypeOf(resource).Size())
log.Printf("proto %d", reflect.TypeOf(*protoPod).Size())
```
output:
```
2020/10/13 13:35:45 update 56
2020/10/13 13:35:45 resource 72
2020/10/13 13:35:45 proto 80
```

From what I can tell it's hard to measure size recursively, so we looking at the size of the `Resource`, `UpdateCacheRequest` and the protoPod is the output of `ptypes.MarshalAny(pod)`.

(208 payload bytes * 5000) / 1024 / 1024 = 0.99 Mb

danielhochman (approved):
ship it! :shipit:

PR #522, opened 2020-10-06
ci: add netlify config for storybook

dschaller:
### Description

Add custom netlify config for storybook builds.

This should only build storybook when there are changes to relevant files. If I'm reading the docs correctly this command will execute from the base directory set in our configuration. **Note**: this will also change the behavior on master to only build production when changes are detected. This is intentional as we do not need to rebuild production every time.

Additionally this adds a few headers to prevent iframing and xss attacks.

### TODO

I will open a similar PR to the docs netlify configuration for optimizing that build as it currently runs even when there aren't any changes.

### DIFF
--- a/frontend/netlify.toml
+++ b/frontend/netlify.toml
@@ -1,6 +1,5 @@
 [build]
-  ignore = "if [[ '$(git rev-parse --abbrev-ref HEAD)' == 'main' ]]; then ROOT='main'; else ROOT='HEAD^'; fi && git diff --quiet '$ROOT' HEAD -- '*.
-ts*' '*.js' 'yarn.lock' '*.html'"
+  ignore = "if [[ '$(git rev-parse --abbrev-ref HEAD)' == 'main' ]]; then ROOT=main; else ROOT=HEAD^; fi && git diff --quiet $ROOT HEAD -- '*.ts*' '*.js' 'yarn.lock'"
 
 [[headers]]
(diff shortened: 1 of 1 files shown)


frontend/netlify.toml
```diff
@@ -0,0 +1,10 @@
+[build]
+  ignore = "git diff --quiet HEAD^ HEAD -- '*.ts*' '*.js*' 'yarn.lock'"
```
danielhochman, line 2: hmm this returns exit code `1` for me on a clean repo, which i think would trigger a build?

dschaller: It's running against the previous commit. Can you check if the prior commit on your checkout has ts or lockfile changes in it?

danielhochman: Ah makes sense.

For PRs is it evaluating a merge commit? Because we need to diff against main not commits within the PR itself.

danielhochman: Last thing this will match .json in addition to .js and .jsx, not sure if intentional.

dschaller: Oh good point. It would evaluate against the previous commit on the PR. I can update this to use main as the base for comparison

dschaller: json is intentional since we might make changes to package.json for storybook builds

danielhochman (approved):
:crossed_fingers:

PR #523, opened 2020-10-06
experimentation: make experiments list view backend-driven

Augustyniak:
### Description

Allow backend to have a bigger control over what (and how) data gets displayed in an experiment list view by introducing a `GetListView` call to the backend. This call returns the list of generic `ListViewItem`s that contains an unique identifier and a map of properties that can be then displayed in the table view.


backend/service/chaos/experimentation/experimentstore/experiment_store.go
```diff
@@ -196,6 +197,67 @@ func (fs *experimentStore) GetExperiments(ctx context.Context, configType string
 	return experiments, nil
 }
 
+func (fs *experimentStore) GetListView(ctx context.Context) ([]*experimentation.ListViewItem, error) {
```
kathan24, line 200: Can we add a TODO here and link to JIRA to make the pagination driven by the server too?

kathan24 (commented):
lgtm

kathan24 (approved):

backend/service/chaos/experimentation/experimentstore/experiment_store.go
Augustyniak: Good idea - I opened a GH issue instead since Jira is not publicly available https://github.com/lyft/clutch/issues/530 :)

PR #532, opened 2020-10-08
experimentation: fix handling of null date case

Augustyniak:
### Description

Correctly handle a case when a given date - end time in the example below - is equal to null. For cases like this, display `unknown` instead of defaulting to 0 epoch time.

| Before | After |
| ----- | ----- |
|Screen Shot 2020-10-07 at 7 27 20 PM|Screen Shot 2020-10-07 at 7 28 48 PM|

### Testing Performed
See screenshots above


kathan24 (approved):
lgtm

PR #540, opened 2020-10-09
experimentation: prefix enum cases

Augustyniak:
### Description

Adding prefixes to enum cases names after reading https://developers.google.com/protocol-buffers/docs/style#enums

### Testing Performed

- started a server experiment
- entered the experiments list view and confirmed that experiments are displayed correctly


Augustyniak:
@danielhochman @dschaller   mind taking a look at this one? I would be curious if you think that adding prefixes is the way to go here

api/chaos/experimentation/v1/experiment.proto
```diff
@@ -10,17 +10,17 @@ option go_package = "experimentationv1";
 message Experiment {
   enum Status {
     // Unspecifiedf Status.
-    UNSPECIFIED = 0;
+    STATUS_UNSPECIFIED = 0;
```
dschaller, line 13: The syntax is a bit redundant here since it would look like `Status.Status_Unspecified`

api/chaos/experimentation/v1/experiment.proto
```diff
@@ -10,17 +10,17 @@ option go_package = "experimentationv1";
 message Experiment {
   enum Status {
     // Unspecifiedf Status.
```
dschaller, line 12: Also, can you fix the docstring while you're here? :)

dschaller:
:/ ideally I'd prefer to avoid the duplication but it seems this is [Bufs recommendation](https://github.com/bufbuild/buf/issues/6#issuecomment-551383510). They are citing [Google's style guide](https://developers.google.com/protocol-buffers/docs/style#enums) which doesn't directly mention this verbose format but does imply it with the example provided.

api/chaos/experimentation/v1/experiment.proto
Augustyniak: yes but it makes more sense in Go/TS layer (i.e., `Experiment_STATUS_UNSPECIFIED`)

api/chaos/experimentation/v1/experiment.proto
Augustyniak: fixed

danielhochman:
Here's the full rationale behind enum prefixes: https://github.com/uber/prototool/blob/a6d064684c011c8482f4218c681a66a64031fb1d/style/README.md#enum-value-names

Basically you can't have two enums at the same level with the same value or proto will fail to compile. So for us two enums at same level would always fail since UNSPECIFIED is always 0. It has to do with generated C++ code and collisions. We would probably never target C++ (and C++11 actually has a fix for the issue) so technically it would never affect us but protoc enforces it nonetheless.

Augustyniak:
@dschaller pls have another look

dschaller (approved):

Issue #302, opened 2020-08-13
audit: record request/response data


danielhochman:
**Description**

Record request and response data for each audit call so that it can be used by other audit sinks to provide richer output.

Complexity [S/M/L]: S

scarlettperry:
closed in https://github.com/lyft/clutch/pull/590

PR #543, opened 2020-10-09
audit: record request/response data in event

scarlettperry:
### Description
Currently, the audit feature only saves these attributes to the database: https://github.com/lyft/clutch/blob/main/api/audit/v1/audit.proto#L48

This PR adds two new attributes to the audit event:  `RequestMetadata` and `ResponseMetadata`, which currently captures the entire request and response body of an API call.

#### Local Testing
Before
```
2020-10-10T21:38:07.287-0400	INFO	logger/logger.go:40	new audit event	{"serviceName": "clutch.service.audit.sink.logger", "event": "occurred_at:{seconds:1602380286 nanos:956180000} event:{username:\"UNKNOWN\" service_name:\"clutch.resolver.v1.ResolverAPI\" method_name:\"Resolve\" type:READ status:{}}"}
```

After - `request_metadata` and `response_metadata` in the audit event

```
2020-10-13T17:01:06.219-0400	INFO	logger/logger.go:40	new audit event	{"serviceName": "clutch.service.audit.sink.logger", "event": "occurred_at:{seconds:1602622862  nanos:629015000}  event:{username:\"UNKNOWN\"  service_name:\"clutch.resolver.v1.ResolverAPI\"  method_name:\"Resolve\"  type:READ  status:{}  request_metadata:{body:{[type.googleapis.com/clutch.resolver.v1.ResolveRequest]:{want:\"type.googleapis.com/clutch.k8s.v1.Pod\"  have:{[type.googleapis.com/clutch.resolver.k8s.v1.PodID]:{name:\"envoy-main-596fb7fbfd-927sc\"  clientset:\"kind-clutch\"  namespace:\"envoy-staging\"}}  limit:1}}}  response_metadata:{body:{[type.googleapis.com/clutch.resolver.v1.ResolveResponse]:{results:{[type.googleapis.com/clutch.k8s.v1.Pod]:{cluster:\"kind-clutch\"  namespace:\"envoy-staging\"  name:\"envoy-main-596fb7fbfd-927sc\"  containers:{name:\"envoy\"  image:\"docker.io/envoyproxy/envoy:v1.14-latest\"  state:RUNNING  ready:true}  node_ip:\"172.18.0.2\"  pod_ip:\"10.244.0.2\"  state:RUNNING  labels:{key:\"app\"  value:\"envoy-main\"}  labels:{key:\"pod-template-hash\"  value:\"596fb7fbfd\"}}}}}}}"}
```

Audit event corresponds to request and response for this flow below:

Screen Shot 2020-10-13 at 5 01 01 PM

Screen Shot 2020-10-13 at 5 01 08 PM

### GitHub Issue
https://github.com/lyft/clutch/issues/302


backend/gateway/meta/meta.go
```diff
@@ -171,3 +174,33 @@ func resolvePattern(message descriptor.Message, pattern *apiv1.Pattern) *auditv1
 
 	return &auditv1.Resource{TypeUrl: pattern.TypeUrl, Id: resourceName}
 }
+
+// APIMetadata returns the API request/response interface as an anypb.Any message.
+func APIMetadata(metadata interface{}) *any.Any {
```
scarlettperry, line 179: Referenced this as a guide: https://github.com/lyft/clutch/blob/main/backend/resolver/resolver.go#L50

backend/service/audit/sql.go
```diff
@@ -167,6 +170,8 @@ type eventDetails struct {
 	Status            status      `json:"status,omitempty"`
 	RequestResources  []*resource `json:"request_resources,omitempty"`
 	ResponseResources []*resource `json:"response_resources,omitempty"`
+	RequestMetadata   *any.Any    `json:"request_metadata,omitempty"`
+	ResponseMetadata  *any.Any    `json:"response_metadata,omitempty"`
```
scarlettperry, line 174: I initially thought I had to convert this back to a type `interface {}` but then realized that I could keep this as type `*any.Any` when I reviewed the backend/api/audit/v1/audit.pb.go file.
```
# backend/api/audit/v1/audit.pb.go 
type RequestEvent struct {
...
	Resources []*Resource `protobuf:"bytes,6,rep,name=resources,proto3" json:"resources,omitempty"`
	// The API request saved as metadata for the event.
	RequestMetadata *any.Any `protobuf:"bytes,7,opt,name=request_metadata,json=requestMetadata,proto3" json:"request_metadata,omitempty"`
	// The API response saved as metadata for the event.
	ResponseMetadata *any.Any `protobuf:"bytes,8,opt,name=response_metadata,json=responseMetadata,proto3" json:"response_metadata,omitempty"`
}
```

danielhochman: unfortunately because proto (and thus `Any`) objects require a special marshaller, you will have to manually marshal them from any first using jsonpb and should not use that type in the database model (which would rely directly on the built-in marshaller).

> [protojson] produces a different output than the standard "encoding/json" package, which does not operate correctly on protocol buffer messages.
source: https://godoc.org/google.golang.org/protobuf/encoding/protojson

from some cursory research, it seems json.RawMessage is commonly used for the database model type JSONB. json.RawMessage is really just a byte array with some helper functions so it can be cast to from the marshaller output, wwhich would look something like
```go
a, _ := ptypes.MarshalAny(pb)
b, _ := protojson.Marshal(a)
r := json.RawMessage(b)
```

backend/gateway/meta/meta.go
danielhochman: see comment below starting on :193

danielhochman (commented):
looks good for a first pass! you identified all the right places to make the modification and plumb the data through.

based on some of the inner workings of proto, there are a few changes we should make. also had a suggestion on the API itself for future expansion.

danielhochman (commented):
i noticed that `func APIBody` and `func apiBodyProto` return nil instead of their respective errors. if for some reason there was a problem with marshalling, it would silently fail currently with limited ways of knowing or debugging what happened. if we return the error instead, we could either log the error and continue or block on the error.

for now i think it's fine to block on the error (i.e. continue bubbling up the error) since it shouldn't happen unless something is seriously wrong.

thoughts?

scarlettperry:
> i noticed that `func APIBody` and `func apiBodyProto` return nil instead of their respective errors. if for some reason there was a problem with marshalling, it would silently fail currently with limited ways of knowing or debugging what happened. if we return the error instead, we could either log the error and continue or block on the error.
> 
> for now i think it's fine to block on the error (i.e. continue bubbling up the error) since it shouldn't happen unless something is seriously wrong.
> 
> thoughts?

@danielhochman thank you for all the feedback! In the latest changes, I return an error in `APIBody ` and `convertAPIBody` which are part of the "write to db" flow. I return a log warning in `apiBodyProto` which is part of the "read from db" flow.

I have a question on logging practices. Initially in  `apiBodyProto` I was going to use the golang log package; however, I noticed we use https://github.com/uber-go/zap. So I make use of this [client](https://github.com/lyft/clutch/blob/ef65d3fe835842d7c49b0a60140dc4e0db0c036d/backend/service/audit/setup.go#L91) type from the audit service and passed it [`RequestEventProto`](https://github.com/lyft/clutch/blob/ef65d3fe835842d7c49b0a60140dc4e0db0c036d/backend/service/audit/sql.go#L200), which has access to the client in [`query`](https://github.com/lyft/clutch/blob/ef65d3fe835842d7c49b0a60140dc4e0db0c036d/backend/service/audit/sql.go#L103), in order to make use of it in `apiBodyProto`. Is this refactor okay?

backend/service/audit/sql.go
```diff
@@ -215,3 +238,29 @@ func convertResources(proto []*auditv1.Resource) []*resource {
...
+}
+
+// Decodes JSON to proto Any message
+func (c *client) apiBodyProto(details json.RawMessage) *any.Any {
+	body := &any.Any{}
+
+	err := protojson.Unmarshal(details, body)
+	if err != nil {
+		c.logger.Warn(
+			"error unmarshall json.RawMessage object to a proto object",
```
danielhochman, line 258:
```suggestion
			"error unmarshal json.RawMessage object to a proto object",
```

danielhochman (commented):
almost, i would bubble the error up one more level and log it there:
- `func apiBodyProto(details json.RawMessage) (*any.Any, error)` (take off of `client`)
- call apiBodyProto in `client.RequestEventProto` catch err and log there
- rename `RequestEventProto` to `requestEventProto` since it shouldn't be exported

alternatively you could pass the logger into requestEventProto and pull it off the `client` as well, so `requestEventProto(logger *zap.Logger, event *event)`. either one is good with me.

danielhochman (approved):
looks great!

backend/service/audit/sql_test.go
```diff
@@ -0,0 +1,26 @@
+package audit
```
dschaller, line 1: This wasn't in the original PR?

PR #555, opened 2020-10-13
experimentation: small clean-up in server-experimentation

Augustyniak:
### Description

1. Extract part of method's implementation into a separate method
1. Slightly modify the formatting of `Target` property
1. Move error logging from a method responsible for transforming experiment configuration to the list of properties (`server_experimentation.go` file) to `Transformer` so that the logging error can be shared between all registered transformations. A transformation returns an error -> Transformer picks it up and logs an error.

### Testing Performed
N/A


backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -77,12 +69,30 @@ func (s *Service) transform(config *experimentstore.ExperimentConfig) ([]*experi
...
+
+func experimentConfigToString(experiment *serverexperimentation.TestConfig) (string, error) {
+	if experiment == nil {
+		return "", errors.New("experiment is nil")
+	}
+
+	var description string
+	switch experiment.GetFault().(type) {
+	case *serverexperimentation.TestConfig_Abort:
+		description = "Abort"
```
mcutalo88, line 90: small nit: you could just return here `return "Abort"` & below.

backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -52,10 +52,11 @@ func New(_ *any.Any, logger *zap.Logger, _ tally.Scope) (service.Service, error)
 		return nil, errors.New("experiment store wrong type")
 	}
 
-	transformer := NewTransformer()
+	sugaredLogger := logger.Sugar()
```
mcutalo88, line 55: we dont use a `sugaredlogger` anywhere else in clutch, just curious why were using it here?

mcutalo88 (approved):
Left a nit and a question but nothing major.

backend/module/chaos/serverexperimentation/server_experimentation.go
Augustyniak: fixed

backend/service/chaos/experimentation/experimentstore/storer.go
Augustyniak: I used it because server experimentation used it - I did not realize that other packages were using a non logger. I'm going to address this in a separate PR I think. Thanks for pointing this one out

PR #556, opened 2020-10-13
aws: topology caching for aws resources

mcutalo88:
### Description
Topology caching for AWS. Currently we will cache all AWS objects that clutch can operate on, this includes ASG's, EC2 Instances and Kinesis Streams.

Also cleaned up a few spelling mistakes in this pr.

### Testing Performed
Unit \ Locally.

### GitHub Issue
This PR makes progress on the Topology API #30 issue.

### DIFF
--- a/backend/clutch-config.yaml
+++ b/backend/clutch-config.yaml
@@ -24,16 +24,6 @@ modules:
   - name: clutch.module.k8s
   - name: clutch.module.kinesis
 services:
-  - name: clutch.service.db.postgres
-    typed_config:
-      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
-      connection:
-        host: 0.0.0.0
-        port: 5432
-        user: clutch
-        ssl_mode: DISABLE
-        dbname: clutch
-        password: clutch
   - name: clutch.service.aws
     typed_config:
       "@type": types.google.com/clutch.config.service.aws.v1.Config
@@ -48,9 +38,6 @@ services:
   - name: clutch.service.k8s
     typed_config:
       "@type": types.google.com/clutch.config.service.k8s.v1.Config
-  - name: clutch.service.topology
-    typed_config:
-      "@type": types.google.com/clutch.config.service.topology.v1.Config
 resolvers:
   - name: clutch.resolver.aws
   - name: clutch.resolver.k8s
--- a/backend/clutch-config.yaml
+++ b/backend/clutch-config.yaml
@@ -24,6 +24,16 @@ modules:
   - name: clutch.module.k8s
   - name: clutch.module.kinesis
 services:
+  - name: clutch.service.db.postgres
+    typed_config:
+      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
+      connection:
+        host: 0.0.0.0
+        port: 5432
+        user: clutch
+        ssl_mode: DISABLE
+        dbname: clutch
+        password: clutch
   - name: clutch.service.aws
     typed_config:
       "@type": types.google.com/clutch.config.service.aws.v1.Config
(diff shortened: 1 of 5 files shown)


backend/clutch-config.yaml
```diff
@@ -24,6 +24,16 @@ modules:
   - name: clutch.module.k8s
   - name: clutch.module.kinesis
 services:
+  - name: clutch.service.db.postgres
```
mcutalo88, line 27: will remove after the review is complete.

backend/service/aws/cache.go
```diff
@@ -0,0 +1,166 @@
...
+	ticker := time.NewTicker(time.Minute * 10)
+	for ; true; <-ticker.C {
+		wg := &sync.WaitGroup{}
+		for name, client := range c.clients {
+			c.log.Info("processing topology objects for region", zap.String("region", name))
+			go c.processAllEC2Instances(ctx, client, wg)
+			go c.processAllAutoScalingGroups(ctx, client, wg)
+			go c.processAllKinesisStreams(ctx, client, wg)
+		}
+		wg.Wait()
```
danielhochman, line 50: i wonder if this is the right approach. do we not want to refresh kinesis streams if instances are taking longer than a tick to complete? with 10 minutes, sure probably not a concern, but what if we want to lower the refresh rate?

mcutalo88: Yeah good question I think I like refactoring this into each resource has its own ticker so a user could more easily control the rate at which we poll for individual resource. I can also leave a comment to make this configurable in the future as well.

backend/service/aws/cache.go
```diff
@@ -0,0 +1,170 @@
...
+
+	go func() {
+		<-ctx.Done()
+		ticker.Stop()
+	}()
+
+	for ; true; <-ticker.C {
+		resourceFunc(ctx, client)
+	}
+}
```
danielhochman, line 63: ticker.Stop() does not close the channel so this will result in a deadlock, need to rethink.

example: https://play.golang.org/p/goI6_5gcBma

danielhochman: this works nicely: https://play.golang.org/p/NMjqyInnB1E

mcutalo88: ah I see now :( alright thanks

mcutalo88: Unable to use `time.Tick()` directly but thats alright I changed it to use `time.NewTicker` instead.

```
service/aws/cache.go:53:12: SA1015: using time.Tick leaks the underlying ticker, consider using it only in endless functions, tests and the main package, and use time.NewTicker here (staticcheck)
	ticker := time.Tick(duration)
```

backend/service/aws/cache.go
```diff
@@ -0,0 +1,172 @@
...
+func (c *client) processRegionTopologyObjects(ctx context.Context) {
+	for name, client := range c.clients {
+		c.log.Info("processing topology objects for region", zap.String("region", name))
+		go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*5), client, c.processAllEC2Instances)
+		go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*10), client, c.processAllAutoScalingGroups)
+		go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*10), client, c.processAllKinesisStreams)
+	}
+}
+
+func (c *client) startTickerForCacheResource(ctx context.Context, duration time.Duration, client *regionalClient, resourceFunc func(context.Context, *regionalClient)) {
```
danielhochman, line 52: let's write a simple test for this fn

danielhochman (approved):
lgtm, let's add a little more test coverage where possible

Issue #40, opened 2020-07-18
ci: automatic license review


dschaller:
**Description**

Add CI to review licensing of Go dependencies. This should be similar to the existing [node dependency CI](https://github.com/lyft/clutch/blob/head/.github/workflows/frontend.yml#L111)

Complexity [S/M/L]: S

PR #558, opened 2020-10-14
ci: review licensing of Go dependencies

scarlettperry:
### Description
This PR adds a CI check to review the `go.mod` file for any unapproved dependencies/licenses not yet tracked in `license_dependency_decisions.yml`. 

The license review is done by https://github.com/pivotal/LicenseFinder, which is used in other Lyft OOS projects ([omnibot](https://github.com/lyft/omnibot/blob/master/.github/workflows/pull_request.yml#L25), [eventbot](https://github.com/lyft/eventbot/blob/master/.github/workflows/pull_request.yml#L25)).

This PR also adds a readme for Clutch owners on how approve new dependencies. License Finder (a Ruby gem) has a [docker image](https://hub.docker.com/r/licensefinder/license_finder/), so individuals don't need to have Ruby set up/the gem installed locally to use the tool. Usage info with exact commands to run is included in the docs.

### Testing Performed
Locally on PR

### GitHub Issue
https://github.com/lyft/clutch/issues/40


.github/workflows/license_review.yml
```diff
@@ -0,0 +1,21 @@
+name: license_review
+on:
+  pull_request:
```
scarlettperry, line 3: thought that perhaps we'd only want to do this check `on pull_request` since deps would be approved before merging to master, but I can add `on: push` too.

scarlettperry:
left some deps unapproved so that you can see it's working 😄

scarlettperry:
do we want to have this workflow isolated like it is now (did it mainly for testing) or pull it in the `backend.yml` workflow?

.github/workflows/license_review.yml
scarlettperry: during standup, we said just `on pull request` is fine.

danielhochman:
i'd like to avoid creating a top-level `config/` dir that doesn't have anything to do with the core clutch config. we should move this into a `tools/license-finder` subdir or similar.

scarlettperry:
> i'd like to avoid creating a top-level `config/` dir that doesn't have anything to do with the core clutch config. we should move this into a `tools/license-finder` subdir or similar.

true, I'll move it

tools/license-finder/config/license_finder.yml
```diff
@@ -0,0 +1,2 @@
+decisions_file: './config/license_dependency_decisions.yml'
```
scarlettperry, line 1: In moving this around, I realized that License Finder is a little inflexible with the default configuration setup. We don't need to have this file but since the Clutch repo architecture isn't the standard for License Finder (i.e. the go.mod file isn't in the root directory and we want to customize where to save the dependency decisions), we'd have to pass a `--decisions-file` and `--project-path` flag to the cli every time.

I found a set up that works but want to explain if things seem confusing:
* License Finder expects the default configurations to be saved in the path `config/license_finder.yml`, so we still need a folder named config.
* The cli looks for `config/license_finder.yml` in your current working directory. That's why `decisions_file:` points to the current working dir and is why the `project_path:` had to be adjusted.

So this set up assumes a user is running the cli commands from the `tools/license-finder` dir instead of the Clutch root dir. However, I think it's more convenient for us to run these command straight from the Clutch root directory so I modified the docker commands to automatically cd into  `tools/license-finder`: https://github.com/lyft/clutch/pull/558/files#diff-69cc77a6488693478c695e4e0d303d1d749ad45f23b632b40cc120001448dc2cR11

If you're interested, the license finder code for a configuration file is here: https://github.com/pivotal/LicenseFinder/blob/a16a19715b530f12431ea05ed761a5a8eef0e573/lib/license_finder/configuration.rb#L7

If anything is confusing, please let me know!

tools/license-finder/config/README.md
```diff
@@ -0,0 +1,31 @@
...
+$ docker run -v $PWD:/scan -it licensefinder/license_finder /bin/bash -lc "cd /scan/tools/license-finder && license_finder"
+
+# Approve a license (any dependency with this license will be approved)
+$ docker run -v $PWD:/scan -it licensefinder/license_finder /bin/bash -lc "cd /scan/tools/license-finder && license_finder permitted_licenses add '<license_to_add>'"
+
+# Approve a dependency (approves just the specific dependency)
+$ docker run -v $PWD:/scan -it licensefinder/license_finder /bin/bash -lc "cd /scan/tools/license-finder && license_finder approvals add '<dependency_to_add>'"
+```
+
+When a license or dependecy is approved, the changes will be automatically added to [`license_dependency_decisions.yml`](./license_dependency_decisions.yml). Commit the file changes to the pull request.
```
danielhochman, line 20:
```suggestion
When a license or dependency is approved, the changes will be automatically added to [`license_dependency_decisions.yml`](./license_dependency_decisions.yml). Commit the file changes to the pull request.
```

.github/workflows/license_review.yml
```diff
@@ -0,0 +1,24 @@
+name: license_review
```
danielhochman, line 1: put comment on top pointing to the awesome README you wrote! if the job fails this might be the first place someone looks.

tools/license-finder/config/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+# Handling unknown licenses (preferred method)
+When License Finder reports that a dependency's license is 'unknown', the license should be manually researched. Then record it with:
+
+```sh
+$ docker run -v $PWD:/scan -it licensefinder/license_finder /bin/bash -lc "cd /scan/tools/license-finder && license_finder licenses add '<dependency>' '<license>'"
+```
+
+Then you can procced with the command above to approve a dependency.
```
danielhochman, line 29:
```suggestion
Then you can proceed with the command above to approve a dependency.
```

tools/license-finder/config/license_finder.yml
```diff
@@ -0,0 +1,4 @@
+# Where the liscense decisions will be saved
```
danielhochman, line 1:
```suggestion
# Where the license decisions will be saved
```

danielhochman (approved):
looks great!

.github/workflows/license_review.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  pull_request:
+    branches:
+     - main
+    paths:
+      - backend/go.mod
```
dschaller, line 7: Should we include `go.sum`?

danielhochman: license finder doesn't support it, so it wouldn't have any effect but it couldn't hurt to add it i guess?

dschaller: ah ok, I guess it wouldn't produce anything meaningful even if it did run.

dschaller (approved):

.github/workflows/license_review.yml
scarlettperry: yeah it doesn't support it but doesn't hurt to add. added the path

PR #562, opened 2020-10-15
chaos: fix missing store error message

dschaller:
### Description

The existing error message implies there is a misconfiguration of the service when in reality the store being misconfigured or missing entirely is the real issue.

### Testing Performed

Manual


mcutalo88 (approved):

danielhochman:
:dash:

PR #563, opened 2020-10-15
storybook: add button story for default and destructive variant

scarlettperry:
### Description
PR adds a story for a button, with two variant types: default and destructive

button_story_3

(Update: removed `color` prop and added `disabled`, `startIcon` in recent commits)

Screen Shot 2020-10-19 at 3 18 07 PM

### Testing Performed
Locally

And ran through one of the workflows with Button components
resize_button_gif

### DIFF
--- a/frontend/packages/core/src/button.stories.tsx
+++ b/frontend/packages/core/src/button.stories.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+import type { Meta } from "@storybook/react";
+
+import type { ButtonProps } from './button'
+
+import { Button } from './button';
+
+export default {
+  title: 'Core/Button/Button',
+  component: Button,
+
+} as Meta;
+
+const Template = (props: ButtonProps) => <Button {...props} />;
+
+export const Default = Template.bind({});
+Default.args = {
+  text: "click here",
+};
+
+export const Destructive  = Template.bind({});
+Destructive.args = {
+  ...Default.args,
+  text: "delete",
+  destructive: true,
+};
--- a/frontend/packages/core/src/button.stories.tsx
+++ b/frontend/packages/core/src/button.stories.tsx
@@ -1,14 +1,15 @@
-import React from 'react';
+import React from "react";
 import type { Meta } from "@storybook/react";
 
-import type { ButtonProps } from './button'
-
-import { Button } from './button';
+import type { ButtonProps } from "./button";
+import { Button } from "./button";
 
 export default {
-  title: 'Core/Button/Button',
+  title: "Core/Button",
   component: Button,
-
+  argTypes: {
+    onClick: { action: "onClick event" },
+  },
 } as Meta;
 
 const Template = (props: ButtonProps) => <Button {...props} />;
@@ -18,9 +19,8 @@ Default.args = {
   text: "click here",
 };
 
-export const Destructive  = Template.bind({});
+export const Destructive = Template.bind({});
 Destructive.args = {
-  ...Default.args,
   text: "delete",
   destructive: true,
 };
--- a/frontend/packages/core/src/button.tsx
+++ b/frontend/packages/core/src/button.tsx
@@ -20,7 +20,7 @@ const StyledButton = styled(MuiButton)`
 `;
 
 interface ButtonProps
-  extends Pick<MuiButtonProps, "endIcon" | "color" | "onClick" | "size" | "type" | "variant"> {
+  extends Pick<MuiButtonProps, "endIcon" | "onClick" | "size" | "type" | "variant"> {
   text: string;
   destructive?: boolean;
 }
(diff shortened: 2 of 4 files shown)


frontend/packages/core/src/button.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import type { Meta } from "@storybook/react";
+
+import type { ButtonProps } from './button'
+
+import { Button } from './button';
+
+export default {
+  title: 'Core/Button/Button',
```
danielhochman, line 9: should this be 
```suggestion
  title: 'Core/Button',
```

not familiar with significance of titles in Storybook

frontend/packages/core/src/button.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+  component: Button,
+
```
dschaller, line 11: can we add an [argType](https://storybook.js.org/docs/react/api/argtypes) here for the onClick handler

frontend/packages/core/src/button.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+const Template = (props: ButtonProps) => <Button {...props} />;
+
+export const Default = Template.bind({});
+Default.args = {
+  text: "click here",
+};
+
+export const Destructive  = Template.bind({});
+Destructive.args = {
+  ...Default.args,
```
dschaller, line 23: You should be able to remove this since we are overriding the only prop from the default args

frontend/packages/core/src/button.stories.tsx
scarlettperry: will update! initially thought we would want a button subfolder for the diff buttons (button, button group, clipboard button)

frontend/packages/core/src/button.stories.tsx
scarlettperry: added!

frontend/packages/core/src/button.stories.tsx
scarlettperry: ooh got it, I now see how this works. fixed!

scarlettperry:
updated screenshots showing the click action

Screen Shot 2020-10-16 at 11 18 49 AM

Screen Shot 2020-10-16 at 11 19 01 AM

frontend/packages/core/src/button.tsx
```diff
@@ -19,7 +19,8 @@ const StyledButton = styled(MuiButton)`
   `}
 `;
 
-interface ButtonProps extends MuiButtonProps {
+interface ButtonProps
+  extends Pick<MuiButtonProps, "endIcon" | "color" | "onClick" | "size" | "type" | "variant"> {
```
scarlettperry, line 23: A few comments/questions:

(1)
Is this the recommended way of auditing props? Saw examples of this pattern: https://sourcegraph.lyft.net/search?q=extends+Pick%3C&patternType=literal#2

(2)
I choose props that we're currently using / plan to use:
* [endIcon](https://github.com/lyft/clutch/blob/799ea1b8314c52b9305d0061122c8708a4f93b91/frontend/packages/wizard/src/wizard.tsx#L207)
* onClick all throughout
[size](https://github.com/lyft/clutch/blob/429fa9527cf4d1386cd1117d214502a7ceaa20b8/frontend/packages/core/src/AppLayout/feedback.tsx#L24)
* [type](https://github.com/lyft/clutch/blob/fb232990d0f852f2a68886ed4d6222008e730baa/frontend/workflows/sourcecontrol/src/create-repository.tsx#L82)
* [variant](https://github.com/lyft/clutch/blob/429fa9527cf4d1386cd1117d214502a7ceaa20b8/frontend/packages/core/src/AppLayout/feedback.tsx#L24)

Are there additional ones we want to pull in?[Button API](https://material-ui.com/api/button/), [ButtonBase API](https://material-ui.com/api/button-base/)

(3)
I didn’t think I’d have to specify the `onClick` since it's a common React event. But if I didn't include it, I got this TS message `Property 'onClick' does not exist on type 'IntrinsicAttributes & ButtonProps & { children?: ReactNode; }'.`

frontend/packages/core/src/button.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
```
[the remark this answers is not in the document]
scarlettperry: added a stories directory in `core/src/stories` but let me know if it would be better to have a `Button/stories` dir

scarlettperry:
Updated the PR description with new gifs.

frontend/packages/core/src/button.tsx
dschaller: (1) Yupp :)
(2) I think the only ones I would add would be disabled and startIcon (purely because we support endIcon)
(3) 👍

dschaller (approved):
lgtm after adding the props in [this comment](https://github.com/lyft/clutch/pull/563#discussion_r507918392)

PR #565, opened 2020-10-15
core: log protos with custom field

danielhochman:
### Description
Spent awhile trying to figure out how to add a hook in the logger to automatically intercept Any fields containing proto messages and fixing them automatically. I think it's possible but couldn't get it to work.

For now this PR introduces a custom field that logs correctly. I also went through and replaced any existing improper usages of zap.Any. Will have to rely on code review for now to ensure people use the correct field.

### Testing Performed
Unit


mcutalo88 (approved):
Nice

PR #569, opened 2020-10-16
storybook: add select story & add CSS fixes

dschaller:
### Description

Add new story for the Select component. This also includes various style fixes and cleanup to Select in order to prevent the label from overlaying on top of the select window when the specified width < label text length.

select

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -7,18 +7,10 @@ import {
 } from "@material-ui/core";
 import styled from "styled-components";
 
-const InputLabel = styled(MuiInputLabel)`
-  ${({ theme }) => `
-  && {
-    color: ${theme.palette.text.primary};
-  }
-  `}
-`;
-
 const FormControl = styled(MuiFormControl)`
   ${({ theme, ...props }) => `
   display: flex;
-  width: 100%;
+  min-width: fit-content;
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/select.tsx
```diff
@@ -72,18 +70,21 @@ const Select: React.FC<SelectProps> = ({
   };
 
   React.useEffect(() => {
-    onChange(options[selectedIdx].value || options[selectedIdx].label);
+    onChange(options[selectedIdx]?.value || options[selectedIdx].label);
```
scarlettperry, line 73: should it be `?.label` all throughout?

scarlettperry:
noticed that the select with label doesn't align with the fields below in the start-experiment workflow
Screen Shot 2020-10-17 at 10 38 22 PM

dschaller:
Awesome catch! Let me take a deeper look. Thanks :)

frontend/packages/core/src/Input/select.tsx
dschaller: every SelectOption should enforce that label is required. Removed the optional chaining of this below

dschaller:
@scarlettperry thanks for catching that issue. It looks like that select isn't meant to be a part of the table fields but instead hovering above. I fixed the CSS to properly set the default width :)

scarlettperry (approved):

PR #571, opened 2020-10-17
serverexperimentation: add support for external upstream services

Augustyniak:
### Description

Add an option for the users of server experimentation package to specify whether a an upstream service that they want to run a fault injection test for is an internal or external service. Why is it needed? By default, rtds part of server experimentation package applies faults on ingress traffic coming to an upstream service. For this to be possible an upstream service needs to be in our control - it has to be an "internal" service.

Now, if upstream service is not controlled by us - it's an "external" service (some 3rd party dependency) - we have no way to apply faults on ingress traffic. For cases like this we want to fallback to the ingestion of faults on egress traffic coming out of a downstream cluster.

ezgif com-gif-maker (14)

| Description | Faults Application | 
| ------ | ------ | 
| Internal Upstream | Faults on Ingress Traffic |
| External Upstream | Fault on Egress Traffic |

The changes to RTDS service will be made in consecutive PRs.

### Testing Performed
N/A (waiting for RTDS changes to be ready)


frontend/workflows/serverexperimentation/src/start-experiment.tsx
```diff
@@ -12,8 +12,58 @@ import {
...
+}
+
+interface RadioControlProps {
+  name: string;
+  label: string;
+  items: RadioControlItem[];
+  onChange: (value: string) => void;
+}
+
+const RadioControl: React.FC<RadioControlProps> = ({ name, label, items, onChange }) => {
```
Augustyniak, line 30: @dschaller do you think I could move this control to Core instead?

kathan24 (approved):
Just reminding that we might need to update existing data with these changes. 
Overall, LGTM

frontend/workflows/serverexperimentation/src/start-experiment.tsx
dschaller: Of course :D

PR #579, opened 2020-10-19
remote triage: filter out unknown data

dschaller:
### Description

Explicitly include certain server information fields in the Envoy Remote Triage workflow.

### Testing Performed

Manual


mcutalo88 (approved):
Looks good we did a code review over VC to improve my understand here a bit.

PR #589, opened 2020-10-20
housekeeping: fix react hook form peer dependency warning

dschaller:
### Description

Fixing react hook form peer dependency warning but including an additional update to hookform/devtools as it's required for the latest react-hook-form package.

### Testing Performed

Manual


scarlettperry:
nice! ty!

scarlettperry (approved):

PR #570, opened 2020-10-17
Revert "audit: record request/response data in event (#543)"

mcutalo88:
This reverts commit e41762549f755f36ac5b72af596a0f3f246083bb.

PR #590, opened 2020-10-20
Revert "Revert "audit: record request/response data in event (#543)""

scarlettperry:
Reverts lyft/clutch#570

### Background
 https://github.com/lyft/clutch/pull/543 broke the workflows when nil was passed to `ptypes.MarshalAny()`. Detailed summary below in **Reproducing the error**.

### Solution
PR:
* does not use the primary resp value in [`eventFromResponse`](https://github.com/lyft/clutch/blob/eec53c69d8c8511707fffae8ea3ff776a2c249ae/backend/middleware/audit/audit.go#L9) if `err != nil`.
* uses the v2 protobuf Go API. As part of this, `anypb.New` is used instead of `ptypes.MarshalAny`. The former gracefully handles typed nils.
* updates the "unmarshallable object" warning log to an error log
* expands unit tests
* adds a new unit test for `eventFromResponse`

### Local Testing
My clutch-config is shared in a PR comment below.

GIF shows a mix of scenarios
newchanges

#### Logger event example

invalid query
`
logger/logger.go:41     new audit event {"serviceName": "clutch.service.audit.sink.logger", "event": {"occurredAt":"2020-10-25T19:13:10.365428Z","event":{"username":"UNKNOWN","serviceName":"clutch.resolver.v1.ResolverAPI","methodName":"Search","type":"READ","status":{"code":3,"message":"did not understand input"},"requestMetadata":{"body":{"@type":"type.googleapis.com/clutch.resolver.v1.SearchRequest","want":"type.googleapis.com/clutch.k8s.v1.Pod","query":"$#@","limit":1}},"responseMetadata":{"body":{}}}}}
`

valid query
`
new audit event {"serviceName": "clutch.service.audit.sink.logger", "event": {"occurredAt":"2020-10-25T19:11:33.431763Z","event":{"username":"UNKNOWN","serviceName":"clutch.k8s.v1.K8sAPI","methodName":"ResizeHPA","type":"UPDATE","status":{},"resources":[{"typeUrl":"clutch.k8s.v1.HPA","id":"kind-clutch/envoy-staging/envoy-main"}],"requestMetadata":{"body":{"@type":"type.googleapis.com/clutch.k8s.v1.ResizeHPARequest","clientset":"kind-clutch","cluster":"kind-clutch","namespace":"envoy-staging","name":"envoy-main","sizing":{"min":7,"max":7}}},"responseMetadata":{"body":{"@type":"type.googleapis.com/clutch.k8s.v1.ResizeHPAResponse"}}}}}
`

### Reproducing the error
#543 broke the workflows when nil was passed to ptypes.MarshalAny().

This happened in cases when:
* err was not nil in [`handler(ctx, req)` ](https://github.com/lyft/clutch/blob/main/backend/middleware/audit/audit.go#L62).
In the example below, I search for an invalid HPA. The err from `handler(ctx, req)` contains `Unable to locate HPA` and resp is a typed nil. However, the Marshall error is surfaced instead b/c a typed nil was passed to `ptypes.MarshalAny()`.
Untitled

* the team also reported seeing the error with valid queries. For example, K8s Resize HPA using **/resolver/search** with a valid query. [Slack message](https://lyft.slack.com/archives/CQLRKE0ER/p1602891127397100?thread_ts=1602888293.379400&cid=CQLRKE0ER).
In the example below, I search for a valid HPA name using resolver/search. I still get an err from `handler(ctx, req)`. The err is `rpc error: code = FailedPrecondition desc = Did not find a 'type.googleapis.com/clutch.k8s.v1.HPA', looked by: name` and resp is a typed nil. Same as above, the Marshall error is surfaced instead b/c a typed nil was passed to `ptypes.MarshalAny()`.
validQ

Reproducing this second scenario has been tricky. I am not sure if I have the right local environment to find a case where valid queries fail in a non-err case / if there were other bugs that overlapped on 10/16.


backend/gateway/meta/meta.go
```diff
@@ -171,3 +173,38 @@ func resolvePattern(message descriptor.Message, pattern *apiv1.Pattern) *auditv1
 
 	return &auditv1.Resource{TypeUrl: pattern.TypeUrl, Id: resourceName}
 }
+
+// isValidInterface will return true if the type is proto.message and value is not nil
+func isValidInterface(body interface{}) bool {
```
scarlettperry, line 178: I took this approach because we know what type we want to process (proto message) and know we don't want to process a proto message with a nil value. A scenario where `body == nil` or is any other type than proto.message, would fall to the default cause.

backend/gateway/meta/meta.go
```diff
@@ -171,3 +173,38 @@ func resolvePattern(message descriptor.Message, pattern *apiv1.Pattern) *auditv1
...
+	switch v := body.(type) {
+	// type we want is proto.message
+	case proto.Message:
+		// want to use a value that is not nil
+		if !reflect.ValueOf(v).IsNil() {
```
scarlettperry, line 183: not sure what case we'd hit this in, but have been wondering if we also want to check if the proto message is empty.

if it's okay to use the v2 message type (saw an import of the package [here](https://github.com/lyft/clutch/blob/main/backend/gateway/log/log.go#L8)), we can use the ProtoReflect().[IsValid](https://godoc.org/google.golang.org/protobuf/reflect/protoreflect#Cardinality.IsValid)() method, which checks if it is an empty object (based on whether the proto message has defined required fields ) and for a nil value

backend/service/audit/storage/sql/sql.go
```diff
@@ -246,3 +279,34 @@ func convertResources(proto []*auditv1.Resource) []*resource {
 	}
 	return resources
 }
+
```
scarlettperry, line 282: for `convertAPIBody`, I added a nil check because it can receive a typed nil value from `meta. APIBody`. Similar thought process for checking for nil in `apiBodyProto`

scarlettperry:
this is what my clutch-config looks like
```
gateway:
  logger:
    pretty: true
    # level: DEBUG
  stats:
    flush_interval: 1s
    log_reporter: {}
  timeouts:
    default: 15s
  middleware:
    - name: clutch.middleware.stats
    - name: clutch.middleware.validate
    - name: clutch.middleware.audit
  listener:
    tcp:
      address: 0.0.0.0
      port: 8080
      secure: false
modules:
  - name: clutch.module.assets
  - name: clutch.module.healthcheck
  - name: clutch.module.resolver
  - name: clutch.module.aws
  - name: clutch.module.k8s
  - name: clutch.module.audit
services:
  - name: clutch.service.db.postgres
    typed_config:
      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
      connection:
        host: 0.0.0.0
        port: 5432
        user: clutch
        ssl_mode: DISABLE
        dbname: clutch
        password: clutch
  - name: clutch.service.audit.sink.logger
    typed_config:
      "@type": types.google.com/clutch.config.service.audit.v1.SinkConfig
  - name: clutch.service.audit
    typed_config:
      "@type": types.google.com/clutch.config.service.audit.v1.Config
      db_provider: clutch.service.db.postgres
      sinks:
        - clutch.service.audit.sink.logger
      filter:
        denylist: True
        rules:
            - field: METHOD
              text: Healthcheck
  - name: clutch.service.aws
    typed_config:
      "@type": types.google.com/clutch.config.service.aws.v1.Config
      regions:
        - us-east-1
        - us-west-2
  - name: clutch.service.k8s
    typed_config:
      "@type": types.google.com/clutch.config.service.k8s.v1.Config
resolvers:
  - name: clutch.resolver.aws
  - name: clutch.resolver.k8s
```

backend/service/audit/storage/sql/sql.go
scarlettperry: thinking on how this flow for processing the metadata can be made simpler so that the nil checks can be removed

backend/gateway/meta/meta.go
scarlettperry: update: RE my first comment, we run into this scenario with https://github.com/lyft/clutch/blob/main/api/k8s/v1/k8s.proto#L232 which is an empty message. this is what is shows up as in the logger event:

`new audit event	{"serviceName": "clutch.service.audit.sink.logger", "event": {"occurredAt":"2020-10-22T16:46:21.423558Z","event":{"username":"UNKNOWN","serviceName":"clutch.k8s.v1.K8sAPI","methodName":"ResizeHPA","type":"UPDATE","status":{},"resources":[{"typeUrl":"clutch.k8s.v1.HPA","id":"kind-clutch/envoy-staging/envoy-main"}],"requestMetadata":{"body":{"@type":"type.googleapis.com/clutch.k8s.v1.ResizeHPARequest","clientset":"kind-clutch","cluster":"kind-clutch","namespace":"envoy-staging","name":"envoy-main","sizing":{"min":7,"max":7}}},"responseMetadata":{"body":{"@type":"type.googleapis.com/clutch.k8s.v1.ResizeHPAResponse"}}}}}` 

empty object != nil

backend/service/audit/storage/sql/sql.go
scarlettperry: when input is type proto message with nil value, `protojson.Marshal()` returns slice [123 125] (which translates to empty obj) and err is nil.

when input is nil, `protojson.Unmarshal()` return empty object and err is nil.

backend/gateway/meta/meta.go
```diff
@@ -135,3 +137,35 @@ func resolvePattern(pb proto.Message, pattern *apiv1.Pattern) *auditv1.Resource
...
+		// want to use a value that is not nil
+		return v.ProtoReflect().IsValid()
+	default:
+		return false
+	}
+}
+
+// APIBody returns a API request/response interface as an anypb.Any message.
+func APIBody(body interface{}) (*anypb.Any, error) {
+	if !isValidInterface(body) {
```
danielhochman, line 155: this check is no longer needed afaict since the any marshaller gracefully handles (*myProtoStruct)(nil).

you would `return nil, nil` from the check below instead for same effect.

backend/middleware/audit/audit.go
```diff
@@ -86,23 +94,39 @@ func (m *mid) eventFromRequest(ctx context.Context, req interface{}, info *grpc.
...
+func (m *mid) eventFromResponse(resp interface{}, err error) (*auditv1.RequestEvent, error) {
 	s := status.Convert(err)
 	if s == nil {
 		s = status.New(codes.OK, "")
 	}
 
+	respBody, err := meta.APIBody(resp)
+	if err != nil {
+		return nil, err
+	}
```
danielhochman, line 123: IMO the biggest mistake we made the first time around was trying to record the body if err != nil. the primary return value should be considered unreliable in the case that an error occurred and discarded.

the rest of the checks are still potentially valuable to prevent a panic in the event that someone decides to `return nil, nil`, though as long as they never do it from middleware it wouldn't matter because `return nil, nil` from a handler would still be (*myProtoType)(nil) which the new any marshaller is fine with.

if err != nil let's ignore the resp altogether.

backend/middleware/audit/audit.go
```diff
@@ -62,7 +66,11 @@ func (m *mid) UnaryInterceptor() grpc.UnaryServerInterceptor {
 		resp, err := handler(ctx, req)
 
 		if id != -1 {
```
danielhochman, line 68: let's add a TODO and/or file an issue to move the response recording into a goroutine so it's async

danielhochman (commented):
a few suggested tweaks based on what we've learned about the v2 APIs and some other thinking from the original issue.

backend/gateway/meta/meta.go
scarlettperry: removed this check and instead do
```
	m, ok := body.(proto.Message)
	if !ok {
		// body is not the type/value we want to process
		return nil, nil
	}
```

backend/middleware/audit/audit.go
scarlettperry: in addition to the above and based on the scenario we tested yesterday (snippet below), the newest changes also do not call [`meta.ResourceNames`](https://github.com/lyft/clutch/blob/eec53c69d8c8511707fffae8ea3ff776a2c249ae/backend/middleware/audit/audit.go#L106) if err != nil. Currently, if resp is an untyped nil, `meta.ResourceNames` will panic with: `panic: interface conversion: interface is nil, not protoreflect.ProtoMessage`.
```
func (m *mid) UnaryInterceptor() grpc.UnaryServerInterceptor {
	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
		return nil, fmt.Errorf("problems")
	}
}
```

scarlettperry: with the latest changes (if UnaryInterceptor returned the snippet above)

scarlettperry:
GIF that captures the latest commit changes

danielhochman (commented):
two quick nits and i'll take a final pass with local testing afterwards! looks good!

scarlettperry:
> two quick nits and i'll take a final pass with local testing afterwards! looks good!

thank you for the suggestions! addressed both

danielhochman (approved):
let's try this again :ship: :tada:

PR #591, opened 2020-10-20
frontend: sanitize dictionaries in metadata table table cell

dschaller:
### Description

Sanitize Table Cell dictionaries before attempting to render their data by casting them to strings.

### Testing Performed

manually tested this correctly renders React elements and dictionaries in existing workflows


danielhochman (commented):
can we write a test for this?

PR #592, opened 2020-10-20
storybook: add hint story

dschaller:
### Description

Add `Hint` story.

hint

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Feedback/stories/hint.stories.tsx
+++ b/frontend/packages/core/src/Feedback/stories/hint.stories.tsx
@@ -8,12 +8,15 @@ export default {
   component: Hint,
 } as Meta;
 
-const Template = (props: { url: string; text: string }) => (
-  <Hint>
-    {props.url && <img src={props.url} />}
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Feedback/stories/hint.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default {
+  title: "Core/Feedback/Hint",
+  component: Hint,
+} as Meta;
+
+const Template = (props: { url: string; text: string }) => (
+  <Hint>
+    {props.url && <img src={props.url} />}
+    <div style={{ padding: "10px" }}>{props.text}</div>
```
danielhochman, line 14: shouldn't the padding be part of the component?

dschaller: The padding is only there because of the image component. In a real world application the elements passed into the Hint would have their own styling which is what this is attempting to demonstrate

scarlettperry (approved):

PR #593, opened 2020-10-20
core: add kubernetes statefulset apis

rithujohn191:
### Description
This adds an API to describe and update kubernetes StatefuSet objects. Similar to deployments objects, the supported fields that can be mutated are labels and annotations.

### Testing Performed
Unit test and local integration testing with a k8s cluster 

```
$ curl -X POST   http://localhost:3000/v1/k8s/updateStatefulSet   -H 'Content-Type: application/json'   -d '{
    "namespace": "envoy-staging",
    "name": "web",
    "fields": {
        "labels": {
             "newlabel": "labelvalue"
         }
    }
}'
```
```
kubectl describe Statefulset -n envoy-staging
Name:               web
Namespace:          envoy-staging
Selector:           app=nginx
Labels:             newlabel=labelvalue
```
The above command updates the StatefulSet by adding the label correctly


backend/service/k8s/statefulset.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		_, err := cs.AppsV1().StatefulSets(cs.Namespace()).Patch(ctx, oldStatefulSet.Name, types.StrategicMergePatchType, patchBytes, metav1.PatchOptions{})
+		return err
+	})
+	return retryErr
+}
+
+func mergeStatefulSetLabelsAndAnnotations(statefulSet *appsv1.StatefulSet, fields *k8sapiv1.UpdateStatefulSetRequest_Fields) {
```
mcutalo88, line 80: To reduce code duplication lets find a more generic way to merge labels and annotations.

backend/service/k8s/statefulset.go
```diff
@@ -0,0 +1,123 @@
...
+			if statefulSet.Spec.Template.ObjectMeta.Annotations == nil {
+				statefulSet.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
+			}
+
+			statefulSet.Spec.Template.ObjectMeta.Annotations[k] = v
+		}
+	}
+}
+
+func generateStatefulSetStrategicPatch(oldStatefulSet, newStatefulSet *appsv1.StatefulSet) ([]byte, error) {
```
mcutalo88, line 106: Same for this lets find a different way to make this more generic.

mcutalo88 (commented):
Just did an initial pass, now that we're adding more resource that can be patched lets try and make some of the code surrounding that code path more generic. As I'm sure will be adding a lot more in the future.

backend/service/k8s/statefulset.go
rithujohn191: reworked this to be a generic strategicPatch method and re-wrote the tests for it

backend/service/k8s/statefulset.go
rithujohn191: re-worked this to use [labels.Merge](https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/labels.go#L114) for a more clean way of doing it with lesser code duplication

backend/service/k8s/helper.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
+)
+
+// GenerateStrategicPatch will return a patch that yields
+// the modified object when applied to the original object,
+// or an error if either of the two objects is invalid.
+func GenerateStrategicPatch(oldBytes, newBytes []byte) ([]byte, error) {
+	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldBytes, newBytes, appsv1.Deployment{})
```
danielhochman, line 12: this is hardcoded to appsv1.Deployment, will it work for other types?

i think a better strategy for this func is to use a signature like:

```go
func GenerateStrategicPatch(original, modified runtime.Object) ([]byte, error)
```

pushing the json marshaling etc down and reflecting on the type within the function

rithujohn191: updated this method to be `GenerateStrategicPatch(original, modified runtime.Object, dataStruct interface{})` to take in the object type as the third argument. This will avoid the need for us to have a long switch statement in the method like:
```
switch original.(type) {
        case *appsv1.Deployment:
            patchBytes, err = strategicpatch.CreateTwoWayMergePatch(oldBytes, newBytes, appsv1.Deployment{})
        case *appsv1.StatefulSet:
            patchBytes, err = strategicpatch.CreateTwoWayMergePatch(oldBytes, newBytes, appsv1.StatefulSet{})
        case *appsv1.DaemonSet:
            patchBytes, err = strategicpatch.CreateTwoWayMergePatch(oldBytes, newBytes, appsv1.DaemonSet{})
....
```

rithujohn191: @danielhochman let me know if you can suggest a cleaner way to initialize the object type thats passed in without the 3rd argument in the method

rithujohn191: Note: re-test manually with the new changes

backend/service/k8s/deployment.go
```diff
@@ -63,9 +63,9 @@ func (s *svc) UpdateDeployment(ctx context.Context, clientset, cluster, namespac
 	}
 
 	newDeployment := oldDeployment.DeepCopy()
-	mergeLabelsAndAnnotations(newDeployment, fields)
+	mergeDeploymentLabelsAndAnnotations(newDeployment, fields)
 
-	patchBytes, err := generateDeploymentStrategicPatch(oldDeployment, newDeployment)
```
danielhochman, line 68: can you delete the `generateDeploymentStrategicPatch` function as well?

backend/service/k8s/statefulset.go
```diff
@@ -0,0 +1,129 @@
...
```
danielhochman, line 112: this func is also no longer used

backend/service/k8s/statefulset.go
danielhochman: if this was reworked i'm not seeing the new code

danielhochman (commented):
looks good overall, minor cleanup questions

backend/service/k8s/statefulset.go
rithujohn191: hmmm...looks like the rebase accidentally removed the changes :(

backend/service/k8s/deployment.go
rithujohn191: removed it and its tests

backend/service/k8s/statefulset.go
rithujohn191: @danielhochman rework can be viewed [here](https://github.com/lyft/clutch/pull/593/commits/ef6abbedb304dcf2e063eaae5eb1784a629d3e2a). Used `labels.Merge` to do this in a more generic way. Since we have to update the labels in 2 locations in the object thought it would be cleaner to let each resource have its own method with lesser code duplicated between the methods.

rithujohn191:
@danielhochman if you have some extra time today could you please review the PR again. All the feedback has been addressed and manually re-tested

danielhochman:
sorry for delay, will review this afternoon after qcon :slightly_smiling_face:

rithujohn191:
@danielhochman thank you. No rush

danielhochman (approved):
lgtm!

PR #601, opened 2020-10-21
audit: localstorage simplification

gumpt:
CI found the tests for this hanging before being killed. Isolated it to a
comparison for the `RequestEvent` returned after an update. The tests were
relying on `map` iteration order during `range` being the same as insertion
order. That's not the case!

This changes that by changing the underlying storage because a map indexed by an
integer has mostly the same interface as a slice. Verified locally with `go test
-count 1000000 ./service/audit/storage/local`.

Still unsure why the tests are hanging on an equality assertion, but that's an
upstream fix and making this green/reducing dev flake comes first.


danielhochman (approved):
lgtm once tests pass, thanks!

PR #602, opened 2020-10-21
gateway: fix hardcoded maximum timeout, timeout for uninterruptible handlers

### LINKED ISSUE #588: backend: gateway ignores configured timeout
**Description**
Clutch gateway does not respect the configured timeout when creating the HTTP server.

**Expected Behavior**
When the backend is initialized, the http server is configured with the timeout values from the config file.

**Actual Behavior**
The timeout for any request is always 15 seconds, regardless of the value specified in the config file.

The specific line of code where gateway ignores the config value: 
https://github.com/lyft/clutch/blob/659c4ead0de004b80b9a5ee8f5bc054bfec025f5/backend/gateway/gateway.go#L247-L248

**Version**
659c4ead0de004b80b9a5ee8f5bc054bfec025f5

danielhochman:
### Description

There was a hardcoded 15 second timeout on the http.Server. That has been modified to read the maximum timeout off of the config and add some buffer to give a chance for the gRPC side of things to time out.

On the gRPC side, timeouts previously were not respected if the handler did not respect context timeouts (e.g. blocking with time.Sleep). This is solved by spawning the handler in a goroutine so the middleware can wait for the response and/or timeout, whichever completes first.

### Testing Performed
Unit, manual.

### GitHub Issue
Fixes #588

### TODOs
- [x] tests for middleware changes


backend/gateway/config.go
```diff
@@ -183,6 +184,27 @@ type validator interface {
...
 }
 
+// Returns maximum timeout, where 0 is considered maximum (i.e. no timeout).
+func computeMaximumTimeout(cfg *gatewayv1.Timeouts) time.Duration {
+	if cfg == nil {
+		return timeouts.DefaultTimeout
+	}
+
+	ret := cfg.Default.AsDuration()
+	for _, e := range cfg.Overrides {
```
mcutalo88, line 194: How many timeout overrides can we have? In this case if we have more than one type of timeout the maximum is always selected? 

If this is only used for max global timeout why not just have an explicit config option for that then?

danielhochman: we can have as many as someone wants to put in. yes, this is just a backstop so the maximum is always selected.

if we added global max (btw this only applies to http.Server so it might be pretty confusing) then we would need to verify that no override exceeds global max, and the user would basically just be performing this algorithm manually and then setting the global max to that value.

mcutalo88: Cool that makes sense, thanks for the explanation.

backend/middleware/timeouts/timeouts.go
```diff
@@ -57,18 +61,47 @@ func (m *mid) getDuration(service, method string) time.Duration {
...
+		// Create a return channel for the goroutine.
+		resultChan := make(chan unaryHandlerReturn)
+		defer close(resultChan)
+
+		// Spawn the handler in a goroutine so we can return early on timeout if it doesn't complete.
+		go func() {
+			resp, err := handler(ctx, req)
+			select {
+			case <-ctx.Done():
+				m.logger.Warn("handler completed after timeout", zap.String("service", service), zap.String("method", method))
```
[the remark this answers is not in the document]
danielhochman: well the handler has to complete to ever get to this point. i.e. if is respecting the context it would return err on timeout.

still i think you're on to something, and maybe this is what you were getting at. if the handler respected `Done()` it would be better to return the handler's error than logging this error.

will think on it a bit more and hopefully get a test up for these cases.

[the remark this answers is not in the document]
mcutalo88: Ah yes true I suppose it must complete to get to this point. I was a bit confused around the wording of the warning message which is why I asked. 

But if err is not nil here then we can also log it and upgrade this warn to an error perhaps.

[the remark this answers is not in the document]
danielhochman: fixed, now using a done channel to track the overall execution of the middleware, which would return after the boost has elapsed, allowing the handler to return its err within the boost period.

backend/middleware/timeouts/timeouts.go
```diff
@@ -57,18 +59,64 @@ func (m *mid) getDuration(service, method string) time.Duration {
...
 			m.logger.Warn("could not parse gRPC method", zap.String("fullMethod", info.FullMethod))
 		}
 
+		// Create a return channel for the goroutine.
+		resultChan := make(chan unaryHandlerReturn)
+		defer close(resultChan)
+
+		// Create a channel to track when the timeout error has already been returned and the return channel is closed.
+		done := make(chan struct{})
+		defer close(done)
```
mcutalo88, line 80: I understand the intent, but I'm not grokking the implementation. Basically we wait for the handler to respond, even if the timeout has exceeded so we can produce a helpful log message understanding why with a hopeful error message.

Two questions:
1. How is this channel closed after the handler is complete?
2. The handler will get a context with a timeout, it might not respect it and in this case we still wait for the handler to complete. Can we leak go routines here?

danielhochman: 1. yes, if the channel is closed the handler goroutine will receive https://golang.org/ref/spec#Close
1. ultimately go does not give you any control over the goroutine. there's no way to interrupt or stop it. so yes it would leak if the function never returns https://tip.golang.org/doc/faq#no_goroutine_id. note: this leak is not new to this implementation since the same leak would be present before in the goroutine running the handler https://github.com/grpc/grpc-go/blob/master/Documentation/concurrency.md#servers

mcutalo88: Nice thanks for the explanation

mcutalo88 (approved):

PR #607, opened 2020-10-22
storybook: add ButtonGroup

scarlettperry:
### Description
PR adds the ButtonGroup story

buttongroup

### Testing Performed
Locally

### TODOs
- [x] Get the onClick event to work

### DIFF
--- a/frontend/packages/core/src/stories/buttongroup.stories.tsx
+++ b/frontend/packages/core/src/stories/buttongroup.stories.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { action } from "@storybook/addon-actions";
 import type { Meta } from "@storybook/react";
 
 import type { ButtonGroupProps } from "../button";
@@ -7,9 +8,6 @@ import { ButtonGroup } from "../button";
 export default {
   title: "Core/Button Group",
   component: ButtonGroup,
-  argTypes: {
-    onClick: { action: "onClick event" },
-  },
 } as Meta;
 
 const Template = (props: ButtonGroupProps) => <ButtonGroup {...props} />;
@@ -19,9 +17,11 @@ Default.args = {
   buttons: [
     {
       text: "Back",
+      onClick: action("onClick event"),
     },
     {
       text: "Next",
+      onClick: action("onClick event"),
     },
   ],
 };
@@ -31,10 +31,12 @@ Destructive.args = {
   buttons: [
     {
       text: "Back",
+      onClick: action("onClick event"),
     },
     {
       text: "Delete",
       destructive: true,
+      onClick: action("onClick event"),
     },
   ],
 };
--- a/frontend/packages/core/src/stories/buttongroup.stories.tsx
+++ b/frontend/packages/core/src/stories/buttongroup.stories.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import type { Meta } from "@storybook/react";
 
 import type { ButtonGroupProps } from "../button";
-import { ButtonGroup, useWizardContext } from "@clutch-sh/core";
+import { ButtonGroup } from "../button";
 
 export default {
   title: "Core/Button Group",
@@ -23,7 +23,7 @@ Default.args = {
     {
       text: "Next",
     },
-  ]
+  ],
 };
 
 export const Destructive = Template.bind({});
@@ -36,5 +36,5 @@ Destructive.args = {
       text: "Delete",
       destructive: true,
     },
-  ]
+  ],
 };
--- a/frontend/packages/core/src/stories/buttongroup.stories.tsx
+++ b/frontend/packages/core/src/stories/buttongroup.stories.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { action } from "@storybook/addon-actions";
+import { action } from "@storybook/addon-actions"; // eslint-disable-line import/no-extraneous-dependencies
 import type { Meta } from "@storybook/react";
 
 import type { ButtonGroupProps } from "../button";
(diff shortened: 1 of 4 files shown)


frontend/packages/core/src/stories/buttongroup.stories.tsx
```diff
@@ -7,9 +8,6 @@ import { ButtonGroup } from "../button";
 export default {
   title: "Core/Button Group",
   component: ButtonGroup,
-  argTypes: {
-    onClick: { action: "onClick event" },
-  },
```
scarlettperry: each button in the [array](https://github.com/lyft/clutch/blob/429fa9527cf4d1386cd1117d214502a7ceaa20b8/frontend/workflows/ec2/src/resize-asg.tsx#L70) has an onClick event, and with this code block, events weren't being emitted when a button was clicked.

did some digging and learned we can manually add an action to a component, which I think is a good approach for this case since we're [creating individual button components](https://github.com/lyft/clutch/blob/be8d04d692bc5084a687fa24cfc8dbf2aecc54ed/frontend/packages/core/src/button.tsx#L49) from the array passed in.

scarlettperry:
Updated gif to show the click events

buttongroupclicks

frontend/packages/core/src/stories/buttongroup.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  title: "Core/Button Group",
+  component: ButtonGroup,
+} as Meta;
+
+const Template = (props: ButtonGroupProps) => <ButtonGroup {...props} />;
+
+export const Default = Template.bind({});
+Default.args = {
+  buttons: [
+    {
```
dschaller, line 18: you should be able to extract out the common entries into an array and share that across both stories

frontend/packages/core/src/stories/buttongroup.stories.tsx
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import { action } from "@storybook/addon-actions";
+import type { Meta } from "@storybook/react";
+
+import type { ButtonGroupProps } from "../button";
+import { ButtonGroup } from "../button";
+
+export default {
+  title: "Core/Button Group",
```
dschaller, line 9: For the Note components I nested the panel under [Notes](https://deploy-preview-607--clutch-components.netlify.app/?path=/story/core-feedback-note-panel--primary) but I think I like this more. What do you think?

scarlettperry: Initially thought to organize it by:
```
Buttons
-- Button
-- Button Group
```
I looked at some Storybook examples, and I think it's a little easier to find/see the variety of components when the stories aren't too nested. For example, I like this [All Components](https://5d559397bae39100201eedc1-gyadsqdnlc.chromatic.com/?path=/story/playground-playground--playground) heading and then the list of stories.

frontend/packages/core/src/stories/buttongroup.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+  component: ButtonGroup,
+} as Meta;
+
+const Template = (props: ButtonGroupProps) => <ButtonGroup {...props} />;
+
+const sharedArgs = [{ text: "Back", onClick: action("onClick event") }, action("onClick event")];
```
scarlettperry, line 15: RE https://github.com/lyft/clutch/pull/607#discussion_r510354273, is this what you had in mind? I researched if there was a Storybook way to do this - tried setting the common entries at the component level (https://storybook.js.org/docs/react/writing-stories/args#component-args) but couldn't figure out how to merge both the component level args and the story-specific args.

frontend/packages/core/src/stories/buttongroup.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { action } from "@storybook/addon-actions"; // eslint-disable-line import/no-extraneous-dependencies
```
scarlettperry, line 2: Got the tslint error `error  '@storybook/addon-actions' should be listed in the project's dependencies. Run 'npm i -S @storybook/addon-actions' to add it  import/no-extraneous-dependencies`. Was unsure why I was getting the lint error for this storybook import and not the others that are in [dev dependencies ](https://github.com/lyft/clutch/blob/4d0b7faab3f2b5cc4c5b6bb9de84eb52b174d393/frontend/package.json#L65).

Came across https://github.com/storybookjs/storybook/issues/1992 and the recommendation was just to silence the ts error. Thoughts?

frontend/packages/core/src/stories/buttongroup.stories.tsx
dschaller: This works. I was thinking about just reusing the back button element and adding in the unique story buttons but having the action is ok as well.

frontend/packages/core/src/stories/buttongroup.stories.tsx
```diff
@@ -0,0 +1,38 @@
+import React from "react";
```
dschaller, line 1: super nit: can we add a hyphen to the name just for readability? e.g `button-group`

frontend/packages/core/src/stories/buttongroup.stories.tsx
dschaller: Let's disable this rule in the eslint config for core to prevent others from having to comment if needed later as well.

something like:
```
    {
      files: ["**/*.stories.*"],
      rules: {
        "import/no-extraneous-dependencies": ["off"],
      }
    },

```

dschaller (approved):

PR #609, opened 2020-10-22
core: add kubernetes delete deployment api

rithujohn191:
### Description
This PR adds an API to delete a Kubernetes deployment object.

### Testing Performed
- Unit test added for the delete deployment method
- Manual testing on Kubernetes cluster to delete an existing deployment

```
clutch-private rjohn$ curl -X POST   http://localhost:3000/v1/k8s/deleteDeployment   -H 'Content-Type: application/json'   -d '{
    "namespace": "envoy-staging",
    "name": "envoy" 
}'             
{}rjohn-mbp151:clutch-private rjohn$ kubectl get deployment -n envoy-staging
No resources found.

```


backend/service/k8s/deployment.go
```diff
@@ -77,6 +77,16 @@ func (s *svc) UpdateDeployment(ctx context.Context, clientset, cluster, namespac
 	return retryErr
 }
 
+func (s *svc) DeleteDeployment(ctx context.Context, clientset, cluster, namespace, name string) error {
+	cs, err := s.manager.GetK8sClientset(clientset, cluster, namespace)
+	if err != nil {
+		return err
+	}
+
+	opts := metav1.DeleteOptions{}
```
rithujohn191, line 86: Open question: do we want to allow users to pass in a flag so they can customize the GracePeriodSeconds to 0 for an immediate delete? otherwise this value defaults to 30s (https://godoc.org/k8s.io/api/core/v1). We don't want to default to 0 because the user could have set `[terminationGracePeriodSeconds`](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2970) at the pod spec level

scarlettperry:
Looks good! I'll let the team give a review as well.

backend/service/k8s/deployment.go
danielhochman: I think that's a great feature to add to the API. Can add it in follow-up PR or this one, up to you.

danielhochman (approved):
lgtm!

backend/service/k8s/deployment.go
rithujohn191: Ok great will get in a follow up PR to address it and update the same for the pod deletion API. It currently defaults to setting it to 0.

PR #610, opened 2020-10-22
experimentation: add additional capabilities to transformer

Augustyniak:
### Description

There are two main changes to `Transformer` here:
- previously the `Transformer` would allow its user to register a transform that took an `ExperimentConfig` as its argument. That worked well for extracting properties such as fault type from a given experiment but it will not work for creating dashboard links. That's because dashboard links require a start and end times and these are a part of `ExperimentRun` and not the `ExperimentConfig`. Add an option for users of the `Transformer` to accept a `Transformation` that takes a run and config as its arguments.
- add an option for the `Transformer` to support multiple `Transformation`s for each message type. Change `map[string]*Transformation{}` to `map[string][]*Transformation{}`

### Testing Performed
Unit Tests


backend/module/chaos/serverexperimentation/server_experimentation.go
```diff
@@ -42,11 +42,11 @@ func New(_ *any.Any, logger *zap.Logger, scope tally.Scope) (module.Module, erro
 }
 
 func (s *Service) Register(r module.Registrar) error {
-	transformation := experimentstore.Transformation{ConfigTypeUrl: "type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig", ConfigTransform: s.transform}
+	transformation := experimentstore.Transformation{ConfigTypeUrl: "type.googleapis.com/clutch.chaos.serverexperimentation.v1.TestConfig", RunTransform: s.transform}
 	return s.storer.RegisterTransformation(transformation)
 }
 
-func (s *Service) transform(config *experimentstore.ExperimentConfig) ([]*experimentation.Property, error) {
+func (s *Service) transform(run *experimentstore.ExperimentRun, config *experimentstore.ExperimentConfig) ([]*experimentation.Property, error) {
```
kathan24, line 49: Whats the use of `run` here?

backend/service/chaos/experimentation/experimentstore/experiment_config.go
```diff
@@ -11,19 +11,12 @@ type ExperimentConfig struct {
...
+	return []*experimentation.Property{
 		{
 			Id:    "config_identifier",
 			Label: "Config Identifier",
 			Value: &experimentation.Property_IntValue{IntValue: int64(ec.id)},
 		},
-	}
-
-	return append(idProperty, properties...), nil
+	}, nil
```
kathan24, line 21: nice

kathan24 (commented):
can we clean up the commit history? there are some commits from external fault injection which seems not related to this PR

backend/module/chaos/serverexperimentation/server_experimentation.go
Augustyniak: it's not used in this specific place, it will be used from within Clutch-Private since we will need to access run's start and end times to create a link to our dashboards. I will try to replace `run` with `_` here.

Augustyniak:
>can we clean up the commit history? there are some commits from external fault injection which seems not related to this PR

I can do this if you feel strongly about it but it's going to be lost anyway since we are `squashing and merging` - external fault injection things are here due to multiple merges of master into this PR.

kathan24 (approved):

PR #624, opened 2020-10-26
storybook: add warning story

dschaller:
### Description

- Add `Warning` story.
- Extracted the `Warning` component into its own file within the feedback folder.
- Fixed a bug where `onClose` was being called twice when warnings were dismissed.

warning

### Testing Performed

manual with story

### DIFF
--- a/frontend/packages/core/src/Feedback/stories/warning.stories.tsx
+++ b/frontend/packages/core/src/Feedback/stories/warning.stories.tsx
@@ -8,7 +8,7 @@ import type { WarningProps } from "../warning";
 import Warning from "../warning";
 
 export default {
-  title: "Core/Warning",
+  title: "Core/Feedback/Warning",
   component: Warning,
 } as Meta;
 
--- a/frontend/packages/core/src/Feedback/stories/warning.stories.tsx
+++ b/frontend/packages/core/src/Feedback/stories/warning.stories.tsx
@@ -1,0 +1,26 @@
+import React from "react";
+// TODO: remove when https://github.com/lyft/clutch/pull/607/files#r512255592 lands
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { action } from "@storybook/addon-actions";
+import type { Meta } from "@storybook/react";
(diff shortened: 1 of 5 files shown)


frontend/packages/core/src/Feedback/stories/warning.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+// TODO: remove when https://github.com/lyft/clutch/pull/607/files#r512255592 lands
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { action } from "@storybook/addon-actions";
+import type { Meta } from "@storybook/react";
+
+import type { WarningProps } from "../warning";
+import Warning from "../warning";
+
+export default {
+  title: "Core/Warning",
```
scarlettperry, line 11: should we move it to  "Core/Feedback/Warning" since it's now part of the feedback folder?

frontend/packages/core/src/Feedback/stories/warning.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const Primary = Template.bind({});
+Primary.args = {
+  message: "Informational warning",
+};
+
+export const CloseAction = Template.bind({});
+CloseAction.args = {
+  ...Primary.args,
+  onClose: action("on-close"),
```
scarlettperry, line 25: I wonder if both stories should have an onClose? just b/c both warnings have an **X**

dschaller: Hmmm this might be a good place to chat about the name. The onClose prop is meant to allow for a side-effect during close but it's not a required prop since we may not want to trigger additional logic when the modal disappears. Let me know if you have ideas on how to clarify this :)

scarlettperry: Thanks for explaining! That makes sense. I think the current name works.

scarlettperry (approved):

PR #625, opened 2020-10-26
k8s: utilize lightweightinformer for topology caching

mcutalo88:
### Description
The motivation of this change is too reduce the memory footprint of the Kubernetes topology caching logic. A major contributor to the memory consumption was the informers cache store which we do not utilize for the Kubernetes topology caching functionality. 

This PR introduces a new type of an informer which sets to reduce the informer cache size by substituting the incoming kubernetes objects with smaller lightweight ones.

During internal testing with large scale Kubernetes deployments we easily saw 15Gb+ of memory usage, when switching to utilize this new `LightweightInfomer` we saw memory usage drop about ~65%.

I pulled out my description from the comments in the code for the reader convenience.

----

The LightweightInformer is an informer thats optimized for memory usage with drawbacks.

The reduction in memory consumption does come at a cost, to achieve this we store small objects
in the informers cache store. We do this by utilizing storing `lightweightCacheObject` instead
of the full Kubernetes object.
`lightweightCacheObject` has just enough metadata for the cache store and DeltaFIFO components to operate normally.

There are drawbacks too using a LightweightInformer and its does not fit all use cases.
For the Topology Caching this type of solution helped to reduce memory footprint significantly
for large scale Kubernetes deployments.

Also to note the memory footprint of the cache store is only part of the story.
While the informers controller is receiving Kubernetes objects it stores that full object in the DeltaFIFO queue.
This queue while processed quickly does store a vast amount objects at any given time and contributes to memory usage greatly.

Drawbacks
- Update resource event handler does not function as expected, old objects will always return nil.
  This is because we dont cache the full k8s object to compute deltas as we are using lightweightCacheObjects instead.

### Testing Performed

Unit test / locally / real deployment in Lyft staging environment.

### DIFF
--- a/backend/service/k8s/lightweightinformer.go
+++ b/backend/service/k8s/lightweightinformer.go
@@ -57,7 +57,6 @@ func NewLightweightInformer(
 		RetryOnError:     false,
 		Process: func(obj interface{}) error {
 			for _, d := range obj.(cache.Deltas) {
-
 				incomeingObjectMeta, err := meta.Accessor(d.Object)
 				if err != nil {
 					return err
--- a/backend/service/k8s/lightweightinformer.go
+++ b/backend/service/k8s/lightweightinformer.go
@@ -11,7 +11,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-type lightweightObj struct {
+type lightWeightCacheObject struct {
 	metav1.Object
 	UID        types.UID
 	Name       string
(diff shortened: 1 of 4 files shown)


backend/service/k8s/lightweightinformer_test.go
```diff
@@ -0,0 +1,68 @@
...
+		&v1.Pod{},
+		time.Minute,
+		informerHandlers,
+	)
+
+	go podInformer.Run(stop)
+
+	// We sleep below to give some time for the informer to get the event and hit a handler
+	fc.Add(&v1.Pod{})
+	time.Sleep(time.Millisecond * 50)
```
mcutalo88, line 58: Looking at this again, I probably dont need the sleeps at all since I wait for the channel to return something anyway

backend/service/k8s/lightweightinformer.go
```diff
@@ -0,0 +1,93 @@
...
+							return err
+						}
+						h.OnAdd(d.Object)
+					}
+				case cache.Deleted:
+					if err := cacheStore.Delete(lightweightObj); err != nil {
+						return err
+					}
+					h.OnDelete(d.Object)
+				}
```
danielhochman, line 88: nit: add default case with err in case they expand the number of events

backend/service/k8s/lightweightinformer.go
```diff
@@ -0,0 +1,93 @@
...
+//
+// Also to note the memory footprint of the cache store is only part of the story.
+// While the informers controller is receiving Kubernetes objects it stores that full object in the DeltaFIFO queue.
+// This queue while processed quickly does store a vast amount of objects at any given time and contributes to memory usage greatly.
+//
+// Drawbacks
+// - Update resource event handler does not function as expected, old objects will always return nil.
+//   This is because we dont cache the full k8s object to compute deltas as we are using lightweightCacheObjects instead.
+func NewLightweightInformer(
+	cs ContextClientset,
```
danielhochman, line 40: var is unused?

danielhochman (approved):
minor changes

backend/service/k8s/lightweightinformer.go
mcutalo88: yes let me clean this up

PR #628, opened 2020-10-27
rtds: scope stats to xds version

snowp:
This scopes the RTDS stats to the xDS version used in order to make it
easier to monitor the individual resources. Also adds a stat for errors
received from xDS clients.

An integration test is added to verify that we're emitting the
appropriate stats for V2 and V3. This test also serves to verify that
we're properly exposing the two protocols.

### DIFF
--- a/backend/module/chaos/serverexperimentation/rtds/rtds_test.go
+++ b/backend/module/chaos/serverexperimentation/rtds/rtds_test.go
@@ -23,31 +23,6 @@ import (
 	"github.com/lyft/clutch/backend/service/chaos/experimentation/experimentstore"
 )
 
-func awaitCounterEquals(t *testing.T, scope tally.TestScope, counter string, value int64) {
-	t.Helper()
-
-	timeout := time.NewTimer(time.Second)
-	for {
-		v := int64(0)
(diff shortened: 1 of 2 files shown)


backend/module/chaos/serverexperimentation/rtds/rtds_test.go
```diff
@@ -0,0 +1,144 @@
...
+	rpc_status "google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc"
+
+	rtdsconfigv1 "github.com/lyft/clutch/backend/api/config/module/chaos/experimentation/rtds/v1"
+	"github.com/lyft/clutch/backend/module/moduletest"
+	"github.com/lyft/clutch/backend/service"
+	"github.com/lyft/clutch/backend/service/chaos/experimentation/experimentstore"
+)
+
+func awaitCounterEquals(t *testing.T, scope tally.TestScope, counter string, value int64) {
```
Augustyniak, line 26: nit: can we move this function to the end of the file? When I open the file I want to see the test case, the test function is of a lower significance I think

Augustyniak (approved):

PR #632, opened 2020-10-27
storybook: add compressed error story

dschaller:
### Description

Add story for CompressedError component. This also includes a component rename from Material and styling fixes.

compressedError

### Testing Performed

manual with story


frontend/packages/core/src/error.tsx
```diff
@@ -47,13 +50,36 @@ const ErrorText = styled(Typography)`
...
-  width: 100%;
 `;
 
-const CompressedError = ({ title, message }) => {
+export interface CompressedErrorProps {
+  title?: string;
+  message: string;
+}
+
+const findBreakpoint = (message: string): number => {
```
mcutalo88, line 63: do we want a unit test for this? or is the storybook enough here?

dschaller: I'll add these later :)

mcutalo88 (approved):

PR #633, opened 2020-10-28
topology: disable topology cache config option

mcutalo88:
### Description
If the user does not specify the `cache` configuration it will not be enabled. This provides more flexibility around clutch deployment configurations and help to separate concerns, as topology caching is a resource intensive operation it is sometimes suitable to deploy and run this functionality in its own instance.

Also fixed a log statement that wasn't using a logger.

### Testing Performed

locally with this configuration below
```
  - name: clutch.service.topology
    typed_config:
      "@type": types.google.com/clutch.config.service.topology.v1.Config
```
```
2020-10-29T09:46:59.891-0700	INFO	gateway/gateway.go:105	registering service	{"serviceName": "clutch.service.topology"}
2020-10-29T09:46:59.891-0700	INFO	topology/topology.go:70	Topology caching is disabled	{"serviceName": "clutch.service.topology"}
```

with caching enabled
```
  - name: clutch.service.topology
    typed_config:
      "@type": types.google.com/clutch.config.service.topology.v1.Config
      cache:
        ttl: 7200s
```
```
2020-10-29T09:46:59.891-0700	INFO	gateway/gateway.go:105	registering service	{"serviceName": "clutch.service.topology"}
2020-10-29T09:46:34.426-0700	INFO	topology/topology.go:83	Topology caching is enabled	{"serviceName": "clutch.service.topology"}
```


api/config/service/topology/v1/topology.proto
```diff
@@ -12,6 +12,8 @@ message Config {
 }
 
 message Cache {
-  // The time to live (ttl) for an item in cache, the default is 1 hour (3600 seconds).
-  google.protobuf.Duration ttl = 1 [ (validate.rules).duration.gt.seconds = 1 ];
+  // This disables the caching functionality completely, all other aspects of the topology will continue to function.
+  bool disable = 1;
+  // The time to live (ttl) for an item in cache, the default is 2 hours (7200 seconds).
+  google.protobuf.Duration ttl = 2 [ (validate.rules).duration.gt.seconds = 1 ];
```
danielhochman, line 18: i wonder if we should use the presence of the cache message to enable and disable it...

currently this would require you to define the TTL even if you have the cache disabled for config validation to pass.

mcutalo88: Strange I didn't have TTL set in my testing locally and config validation passed. But  I like your idea better and will change it up.

backend/service/topology/topology.go
```diff
@@ -66,6 +66,11 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
 		scope:  scope,
 	}, nil
```
danielhochman, line 67: nit: can you clean this up actually, should just be c := &client{...} then return c, nil

weird to allocate your own nil err

danielhochman (approved):

backend/service/topology/topology.go
mcutalo88: Idk what i was thinking here, i think previously this was just a return statement

PR #649, opened 2020-11-02
authn: store provider refresh token in db

danielhochman:
### Description
Store the encrypted provider refresh token in the database as part of the new refresh token flow.

### Testing Performed
Unit w/ sqlmock

### TODOs
- [x] tests
- [x] update queries with new schema


backend/service/authn/sql.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type repository struct {
+	db *sql.DB
+}
+
+func newRepository() (*repository, error) {
+	svc, ok := service.Registry[postgres.Name]
+	if !ok {
+		return nil, fmt.Errorf("no database registered")
```
rithujohn191, line 24: are we planning to support other databases in the future? If yes, it might be helpful to display the name of the db in the error message i.e fmt.Errorf("%s database not registered", postgres.Name)?

backend/service/authn/sql.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	pg, ok := svc.(postgres.Client)
+	if !ok {
+		return nil, fmt.Errorf("database in does not implement the required interface")
```
rithujohn191, line 29:
```suggestion
		return nil, fmt.Errorf("database does not implement the required interface")
```

backend/cmd/migrate/migrations/000007_authn_users_create.up.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE authn_users(
+  id text PRIMARY KEY,
+  provider_refresh_token bytea
```
rithujohn191, line 3: Do we not want to store the expiry and the access token here as well?

danielhochman: good question. was going to wait on storing the access token for when we plan to use it. right now we are just focused on the silent refresh flow. i haven't quite decided how we're going to store service tokens yet either, e.g. we will feasibly have an okta, github token and maybe more for each user so we will probably have multiple rows for each user for that purpose.

i think i will rename this to `refresh_token` and make a composite primary key with another string field that holds the name of the provider in anticipation of later changes with multiple tokens stored.

the expiry is embedded in the access token's claims so there's no need to store it separately unless it needs to be indexed and range queryable for some purpose.

rithujohn191 (approved):

backend/service/authn/sql_test.go
```diff
@@ -0,0 +1,28 @@
...
+func TestCreateOrUpdateUser(t *testing.T) {
+	db, mock, err := sqlmock.New(sqlmock.QueryMatcherOption(sqlmock.QueryMatcherEqual))
+	assert.NoError(t, err)
+	defer db.Close()
+
+	id := "foo@example.com"
+	refreshToken := []byte("abdcdefgzzz")
+
+	r := &repository{db: db}
+	mock.ExpectQuery(createOrUpdateUser).
```
mcutalo88, line 20: I believe your missing an assert for mock at the end of your test, `mock.ExpectationsWereMet()`.

For example: https://github.com/DATA-DOG/go-sqlmock/blob/master/examples/basic/basic_test.go#L28-L31

mcutalo88 (approved):

danielhochman:
ptal #platform-tools

danielhochman:
@mcutalo88 @rithujohn191 ready for re-review

rithujohn191:
not directly related to this PR but in the future do we plan on adding support for different databases like SQLite?

backend/service/authn/sql.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	pg, ok := svc.(postgres.Client)
+	if !ok {
+		return nil, fmt.Errorf("database does not implement the required interface")
+	}
+
+	return &repository{db: pg.DB()}, nil
+}
+
```
rithujohn191, line 41: I see we dont' have a delete or remove function, whats our plan for token revocation?

rithujohn191 (approved):
Looks good and thanks for the thorough tests. Just a few question which can be addressed as follow up PRs.

danielhochman:
@rithujohn191 it's hard to say. most sql in clutch so far is behind a separate interface (i.e. `repository` here) so postgres could be substituted with a different database with minimal refactoring. however, i don't think we will add support as there's little benefit. someone else would need to come along who is interested in funding that work.

backend/service/authn/sql.go
danielhochman: good question. i view revocation as the responsibility of the provider (okta in this case). when we add clutch refresh (and store the clutch-issued tokens in the db) we can add logout and revocation.

backend/service/authn/crypto.go
```diff
@@ -38,6 +38,10 @@ func newCryptographer(passphrase string) (*cryptographer, error) {
 
 // Return decrypted bytes using input with nonce as the prefix and ciphertext as remaining bytes.
 func (c *cryptographer) Decrypt(b []byte) ([]byte, error) {
+	if len(b) < c.gcm.NonceSize()+1 {
```
mcutalo88, line 41: whats the `+1` for? maybe leave a comment here to clarify?

mcutalo88 (approved):
Just one comment for clarification purposes but otherwise it looks good.

Issue #415, opened 2020-09-16
core: all purpose modal view


Augustyniak:
A control that allows to define a view that's displayed modally - on top of the existing screen/page. 

<img width="986" alt="Screen Shot 2020-09-15 at 7 19 10 PM" src="https://user-images.githubusercontent.com/4410346/93284880-7cd26e80-f788-11ea-954f-a02ce15c8b83.png">

This issue is related to https://github.com/lyft/clutch/issues/5

Complexity [S/M/L]:
S

danielhochman:
I put a request for the designer to mock one up.

In the meantime have you played at all with https://material-ui.com/components/dialogs/?

danielhochman:
Here are the new designs:

![Screenshot from 2020-09-22 12-15-28](https://user-images.githubusercontent.com/4712430/93915257-5dc66600-fccd-11ea-8ec9-8259b305c746.png)

Once we get Storybook up and running (hopefully this week) we will be able to introduce this component as it appears above.

Augustyniak:
Thanks for sharing these - they look nice. 

Frankly speaking, I'm a little bit hesitant to adding new custom components to Clutch since I ran into multiple styling related issue when I tried to add my own datetime picker (somehow related to https://github.com/lyft/clutch/issues/414) or when I tried to add a custom implementation of `Select` (related to https://github.com/lyft/clutch/issues/433). 

@dschaller did make a few styling related fixes in Clutch so maybe my worries are not valid anymore but ideally I would prefer to use a Clutch provided control here instead of introducing a custom one.

dschaller:
closed in https://github.com/lyft/clutch/pull/674

PR #653, opened 2020-11-04
serverexperimentation: combine all steps of start-experiment into one step

Augustyniak:
### Description

- Combine all steps of start-experiment flow (3 of them currently) into one step.
- Add a modal dialog that asks for the confirmation when a user clicks a button that's supposed to start an experiment.
- Move a user to experiment details view when a new experiment is created.

| Before | After |
| ------ | ------ |
|ezgif com-gif-maker (20)|ezgif com-gif-maker (21)|

### Testing Performed
Created and start a few server experiments with the use of new UI.

### DIFF
--- a/frontend/workflows/serverexperimentation/src/dialog.tsx
+++ b/frontend/workflows/serverexperimentation/src/dialog.tsx
@@ -1,0 +1,42 @@
+import React from "react";
+import Button from "@material-ui/core/Button";
+import MuiDialog from "@material-ui/core/Dialog";
+import DialogActions from "@material-ui/core/DialogActions";
+import DialogContent from "@material-ui/core/DialogContent";
+import DialogTitle from "@material-ui/core/DialogTitle";
+import Typography from "@material-ui/core/Typography";
+
+interface ButtonProps {
+  label: string;
+  onAction: () => void;
+}
+
+interface DialogProps {
+  title: string;
+  content: string;
+  open: boolean;
+  onClose: () => void;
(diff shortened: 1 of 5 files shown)


frontend/workflows/serverexperimentation/src/dialog.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface DialogProps {
+  title: string;
+  content: string;
+  open: boolean;
+  onClose: () => void;
+  buttons: ButtonProps[];
+}
+
+const Dialog: React.FC<DialogProps> = ({ title, content, open, onClose, buttons }) => {
```
Augustyniak, line 22: @dschaller Per https://github.com/lyft/clutch/issues/415 I know that you want to introduce a dialog/modal component in Clutch at some point. Do you think that there would be any value in me moving the dialog defined in here to Clutch Core? I know that it does NOT look the way you would like it to look eventually but if we have an abstraction for dialogs in core you can change it later on (its internal implementation) to match your designs accordingly. We could keep the public API pretty limited (what I outlined above is almost a minimum I guess).

Let me know what you think.

donyu (approved):

frontend/workflows/serverexperimentation/src/dialog.tsx
dschaller: @Augustyniak I think this is a great candidate for core. We can take in the relevant props and then render children within the actions

PR #654, opened 2020-11-05
storybook: add loading story

dschaller:
### Description

Add loading story and a few updates to styling & props:
- Loadable will not longer center it's children by default. This allows each component to determine its own styling, in this case the resolver needed to be updated for this change in default behavior.
- The `variant` prop was turned into a flag on the loadable component.

loading

### Testing Performed

Manual with story and local clutch build


scarlettperry:
Locally (`make dev-mock`), the ec2 terminate instance and k8s resize hpa gets stuck on the loading component 🤔. Not seeing this one the main branch

dschaller:
Ok let me take a look. Very odd

PR #655, opened 2020-11-05
experimentation: consistent page layout

Augustyniak:
### Description
Use the same page layout with a heading positioned in the top-left corner in all of the screens that are implemented as part of `experimentation` and `serverexperimentation` packages.

| Before | After |
| ----- | ----- |
|Screen Shot 2020-11-04 at 7 16 28 PM|Screen Shot 2020-11-04 at 6 54 54 PM|
|Screen Shot 2020-11-04 at 7 16 36 PM|Screen Shot 2020-11-04 at 7 17 54 PM|

### Testing Performed
Went through all of the modified screens.

### DIFF
--- a/frontend/workflows/experimentation/tsconfig.json
+++ b/frontend/workflows/experimentation/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "composite": true,
     "outDir": "./dist",
     "rootDir": "./src"
   },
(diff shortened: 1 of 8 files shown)


murki (approved):
neat

donyu (approved):

frontend/workflows/experimentation/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "composite": true,
```
Augustyniak, line 4: This is needed due to some weird JS compilation issues - we use the same attribute in other shared packages such as clutch core too so I think it should be fine

PR #656, opened 2020-11-05
frontend: add types for JWT Token

dschaller:
### Description

jwt-decode released v3.1.1 this morning which included [type support](https://github.com/auth0/jwt-decode/commit/c9eaa4f880bb3638eddc631371010f7022436b10). `jwtDecode` now returns unknown and therefore fails typescript builds.

This defines the expected token props we rely on.

### Testing Performed

manually built

### DIFF
--- a/frontend/packages/core/src/AppLayout/user.tsx
+++ b/frontend/packages/core/src/AppLayout/user.tsx
@@ -33,6 +33,10 @@ const Initials = styled(Typography)`
   `}
 `;
 
+interface JwtToken {
+  sub: string;
+}
+
 const userId = (): string => {
   // Check JWT token for subject and display if available.
   const token = Cookies.get("token");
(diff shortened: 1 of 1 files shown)


rithujohn191 (approved):
🥳

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -33,6 +33,10 @@ const Initials = styled(Typography)`
   `}
 `;
 
+interface JwtToken {
```
mcutalo88, line 36: My typescript skills are pretty weak but do they have an interface which we can just use? or do we have to define it ourselves?

dschaller: Not right now. [The existing type definitions](https://github.com/auth0/jwt-decode/blob/master/index.d.ts) return unknown but I can see us adding at least the [registered claim names](https://tools.ietf.org/html/rfc7519#section-4.1) as optional.

mcutalo88 (approved):

dschaller:
We can remove this once https://github.com/auth0/jwt-decode/pull/107 lands and is released

PR #666, opened 2020-11-09
core: message redaction for auditing

danielhochman:
### Description

Adds a "redacted" message annotation which audit respects on request/response objects in order to log a redacted placeholder.

This prevents any sensitive data from being logged in auditing.

Another possibility would be moving the annotation to individual fields, but it's significantly more complicated.

### Testing Performed
TODO

### TODOs
- [x] Unit testing


mcutalo88 (commented):
Nice looks good, just needs some tests.

backend/gateway/meta/meta_test.go
```diff
@@ -118,6 +120,23 @@ func TestResourceNames(t *testing.T) {
 	}
 }
 
+func TestIsRedacted(t *testing.T) {
+	assert.True(t, IsRedacted(&authnv1.CallbackResponse{}))
+	assert.False(t, IsRedacted(&healthcheckv1.HealthcheckRequest{}))
+}
+
+func TestAPIBodyRedaction(t *testing.T) {
```
dschaller, line 128: Should we have a test for login response as well just to be sure the field isn't removed in the future?

dschaller (approved):
LGTM, pending question

backend/gateway/meta/meta_test.go
danielhochman: sure, i'll add tests to the auditing module itself to check this

mcutalo88 (approved):

PR #670, opened 2020-11-11
core: disable_audit annotation for healthcheck

danielhochman:
### Description
Add method annotation which prevents auditing from occurring.

### Testing Performed
Unit.


mcutalo88 (approved):
nice tests

PR #671, opened 2020-11-11
Experimentation: make UI links clickable

kathan24:
### Description

Making the links in the UI clickable with Material UI Links. 

### Testing Performed

Before
Screen Shot 2020-11-16 at 10 30 16 AM

After
Screen Shot 2020-11-16 at 2 17 06 PM

### GitHub Issue

Fixes #


frontend/workflows/experimentation/src/view-experiment-run.tsx
```diff
@@ -82,14 +82,20 @@ const ViewExperimentRun: React.FC<BaseWorkflowProps> = ({ heading }) => {
...
+            {experiment.properties.items.map(property =>
+              property.urlValue !== undefined ? (
+                <Link maxWidth="500px" href={property.urlValue}>
+                  {property.label}
+                </Link>
+              ) : (
+                <TextField
+                  key={property.label}
+                  label={property.label}
+                  defaultValue={propertyToString(property)}
```
Augustyniak, line 94: Should we update `propertyToString` method to take care of URL case too?

kathan24: Just looked at the function. 
Since `urlValue` will always be a string, so I am not sure if we need it because all it will do is return the value. Also, dashboard links should never be `undefined`. 
Let me know if there is anything else

Augustyniak: I think that we should not be assuming that for every use of `urlValue` we will always have a corresponding string values. It's true for now - as you mentioned, for dashboards links it's always the case - but nothing prevents somebody else from adding a urlValue that does NOT have a corresponding string value. That's what makes me think that we should update `propertyToString` and `compareProperties` methods.

Augustyniak: thanks - can we also update `compareProperties` method?

Augustyniak:
that's kind of nit picking from my side but:
1. could we increase the font size of links (so that it's the same as the font size used by text fields) 
1. add some padding to them to make them "fit better" into the list of properties that's displayed in experiment details page?

frontend/packages/core/src/link.tsx
```diff
@@ -4,15 +4,23 @@ import { Link as MuiLink } from "@material-ui/core";
 import styled from "styled-components";
 
 const StyledLink = styled(MuiLink)`
-  padding-left: 5px;
+  ${({ ...props }) => `
+  padding-top: 10px;
+  font-size: 16px;
```
Augustyniak, line 9: not sure whether it's a good idea to change the font-size and padding for all links - can we just do this for our own link (the one in experimentation details view)? There are other places in Clutch that use `Link`s and they assume a lower font-size I guess?

kathan24: Well, Derek made changes here so I think he is the best person to ask. 
@dschaller what do you suggest here?

Augustyniak: If Derek is OK you get +1 from me but I'm afraid that this will break the formatting of some of the existing links including the ones that we have in squeeze tests.

Augustyniak (approved):

frontend/workflows/experimentation/src/view-experiment-run.tsx
Augustyniak: ping

frontend/packages/core/src/link.tsx
Augustyniak: IMO font-size should be configurable on the caller site. Padding could also be added on the caller-site as needed since there are valid cases for when we will probably do not want to have any padding

dschaller: I didn't change the font size or padding top. The changes I made should have been backwards compatible with existing links since I only set a width if one is set as a prop. When I was testing this I didn't see a lot of things using links. At a min we shouldn't drop padding-left since this ensures links don't run into the end of another word.

frontend/workflows/experimentation/src/view-experiment-run.tsx
kathan24: added changes to `compareProperties` and `propertyToString`

frontend/packages/core/src/link.tsx
kathan24: added `padding-left` back

dschaller (commented):
Probably ok for now. We can revisit the API for links in the future. Can you add a story for Links so we can try this out?

kathan24:
@dschaller there is already a storybook story for [Links](https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/stories/link.stories.tsx)

dschaller (approved):

kathan24:
🚀

PR #683, opened 2020-11-16
ui v2: add roboto font

danielhochman:
### Description
Adds Roboto to HTML for the main application and storybook.

Uses `<link>` directives from https://fonts.google.com/specimen/Roboto?query=roboto&sidebar.open=true&selection.family=Roboto:wght@300;400;500;700

### DIFF
--- a/frontend/packages/app/public/index.html
+++ b/frontend/packages/app/public/index.html
@@ -4,8 +4,8 @@
     <meta charset="utf-8" />
     <link rel="icon" href="%PUBLIC_URL%/icon.svg" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
-    <link rel="preconnect" href="https://fonts.gstatic.com">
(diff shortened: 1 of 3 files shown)


frontend/packages/app/public/index.html
```diff
@@ -4,6 +4,8 @@
     <meta charset="utf-8" />
     <link rel="icon" href="%PUBLIC_URL%/icon.svg" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" />
```
dschaller, line 7: Can you add this in the [scaffolding](https://github.com/lyft/clutch/blob/main/tools/scaffolding/templates/gateway/frontend/public/index.html) as well?

dschaller (approved):
looks good pending scaffolding comment

PR #686, opened 2020-11-18
ui v2: checkbox

scarlettperry:
### Description
PR
* adds new designs for the checkbox component
* adds checkbox story

design is restyling the checkbox, so for now, added the design for what's highlighted in the dotted border. 

### Testing Performed
storybook

checbox

### DIFF
--- a/frontend/packages/core/src/Input/checkbox.tsx
+++ b/frontend/packages/core/src/Input/checkbox.tsx
@@ -1,4 +1,6 @@
 import React from "react";
+import styled from "@emotion/styled";
+import type { CheckboxProps } from "@material-ui/core";
 import {
   Checkbox as MuiCheckbox,
   FormControl as MuiFormControl,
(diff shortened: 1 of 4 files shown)


frontend/packages/core/src/Input/stories/checkbox-panel.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
```
scarlettperry, line 1: i just renamed the file and updated the imports.

frontend/packages/core/src/Input/stories/checkbox-panel.stories.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
```
scarlettperry, line 1: i just renamed the file and updated the imports.

frontend/packages/core/src/Input/checkbox.tsx
```diff
@@ -1,4 +1,6 @@
 import React from "react";
```
danielhochman, line 1: notice the name prop doesn't appear in storybook unless the import is like this

```suggestion
import * as React from "react";
```

frontend/packages/core/src/Input/checkbox.tsx
```diff
@@ -7,17 +9,85 @@ import {
...
+const Icon = styled.div(
+  {
+    borderRadius: "2px",
+    boxSizing: "border-box",
+    height: "24px",
+    width: "24px",
+  },
+  props => ({
+    border: props["data-disabled"] ? "1px solid #e7e7ea" : "1px solid #6e7083",
+  })
```
danielhochman, line 55: this could be styled from the parent without having to plumb props. IMO it's cleaner but we haven't written a style guide or made firm decisions here so it can be left as-is.

danielhochman (approved):
lgtm

danielhochman:
how will we render the 16px variant?

scarlettperry:
> how will we render the 16px variant?

was thinking we could pass a size prop? and toggle b/w the two sizes.

PR #690, opened 2020-11-19
logging: log 5xx server errors

rithujohn191:
### Description
Log 5xx HTTP errors in the server code along with the response body

This PR updated the mux.New method to take in a logger as a parameter and looks for a 5xx error in the `copyHTTPResponse` method that is invoked when `StatusCode != http.StatusOK`

### DIFF
--- a/backend/gateway/mux/mux.go
+++ b/backend/gateway/mux/mux.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -13,6 +14,7 @@ import (
 	"strings"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"go.uber.org/zap"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 	"google.golang.org/grpc"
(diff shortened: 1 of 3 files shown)


backend/gateway/mux/mux.go
```diff
@@ -35,14 +37,21 @@ type assetHandler struct {
...
+func copyHTTPResponse(resp *http.Response, w http.ResponseWriter, logger *zap.Logger) {
 	for key, values := range resp.Header {
 		for _, val := range values {
 			w.Header().Add(key, val)
 		}
 	}
+	if resp.StatusCode >= 500 && resp.StatusCode <= 599 {
+		body, _ := ioutil.ReadAll(resp.Body)
+		if len(body) > 0 {
+			logger.Error("HTTP 5xx error:", zap.Int("status code", resp.StatusCode), zap.String("response body", string(body)))
```
rithujohn191, line 52: left the body as a string for now since we do not have an expected response struct defined for us to call json.Unmarshal on

danielhochman (commented):
`copyHTTPresponse` is only called when the asset middleware is serving the response. 

thinking through it a bit more we should also move this to the gRPC side of things. we can make this a normal `middleware` component (1st PR). we can call it the `access_log` middleware and we can give it an access code filter for logging similar to envoy. then we can integrate it (2nd PR) similar to how the timeout middleware works, with instantiation by the gateway.

rithujohn191:
closing this PR to implement logging as a middleware component

PR #691, opened 2020-11-19
ui v2: add stepper component

dschaller:
### Description

Add a stepper component. I've improvised with an error icon for now. We can update once we get finalized designs.

Once we settle on colors we can hoist these up into the palette

stepper

### Testing Performed

manual via story


scarlettperry (approved):
nice!!! 🎉

frontend/packages/core/src/stepper.tsx
```diff
@@ -0,0 +1,186 @@
...
+          <Button onClick={handleReset} text="Reset" variant="neutral" />
+        ) : (
+          <div>
+            <Button
+              disabled={activeStep === 0}
+              onClick={handleBack}
+              text="Back"
+              variant="neutral"
+            />
+            <Button onClick={handleNext} text={activeStep === stepCount ? "Finish" : "Next"} />
```
danielhochman, line 178: should the controls be a part of the story (actions) and not the component itself?

dschaller: I figured as we use this we can shape how to hook into the stepper. My idea at the moment was to have a `onSubmit` for each step that triggers a callback during the transition phase but I think plugging this into a real world use case will flush out if that makes sense.

frontend/packages/core/src/stepper.tsx
```diff
@@ -0,0 +1,153 @@
...
+const CheckIcon = styled(MuiCheckIcon)((props: { font: string }) => ({
+  fill: props.font,
+  padding: "8px",
+}));
+
+const ClearIcon = CheckIcon.withComponent(PriorityHighIcon);
+
+const StepConnector = styled(MuiStepConnector)((props: { completed?: boolean }) => ({
+  ".MuiStepConnector-line": {
+    height: "5px",
```
scarlettperry, line 44: i think the height on figma is 7px

scarlettperry: oh it depends on which bar you read the height from. the bar example in the last row specifies 5px

frontend/packages/core/src/stepper.tsx
```diff
@@ -0,0 +1,153 @@
...
+const Step: React.FC<StepProps> = ({ children }) => <>{children}</>;
+
+export interface StepperProps {
+  activeStep: number;
+  children?: React.ReactElement<StepProps>[] | React.ReactElement<StepProps>;
+}
+
+const Stepper = ({ activeStep, children }: StepperProps) => (
+  <div>
+    <MuiStepper style={{ padding: "0" }} activeStep={activeStep + 1} connector={<StepConnector />}>
```
danielhochman, line 124: can we push the inline styles up to the parent div and use selectors?

frontend/packages/core/src/stepper.tsx
```diff
@@ -0,0 +1,153 @@
...
+        return (
+          <MuiStep key={step.label} style={{ padding: "0" }}>
+            <StepLabelIcon icon={<StepIcon {...stepProps} />} />
+          </MuiStep>
+        );
+      })}
+    </MuiStepper>
+    <Grid style={{ padding: "16px 0" }} container justify="space-between">
+      {React.Children.map(children, (step: any, idx: number) => (
+        <StepLabel item data-active={idx === activeStep}>
```
danielhochman, line 145: thinking about it more what about using a class name instead of data-.

frontend/packages/core/src/stepper.tsx
danielhochman: i think it is supposed to be 7px, left a comment on figma for clarification though

danielhochman (approved):
lgtm, discretionary nits

frontend/packages/core/src/stepper.tsx
dschaller: I kind of like using props here since it makes it clear what we are specifying compared to what Material is injecting

danielhochman: though data won't cascade for nested selectors, i think in general we should standardize on classes. we can use a prefix to distinguish from Mui if needed.

note: (comment after merge, not a blocker)

dschaller: 👍  going to make a cleanup task and add some of the standardizing bits we have talked about to it

PR #696, opened 2020-11-19
UI V2: header

scarlettperry:
### Description
This PR
* updates the Header, Logo, and UserInformation components
* adds the notification icon (created a Notifications component for when we implement that feature)
* adds a Header, Notifications, and UserInformation story

### Testing Performed
Created the components in storybook
Ran `make frontend-lint` and `make frontend-test` without errors

Before
Screen Shot 2020-11-19 at 3 30 34 PM

New (reflects latest commits)
Screen Shot 2020-12-02 at 7 16 35 PM

Untitled 5


frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -76,12 +64,14 @@ const Header: React.FC = () => {
...
             <Logo />
           </Link>
-          <Divider orientation="vertical" flexItem />
-          <Title variant="h5">clutch</Title>
+          <Title>clutch</Title>
           <Box />
           <SearchField />
           <Box />
-          <UserInformation />
+          <Grid container alignItems="center" justify="flex-end">
```
scarlettperry, line 71: since the designs have the search bar, notification icon, and user info grouped to the right, I added this grid component here.

scarlettperry:
hi @dschaller, mind giving this a first pass review? Still need to run the tests locally to see what tests to update.

frontend/packages/core/src/AppLayout/logo.tsx
```diff
@@ -11,8 +11,9 @@ const rotate = keyframes`
 `;
 
 const StyledSvg = styled.svg`
-  height: 40px;
-  width: 40px;
+  height: 32px;
```
scarlettperry, line 14: figma designs specify 32px for width and height but it looks smaller in storybook than it does on the design.
here is what it looks like at the original size (40px)
Screen Shot 2020-11-19 at 6 12 34 PM

frontend/packages/core/src/AppLayout/logo.tsx
```diff
@@ -11,8 +11,9 @@ const rotate = keyframes`
...
+  width: 32px;
+  margin-right: 16px;
```
scarlettperry, line 16: based the space between the logo and clutch from the design

scarlettperry:
I get errors when I run `make frontend test` but they look unrelated. Example:
```
@clutch-sh/core: FAIL src/tests/button.test.tsx
@clutch-sh/core:   ● Test suite failed to run
@clutch-sh/core:     /Users/sperry/go/src/github.com/lyft/clutch/frontend/packages/core/node_modules/@clutch-sh/tools/jest.setup.js:1
@clutch-sh/core:     ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import Enzyme from "enzyme";
@clutch-sh/core:                                                                                              ^^^^^^
@clutch-sh/core:     SyntaxError: Cannot use import statement outside a module
@clutch-sh/core:       at ScriptTransformer._transformAndBuildScript (../../node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
@clutch-sh/core:       at ScriptTransformer.transform (../../node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
@clutch-sh/core:           at Array.forEach (<anonymous>)
```

I see the same on the main UIV2 branch.

scarlettperry:
for the avatar hover design, for now do we just want to show the avatar icon again and the user's email since we have that info?

frontend/packages/core/src/AppLayout/notifcations.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <StyledNotificationsIcon
+        ref={anchorRef}
+        edge="end"
+        aria-controls={open ? "notifcation-options" : undefined}
+        aria-haspopup="true"
+        onClick={handleToggle}
+      >
+        <NotificationsIcon />
+      </StyledNotificationsIcon>
+      <Popper open={open} anchorEl={anchorRef.current} role={undefined} transition>
```
scarlettperry, line 52: unused at the moment, but thought it'd be good to have this set up

frontend/packages/core/src/AppLayout/notifcations.tsx
```diff
@@ -0,0 +1,74 @@
...
+  IconButton,
+  MenuList,
+  Paper,
+  Popper,
+} from "@material-ui/core";
+import NotificationsIcon from "@material-ui/icons/Notifications";
+
+const StyledNotificationsIcon = styled(IconButton)`
+  color: #ffffff;
+  margin-right: 1rem;
```
scarlettperry, line 16: I asked Kanchan for the hover and pressed states

frontend/packages/core/src/AppLayout/stories/header.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { MemoryRouter } from "react-router";
+import type { Meta } from "@storybook/react";
+
+import { ApplicationContext } from "../../Contexts/app-context";
+import Header from "../header";
+
+export default {
+  title: "Core/AppLayout/Header",
+  component: Header,
+  decorators: [
```
scarlettperry, line 11: added these decorators b/c the Header imports components that make use of `Link` and `React.createContext`. without it, the Header story fails with errors such as `useHref() may be used only in the context of a <Router> component` and `Cannot read property 'workflows' of undefined`

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -68,20 +55,22 @@ const Header: React.FC = () => {
 
   return (
     <>
-      <AppBar position="static" color="secondary">
+      <AppBar position="static" color="secondary" elevation={0}>
```
scarlettperry, line 58: `elevation` removes the box shadow 
https://stackoverflow.com/questions/34550593/how-to-disable-box-shadows-globally-for-all-material-ui-react-components

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -76,34 +102,52 @@ const UserInformation: React.FC = () => {
...
-      <Popper open={open} anchorEl={anchorRef.current} role={undefined} transition>
+      <Popper
+        open={open}
+        anchorEl={anchorRef.current}
+        role={undefined}
+        transition
+        popperOptions={{
+          modifiers: {
+            offset: {
+              offset: "-105,0",
```
scarlettperry, line 123: the popper has this css property `transform: translate3d` which sets its position. i changed where the popper ends so that it better matches the figma design

https://stackoverflow.com/questions/60122232/how-do-you-position-material-ui-popper-in-the-bottom-right-corner-of-the-browser

frontend/packages/core/src/AppLayout/notifcations.tsx
```diff
@@ -0,0 +1,74 @@
+import React from "react";
```
danielhochman, line 1: filename is misspelled

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -68,20 +55,22 @@ const Header: React.FC = () => {
...
         <Toolbar>
           <MenuButton onClick={openDrawer} edge="start" color="primary" data-qa="menuBtn">
             <MenuIcon />
           </MenuButton>
```
danielhochman, line 62: this hamburger menu does not exist anymore afaict, it's a fixed sidebar

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -68,20 +55,22 @@ const Header: React.FC = () => {
...
-          <Divider orientation="vertical" flexItem />
-          <Title variant="h5">clutch</Title>
+          <Title>clutch</Title>
           <Box />
           <SearchField />
           <Box />
-          <UserInformation />
+          <Grid container alignItems="center" justify="flex-end">
+            <Notifications />
+            <UserInformation />
```
danielhochman, line 72: the avatar is not vertically aligned

danielhochman: same for clutch logo actually, slightly offset  up

danielhochman (commented):
quick pass, looking good

frontend/packages/core/src/AppLayout/header.tsx
scarlettperry: yep! going to work on the navbar changes in a separate PR so kept it for now

frontend/packages/core/src/AppLayout/header.tsx
scarlettperry: fixed both. the avatar had a bottom padding and the logo wasn't aligned in its container.
Screen Shot 2020-12-01 at 10 56 04 AM

frontend/packages/core/src/AppLayout/notifcations.tsx
scarlettperry: I added the notfication states. Figma designs set the padding to 12px but it looked too large on the header. Example:
Screen Shot 2020-12-01 at 2 10 35 PM

So I brought it down to 8px.

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -31,15 +22,11 @@ const MenuButton = styled(IconButton)`
...
-const Divider = styled(MuiDivider)`
-  ${({ theme }) => `
-  background-color: ${theme.palette.primary.main};
-  margin: 16px 8px;
-  `}
+  margin-right: 1.563rem;
+  font-weight: bold;
+  font-size: 1.25rem;
+  color: #ffffff;
+  opacity: 0.87;
```
dschaller, line 29: Emotion supports dicts as parameters here which gives us type support. Something like
```
const Title = styled(Typography)({
  marginRight: "1.563rem",
  ...
});

File: frontend/packages/core/src/AppLayout/header.tsx (hunk 11 above)
Comment (dschaller) [reply frontend/packages/core/src/AppLayout/header.tsx:72]: I think the avatar is supposed to have a opaque white border

File: frontend/packages/core/src/AppLayout/user.tsx [hunk 13]
```diff
@@ -1,36 +1,61 @@
...
 const UserPhoto = styled(IconButton)`
-  ${({ theme }) => `
-  padding: 1px 0 ${theme.spacing(0.5)}px 12px;
-  margin-right: ${theme.spacing(0.5)}px;
-  `}
+  padding: 0.063rem 0rem 0rem 0.75rem;
+  margin-right: 0.25rem;
 `;
 
 const Avatar = styled(MuiAvatar)`
```
dschaller, line 25: I think the avatar is supposed to have a opaque white border

frontend/packages/core/src/AppLayout/notifications.tsx
```diff
@@ -0,0 +1,82 @@
...
+  Paper,
+  Popper,
+} from "@material-ui/core";
+import NotificationsIcon from "@material-ui/icons/Notifications";
+
+const StyledNotificationsIcon = styled(IconButton)`
+  color: #ffffff;
+  margin-right: 1rem;
+  border-radius: 8px;
+  padding: 8px;
```
dschaller, line 18:
```suggestion
  margin: 8px;
  padding: 12px;
```

frontend/packages/core/src/AppLayout/notifications.tsx
```diff
@@ -0,0 +1,82 @@
...
+              </ClickAwayListener>
+            </Paper>
+          </Grow>
+        )}
+      </Popper>
+    </Box>
+  );
+};
+
+export default Notifications;
```
dschaller, line 82: It might be useful to have a notifications story so we can test the whole component (e.g. the popper and such) in isolation from the header.

frontend/packages/core/src/AppLayout/notifications.tsx
```diff
@@ -0,0 +1,82 @@
...
+  padding: 8px;
+  &:hover {
+    background: #2d3db4;
+  }
+  &:active {
+    background: #2938a5;
+  }
+`;
+
+const Notifications: React.FC = () => {
```
dschaller, line 27: We might want to provide an interface to pass in notifications here to render in the popper

frontend/packages/core/src/AppLayout/notifications.tsx
```diff
@@ -0,0 +1,82 @@
...
+        {({ TransitionProps, placement }) => (
+          <Grow
+            {...TransitionProps}
+            style={{ transformOrigin: placement === "bottom" ? "center top" : "center bottom" }}
+          >
+            <Paper>
+              <ClickAwayListener onClickAway={handleClose}>
+                <MenuList
+                  autoFocusItem={open}
+                  id="notifcation-options"
```
dschaller, line 70:
```suggestion
```

frontend/packages/core/src/AppLayout/notifications.tsx
```diff
@@ -0,0 +1,82 @@
...
+        aria-haspopup="true"
+        onClick={handleToggle}
+      >
+        <NotificationsIcon />
+      </StyledNotificationsIcon>
+      <Popper open={open} anchorEl={anchorRef.current} role={undefined} transition>
+        {({ TransitionProps, placement }) => (
+          <Grow
+            {...TransitionProps}
+            style={{ transformOrigin: placement === "bottom" ? "center top" : "center bottom" }}
```
dschaller, line 64: Emotion supports [prop evaluation](https://emotion.sh/docs/styled#changing-based-on-props) that you could use here

frontend/packages/core/src/AppLayout/logo.tsx
```diff
@@ -11,11 +11,13 @@ const rotate = keyframes`
 `;
 
 const StyledSvg = styled.svg`
-  height: 40px;
-  width: 40px;
+  height: 2rem;
+  width: 2rem;
+  margin-right: 1rem;
```
dschaller, line 16: In the designs I think this is split between the title and the logo. Mostly a nit for now as I don't think it'll impact anything but the margin on left and right of the title is 8px and then right margin here is 8px

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -76,34 +109,52 @@ const UserInformation: React.FC = () => {
...
-    <Grid container alignItems="center" justify="flex-end">
-      <div>{userId()}</div>
+    <Box>
       <UserPhoto
         ref={anchorRef}
         edge="end"
         aria-controls={open ? "account-options" : undefined}
         aria-haspopup="true"
-        onClick={handleToggle}
+        onMouseEnter={handleToggle}
```
dschaller, line 118: Did we want this menu to present itself on mouseover? It's a little confusing to have this instead of an explicit click

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -76,34 +109,52 @@ const UserInformation: React.FC = () => {
...
-      <Popper open={open} anchorEl={anchorRef.current} role={undefined} transition>
+      <Popper
+        open={open}
+        anchorEl={anchorRef.current}
+        role={undefined}
+        transition
+        popperOptions={{
+          modifiers: {
+            offset: {
+              offset: "-115,5",
```
dschaller, line 130: The popper should be right aligned with the user photo 
Screen Shot 2020-12-01 at 11 43 58 AM

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: the figma designs mention the avatar popper opens on hover (or maybe I misunderstood "Avatar Hover List" to mean that), so i changed it to onMouseEnter. But I agree, onClick is better. Will change it back

frontend/packages/core/src/AppLayout/header.tsx
scarlettperry: updated all to use dicts as parameters

frontend/packages/core/src/AppLayout/logo.tsx
scarlettperry: thanks for catching this. after zooming in on the design, I think the logo has a margin left and right @ 8px and the title has a margin left @ 8px.



i left a to-do to remove the margin right in my PR with the V2 search bar design, since it'll be aligned to the right of the header.

frontend/packages/core/src/AppLayout/notifications.tsx
scarlettperry: do you mean something as simple as an interface with just `text: string` for now?

frontend/packages/core/src/AppLayout/notifications.tsx
scarlettperry: I pushed up a notifications story. At the moment, we won't see the popper b/c we don't have menu items.

frontend/packages/core/src/AppLayout/notifications.tsx
scarlettperry: I added it b/c I saw it included [here](https://github.com/lyft/clutch/blob/f9eb24f37b203c67037fa4071020e5511efa2563/frontend/packages/core/src/AppLayout/user.tsx#L100) and in the material ui examples. I removed it locally tho and it didn't break anything but also unsure if it's b/c of an edge case that I didn't hit. From quick research, seems like it's a common property included to create the relationship b/w a component and an expanded panel

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: Is this better? i moved it more to the right. i haven't had luck getting the alignment correct with normal css properties, and I found this workaround https://github.com/lyft/clutch/pull/696#discussion_r532778102.
Screen Shot 2020-12-02 at 12 47 16 AM

frontend/packages/core/src/AppLayout/notifications.tsx
scarlettperry: updated the hover states for notification and avatar. the 12px padding still looks to big in relation to the header (isolated it looks fine) so kept it at 8px

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: also I left a note on figma that the white-ish avatar background doesn't translate well on the menu

frontend/packages/core/src/AppLayout/logo.tsx
```diff
@@ -1,5 +1,6 @@
 import React from "react";
-import styled, { keyframes } from "styled-components";
+import { keyframes } from "@emotion/core";
```
scarlettperry, line 2: the rotate animation didn't work when I imported keyframes from `'@emotion/react'`. came across this https://stackoverflow.com/questions/55119780/react-emotion-animating-a-div-with-emotionjs which solved the issue.

frontend/packages/core/src/AppLayout/notifications.tsx
dschaller: Awesome! I was thinking that we could have an interface for the component where it takes in the notifications that we want to render. I know we don't have designs for this yet so feel free to punt but ideally we pass in notifications that would allow us to test the component. Not blocking at this point but something to keep in mind.

frontend/packages/core/src/AppLayout/notifications.tsx
dschaller: 12 looks good to me. We can always adjust when we move the search bar and remove the hamburger. I think for desktop we have the space though :)

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -1,46 +1,34 @@
...
-const AppBar = styled(MuiAppBar)`
-  ${({ theme }) => `
-  min-width: fit-content;
-  background-color: ${theme.palette.secondary.main};
-  color: ${theme.palette.primary.main};
-  min-width: fit-content;
-  `}
-`;
+const AppBar = styled(MuiAppBar)({
+  minWidth: "fit-content",
```
dschaller, line 14: You can leave a todo here since we don't design for mobile at the moment but on smaller devices the layout here doesn't stay consistent.
Screen Shot 2020-12-02 at 8 19 52 PM

frontend/packages/core/src/AppLayout/logo.tsx
```diff
@@ -10,13 +11,16 @@ const rotate = keyframes`
...
-  width: 40px;
-  &:hover {
-    animation: ${rotate} 5s linear;
-  }
-`;
+const StyledSvg = styled.svg({
+  height: "32px",
+  width: "32px",
+  marginLeft: "8px",
+  marginRight: "8px",
```
dschaller, line 18: Looking at the designs I think this should be
```suggestion
  height: "48px",
  width: "48px",
  padding: "8px",
```

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -1,46 +1,34 @@
...
-const Divider = styled(MuiDivider)`
-  ${({ theme }) => `
-  background-color: ${theme.palette.primary.main};
-  margin: 16px 8px;
-  `}
-`;
+// TODO (sperry): remove marginRight in new search bar design
+const Title = styled(Typography)({
+  marginRight: "25px",
+  marginLeft: "8px",
```
dschaller, line 26: I think the designs have this as 
```suggestion
  margin: "12px 8px",
```
but we can expand the right for the search bar to be
```suggestion
  margin: "12px 25px 12px 8px",
```

frontend/packages/core/src/AppLayout/logo.tsx
dschaller: can you import this from `@emotion/react` instead since we already depend on that? Looking at the [emotion docs](https://emotion.sh/docs/keyframes) it appears to be the recommendation

frontend/packages/core/src/AppLayout/notifications.tsx
dschaller: Resolving in lieu of https://github.com/lyft/clutch/pull/696#discussion_r534655596

frontend/packages/core/src/AppLayout/notifications.tsx
dschaller: ah yeah you are right. Sorry about that. It does create a relationship between the modal and the button

frontend/packages/core/src/AppLayout/stories/notifications.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import Notifications from "../notifications";
+
+export default {
+  title: "Core/AppLayout/Notifications",
+  component: Notifications,
+  parameters: {
+    backgrounds: {
+      default: "clutch",
+      values: [{ name: "clutch", value: "#131C5F" }],
```
dschaller, line 12:
```suggestion
      values: [{ name: "header blue", value: "#131C5F" }],
```

frontend/packages/core/src/AppLayout/stories/notifications.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+    },
+  },
+} as Meta;
+
+export const Primary: React.FC<{}> = () => <Notifications />;
```
dschaller, line 17: We should probably stick with using templates and exporting the bindings

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -1,37 +1,79 @@
...
+const Avatar = styled(MuiAvatar)({
+  backgroundColor: "#dce7f4",
+  height: "28px",
+  width: "28px",
+});
+
+const AvatarBackdrop = styled(MuiAvatar)({
+  backgroundColor: "#f6faff",
+  height: "32px",
+  width: "32px",
```
dschaller, line 39:
```suggestion
  border: "2px solid #f6faff",
  borderRadius: "50%",
```

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -53,6 +95,16 @@ const userId = (): string => {
   return subject;
 };
 
+const UserAvatar: React.FC = () => {
+  return (
+    <AvatarBackdrop>
+      <Avatar>
+        <Initials>{userId().slice(0, 2).toUpperCase()}</Initials>
+      </Avatar>
+    </AvatarBackdrop>
```
dschaller, line 104:
```suggestion
    <Avatar>
      <Initials>{userId().slice(0, 2).toUpperCase()}</Initials>
    </Avatar>
```

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -53,6 +95,16 @@ const userId = (): string => {
...
+  );
+};
```
dschaller, line 106:
```suggestion
const UserAvatar: React.FC = () => (
  <Avatar>
    <Initials>{userId().slice(0, 2).toUpperCase()}</Initials>
  </Avatar>
);
```

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -1,37 +1,79 @@
...
+  border: "1px solid #e2e2e6",
+  boxShadow: "0px 5px 15px rgba(53, 72, 212, 0.2)",
+});
+
+const UserProfileMenuItem = styled(MuiMenuItem)({
+  "&:focus": {
+    background: "transparent",
+  },
+  "&:hover": {
+    background: "transparent",
```
dschaller, line 60: I thought we were going to use the hover/active states from the select for this as well.

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -76,34 +128,52 @@ const UserInformation: React.FC = () => {
...
         aria-haspopup="true"
         onClick={handleToggle}
       >
-        <Avatar>
-          <Initials variant="h6">{userId().slice(0, 2).toUpperCase()}</Initials>
-        </Avatar>
+        <UserAvatar />
       </UserPhoto>
-      <Popper open={open} anchorEl={anchorRef.current} role={undefined} transition>
+      <Popper
```
dschaller, line 141: We might want to look into using popover for the user menu. Blocking scrolling and clicks might be reasonable here. Definitely investigate this later though

frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -76,34 +128,52 @@ const UserInformation: React.FC = () => {
...
+        anchorEl={anchorRef.current}
+        role={undefined}
+        transition
+        popperOptions={{
+          modifiers: {
+            offset: {
+              offset: "-116, -112",
+            },
+          },
+        }}
```
dschaller, line 152: We probably shouldn't inline style here but just for simplicity I think you can achieve the same thing with the following:
```suggestion
        placement="bottom-end"
        style={{
          padding: "0 12px",
          marginLeft: "12px"
        }}
```

The addition margin on the left is due to the `edge="end"` on the `UserPhoto`. Looking at the story, if you remove the grid wrapper around the user information the current logic breaks because the offset it hard coded. With these changes you can remove that wrapper and it'll still render with correct alignment when opening.

dschaller:
Overall looks really good! A few more comments here and we should be good to go. Let's get this merged into the feature branch and open follow up PRs to drill into what the notification and user menus should look like with content populated. Feel free to punt on things like the hover and active states until then as well.

frontend/packages/core/src/AppLayout/notifications.tsx
scarlettperry: okay got it! will leave a to-do, I'll follow up to add an interface. I think the metadata table interface would be a good one to pattern

frontend/packages/core/src/AppLayout/header.tsx
scarlettperry: got it, will leave a to-do and follow up!

frontend/packages/core/src/AppLayout/logo.tsx
scarlettperry: updated!

frontend/packages/core/src/AppLayout/header.tsx
scarlettperry: updated

frontend/packages/core/src/AppLayout/stories/notifications.stories.tsx
scarlettperry: updated, addressed it on user story as well

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: oh nice! i tried a variation of this but couldn't get it right. thank you, updated

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: i thought we were adding hover states to the list items that follow, but not to the user profile section (like the github profile menu). but i'll follow up to add an interface to the user component so we can render out more menu items and test how it all looks together. will leave a to-do

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: okay! will leave a to-do to investigate

frontend/packages/core/src/AppLayout/user.tsx
scarlettperry: thank you! this is so much better. fixed

frontend/packages/core/src/AppLayout/logo.tsx
scarlettperry: oh it works now. i probs had other changes at the time that broke the animation when i first imported it from emotion/react.

updated and removed the dep from package.json.

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -68,20 +56,22 @@ const Header: React.FC = () => {
 
   return (
     <>
-      <AppBar position="static" color="secondary">
+      <AppBar position="static" color="secondary" elevation={0}>
```
danielhochman, line 59: has no effect with CSS overrides

```suggestion
      <AppBar position="static" elevation={0}>
```

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -1,46 +1,34 @@
...
-  margin: 16px 8px;
-  `}
-`;
+// TODO (sperry): remove marginRight in new search bar design
+const Title = styled(Typography)({
+  margin: "12px 25px 12px 8px",
+  fontWeight: "bold",
+  fontSize: "20px",
+  color: "#ffffff",
+  opacity: "0.87",
```
danielhochman, line 30: make opacity part of the font color

```suggestion
  color: "rgba(255, 255, 255, 0.87)",
```

danielhochman (approved):
tiny nits, lgtm otherwise

we will need a way to disable the notifications functionality via a prop since we don't support it yet, but we can follow up with that later.

dschaller (approved):

PR #699, opened 2020-11-20
serverexperimentation: new proto model for server experiments

Augustyniak:
### Description
Introduce a new model that will be used for all of the server experiments. For now, we introduce the new model alongside the old one, we will follow up with the removal of the old one once the migration is completed.

### Testing performed
N/A


Augustyniak:
FYI @donyu

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+}
+
+message Latency {
+  FaultTargeting fault_targeting = 1 [ (validate.rules).message.required = true ];
+  FaultPercentage fault_parcentage = 2 [ (validate.rules).message.required = true ];
+  LatencyDuration latency_duration = 3 [ (validate.rules).message.required = true ];
+}
+
+message FaultPercentage {
+  float percentage = 1 [ (validate.rules).float = {gt : 0.0, lte : 100.0} ];
```
Augustyniak, line 82: alternatively we can make it an int or sth like this

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+}
+
+message AbortStatus {
+  int32 http_status_code = 1 [ (validate.rules).int32 = {gt : 99, lt : 600} ];
```
Augustyniak, line 86: we can mate it uint32 but I decided to go with int32 and validation rules - what do you think @snowp ?

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+}
+
+message LatencyDuration {
+  int32 duration_ms = 1 [ (validate.rules).int32.gt = 0 ];
```
Augustyniak, line 90: same as above - we could go with an uint32 here instead

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
   // Faults are injected to ingesss traffic by an upstream service(s).
   FAULTINJECTIONCLUSTER_UPSTREAM = 2;
 }
+
+message Config {
```
snowp, line 60: Maybe call this `HttpFaultConfig` or something that captures that this is HTTP? We assume HTTP deeper in with the http_abort_status, so I don't think this would be protocol agnostic

api/chaos/serverexperimentation/v1/serverexperimentation.proto
snowp: I would opt for unsigned unless there is a reason to keep it signed

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+  }
+}
+
+message SingleCluster {
+  string name = 1 [ (validate.rules).string = {min_bytes : 1} ];
+}
+
+message PartialSingleCluster {
+  string name = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  float percentage = 2 [ (validate.rules).float = {gt : 0.0, lte : 100.0} ];
```
snowp, line 137: Making this float gives the impression that we have float precision here, but I think realistically we'll have integer precision with a fixed denumerator. Might make sense to encode this here and for the other percent fields? Ie either let it specify both numerator/denumerator or a single uint32 for the numerator with a documented denumerator

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+  oneof enforcer {
+    option (validate.required) = true;
+
+    UpstreamEnforcing upstream = 1;
+    DownstreamEnforcing downstream = 2;
+  }
+}
+
+message UpstreamEnforcing {
+  oneof upstream {
```
snowp, line 103: i would name these oneof names `*_type`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
snowp: Again I'd prefer uint32 if possible. 

Maybe call this `fixed_duration_ms`? Just in anticipation of more dynamic latency injection options

snowp (changes requested):

api/chaos/serverexperimentation/v1/serverexperimentation.proto
Augustyniak: What if we start supporting GRPC status code going forward? One of the approaches would be to add them as part of existing `latency`/`abort` message. Do you think that renaming `Config` to `HttpFaultConfig` still makes sense for this scenario?

api/chaos/serverexperimentation/v1/serverexperimentation.proto
Augustyniak: > a single uint32 for the numerator with a documented denumerator
sounds good, will update. I agree, realistically speaking there is not a lot of value in us supporting float precision

api/chaos/serverexperimentation/v1/serverexperimentation.proto
snowp: I don't think it makes a whole lot sense to try to treat gRPC and HTTP as two different things here: gRPC is a RPC mechanism on top of HTTP/2, and so "HTTP faults" like latency trivially apply to gRPC as well. I think if we think of gRPC configuration as a way to specialize the subset of HTTP faults that apply to gRPC requests then we can happily add gRPC fields to a HTTPFault config

api/chaos/serverexperimentation/v1/serverexperimentation.proto
Augustyniak: Great idea

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+  oneof fault {
+    option (validate.required) = true;
+
+    Abort abort = 1;
+    Latency latency = 2;
+  }
+}
+
+message Abort {
+  FaultTargeting fault_targeting = 1 [ (validate.rules).message.required = true ];
```
Augustyniak, line 70: any opinion on removing `fault_` from the name of the field here? @snowp

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,83 @@ enum FaultInjectionCluster {
...
+  FaultPercentage fault_parcentage = 2 [ (validate.rules).message.required = true ];
```
Augustyniak, line 71: any opinion on removing `fault_` from the name of the field here? @snowp

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+    // The abort fault.
+    AbortFault abort_fault = 1;
+    // The latency fault.
+    LatencyFault latency_fault = 2;
+  }
+}
+
+// The definition of an abort fault.
+message AbortFault {
+  // The targeting of the fault describing what requests faults should be applied to.
```
snowp, line 74: I would try to phrase this more about "what requests are being considered for faults", so that it makes more sense with the fault_percentage field

api/chaos/serverexperimentation/v1/serverexperimentation.proto
snowp: sgtm

api/chaos/serverexperimentation/v1/serverexperimentation.proto
snowp: sgtm. `percentage`, not `parcentage`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+    option (validate.required) = true;
+
+    // Enforce a fault on an upstream.
+    UpstreamEnforcing upstream = 1;
+    // Enfore a fault on a downstream.
+    DownstreamEnforcing downstream = 2;
+  }
+}
+
+/// Enforce faults on an upstream.
```
snowp, line 105: Remove one of the `/`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+  }
+}
+
+// A single cluster.
+message SingleCluster {
+  // The name of a cluster.
+  string name = 1 [ (validate.rules).string = {min_bytes : 1} ];
+}
+
+// A partial single cluster - the part of a single cluster.
```
snowp, line 151: I would explain a bit more what this means, it's not immediately obvious. Maybe speak to how this allows for targeting a specific subset of the cluster?

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+}
+
+// The percentage controlling what portion of a given cluster should be applying faults.
+message ClusterPercentage {
+  // The numerator of a percentage with a fixed denumerator equal to 100
+  // (i.e. percentage equal to 50 results in 50/100 = 50%)
+  uint32 percentage = 1 [ (validate.rules).uint32 = {gt : 0, lte : 100} ];
+}
+
+// The fault percentage controlling how often a given fault should be applied.
```
snowp, line 166: "Often" to me implies a temporal aspect, I would phrase this is `what percentage of requests should have the fault applied`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+message FaultPercentage {
+  // The numerator of a percentage with a fixed denumerator equal to 100
+  // (i.e. percentage equal to 50 results in 50/100 = 50%)
+  uint32 percentage = 1 [ (validate.rules).uint32 = {gt : 0, lte : 100} ];
+}
+
+// The abort status to apply as part of an abort fault.
+message FaultAbortStatus {
+  // The abort HTTP status.
```
snowp, line 175: `// The HTTP status code to insert when applying an abort fault.`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+  uint32 http_status_code = 1 [ (validate.rules).uint32 = {gt : 99, lt : 600} ];
+}
+
+// The latency duration to apply as part of a latency fault.
+message FaultLatencyDuration {
+  // The fixed latency duration in milliseconds.
```
snowp, line 181: `to inject`/`to apply`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+message SingleCluster {
+  // The name of a cluster.
+  string name = 1 [ (validate.rules).string = {min_bytes : 1} ];
+}
+
+// A partial single cluster - the part of a single cluster.
+message PartialSingleCluster {
+  // The name of a cluster.
+  string name = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  // The percentage controlling what part of a cluster should be included.
```
snowp, line 155: Here I would speak to how this ought to be a consistent partition of the cluster, just to give some indication for how this works

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+  // A downstream responsible for enforcing a fault.
+  oneof downstream_type {
+    option (validate.required) = true;
+
+    // A single downstream cluster.
+    SingleCluster downstream_cluster = 2;
+  }
+}
+
+// A single cluster.
```
snowp, line 145: Maybe expand this to something like `A single cluster that is partaking in the fault injection`

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+message LatencyFault {
+  // The targeting of the fault describing where the faults should be applied.
+  FaultTargeting fault_targeting = 1 [ (validate.rules).message.required = true ];
+  // The percentage of requests the fault should be applied to.
+  FaultPercentage fault_parcentage = 2 [ (validate.rules).message.required = true ];
+  // The latency duration to apply.
+  FaultLatencyDuration fault_latency_duration = 3 [ (validate.rules).message.required = true ];
+}
+
+// The fault targeting that describes a part of the system that's responsible for enforcing a fault.
```
snowp, line 92: This could use some high level explanation about why this structure is used; how it allows us to support various forms of targeting criteria based on whether we're having the upstream or the downstream enforce the faults.

api/chaos/serverexperimentation/v1/serverexperimentation.proto
```diff
@@ -56,3 +56,128 @@ enum FaultInjectionCluster {
...
+  // The percentage of requests the fault should be applied to.
+  FaultPercentage fault_parcentage = 2 [ (validate.rules).message.required = true ];
+  // The abort status to apply.
+  FaultAbortStatus fault_abort_status = 3 [ (validate.rules).message.required = true ];
+}
+
+// The definition of a latency fault.
+message LatencyFault {
+  // The targeting of the fault describing where the faults should be applied.
+  FaultTargeting fault_targeting = 1 [ (validate.rules).message.required = true ];
```
snowp, line 85: Thinking about this a bit, should this be top level? It seems generic enough that I don't think it would ever be different between fault and latency?

Augustyniak: yea, probably it' fine to move it higher - it will be less flexible but I cannot think of a use case for this flexibility so lets just move it higher

snowp (approved):
This looks great, thanks for iterating!

PR #723, opened 2020-12-01
ui v2: horizontal rule

danielhochman:
### Description
Adds a horizontal rule that accepts children for display in the middle. 

Screenshot from 2020-12-01 11-26-30

### Testing Performed
Storybook

### DIFF
--- a/frontend/packages/core/src/stories/horizontal-rule.stories.tsx
+++ b/frontend/packages/core/src/stories/horizontal-rule.stories.tsx
@@ -1,0 +1,27 @@
+import * as React from "react";
+import type { Meta } from "@storybook/react";
+
+import type { HorizontalRuleProps } from "../horizontal-rule";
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/stories/horizontal-rule.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  children: (
+    <>
+      <ErrorIcon />
+      &nbsp;Proceed with caution
+    </>
+  ),
+};
+
+export const WithoutText = Template.bind({});
+WithoutText.args = {};
```
dschaller, line 31:
```suggestion
```

dschaller:
I'm not seeing `children` show up as a control in this story even with the export

danielhochman:
afaict storybook filters out the `children` prop unless you make it an explicit control of some kind

dschaller (approved):

PR #728, opened 2020-12-02
ui v2: radio button

dschaller:
### Description

Add new radio button component and update radio group to use it.

radio

### Testing Performed

storybook

### DIFF
--- a/frontend/packages/core/src/Input/radio.tsx
+++ b/frontend/packages/core/src/Input/radio.tsx
@@ -21,10 +21,10 @@ const StyledRadio = styled(MuiRadio)(
 
 const Icon = styled.div(
   {
-    height: "1.5rem",
-    width: "1.5rem",
+    height: "24px",
+    width: "24px",
     border: "1px solid rgba(13, 16, 48, 0.38)",
-    borderRadius: "6.25rem",
+    borderRadius: "100px",
     boxSizing: "border-box",
   },
   props => ({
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/Input/radio.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { RadioProps as MuiRadioProps } from "@material-ui/core";
+import { Radio as MuiRadio } from "@material-ui/core";
+
+const StyledRadio = styled(MuiRadio)(
+  {
+    ".MuiIconButton-label": {
+      height: "24px",
```
scarlettperry, line 9: nit: should we use rem throughout?

scarlettperry:
should we ask for updated radio designs? if it's planned to match the new checkbox designs, then there will be additional color and size props and the states will be different.

dschaller:
I believe Kanchan is working on updated designs this week. I figured we could get these in and modify based on the new designs as needed.

frontend/packages/core/src/Input/radio.tsx
dschaller: Daniel and I chatted about this Monday and I think we are going to stick with pixels. There used to be much more of a reason to use rem to prevent scaling issues but I think most modern browsers handle this for us now.

scarlettperry: ah ok! will update my PR

scarlettperry (approved):

frontend/packages/core/src/Input/radio.tsx
```diff
@@ -0,0 +1,74 @@
...
+  props => ({
+    "&:hover > .MuiIconButton-label > div": {
+      border: props.checked ? "1px solid #283CD2" : "1px solid #2E45DC",
+    },
+  })
+);
+
+const Icon = styled.div(
+  {
+    height: "1.5rem",
```
scarlettperry, line 24: nit: since we're going with px, perhaps the few `rems` throughout should be converted to px.

PR #729, opened 2020-12-02
ui v2: restyle metadata table

dschaller:
### Description

Restyle metadata table.

Screen Shot 2020-12-01 at 8 49 59 PM

### Testing Performed

storybook

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -81,7 +81,6 @@ const KeyCellContainer = styled(TableCell)({
 interface KeyCellProps {
   data: IdentifiableRowData;
   size: Size;
-  isLast?: boolean;
 }
 
 const KeyCell: React.FC<KeyCellProps> = ({ data, size }) => {
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -10,14 +10,13 @@ import {
   TableCell as MuiTableCell,
   TableContainer as MuiTableContainer,
   TableRow,
-  Typography,
 } from "@material-ui/core";
 import _ from "lodash";
 import type { Schema } from "yup";
 import { object } from "yup";
 
 import { useWizardContext } from "../Contexts";
-import TextField from "../Input/text-field";
+import { TextField } from "../Input/text-field";
 
 interface RowData {
   input?: {
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -3,7 +3,6 @@ import { useForm } from "react-hook-form";
 import styled from "@emotion/styled";
 import { DevTool } from "@hookform/devtools";
 import { yupResolver } from "@hookform/resolvers/yup";
-import type { Size } from "@material-ui/core";
 import {
   Table as MuiTable,
   TableBody as MuiTableBody,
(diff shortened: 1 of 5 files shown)


frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -145,22 +210,29 @@ const MetadataTable: React.FC<MetadataTableProps> = ({
...
+      <Table>
         <TableBody>
-          {rows.map((row: IdentifiableRowData) => {
+          {rows.map((row: IdentifiableRowData, idx: number) => {
+            let position = "middle" as RowPosition;
+            if (idx === 0) {
+              position = "first";
+            } else if (idx === rows.length - 1) {
+              position = "last";
+            }
```
danielhochman, line 223: can't we do this with css selectors `:first-child` and `:last-child`?

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -103,18 +137,17 @@ const EditableRow: React.FC<EditableRowProps> = ({
   );
 };
 
-interface MetadataTableProps {
+export interface MetadataTableProps {
   data: RowData[];
   onUpdate?: (id: string, value: unknown) => void;
   variant?: Size;
```
danielhochman, line 143: is there a use case for this prop? i didn't see it as part of any of the existing designs

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -26,33 +32,69 @@ interface IdentifiableRowData extends RowData {
...
+const KeyCellContainer = styled(TableCell)({
+  width: "45%",
+  background: "rgba(13, 16, 48, 0.03)",
+  fontWeight: 500,
+});
+
 interface KeyCellProps {
   data: IdentifiableRowData;
   size: Size;
+  isLast?: boolean;
```
danielhochman, line 84: i think this is obsolete now

```suggestion
```

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -26,33 +32,69 @@ interface IdentifiableRowData extends RowData {
...
+    "*:first-child": {
+      borderBottomLeftRadius: "3px",
+    },
+    "*:last-child": {
+      borderBottomRightRadius: "3px",
+    },
+  },
+});
+
+const TableCell = styled(MuiTableCell)({
```
danielhochman, line 69: figma cells are ~5px less in height than these

frontend/packages/core/src/Table/stories/metadata-table.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+        }}
+      >
+        {story()}
+      </WizardContext.Provider>
+    ),
+  ],
+  component: MetadataTable,
+} as Meta;
+
+const Template = (props: MetadataTableProps) => (
```
danielhochman, line 30: this is rendering in storybook code preview as `<[object Object] value={() => {}}>` for some reason

danielhochman (commented):
nits

frontend/packages/core/src/Table/metadata-table.tsx
dschaller: I wasn't sure if we should keep it or not. My main thought was if we wanted to embed this within a full page dashboard and make it smaller for informational purposes. I can remove for now though.

frontend/packages/core/src/Table/stories/metadata-table.stories.tsx
danielhochman: oh something to do with the decorator?

dschaller: This is the application context wrapper. It looks like the context provider is rendering as the return value (e.g. an object in this case).

danielhochman: ok not a blocker

danielhochman (approved):

PR #731, opened 2020-12-02
ui v2: restyle dialog component

dschaller:
### Description

restyle dialog

Screen Shot 2020-12-01 at 10 48 34 PM

Screen Shot 2020-12-01 at 10 48 38 PM

### Testing Performed

storybook

### Todo
- [x] hook up the close icon

### DIFF
--- a/frontend/packages/core/src/Input/text-field.tsx
+++ b/frontend/packages/core/src/Input/text-field.tsx
@@ -20,11 +20,13 @@ const BaseTextField = ({ InputProps, InputLabelProps, ...props }: MuiStandardTex
 );
 
 const StyledTextField = styled(BaseTextField)({
+  padding: "8px 0",
   ".MuiInputLabel-root": {
     fontSize: "14px",
     fontWeight: 500,
     transform: "scale(1)",
     marginLeft: "2px",
+    padding: "8px 0",
   },
 
   ".MuiInputLabel-root, .MuiInputLabel-root.Mui-focused": {
--- a/frontend/packages/core/src/dialog.tsx
+++ b/frontend/packages/core/src/dialog.tsx
@@ -1,27 +1,50 @@
-import React from "react";
-import MuiDialog from "@material-ui/core/Dialog";
-import DialogActions from "@material-ui/core/DialogActions";
-import DialogContent from "@material-ui/core/DialogContent";
-import DialogTitle from "@material-ui/core/DialogTitle";
-import Typography from "@material-ui/core/Typography";
+import * as React from "react";
+import styled from "@emotion/styled";
+import {
+  Dialog as MuiDialog,
+  DialogActions as MuiDialogActions,
+  DialogContent as MuiDialogContent,
+  DialogTitle as MuiDialogTitle,
+} from "@material-ui/core";
+import CloseIcon from "@material-ui/icons/Close";
+
+const DialogTitle = styled(MuiDialogTitle)({
+  display: "flex",
+  justifyContent: "space-between",
+  fontSize: "16px",
+  padding: "16px 16px 0 16px",
+  fontWeight: 400,
+});
+
+const DialogContent = styled(MuiDialogContent)({
+  padding: "8px 16px",
+  fontSize: "12px",
+  fontWeight: 400,
+  color: "rgba(13, 16, 48, 0.6)",
+});
+
+const DialogActions = styled(MuiDialogActions)({
+  borderTop: "1px solid rgba(13, 16, 48, 0.12)",
+  padding: "0 8px",
+});
 
 export interface DialogProps {
   title: string;
-  content: string;
+  children: React.ReactNode;
+  actions: React.ReactNode;
   open: boolean;
   onClose: () => void;
 }
 
-const Dialog: React.FC<DialogProps> = ({ title, content, open, onClose, children }) => {
-  return (
-    <MuiDialog open={open} onClose={onClose}>
-      <DialogTitle>{title}</DialogTitle>
-      <DialogContent>
-        <Typography gutterBottom>{content}</Typography>
-      </DialogContent>
-      <DialogActions>{children}</DialogActions>
-    </MuiDialog>
-  );
-};
+export const Dialog = ({ title, children, open, onClose, actions }: DialogProps) => (
+  <MuiDialog open={open} onClose={onClose}>
+    <DialogTitle disableTypography>
+      {title}
+      <CloseIcon />
+    </DialogTitle>
+    <DialogContent>{children}</DialogContent>
+    <DialogActions>{actions}</DialogActions>
+  </MuiDialog>
+);
 
 export default Dialog;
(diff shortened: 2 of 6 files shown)


frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+} from "@material-ui/core";
+import CloseIcon from "@material-ui/icons/Close";
 
-export interface DialogProps {
+const DialogTitle = styled(MuiDialogTitle)({
+  display: "flex",
+  justifyContent: "space-between",
+  fontSize: "16px",
+  padding: "16px 16px 0 16px",
+  fontWeight: 400,
```
danielhochman, line 18: in Figma it's Roboto Medium 20px

```suggestion
  fontSize: "20px",
  fontWeight: 500,
  padding: "16px 16px 0 16px"
```

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+  color: "#000000",
```
danielhochman, line 19: from figma

```suggestion
  color: "#0d1030",
```

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+});
+
+const IconButton = styled(MuiIconButton)({
+  height: "16px",
+  width: "16px",
+  padding: "0",
+  color: "#000000",
```
danielhochman, line 26:
```suggestion
  color: "#0d1030",
```

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+});
+
+const DialogContent = styled(MuiDialogContent)({
+  padding: "8px 16px",
+  fontSize: "12px",
```
danielhochman, line 31: from figma

```suggestion
  fontSize: "16px",
```

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+  fontWeight: 400,
+  color: "rgba(13, 16, 48, 0.6)",
+});
+
+const DialogActions = styled(MuiDialogActions)({
+  borderTop: "1px solid rgba(13, 16, 48, 0.12)",
```
danielhochman, line 37: hmm should we use `<Divider>` or just add the border to actions like this? (asking for `<Accordion>`)

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+  padding: "0 8px",
+});
+
+export interface DialogProps extends Pick<MuiDialogProps, "open"> {
   title: string;
-  content: string;
-  open: boolean;
+  children: React.ReactNode;
+  actions: React.ReactNode;
   onClose: () => void;
```
danielhochman, line 45: i'm not sure that onClose should ever do anything other than dismiss the dialog, that seems like a bad UX pattern to allow devs to tie behavior to onClose. any use case in mind?

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
-      <DialogTitle>{title}</DialogTitle>
-      <DialogContent>
-        <Typography gutterBottom>{content}</Typography>
-      </DialogContent>
-      <DialogActions>{children}</DialogActions>
-    </MuiDialog>
-  );
-};
+export const Dialog = ({ title, children, open, onClose, actions }: DialogProps) => (
+  <MuiDialog open={open} onClose={onClose}>
```
danielhochman, line 49: Dialog has `border: 1px solid rgba(13, 16, 48, 0.1);` in Figma

and the box shadow doesn't match Figma's `box-shadow: 0px 10px 24px rgba(35, 48, 143, 0.3);`

frontend/packages/core/src/dialog.tsx
```diff
@@ -1,27 +1,61 @@
...
+  borderTop: "1px solid rgba(13, 16, 48, 0.12)",
+  padding: "0 8px",
+});
+
+export interface DialogProps extends Pick<MuiDialogProps, "open"> {
   title: string;
-  content: string;
-  open: boolean;
+  children: React.ReactNode;
+  actions: React.ReactNode;
```
danielhochman, line 44: we should standardize, either use these richer API where you pass components to be rendered nested vs just using children for everything and passing in Actions as part of the children (this same thing applies for accordion). i think we just need to some quick pro/con on it.

my main thought in potentially choosing the children-only API is that it reads more literally despite being a little more complicated for the user (but that's what stories are for, or maybe an invariant check on top of it), i.e. title > content > buttons, whereas this API reads title > buttons > content but renders as title > content > buttons

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -20,11 +20,13 @@ const BaseTextField = ({ InputProps, InputLabelProps, ...props }: MuiStandardTex
 );
 
 const StyledTextField = styled(BaseTextField)({
+  padding: "8px 0",
```
danielhochman, line 23: another decision point here, should children have inherent padding or should the parent pad it's children (e.g. something like a `& > *` selector in containers), not sure what's best practice

frontend/workflows/serverexperimentation/src/start-experiment.tsx
```diff
@@ -210,17 +210,22 @@ const StartExperiment: React.FC<StartExperimentProps> = ({
       />
       <Dialog
         title="Experiment Start Confirmation"
-        content="Are you sure you want to start an experiment? The experiment will start immediately and you will be moved to experiment details view page."
         open={experimentData !== undefined}
         onClose={() => setExperimentData(undefined)}
+        actions={
+          <>
```
dschaller, line 216: ButtonGroup instead to apply margin

frontend/packages/core/src/stories/dialog.stories.tsx
```diff
@@ -13,17 +14,47 @@ export default {
...
+        <>
+          <Button text="Back" variant="neutral" />
+          <Button text="Continue" />
+        </>
+      }
+      onClose={onClose}
+      open={isOpen}
+      {...props}
+    >
+      {children}
```
dschaller, line 34:
```suggestion
      {children}
      <DialogActions>
        <Button text="Back" variant="neutral" />
        <Button text="Continue" />
      </DialogActions>
```

frontend/packages/core/src/stories/dialog.stories.tsx
```diff
@@ -13,17 +14,47 @@ export default {
...
+    setIsOpen(!isOpen);
+  };
+  return (
+    <Dialog
+      actions={
+        <>
+          <Button text="Back" variant="neutral" />
+          <Button text="Continue" />
+        </>
+      }
```
dschaller, line 29:
```suggestion
```

frontend/packages/core/src/dialog.tsx
dschaller: I think for now I'll leave this as a border. It makes it slightly easier for purposes of the children. e.g. if we used a divider we would need to wrap the DialogActions children to prepend the divider to them

danielhochman (approved):
beautiful

PR #753, opened 2020-12-07
frontend: dynamic api support

### LINKED ISSUE #650: resolver: front-end resolution of non-core objects
**Description**

The front-end resolver component resolves the returned objects from the resolve/search call to the protobuf.js object.

However, this currently only resolves objects from the core package `@clutch-sh/api`. If we have a custom package e.g. `my-custom-api-package` that we build in a custom gateway, the `.fromObject` call will fail as the object is not in the core package.

This results in the error: ` 'Cannot read property 'fromObject' of undefined'.`, which comes from

```jsx
import * as $pbclutch from "@clutch-sh/api";
```

and then

https://github.com/lyft/clutch/blob/5ce6456b376edce0517acf4b8a0a5049512d4ed9/frontend/packages/core/src/Resolver/fetch.tsx#L63

We need to have some way to configure the resolver to also attempt `fromObject` for other packages.

Slack thread with original report: https://lyftoss.slack.com/archives/C015UJ6LED9/p1604147269019800

Complexity [S/M/L]: S

rithujohn191:
### Description
This PR addresses https://github.com/lyft/clutch/issues/650. It enables users to provide a custom package to the front-end resolver via an environment variable. In the long run we should move this to the clutch config but in the meantime this change will unblock users that need to resolves objects that are not in the core package `@clutch-sh/api`

### Testing Performed
Local testing: spun up the frontend and backend and tested the k8s flow. No errors were observed.

### GitHub Issue
Fixes #650

### TODOs
Documentation on how to include custom packages for the front-end resolver


frontend/packages/core/src/Resolver/fetch.tsx
```diff
@@ -1,15 +1,15 @@
 import type { clutch as IClutch } from "@clutch-sh/api";
-import * as $pbclutch from "@clutch-sh/api";
 import _ from "lodash";
 
 import { client, parseErrorMessage } from "../network";
 
 const fetchResourceSchemas = async (type: string): Promise<IClutch.resolver.v1.Schema[]> => {
+  const pbclutch = await require(`${process.env.REACT_APP_API_PACKAGE || "@clutch-sh/api"}`);
```
dschaller, line 7: This is great!

Can we read this in from the clutch config file instead of a process env to keep (mostly) everything in the same place?

danielhochman:
we need to support multiple API packages at the same time, e.g. core and -private. this appears to only allow one API package unless i'm misinterpreting.

rithujohn191:
closing this PR for now. If we need to support dynamic imports we need to update the webpack config to explicitly allow this which is not considered a best practice.

PR #754, opened 2020-12-07
ui v2: accordion group

danielhochman:
### Description
Add a mutually exclusive expanded accordion group.

### Testing Performed
Storybook.

### DIFF
--- a/frontend/packages/core/src/accordion.tsx
+++ b/frontend/packages/core/src/accordion.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import styled from "@emotion/styled";
-import type { AccordionProps as MuiAccordionProps } from "@material-ui/core";
+import { AccordionProps as MuiAccordionProps, useControlled } from "@material-ui/core";
 import {
   Accordion as MuiAccordion,
   AccordionActions as MuiAccordionActions,
@@ -77,7 +77,7 @@ export const StyledAccordionSummary = styled(AccordionSummaryBase)({
   },
 });
 
-export interface AccordionProps extends Pick<MuiAccordionProps, "defaultExpanded"> {
+export interface AccordionProps extends Pick<MuiAccordionProps, "defaultExpanded" | "expanded"> {
   title?: string;
   collapsible?: boolean;
   children: React.ReactNode[];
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/accordion.tsx
```diff
@@ -77,36 +78,92 @@ export const StyledAccordionSummary = styled(AccordionSummaryBase)({
...
+  ".MuiAccordion-root:before": {
+    display: "none",
+  },
+});
+
+export interface AccordionProps extends Pick<MuiAccordionProps, "defaultExpanded" | "expanded"> {
   title?: string;
   collapsible?: boolean;
-  children: React.ReactNode[];
+  children: React.ReactNode | React.ReactNode[];
```
dschaller, line 98: `ReactNode` should accept an array of ReactNodes so I think you can do:
```suggestion
  children: React.ReactNode;
```

frontend/packages/core/src/accordion.tsx
```diff
@@ -77,36 +78,92 @@ export const StyledAccordionSummary = styled(AccordionSummaryBase)({
...
+      <StyledAccordionSummary expanded={expanded} collapsible={collapsible} onClick={handleClick}>
         {title}
       </StyledAccordionSummary>
       {children}
     </StyledAccordion>
   );
 };
 
+export interface AccordionGroupProps {
+  children: React.ReactElement | React.ReactElement[];
```
dschaller, line 142:
```suggestion
  children: React.ReactElement;
```

dschaller (approved):

frontend/packages/core/src/accordion.tsx
danielhochman:
```
This JSX tag's 'children' prop expects a single child of type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>', but multiple children were provided.ts(2746)
```

danielhochman: (worked for node but not for element)

dschaller: ah sorry I misread this as `ReactNode`

PR #768, opened 2020-12-08
github: add function for retrieving repository information

angethuy:
### Description
This pull request adds a function `GetRepository` to the Github service that takes a repository name and owner and returns repository information. Currently returns repo name, owner, and default branch name but can be expanded to expose additional repository info as needed.

### Testing Performed
Manual local test and additional unit test cases.


danielhochman (commented):
nice. let's add this to a GetRepository method that returns the default branch as one of its fields.

backend/service/github/github.go
```diff
@@ -56,6 +56,23 @@ type RemoteRef struct {
...
+// RepositoryOwner represents the entity that owns a repo
+type RepositoryOwner struct {
+	Login string
+}
+
+// Repository contains information about a requested repository.
+type Repository struct {
+	Name          string
+	Owner         RepositoryOwner
+	DefaultBranch BranchRef
```
danielhochman, line 73: nit: flatten to string, we don't need to adhere to graphQL abstractions in our internal objects

danielhochman (commented):
looks good after comment

danielhochman (approved):
:tada:

PR #769, opened 2020-12-09
k8s: option to receive informer updates

mcutalo88:
### Description
For large Kubernetes deployments reacting to all informer events can put a lot of load on the datastore, the motivation behind this change is to reduce load.

Currently we handle all updates to pods for the topology caching but we dont currently need that level of granularity, requirements can change in the future and we can think about making these configuration options if use cases arise.

### Testing Performed

This was tested locally.


danielhochman (approved):
lgtm with additional documentation

PR #774, opened 2020-12-10
aws: configuration for client retries

mcutalo88:
### Description
The motivation for these changes are due to the account wide rate limit specifically for [DescribeStreamSummary](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_DescribeStreamSummary.html), which is 20rps for an aws account.

> 2020-12-11T11:12:28.126-0800	ERROR	aws/cache.go:149	unable to describe kinesis stream	{"serviceName": "clutch.service.aws", "error": "LimitExceededException: Rate exceeded for stream [some-stream-name-here] under account 173840052742."}

We were hitting these rate limits pretty easily with topology caching enabled, here are the things we are doing to prevent that from being an issue in the future.

* The aws client has a [DefaultRetryer](https://docs.aws.amazon.com/sdk-for-go/api/aws/client/#DefaultRetryer) which will utilize.
* Increase the timer for Kinesis topology caching, to reduce overall calls to this API.

### Testing Performed

For local testing I started several `processAllKinesisStreams` threads, increasing overall RPS to the `DescribeStreamSummary` API. By doing this I was able to easily reproduce the rate limit issue.

This was the configuration I used to reproduce.
```
  - name: clutch.service.aws
    typed_config:
      "@type": types.google.com/clutch.config.service.aws.v1.Config
      regions:
        - us-east-1
        - us-west-2
  - name: clutch.service.topology
    typed_config:
      "@type": types.google.com/clutch.config.service.topology.v1.Config
      cache:
        ttl: 7200s
```

With the below configuration I was unable to hit any rate limiting errors like the one above.
```
  - name: clutch.service.aws
  - name: clutch.service.aws
    typed_config:
      "@type": types.google.com/clutch.config.service.aws.v1.Config
      regions:
        - us-east-1
        - us-west-2
      client_config:
        retries: 10
  - name: clutch.service.topology
    typed_config:
      "@type": types.google.com/clutch.config.service.topology.v1.Config
      cache:
        ttl: 7200s
```


api/config/service/aws/v1/aws.proto
```diff
@@ -8,4 +8,10 @@ import "validate/validate.proto";
 
 message Config {
   repeated string regions = 1 [ (validate.rules).repeated = {min_items : 1} ];
+  AWSClient aws_client = 2;
```
danielhochman, line 11: nit: stutters with package name
```suggestion
  ClientConfig client_config = 2;
```

danielhochman (commented):
was the lack of rate limiting when testing this branch due to the increased ticker duration or the retry?

in general 50 retries with no backoff at all is a bad idea and could DoS the API or exacerbate rate limiting issues

mcutalo88:
I was under the assumption that the [DefaultRetryer](https://docs.aws.amazon.com/sdk-for-go/api/aws/client/#DefaultRetryer) would be utilized, which has exponential backoff amongst several other things. Ill do another pass through the aws client to verify this is the case.

During testing I exacerbate the ratelimit by reducing the timer on `processAllKinesisStreams` while also invoking `processAllKinesisStreams` several times to increase concurrency, for example this is what my testing looked like.

```
	for name, client := range c.clients {
		c.log.Info("processing topology objects for region", zap.String("region", name))
		// go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*5), client, c.processAllEC2Instances)
		// go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*10), client, c.processAllAutoScalingGroups)
		go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*30), client, c.processAllKinesisStreams)		 
                 go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*30), client, c.processAllKinesisStreams)
		go c.startTickerForCacheResource(ctx, time.Duration(time.Minute*30), client, c.processAllKinesisStreams)
	}
```

danielhochman:
unless i'm misreading, for all backoff params the docs say something to the effect of
```
    // If not set, the value is 0.
```

mcutalo88:
Alright I fixed this up and the aws client should be setup properly now. I ran all my test locally again and validated it was working as expected. I also created https://github.com/lyft/clutch/issues/790 as a followup for a more generic solution.

danielhochman (approved):

PR #776, opened 2020-12-10
ui v2: add end adornment support to text fields

dschaller:
### Description

Add support to text fields for end adornment

Screen Shot 2020-12-10 at 2 44 43 PM

### Testing Performed

In dev app and storybook

### DIFF
--- a/frontend/packages/core/src/Input/text-field.tsx
+++ b/frontend/packages/core/src/Input/text-field.tsx
@@ -50,6 +50,12 @@ const StyledTextField = styled(BaseTextField)({
 
   ".MuiInputBase-root.Mui-focused": {
     borderColor: "#3548d4",
+    ".MuiButtonBase-root": {
+      backgroundColor: "#3548D4",
+      "*": {
+        color: "#FFFFFF"
+      },
+    },
   },
 
   ".MuiInputBase-root.Mui-disabled": {
@@ -95,6 +101,18 @@ const IconButton = styled(MuiIconButton)({
   backgroundColor: "#E7E7EA",
   borderBottomRightRadius: "3px",
   borderTopRightRadius: "3px",
+  "&:hover": {
+    backgroundColor: "#2D3DB4",
+    "*": {
+      color: "#FFFFFF"
+    },
+  },
+  "&:active": {
+    backgroundColor: "#2938A5",
+    "*": {
+      color: "#FFFFFF",
+    },
+  },
 });
 
 export interface TextFieldProps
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -113,6 +119,7 @@ export interface TextFieldProps
     >,
     Pick<MuiInputProps, "readOnly"> {
   onReturn?: () => void;
+  endAdornment?: React.ReactElement | React.ReactElement[];
```
danielhochman, line 122:
```suggestion
    Pick<MuiInputProps, "readOnly" | "endAdornment"> {
  onReturn?: () => void;
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -91,6 +90,13 @@ const StyledTextField = styled(BaseTextField)({
   },
 });
 
+const IconButton = styled(MuiIconButton)({
+  borderRadius: "0",
+  backgroundColor: "#E7E7EA",
+  borderBottomRightRadius: "3px",
+  borderTopRightRadius: "3px",
+});
```
danielhochman, line 98: what about focus, hover, etc?

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -153,7 +161,10 @@ export const TextField = ({
       onBlur={onChange}
       error={error}
       helperText={helpText}
-      InputProps={{ readOnly }}
+      InputProps={{
+        readOnly,
+        endAdornment: endAdornment && <IconButton type="submit">{endAdornment}</IconButton>,
```
danielhochman, line 166: at first glance i wonder if we should take IconButton instead of wrapping it ourselves for further customization, this seems fine though.

frontend/packages/core/src/Input/text-field.tsx
dschaller: 👍  I was going to follow up with the implementation of icon only buttons but I can add those states here and extract later

frontend/packages/core/src/Input/text-field.tsx
dschaller: I had that thought too but I can't imagine a situation where we wouldn't want an icon button. Also, the styling applied is overriding some of the icon buttons styles which by default provide a border radius of 50% and have padding

danielhochman (approved):

PR #779, opened 2020-12-11
topology: search api protos

mcutalo88:
### Description
Paginated API Protos, I fully implemented a working solution with all options in #791 to workout some of the kinks and to make sure the structure was viable.

Here are all the fields for the reader convenience. 
```javascript
{
    "limit": 2,
    "page_token": 0,
    "sort": {
        "direction": "ASCENDING",
        "field": "column.id"
        // OR
        "field": "metadata.field1.field2"
    },
    "filter": {
        "search": {
            "field": "column.id"
            // OR
            "field": "metadata.field1.field2"
            "text": "i-00"
        },
        "typeUrl": "type.googleapis.com/clutch.aws.ec2.v1.Instance",
        "metadata": {
            "label": "value"
        }
    }
}
```

### Testing Performed

Full integration #791. 

### GitHub Issue

This work makes progress on #30


dschaller (approved):

api/topology/v1/topology_api.proto
```diff
@@ -31,6 +40,54 @@ message GetTopologyResponse {
...
+    }
+
+    Search search = 1;
+    string type_url = 2;
+    // Any valid metadata on the cache Resource object
+    map<string, string> metadata = 3;
+  }
+
+  Sort sort = 1;
+  string page_token = 2;
```
dschaller, line 81: Chatted IRL - only thing to follow up on is clarity around the default behavior of the page token here

api/topology/v1/topology_api.proto
```diff
@@ -31,6 +40,57 @@ message GetTopologyResponse {
...
+    // Any valid metadata on the cache Resource object
+    map<string, string> metadata = 3;
+  }
+
+  Sort sort = 1;
+  // Currently page_token specifies the page number you wish to request.
+  // The rational behind the naming is we might changes this to a cursor implentation
+  // in the future and did not want to break existing implementations of the API.
+  string page_token = 2;
+  int64 limit = 3;
```
danielhochman, line 85: do limit and page size need to be different?

also limit could be uint32

api/topology/v1/topology_api.proto
```diff
@@ -31,6 +40,57 @@ message GetTopologyResponse {
...
+
+    Search search = 1;
+    string type_url = 2;
+    // Any valid metadata on the cache Resource object
+    map<string, string> metadata = 3;
+  }
+
+  Sort sort = 1;
+  // Currently page_token specifies the page number you wish to request.
+  // The rational behind the naming is we might changes this to a cursor implentation
```
danielhochman, line 82:
```suggestion
  // The rationale behind the naming is we might changes this to a cursor implentation
```

api/topology/v1/topology_api.proto
danielhochman: for reference: https://cloud.google.com/apis/design/design_patterns#list_pagination

api/topology/v1/topology_api.proto
```diff
@@ -31,6 +40,57 @@ message GetTopologyResponse {
   repeated QueryResult results = 1;
 }
 
+message SearchTopologyRequest {
+  message FieldSelector {
+    enum Column {
+      UNSPECIFIED = 0;
+      ID = 1;
+    }
```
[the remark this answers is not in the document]
danielhochman: and maybe this could just be a general field mask for selector, so `metadata.level1.fieldname` or `id` at the top level rather than differentiating. just a thought, you probably have thought more of implications here.

api/topology/v1/topology_api.proto
```diff
@@ -31,6 +40,57 @@ message GetTopologyResponse {
...
+    oneof field {
+      Column column = 1;
+      // if metadata is nested specific the full path like so
+      // meatadata = 'level1.level2.level3.fieldname'
```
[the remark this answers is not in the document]
mcutalo88: lol

api/topology/v1/topology_api.proto
mcutalo88: Good question, I suppose they dont need to be it really depends on our table configuration when / if we look at the cursor implantation but I think we can just change it to ints for now.

api/topology/v1/topology_api.proto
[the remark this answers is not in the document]
mcutalo88: Ah yeah true i was mostly trying to work around the field mask selector in this case, an easy way to delineate columns vs metadata. I like the field mask approach, let me give that a try.

danielhochman (approved):
:mag:

PR #781, opened 2020-12-11
ui v2: restyle terminate instance

dschaller:
### Description

Restyle the terminate instance workflow.

This doesn't include all the final touches. This is just the major style changes. There are still a few minor things to fix along the way include:

- The accordion group has a top border that is rendering when it's a child
- The metadata keys on the confirmation page need to converted to either hard coded strings or normalized
- Fix breakpoints for responsive devices

terminstancev2

### Testing Performed

manually

### DIFF
--- a/frontend/packages/app/src/clutch.config.js
+++ b/frontend/packages/app/src/clutch.config.js
@@ -4,7 +4,12 @@ module.exports = {
       trending: true,
       componentProps: {
         resolverType: "clutch.aws.ec2.v1.Instance",
-        note: "The instance may take several minutes to shut down.",
+        notes: [
+          {
+            severity: "info",
+            text: "The instance may take several minutes to shut down.",
+          },
+        ],
       },
     },
     rebootInstance: {
--- a/frontend/packages/app/src/clutch.config.js
+++ b/frontend/packages/app/src/clutch.config.js
@@ -4,6 +4,7 @@ module.exports = {
       trending: true,
       componentProps: {
         resolverType: "clutch.aws.ec2.v1.Instance",
+        note: "The instance may take several minutes to shut down.",
       },
     },
     rebootInstance: {
--- a/frontend/packages/core/src/accordion.tsx
+++ b/frontend/packages/core/src/accordion.tsx
@@ -36,10 +36,6 @@ const StyledAccordion = styled(MuiAccordion)({
     padding: "8px",
     fontSize: "16px",
   },
-
-  ".MuiAccordionActions-root": {
-    padding: "0",
-  },
 });
 
 const AccordionSummaryBase = ({ children, collapsible, expanded, ...props }) => {
@@ -89,7 +85,7 @@ const StyledAccordionGroup = styled.div({
     marginBottom: "16px",
   },
 
-  ".MuiAccordion-root:before": {
+  "&.MuiAccordion-root:before": {
     display: "none",
   },
 });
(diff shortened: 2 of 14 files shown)


frontend/packages/core/src/accordion.tsx
```diff
@@ -89,7 +85,7 @@ const StyledAccordionGroup = styled.div({
     marginBottom: "16px",
   },
 
-  ".MuiAccordion-root:before": {
+  "&.MuiAccordion-root:before": {
```
danielhochman, line 88: not sure this is correct, do you remember why you made this change? also since this is a styled div the root class will never have the root class i think

originally the class was to remove a top divider element that accordiongroup children `[1:]` get

frontend/packages/app/src/clutch.config.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
       trending: true,
       componentProps: {
         resolverType: "clutch.aws.ec2.v1.Instance",
+        note: "The instance may take several minutes to shut down.",
```
danielhochman, line 7: why not array of notes into note panel? for consistency

the confirmation design in figma is missing the incorporation of severity which we will end up needing

frontend/workflows/ec2/src/terminate-instance.tsx
```diff
@@ -42,20 +43,28 @@ const InstanceDetails: React.FC<WizardChild> = () => {
...
     Object.keys(instance.tags).forEach(key => {
-      data.push({ name: key, value: instance.tags[key] });
+      metadata.push({ name: key, value: instance.tags[key] });
     });
   }
 
   return (
     <WizardStep error={resourceData.error} isLoading={resourceData.isLoading}>
       <MetadataTable data={data} />
+      <Accordion title="Metadata">
```
danielhochman, line 56: nit wdyt about
```suggestion
      <Accordion title="Tags">
```

danielhochman (commented):
looks awesome! a few minor nits/qs

scarlettperry:
Screen Shot 2020-12-14 at 2 51 13 PM

Screen Shot 2020-12-14 at 3 02 26 PM

can the error message be stretched to match the width of the parent?

frontend/workflows/ec2/src/terminate-instance.tsx
danielhochman: defer this comment for now

dschaller:
@scarlettperry yeah we have to revisit the error message css entirely since it appears to be fixed in max width :)

frontend/packages/app/src/clutch.config.js
dschaller: This was to match the designs in figma. I figured all the config notes would move to this interface in the future. I wasn't sure if going forward we would have these be a list or a single string to match figma

frontend/packages/core/src/accordion.tsx
dschaller: Ah I can drop this since I don't think it worked anyway. Right now we have an issue on the second step of the workflow where the before is still rendering if the accordion is a child. Need to figure out a fix for that but it can happen in a subsequent PR

frontend/packages/app/src/clutch.config.js
danielhochman: per discussion earlier, let's go back to a list w sev since the designer will be updating the designs to incorporate the possibility of multiple notes.

danielhochman (commented):
lgtm pending comment

danielhochman (approved):

PR #784, opened 2020-12-13
ui v2: fix searchfield bugs

scarlettperry:
## Description

PR fixes 2 bugs for the header searchfield component:

(1) The [result component](https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/AppLayout/search.tsx#L90) has the onClick event to navigate to a workflow. The issue was that its Grid did not match the parent height (see gif). So if you did not click exactly within the result Grid, you were not navigated to a workflow.

Solution: Have the result Grid inherit the height of its parent (autocomplete option container)

(2) If you chose a workflow by pressing the enter/return key, the header search bar would remain open. The expected behavior is for the searchbar to collapse back to the search icon when a result option is selected, like the mouse click behavior.

Solution: Add a `onKeyDown` event to update the `open` state

## Testing Performed
Local app, Storybook

### DIFF
--- a/frontend/packages/core/src/AppLayout/search.tsx
+++ b/frontend/packages/core/src/AppLayout/search.tsx
@@ -60,6 +60,7 @@ const InputField = styled(TextField)({
   },
 });
 
+// search's result options container
 const ResultGrid = styled(Grid)({
   height: "inherit",
 })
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/AppLayout/search.tsx
```diff
@@ -239,6 +245,13 @@ const SearchField: React.FC = () => {
     setOpen(false);
   };
 
+  function handleListKeyDown(event) {
```
danielhochman, line 248: add brief comment to explain rationale

frontend/packages/core/src/AppLayout/search.tsx
```diff
@@ -54,13 +54,18 @@ const InputField = styled(TextField)({
...
       },
     },
   },
 });
 
+// search's result options container
+const ResultGrid = styled(Grid)({
+  height: "inherit",
+  padding: "12px 16px 12px 16px",
+})
```
danielhochman, line 67: didn't check locally but this might need a lint (i think maybe missing a semicolon)

danielhochman (approved):
minor comments

PR #789, opened 2020-12-14
frontend: allow override of resolver api package

dschaller:
### Description

Allow workflows to specify optional Resolver prop to override the API proto used to resolve schema type lookups and response objects against.

### Testing Performed

Manual

### GitHub Issue

Fixes #

### TODOs

- [ ] Test against mock server

### DIFF
--- a/frontend/packages/core/src/Resolver/fetch.tsx
+++ b/frontend/packages/core/src/Resolver/fetch.tsx
@@ -4,13 +4,18 @@ import _ from "lodash";
 
 import { client, parseErrorMessage } from "../network";
 
-const fetchResourceSchemas = async (type: string): Promise<IClutch.resolver.v1.Schema[]> => {
+const fetchResourceSchemas = async (type: string, apiProto?: any): Promise<IClutch.resolver.v1.Schema[]> => {
   const response = await client.post("/v1/resolver/getObjectSchemas", {
     type_url: `type.googleapis.com/${type}`,
   });
-  return response.data.schemas.map((schema: object) =>
-    $pbclutch.clutch.resolver.v1.Schema.fromObject(schema)
-  );
+  return response.data.schemas.map((schema: object) => {
+    if (apiProto) {
+      try {
+        return apiProto.clutch.resolver.v1.Schema.fromObject(schema);
+      } catch {}
+    }
+    return $pbclutch.clutch.resolver.v1.Schema.fromObject(schema);
+  });
 };
 
 export interface ResolutionResults {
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Resolver/fetch.tsx
```diff
@@ -4,13 +4,21 @@ import _ from "lodash";
 
 import { client, parseErrorMessage } from "../network";
 
-const fetchResourceSchemas = async (type: string): Promise<IClutch.resolver.v1.Schema[]> => {
+const fetchResourceSchemas = async (
```
danielhochman, line 7: the Schema object will always come from core package, there is no need to override or try both. it's only in `resolveResource` that it's a problem.

rithujohn191:
tested this end to end with a custom workflow that passes in a custom apiProto. Works as expected 👍

frontend/packages/core/src/Resolver/fetch.tsx
```diff
@@ -55,12 +55,17 @@ const resolveResource = async (
     [key: string]: any;
   },
   onResolve: (resultObjects: any[], failureMessages: string[]) => void,
-  onError: (message: string) => void
+  onError: (message: string) => void,
+  apiProto?: any
```
rithujohn191, line 59: are we planning to later expand this to take in multiple apis?

danielhochman, line 59: also var name is confusing, it's not the proto (object) it's the package

frontend/packages/core/src/Resolver/fetch.tsx
```diff
@@ -55,12 +55,17 @@ const resolveResource = async (
...
 ) => {
   const resolver = fields?.query !== undefined ? resolveQuery : resolveFields;
   return resolver(type, limit, fields)
     .then(({ results, failures }) => {
-      const resultObjects = results.map(result => _.get($pbclutch, type).fromObject(result));
+      let pbClutch = _.get($pbclutch, type);
+      if (apiProto) {
+        pbClutch = _.get(apiProto, type);
+      }
```
danielhochman, line 67: we should try catch, attempt to resolve from the override but fall back to core

danielhochman (commented):
looks good, minor comments

frontend/packages/core/src/Resolver/fetch.tsx
danielhochman: eh maybe not, i'm okay with this actually

frontend/packages/core/src/Resolver/fetch.tsx
dschaller: We can modify this later when we come up with a more wholistic solution

frontend/packages/core/src/Resolver/fetch.tsx
```diff
@@ -55,12 +55,17 @@ const resolveResource = async (
...
   onResolve: (resultObjects: any[], failureMessages: string[]) => void,
-  onError: (message: string) => void
+  onError: (message: string) => void,
+  apiPackage?: any
 ) => {
   const resolver = fields?.query !== undefined ? resolveQuery : resolveFields;
   return resolver(type, limit, fields)
     .then(({ results, failures }) => {
-      const resultObjects = results.map(result => _.get($pbclutch, type).fromObject(result));
+      let pbClutch = _.get($pbclutch, type);
```
danielhochman, line 64: add comment

frontend/packages/core/src/Resolver/index.tsx
```diff
@@ -48,9 +48,16 @@ interface ResolverProps {
   searchLimit: number;
   onResolve: (data: { results: object[]; input: object }) => void;
   variant?: "dual" | "query" | "schema";
+  apiPackage?: any;
```
danielhochman, line 51: add comment, also is there a type we can provide to avoid the confusion of passing string of package name vs package obj itself?

danielhochman (approved):
lgtm with some additional comments for the future

rithujohn191 (approved):
LGTM. Do we want to add docs on how users can make use of this now or do we want to wait till we ship the global resolution approach?

dschaller:
@rithujohn191 I added a docstring to the type which should be enough for now

PR #794, opened 2020-12-16
ui v2: restyle resize workflows

scarlettperry:
### Description
PR
* adds style updates made in https://github.com/lyft/clutch/pull/781 to the resize workflows (EC2 & K8S)
* updates MutableRow to have disabled text field and icon
* updates EC2 term instance confirmation page to use `useDataLayout.displayValue`
* adds MutableRow story

Resize ASG
asg

Resize HPA
hpa

Screen Shot 2020-12-16 at 2 49 26 PM

### Testing Performed
Locally

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -67,7 +67,7 @@ const TableCell = styled(MuiTableCell)({
   fontSize: "14px",
   fontWeight: "normal",
   height: "48px",
-  padding: "0 16px",
+  padding: "15px 16px 15px 16px",
 });
 
 const KeyCellContainer = styled(TableCell)({
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -4,12 +4,14 @@ import styled from "@emotion/styled";
 import { DevTool } from "@hookform/devtools";
 import { yupResolver } from "@hookform/resolvers/yup";
 import {
+  Grid,
   Table as MuiTable,
   TableBody as MuiTableBody,
   TableCell as MuiTableCell,
   TableContainer as MuiTableContainer,
   TableRow,
 } from "@material-ui/core";
+import ChevronRightIcon from "@material-ui/icons/ChevronRight";
 import _ from "lodash";
 import type { Schema } from "yup";
 import { object } from "yup";
@@ -114,18 +116,22 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
     <TableRow key={data.id}>
       <KeyCell data={data} />
       <TableCell>
-        <TextField
-          id={data.id}
-          name={data.name}
-          defaultValue={data.value}
-          type={data?.input?.type}
-          onChange={onUpdate}
-          onReturn={onReturn}
-          onFocus={onUpdate}
-          inputRef={validation.register}
-          helperText={error?.message || ""}
-          error={!!error || false}
-        />
+        <Grid container alignItems="center" wrap="nowrap" spacing={2}>
+          <TextField disabled id={data.id} name={data.name} defaultValue={data.value} />
+          <ChevronRightIcon />
+          <TextField
+            id={data.id}
+            name={data.name}
+            defaultValue={data.value}
+            type={data?.input?.type}
+            onChange={onUpdate}
+            onReturn={onReturn}
+            onFocus={onUpdate}
+            inputRef={validation.register}
+            helperText={error?.message || ""}
+            error={!!error || false}
+          />
+        </Grid>
       </TableCell>
     </TableRow>
   );
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -4,7 +4,7 @@ import styled from "@emotion/styled";
 import { DevTool } from "@hookform/devtools";
 import { yupResolver } from "@hookform/resolvers/yup";
 import {
-  Grid,
+  Grid as MuiGrid,
   Table as MuiTable,
   TableBody as MuiTableBody,
   TableCell as MuiTableCell,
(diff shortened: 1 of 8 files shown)


frontend/workflows/ec2/src/resize-asg.tsx
```diff
@@ -87,26 +87,24 @@ const GroupDetails: React.FC<WizardChild> = () => {
...
-      <NotePanel notes={notes} />
+      <Confirmation action="Resize">{notes && formattedNotes}</Confirmation>
+      <MetadataTable
+        data={[
+          { name: "Name", value: group.name },
+          { name: "New Min Size", value: group.size.min },
+          { name: "New Max Size", value: group.size.max },
+          { name: "New Desired Size", value: group.size.desired },
+        ]}
+      />
```
scarlettperry, line 102: didn't add the accordion since we don't have metadata like annotations/labels for an [ASG](https://github.com/lyft/clutch/blob/main/api/aws/ec2/v1/ec2.proto#L67)

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -65,7 +67,7 @@ const TableCell = styled(MuiTableCell)({
   fontSize: "14px",
   fontWeight: "normal",
   height: "48px",
-  padding: "0 16px",
```
scarlettperry, line 68: on figma, the table cells have a padding top/bottom 15px, left/right 16px

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -65,7 +67,7 @@ const TableCell = styled(MuiTableCell)({
...
+  padding: "15px 16px 15px 16px",
```
scarlettperry, line 70: based on figma design for table cell

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -114,18 +116,22 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
...
-          type={data?.input?.type}
-          onChange={onUpdate}
-          onReturn={onReturn}
-          onFocus={onUpdate}
-          inputRef={validation.register}
-          helperText={error?.message || ""}
-          error={!!error || false}
-        />
+        <Grid container alignItems="center" wrap="nowrap" spacing={2}>
+          <TextField disabled id={data.id} name={data.name} defaultValue={data.value} />
```
scarlettperry, line 120: the text field widths are larger than figma design. i can override by giving them fix widths but unsure if that's best practice

dschaller: Let's override these for now to match the designs

frontend/packages/core/src/Table/metadata-table.tsx
dschaller: This one is hard because it depends on which cell you are looking at. Some of them have top & bottom padding but others don't. I think for now we shouldn't have any and just center the items vertically

frontend/workflows/ec2/src/resize-asg.tsx
```diff
@@ -68,10 +66,12 @@ const GroupDetails: React.FC<WizardChild> = () => {
         ]}
       />
       <ButtonGroup
+        justify="flex-end"
```
dschaller, line 69: ah I just landed a PR to change the button group interface. You'll probably need to pick that up :)

frontend/workflows/ec2/src/terminate-instance.tsx
```diff
@@ -79,15 +79,18 @@ const InstanceDetails: React.FC<WizardChild> = () => {
...
-  const confirmationData = Object.keys(configData).map(key => {
-    return { name: key, value: configData[key] };
-  });
-  const formattedNotes = notes.map(note => <div>{note.text}</div>)
+  const instance = useDataLayout("resourceData").displayValue();
+  const formattedNotes = notes?.map(note => <div>{note.text}</div>);
+
+  const data = [
+    { name: "Instance ID", value: instance.instanceId },
+    { name: "Region", value: instance.region },
```
dschaller, line 87: Thank you!

frontend/workflows/k8s/src/resize-hpa.tsx
```diff
@@ -65,11 +80,18 @@ const HPADetails: React.FC<WizardChild> = () => {
           { name: "Cluster", value: hpa.cluster },
         ]}
       />
+      <Accordion title="Metadata">
```
dschaller, line 83: Probably want to wrap this to only display it if we have metadata

frontend/workflows/ec2/src/resize-asg.tsx
```diff
@@ -87,26 +87,24 @@ const GroupDetails: React.FC<WizardChild> = () => {
...
-            { name: "New Max Size", value: group.size.max },
-            { name: "New Desired Size", value: group.size.desired },
-          ]}
-        />
-      </Paper>
-      <NotePanel notes={notes} />
+      <Confirmation action="Resize">{notes && formattedNotes}</Confirmation>
+      <MetadataTable
+        data={[
+          { name: "Name", value: group.name },
```
dschaller, line 97: Don't worry too much about this as part of this PR but I know @danielhochman mentioned adding the previous values into this screen at some point with Kanchan. We should probably follow up about that to make sure we are ok not showing those.

scarlettperry: ah okay sounds good! will follow up
I'll also leave a to-do for a reminder

frontend/packages/core/src/Table/metadata-table.tsx
scarlettperry: okay sounds good. i reverted to what we currently have (`padding: "0 16px"`) bc didn't look good w/o some right padding 

ex w/o the right padding
Screen Shot 2020-12-16 at 5 52 15 PM

scarlettperry:
addressed the feedback. this is how the mutable rows look now. for the disabled text field, I also had to remove the default padding b/c at the smaller width, you couldn't see the input value.

Screen Shot 2020-12-16 at 7 07 52 PM

frontend/workflows/k8s/src/resize-hpa.tsx
```diff
@@ -40,6 +41,20 @@ const HPADetails: React.FC<WizardChild> = () => {
     hpaData.updateData(key, value);
   };
 
+  const metadata = [];
+
+  if (hpa.annotations) {
+    Object.keys(hpa.annotations).forEach(key => {
```
danielhochman, line 47: nit: recommend lodash forEach

danielhochman (approved):
lgtm, one nit

frontend/workflows/k8s/src/resize-hpa.tsx
scarlettperry: @danielhochman , jst want to triple check this is what you had in mind https://github.com/lyft/clutch/pull/794/commits/6c86a0a9f48c588c9d4a20a05a790fc292603443.

could have also done `_.keys(something).forEach(key => {`

danielhochman: yep looks good!

PR #795, opened 2020-12-16
ui v2: restyle delete pod workflow

dschaller:
### Description

Restyle delete pod workflow

deletePodV2

### Testing Performed

Manual

### DIFF
--- a/frontend/packages/core/src/Resolver/input.tsx
+++ b/frontend/packages/core/src/Resolver/input.tsx
@@ -3,6 +3,7 @@ import type { UseFormMethods } from "react-hook-form";
 import { useForm } from "react-hook-form";
 import type { clutch } from "@clutch-sh/api";
 import SearchIcon from "@material-ui/icons/Search";
+import styled from "@emotion/styled";
 
 import {
   Accordion,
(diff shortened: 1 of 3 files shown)


danielhochman (approved):

frontend/packages/core/src/accordion.tsx
```diff
@@ -159,8 +159,13 @@ export const AccordionGroup = ({ children, defaultExpandedIdx }: AccordionGroupP
 };
 
 const StyledAccordionDetails = styled(MuiAccordionDetails)({
+  display: "flex",
+  flexWrap: "wrap",
+  boxSizing: "border-box",
   "> *": {
     padding: "8px 8px",
+    flex: "1 50%",
+
```
danielhochman, line 168: dont these belong in the resolver or form somehow? i don't think applying them to all accordiondetails makes sense

dschaller: Good point. I think probably moving to the resolver makes more sense.

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -55,6 +56,12 @@ interface SchemaResolverProps extends Pick<AccordionProps, "expanded" | "onClick
   submitHandler: any;
 }
 
+const SchemaDetails = styled(AccordionDetails)({
+  "> *": {
+    flex: "1 50%",
+  },
+});
```
danielhochman, line 63: last q, why this as opposed to nesting a <Grid> in AccordionDetails

danielhochman (approved):

frontend/packages/core/src/Resolver/input.tsx
dschaller: We would have to know which breakpoints to apply to each child based on the total number instead of letting flex do it for us.

PR #796, opened 2020-12-16
ui v2: resolver refactor and other polish

danielhochman:
### Description
Moving the QueryResolver < form > element into the child similar to the SchemaResolver.

Had to change some margin rules around. Need to add it back.

Also significant changes to stepper to fix alignment by recombining labels and icons.

Screenshot from 2020-12-17 15-40-16

### Testing Performed
dev

### TODO
- [ ] loadable paper
- [ ] fix extra width on first and last stepper element
- [ ] refactor error display

### DIFF
--- a/frontend/packages/core/src/error.tsx
+++ b/frontend/packages/core/src/error.tsx
@@ -41,8 +41,7 @@ const Error: React.FC<ErrorProps> = ({ message, onRetry }) => {
 };
 
 const Collapse = styled(MuiCollapse)`
-  margin-top: 10px;
-  width: 45%;
+  display: none;
 `;
 
 const ErrorText = styled(Typography)`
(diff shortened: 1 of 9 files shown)


frontend/packages/core/src/error.tsx
```diff
@@ -40,10 +40,7 @@ const Error: React.FC<ErrorProps> = ({ message, onRetry }) => {
   );
 };
 
-const Collapse = styled(MuiCollapse)`
-  margin-top: 10px;
-  width: 45%;
-`;
+const Collapse = styled(MuiCollapse)``;
```
scarlettperry, line 43: nit: can be removed & import updated

scarlettperry:
lgtm & tested locally

PR #797, opened 2020-12-16
ui v2: button group

dschaller:
### Description

Update button group to take `Button` as children instead of using a prop.

### Testing Performed

Manual and storybook

### DIFF
--- a/frontend/packages/core/src/button.tsx
+++ b/frontend/packages/core/src/button.tsx
@@ -100,15 +100,13 @@ const ButtonGroupContainer = styled(Grid)({
 });
 
 export interface ButtonGroupProps {
-  buttons: ButtonProps[];
+  children: React.ReactElement<ButtonProps> | React.ReactElement<ButtonProps>[];
   justify?: GridJustification;
 }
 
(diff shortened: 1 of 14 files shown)


frontend/packages/core/src/button.tsx
```diff
@@ -100,15 +100,13 @@ const ButtonGroupContainer = styled(Grid)({
...
   justify?: GridJustification;
 }
 
-const ButtonGroup: React.FC<ButtonGroupProps> = ({ buttons, justify = "center" }) => (
+const ButtonGroup = ({ children, justify = "center" }: ButtonGroupProps) => (
   <ButtonGroupContainer container justify={justify}>
-    {buttons.map(button => (
-      <Button key={button.text} {...button} />
-    ))}
+    {React.Children.map(children, child => child)}
```
danielhochman, line 109:
```suggestion
    {children}
```

danielhochman (approved):
pending comment

frontend/packages/core/src/button.tsx
dschaller: doh thanks!

PR #804, opened 2020-12-17
gateway: integrate accesslog middleware

rithujohn191:
### Description
This PR instantiates the accesslog middleware component in the gateway. It allows filtered logging based on status code filters provided in the backend config. By default it logs all gRPC status codes

https://github.com/lyft/clutch/pull/772

### Testing Performed
Local testing:

When status code filter = 0
```
2020-12-17T15:44:28.988-0800	INFO	accesslog/accesslog.go:83	gRPC	{"service": "clutch.resolver.v1.ResolverAPI", "method": "GetObjectSchemas", "statusCode": 0, "status": "OK"}
```

### TODOs
Add docs on use of accesslog filters


backend/gateway/gateway.go
```diff
@@ -134,11 +135,17 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
 		resolver.Registry[resolverCfg.Name] = res
 	}
 
+	accesslog, err := accesslog.New(cfg.Gateway.Accesslog, logger, scope)
```
danielhochman, line 138: let's wrap this is in
```go
if cfg.Gateway.Accesslog != nil
```
so it's not enabled by default logging everything

rithujohn191: Oh I thought we were going to default to logging everything.
So if a user wants to enable logging for all the codes they would have to list all the codes in the config?

rithujohn191: If that's the case then should we have a default config in place in the backend clutch-config? Something like log all 5xx and 4xx?

danielhochman: with nil check to enable access logging (and logging everything) they would pass an empty dict for the key,
`accesslog: {}`

default 5xx logging would also acceptable behavior

rithujohn191: Done :) and added default 5xx logging https://github.com/lyft/clutch/pull/804/files#diff-414d3d9d7c6a8dc5868d9121ddb15d30d16564a3a459c66f455129970e5c7ee1R5

backend/gateway/gateway.go
```diff
@@ -134,11 +135,21 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
 		resolver.Registry[resolverCfg.Name] = res
 	}
 
+	var interceptors []grpc.UnaryServerInterceptor
+	if cfg.Gateway.Accesslog != nil {
+		accesslog, err := accesslog.New(cfg.Gateway.Accesslog, logger, scope)
+		if err != nil {
+			logger.Fatal("could not create accesslog config", zap.Error(err))
```
danielhochman, line 142:
```suggestion
			logger.Fatal("could not create accesslog interceptor", zap.Error(err))
```

danielhochman (approved):

PR #805, opened 2020-12-18
rtds: add support for configuring resource TTL

snowp:
### Description
Adds support for configuring a TTL and heartbeat interval for RTDS resources. This should allow Clutch to more safely set RTDS values without worrying that a Clutch outage will prolong an experiment: when TTLs are configured, if Clutch becomes
unavailable the RTDS layers will revert themselves after the TTL as expired.

### Testing Performed
Integration testing to verify that go-control-plane is configured correctly

### GitHub Issue

Fixes #


kathan24 (approved):
lgtm

snowp:
@lyft/clutch-admin mind giving this a look? I think it requires admin approval for the dependency bump

mcutalo88 (approved):
Approving the go mod changes.

PR #808, opened 2020-12-18
ui v2: right align button groups by default

dschaller:
### Description

Right justify all button groups on details pages.

### DIFF
--- a/frontend/packages/core/src/button.tsx
+++ b/frontend/packages/core/src/button.tsx
@@ -104,7 +104,7 @@ export interface ButtonGroupProps {
   justify?: GridJustification;
 }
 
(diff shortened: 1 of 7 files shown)


danielhochman:
can we make this the default for ButtonGroup?

dschaller:
We can but I'd think that by default a button group would grow from the left or center.

danielhochman:
in clutch i'm pretty sure all buttons are right aligned though, aren't they?

another way to do it would be to apply the flex property from the step to any children with the button groups class

dschaller:
I think for the most part yes. I guess the one place I'm thinking of it the experimentation code that might have centered buttons at the moment but we can shift these to default to right since it'll match the workflows. Will change the default

danielhochman (approved):

PR #809, opened 2020-12-18
ui v2: fix header searchfield bug

scarlettperry:
### Description

The error was coming from here:
https://github.com/lyft/clutch/blob/UIV2/frontend/packages/core/src/AppLayout/search.tsx#L136

Since we toggle between rendering the search icon or search bar, we need to add a null check for `searchRef.current`. Otherwise, once the search bar is collapsed, the `handleKeyPress` will still allow the `\` to focus the input field which isn't rendered.

Open to other suggestions for fixing the bug! 

Steps to reproduce bug on ui v2 master:
* open searchbar
* click into it
* close out of the searchbar
* press `/` key

Screen Shot 2020-12-18 at 5 11 57 PM


scarlettperry:
sorry for the confusing branch name. still thinking about the sidebar PR 😂

mcutalo88 (approved):
Choice!

PR #810, opened 2020-12-18
ui v2: tabs

dschaller:
### Description

Introduce tab & tabs components

tabs

### Testing Performed

Storybook

### DIFF
--- a/frontend/packages/core/src/tab.tsx
+++ b/frontend/packages/core/src/tab.tsx
@@ -31,7 +31,7 @@ const StyledTab = styled(MuiTab)({
   },
   "&:active": {
     color: "rgba(13, 16, 48, 0.6)",
-    backgroundColor: "rgba(219, 219, 224, 1)",
+    backgroundColor: "#DBDBE0",
   },
 });
 
--- a/frontend/packages/core/src/tab.tsx
+++ b/frontend/packages/core/src/tab.tsx
@@ -1,0 +1,69 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { TabProps as MuiTabProps, TabsProps as MuiTabsProps } from "@material-ui/core";
+import { Tab as MuiTab, Tabs as MuiTabs } from "@material-ui/core";
+
+const StyledTab = styled(MuiTab)({
+  minWidth: "111px",
+  height: "46px",
+  padding: "0",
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/tab.tsx
```diff
@@ -0,0 +1,69 @@
...
+import styled from "@emotion/styled";
+import type { TabProps as MuiTabProps, TabsProps as MuiTabsProps } from "@material-ui/core";
+import { Tab as MuiTab, Tabs as MuiTabs } from "@material-ui/core";
+
+const StyledTab = styled(MuiTab)({
+  minWidth: "111px",
+  height: "46px",
+  padding: "0",
+  color: "rgba(13, 16, 48, 0.6)",
+  borderBottom: "3px solid #E7E7EA",
```
scarlettperry, line 11: it looks a little thicker on figma. i think it should be 4px
```suggestion
  borderBottom: "4px solid #E7E7EA",
```

frontend/packages/core/src/tab.tsx
```diff
@@ -0,0 +1,69 @@
...
+    color: "#3548D4",
+    backgroundColor: "#EBEDFB",
+  },
+  "&:focus-within": {
+    color: "#3548D4",
+    backgroundColor: "#EBEDFB",
+  },
+  "&:active": {
+    color: "rgba(13, 16, 48, 0.6)",
+    backgroundColor: "rgba(219, 219, 224, 1)",
```
scarlettperry, line 34:
```suggestion
    backgroundColor: "#DBDBE0",
```

scarlettperry:
for the single tab story, noticed that the selected tab doesn't have the bottom blue border

frontend/packages/core/src/tab.tsx
scarlettperry: actually nvm, it's correct. on figma it's only the active tab that has the thicker border

frontend/packages/core/src/tab.tsx
dschaller: This is the rgb value :)

dschaller:
yeah, the marker is only available on tabs. We can force the bottom border on selected but I think it doesn't make sense to since we already have it from the marker when activated.

PR #813, opened 2020-12-21
ui v2: trim long group headings

dschaller:
### Description

Implement stop gap solution to prevent text overflow on drawer. Following up on longer term solution with design [here](https://lyftoss.slack.com/archives/C015UJ6LED9/p1608581951140400)

**Before**
Screen Shot 2020-12-21 at 12 29 04 PM

**After**
Screen Shot 2020-12-21 at 12 28 52 PM

### Testing Performed

manual
### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/AppLayout/drawer.tsx
+++ b/frontend/packages/core/src/AppLayout/drawer.tsx
@@ -151,6 +151,12 @@ const Group: React.FC<GroupProps> = ({
   if (React.Children.count(children) === 0) {
     return null;
   }
+  // TODO (dschaller): revisit how we handle long groups once we have designs.
+  // n.b. this is a stop-gap solution to prevent long groups from looking unreadable.
+  let formattedHeading = heading;
+  if (heading.length > 11) {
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/AppLayout/drawer.tsx
```diff
@@ -151,6 +151,12 @@ const Group: React.FC<GroupProps> = ({
   if (React.Children.count(children) === 0) {
     return null;
   }
+  // TODO (dschaller): revisit how we handle long groups once we have designs.
+  // n.b. this is a stop-gap solution to prevent long groups from looking unreadable.
+  let formattedHeading = heading;
+  if (heading.length > 11) {
+    formattedHeading = `${heading.substring(0, 10)}...`;
+  }
```
lyft-tla, line 159: Nit: This could probably be a ternary, but this is very readable too.

dschaller: 👍  I figured making it explicit would call attention to it for refactoring later :)

PR #842, opened 2020-12-31
ui v2: add alert component

dschaller:
### Description

Add alert component.

Once this lands we can remove/update the notes and error components.

### Testing Performed

manual and storybook


danielhochman:
icon and content not vertically aligned. i think using line height would work in this case to get the first line of text or title and the icon to line up.

Screenshot from 2020-12-31 12-00-13

danielhochman (approved):
minor alignment issue but can be fixed later even

dschaller:
👍  when you say they don't align vertically do you mean the text is slightly offset? They don't align in the designs either but we can update both. I know those designs are still in progress.

PR #856, opened 2021-01-06
ui v2: change textfield icon button color

scarlettperry:
### Description
Address bug bash action item
* updates the text field icon button to be blue

Screen Shot 2021-01-05 at 11 51 35 PM

### Testing Performed
Locally

### DIFF
--- a/frontend/packages/core/src/Input/text-field.tsx
+++ b/frontend/packages/core/src/Input/text-field.tsx
@@ -51,12 +51,6 @@ const StyledTextField = styled(BaseTextField)({
 
   ".MuiInputBase-root.Mui-focused": {
     borderColor: "#3548d4",
-    ".MuiButtonBase-root": {
-      backgroundColor: "#3548D4",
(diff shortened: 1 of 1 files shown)


rithujohn191 (approved):

dschaller:
This might just be me but it looks a bit odd with the border color against the blue button being inset

scarlettperry:
> This might just be me but it looks a bit odd with the border color against the blue button being inset

hmm yea I see that odd illusion with the gray input border. To make it look better, I played with adding a blue border around 
 the button to see if that helped, but that didn't work.

dschaller:
We could look at making the border a gradient which could help but that might also be odd. We could also just extract the search icon for now or maybe invert the colors such that the button is white and the icon is blue

scarlettperry:
we're going to keep these changes for now
slack sync: https://lyftoss.slack.com/archives/C015UJ6LED9/p1610127793002100

PR #858, opened 2021-01-06
resolver: use common response object

danielhochman:
### Description
Create a common response object so we get common error handling, truncation of extra results, etc.

Before there was a bug where Search properly removed partial failures with a limit met but Resolve did not.

Moving the code to a common object will ensure both calls are handled the same way.

### Testing Performed
Unit, manual.

### TODOs
- [x] Unit tests

### DIFF
--- a/backend/module/resolver/response.go
+++ b/backend/module/resolver/response.go
@@ -1,0 +1,76 @@
+package resolver
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/golang/protobuf/ptypes"
+	"github.com/golang/protobuf/ptypes/any"
+	statuspb "google.golang.org/genproto/googleapis/rpc/status"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/lyft/clutch/backend/resolver"
+)
+
+func newResponse() *response {
+	return &response{
+		Results:         []*any.Any{},
+		PartialFailures: []*statuspb.Status{},
+	}
+}
+
+// Generic object to handle common operations for SearchResponse and ResolveResponse.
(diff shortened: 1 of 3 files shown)


backend/module/resolver/response.go
```diff
@@ -0,0 +1,76 @@
...
+
+	for _, failure := range results.PartialFailures {
+		r.PartialFailures = append(r.PartialFailures, failure.Proto())
+	}
+
+	return nil
+}
+
+func (r *response) isError(wanted string, searchedSchemas []string) error {
+	if len(r.Results) > 0 {
```
rithujohn191, line 63: Is there ever a case where len(r.Results) > 0 and len(r.PartialFailures) > 0 or do the methods strictly not return any `Result` if there is a `PartialFailure`?

danielhochman: yes, that case is addressed by the truncate code which runs before isError. if the limit is met the errors are discarded, if the limit is not met the errors are preserved. in that case isError is false, and the API will return the results and partial failures together.

rithujohn191 (approved):

backend/module/resolver/response_test.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func TestIsError(t *testing.T) {
+	wanted := "type.googleapis.com/clutch.foo"
+	schemas := []string{"foo", "bar"}
+
+	resp := newResponse()
+
+	// No results, no failures = NotFound.
+	{
```
mcutalo88, line 77: Didn't know you could short hand a closure this way.

mcutalo88 (approved):
Looks good, nice tests.

PR #861, opened 2021-01-06
ui v2: add titles to workflows details tables

dschaller:
### Description

Add titles to the metadata tables for ec2 and k8s workflows.

### Testing Performed

manual


scarlettperry (approved):

scarlettperry:
I think it might be a good idea to create a Typography core component and we have designs for it. I can follow up on that

dschaller:
Yeah I think that makes sense. I don't think we have a common heading at the moment but we should get one from design and add it.

PR #863, opened 2021-01-07
ui V2: fix accordion overflow

scarlettperry:
### Description
Fixes issue when accordion overflows its parent container, which happens when there are long strings of text that do not have an acceptable break (ie space, hyphen).

before
Screen Shot 2021-01-07 at 4 58 06 PM

Screen Shot 2021-01-07 at 4 58 30 PM

now
Screen Shot 2021-01-07 at 4 59 08 PM

### Testing Performed
Locally and checked stories to make sure it didn't break other functionality/designs

### DIFF
--- a/frontend/packages/core/src/accordion.tsx
+++ b/frontend/packages/core/src/accordion.tsx
@@ -17,6 +17,7 @@ const StyledAccordion = styled(MuiAccordion)({
   boxShadow: "none",
   border: "1px solid transparent",
   boxSizing: "content-box",
+  maxWidth: "100%",
 
   "&.Mui-expanded": {
     boxShadow: "0px 4px 6px rgba(53, 72, 212, 0.2)",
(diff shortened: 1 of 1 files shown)


danielhochman (approved):

frontend/packages/core/src/accordion.tsx
```diff
@@ -17,6 +17,8 @@ const StyledAccordion = styled(MuiAccordion)({
   boxShadow: "none",
   border: "1px solid transparent",
   boxSizing: "content-box",
+  maxWidth: "100%",
+  overflowWrap: "anywhere",
```
scarlettperry, line 21: breaks the string at an arbitrary point or where there is a natural break. otherwise, the cells in the row won't share the space evenly. example

Screen Shot 2021-01-07 at 5 13 42 PM

scarlettperry:
@danielhochman it's ready for another look

danielhochman (approved):
great!

PR #864, opened 2021-01-07
ui v2: use new alert for note panel and fix layout max width

dschaller:
### Description

Update Note and NotePanel to use new alert component. Fix max width of alerts and overflow.

Screen Shot 2021-01-07 at 12 03 35 PM

### Testing Performed

manual


danielhochman (approved):
the spacing on the second example is pretty rough the icon should probably float in the margin rather than affecting the centering of content

dschaller:
👍  that should be controlled in the content being passed in (in this case it's in a private workflow)

PR #868, opened 2021-01-08
ui v2: reduce button group margin

dschaller:
### Description

Reduce button group margin

Screen Shot 2021-01-08 at 10 40 14 AM
Screen Shot 2021-01-08 at 10 39 25 AM

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/button.tsx
+++ b/frontend/packages/core/src/button.tsx
@@ -100,11 +100,11 @@ const ButtonGroupContainer = styled(Grid)(
   props =>
     props["data-border"] === "bottom"
       ? {
-          marginBottom: "24px",
+          marginBottom: "12px",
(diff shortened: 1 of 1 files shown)


danielhochman:
ideally i think these 3 margins would be equal?

image

dschaller:
I think they are intentionally not equal in figma. We can make the margin between the border and the content above/below equal to the padding for the buttons but only in context of the button group. E.g. there will still be additional spacing due to the parent and sibling styled. For instance the screenshot you provided has padding applied from the parent which is why the bottom has so much space.

danielhochman:
makes sense, the biggest issue was 2 being too small but in your latest screenshot it looks better

danielhochman (approved):

PR #870, opened 2021-01-09
ui v2: warning invoke onClose before exiting

dschaller:
### Description

invoke onClose before the snackbar exits. This allows for the wizard to reset warnings after they have been dismissed.

### Testing Performed

manaul


danielhochman:
can you give a bit more detail on the use case here? seems like the data flow for a toast/snackbar should be one way

dschaller:
Currently we maintain a list of global warnings that are shown as toasts on the app. To avoid duplicate toasts being presented at once we ensure that these warnings are unique by only inserting new ones. However, once the toast has cleared from the screen we need a way to signal to the parent that it can send the same warning again.

danielhochman (approved):

PR #893, opened 2021-01-12
topology: paginated api

mcutalo88:
### Description

Adds the paginated API implantation to search against the `topology_cache` table. This is foundational work that will power many up coming features such as auto complete.

### Testing Performed

Ive done a lot of local testing as well as Unit, there are many permutations and I've tried to capture them all in unit tests.

### GitHub Issue

This makes progress on https://github.com/lyft/clutch/issues/30.


backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,115 @@
...
+	metadata = "metadata."
+)
+
+func paginatedQueryBuilder(
+	filter *topologyv1.SearchTopologyRequest_Filter,
+	sort *topologyv1.SearchTopologyRequest_Sort,
+	pageToken,
+	limit int,
+) (sq.SelectBuilder, error) {
+	queryLimit := 100
```
rithujohn191, line 23: Define this as a global const value?

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,115 @@
...
+	return metadataQuery
+}
+
+func getDirection(direction string) string {
+	switch direction {
+	case topologyv1.SearchTopologyRequest_Sort_ASCENDING.String():
+		return "ASC"
+	case topologyv1.SearchTopologyRequest_Sort_DESCENDING.String():
+		return "DESC"
+	case topologyv1.SearchTopologyRequest_Sort_UNSPECIFIED.String():
```
rithujohn191, line 108: nit: since we default to "ASC" in the next line, we can get rid of the `UNSPECIFIED` case all together

backend/service/topology/querybuilder_test.go
```diff
@@ -0,0 +1,233 @@
...
+			expect:    "SELECT id, data, metadata FROM topology_cache ORDER BY ID ASC LIMIT 100 OFFSET 0",
+		},
+		{
+			id:        "Change PageToken and Limits",
+			filter:    &topologyv1.SearchTopologyRequest_Filter{},
+			sort:      &topologyv1.SearchTopologyRequest_Sort{},
+			pageToken: 10,
+			limit:     5,
+			expect:    "SELECT id, data, metadata FROM topology_cache ORDER BY ID ASC LIMIT 5 OFFSET 50",
+		},
```
rithujohn191, line 36: Should we also add a case for pageToken: 0 & limit !=0?

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,115 @@
...
+	return query, nil
+}
+
+func filterQueryBuilder(query sq.SelectBuilder, f *topologyv1.SearchTopologyRequest_Filter) sq.SelectBuilder {
+	if f.Search != nil && len(f.Search.Field) > 0 {
+		if strings.HasPrefix(f.Search.Field, column) {
+			query = query.Where(sq.Like{strings.TrimPrefix(f.Search.Field, column): fmt.Sprintf("%%%s%%", f.Search.Text)})
+		} else if strings.HasPrefix(f.Search.Field, metadata) {
+			mdQuery := convertMetadataToQuery(strings.TrimPrefix(f.Search.Field, metadata))
+			query = query.Where(sq.Like{mdQuery: fmt.Sprintf("%%%s%%", f.Search.Text)})
```
lyft-tla, line 51: What do you think about using [ilike](https://sourcegraph.com/github.com/Masterminds/squirrel/-/blob/expr.go#L272) for insensitive searching?

lyft-tla: Also, I think you got an extra percent in `%%%s%%`

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,115 @@
...
+
+func paginatedQueryBuilder(
+	filter *topologyv1.SearchTopologyRequest_Filter,
+	sort *topologyv1.SearchTopologyRequest_Sort,
+	pageToken,
+	limit int,
+) (sq.SelectBuilder, error) {
+	queryLimit := 100
+	if limit > 0 {
+		queryLimit = limit
```
lyft-tla, line 25: What do you think about having an upper bound for the limit?

backend/service/topology/querybuilder.go
mcutalo88: I could an option to toggle case sensitive search, so the caller can specify the exact behavior. Since that requires a proto change I can do that in a follow up PR.

No extra precent here, just escaping for fmt.
https://golang.org/pkg/fmt/#hdr-Printing
> %%	a literal percent sign; consumes no value

backend/service/topology/querybuilder_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+func TestSortQueryBuilder(t *testing.T) {
+	testCases := []struct {
+		id     string
+		input  *topologyv1.SearchTopologyRequest_Sort
+		expect string
+	}{
+		{
+			id:     "Default to ID ASC",
+			input:  &topologyv1.SearchTopologyRequest_Sort{},
```
rithujohn191, line 136: if only Direction is specified and not the field, is the field expected to default to ID? or is that not a valid input?

mcutalo88: currently its a required field https://github.com/lyft/clutch/blob/main/api/topology/v1/topology_api.proto#L56

backend/service/topology/querybuilder.go
lyft-tla: Probably not necessary to support it immediately.

Yeah, my mistake with the %%%

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,115 @@
...
+	query = filterQueryBuilder(query, filter)
+	query = sortQueryBuilder(query, sort)
+
+	return query, nil
+}
+
+func filterQueryBuilder(query sq.SelectBuilder, f *topologyv1.SearchTopologyRequest_Filter) sq.SelectBuilder {
+	if f.Search != nil && len(f.Search.Field) > 0 {
+		if strings.HasPrefix(f.Search.Field, column) {
+			query = query.Where(sq.Like{strings.TrimPrefix(f.Search.Field, column): fmt.Sprintf("%%%s%%", f.Search.Text)})
```
danielhochman, line 48: i'm fairly certain you cannot trust user input on the left side of the these clauses without the risk of injection. only the right hand side is converted to an argument which is safe from injection.

backend/service/topology/querybuilder.go
lyft-tla: @rithujohn191, why do you think this should be global? I might be missing it but I don't think queryLimit is used outside this main func.

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,115 @@
...
+
+const (
+	column   = "column."
+	metadata = "metadata."
+)
+
+func paginatedQueryBuilder(
+	filter *topologyv1.SearchTopologyRequest_Filter,
+	sort *topologyv1.SearchTopologyRequest_Sort,
+	pageToken,
```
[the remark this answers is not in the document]
mcutalo88: In the proto PR for this you had suggested it be an int? Which is when i changed it from a string -> int.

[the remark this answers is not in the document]
danielhochman: suggested limit be uint, not page_token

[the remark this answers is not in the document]
mcutalo88: https://github.com/lyft/clutch/pull/779#discussion_r556047961

backend/service/topology/querybuilder.go
mcutalo88: Ill leave a todo and follow up.

backend/service/topology/querybuilder.go
rithujohn191: I suggested it so it could be used in the test file as well and its usually considered best practice to declare default values as consts so they can be reused

backend/service/topology/querybuilder.go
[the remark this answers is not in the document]
danielhochman: sorry for the confusion. i think i may have started my review on a version with both a page_size and limit variable that was force pushed over? anyways it's not important. the page token should be a string so that you can encode arbitrary data in it if necessary in the future. per google's page it also allows backwards/forwards compatibility. "Page token contents should be a url-safe base64 encoded protocol buffer. This allows the contents to evolve without compatibility issues."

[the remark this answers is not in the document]
mcutalo88: It's fine I changed it back, no big deal.

backend/service/topology/querybuilder.go
mcutalo88: Let me do some more testing and add some sql injection tests as well.

backend/service/topology/querybuilder.go
mcutalo88: Yeah totally sounds good, i'll say 1000 for now?

backend/service/topology/querybuilder.go
danielhochman: did some poking around and i think for JSONB columns you can use the contains operator to get around this https://stackoverflow.com/a/62895355. though on second look i don't think that will work the LIKE % syntax.

different use case and methodology https://stackoverflow.com/a/45849743 but i think it's probably going to look something like that using the JSON path language and passing a expression built from user input as an argument.

that led me to some really good details here: https://blog.sql-workbench.eu/post/json-path/, especially under `More complex queries`. it does mention some impact on the query plan and perf implications but i'm not sure it would matter even with row counts on the order of thousands where the entire set probably fits in the DB's memory.

maybe even another option is to only allow alphanumeric values as left-hand side input, which i think are impossible to perform injection attacks with. or some other form of escaping. more research and tinkering required.

backend/service/topology/querybuilder.go
danielhochman: fmt.Sprint is a little cleaner and easier to read maybe, `fmt.Sprint("%", f.Search.Text, "%"). doesn't really matter if there are tests though.

api/topology/v1/topology_api.proto
```diff
@@ -23,7 +23,7 @@ service TopologyAPI {
     option (clutch.api.v1.action).type = READ;
   }
 
-  rpc SearchTopology(SearchTopologyRequest) returns (SearchTopologyResponse) {
+  rpc Search(SearchTopologyRequest) returns (SearchTopologyResponse) {
```
danielhochman, line 26: nit: SearchRequest, SearchResponse

backend/service/topology/querybuilder.go
mcutalo88: Nice some of the metadata queries would be much easier to construct with some of this, I'm going to incorporate some of this. 

Ive also found a way to parameterize all of these user inputs with squirrel by using [Expressions](https://godoc.org/github.com/Masterminds/squirrel#Expr).

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,141 @@
...
+		query = sortQueryBuilder(query, sort)
+	}
+
+	return query, nil
+}
+
+func filterQueryBuilder(query sq.SelectBuilder, f *topologyv1.SearchRequest_Filter) (sq.SelectBuilder, error) {
+	if f == nil {
+		return query, nil
+	}
```
scarlettperry, line 68: nit: don't need to do this check again since it's done on line 51

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if f.Search != nil && len(f.Search.Field) > 0 {
```
scarlettperry, line 70: is this check needed? seems like Search will always have data b/c of the validation rules https://github.com/lyft/clutch/blob/main/api/topology/v1/topology_api.proto#L60. but i could be misreading / this is best-practice

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,141 @@
...
+		query = query.Where(sq.Expr("metadata @> ?::jsonb", metadataJson))
+	}
+
+	return query, nil
+}
+
+func sortQueryBuilder(query sq.SelectBuilder, s *topologyv1.SearchRequest_Sort) sq.SelectBuilder {
+	if s == nil {
+		return query
+	}
```
scarlettperry, line 97: similar comment as https://github.com/lyft/clutch/pull/893/files?file-filters%5B%5D=.go&file-filters%5B%5D=.js&file-filters%5B%5D=.proto&file-filters%5B%5D=.ts#r558411837

backend/service/topology/querybuilder.go
mcutalo88: Yeah the proto only generates validation for the gateway, but for unit level testing i'll still need to check again. Which is a bit redundant but necessary IMO to validate the function does not panic with malformed / missing data.

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,141 @@
...
+
+	if s.Direction > topologyv1.SearchRequest_Sort_UNSPECIFIED && len(s.Field) > 0 {
```
danielhochman, line 99: `if s.Direction > topologyv1.SearchRequest_Sort_UNSPECIFIED`
i thought this would default to ASC anyways? otherwise this validation belongs in the proto

backend/service/topology/querybuilder_test.go
```diff
@@ -0,0 +1,243 @@
...
+		sql, _, err := output.ToSql()
+		assert.NoError(t, err)
+		assert.Equal(t, test.expect, sql)
+	}
+}
+
+func TestConvertMetadataToQuery(t *testing.T) {
+	testCases := []struct {
+		id     string
+		input  string
```
danielhochman, line 190: add test for empty string input? i think it can happen if someone passes `column.` as field.

backend/service/topology/topology.go
```diff
@@ -93,6 +95,58 @@ func (c *client) GetTopology(ctx context.Context) error {
...
+	defer rows.Close()
+
+	results := []*topologyv1.Resource{}
+	for rows.Next() {
+		var id string
+		var data []byte
+		var metadata []byte
+
+		if err := rows.Scan(&id, &data, &metadata); err != nil {
+			c.log.Error("Error scaning row", zap.Error(err))
```
danielhochman, line 122:
```suggestion
			c.log.Error("Error scanning row", zap.Error(err))
```

danielhochman (commented):
nice, getting there

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,141 @@
...
+		direction := getDirection(s.Direction)
+
+		if strings.HasPrefix(s.Field, column) {
+			query = query.OrderByClause(fmt.Sprintf("? %s", direction), strings.TrimPrefix(s.Field, column))
+		} else if strings.HasPrefix(s.Field, metadata) {
+			mdQuery := convertMetadataToQuery(strings.TrimPrefix(s.Field, metadata))
+			query = query.OrderBy(fmt.Sprintf("%s %s", mdQuery, direction))
```
danielhochman, line 106: pretty sure this one is still vulnerable to injection

backend/clutch-config.yaml
```diff
@@ -34,7 +34,18 @@ modules:
...
+  - name: clutch.service.db.postgres
+    typed_config:
+      "@type": types.google.com/clutch.config.service.db.postgres.v1.Config
+      connection:
+        host: 0.0.0.0
+        port: 5432
+        user: clutch
+        ssl_mode: DISABLE
+        dbname: clutch
+        password: clutch
```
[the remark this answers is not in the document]
mcutalo88: Yes of course I shall, just easier to iterate.

backend/service/topology/querybuilder.go
mcutalo88: Yeah looks like i missed this one, ty.

danielhochman: did a quick proof-of-concept

```go
sb := sq.StatementBuilder.PlaceholderFormat(sq.Dollar).Select("*").From("topology_cache")
mdQuery := convertMetadataToQuery(strings.TrimPrefix("column.bar' ASC; DROP TABLE users; SELECT * ORDER BY 'baz", "column."))
sb = sb.OrderBy(fmt.Sprintf("%s %s", mdQuery, "ASC"))
sql, _, _ := sb.ToSql()
fmt.Println(sql)
```

`>>> SELECT * FROM topology_cache ORDER BY metadata->>'bar' ASC; DROP TABLE users; SELECT * ORDER BY 'baz' ASC`

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,141 @@
...
+	if f == nil {
+		return query, nil
+	}
+
+	if f.Search != nil && len(f.Search.Field) > 0 {
+		if strings.HasPrefix(f.Search.Field, column) {
+			query = query.Where(sq.Expr("quote_literal(?) LIKE ?", sq.Expr(strings.TrimPrefix(f.Search.Field, column)), fmt.Sprintf("%%%s%%", f.Search.Text)))
+		} else if strings.HasPrefix(f.Search.Field, metadata) {
+			mdQuery := convertMetadataToQuery(strings.TrimPrefix(f.Search.Field, metadata))
+			query = query.Where(sq.Expr("quote_literal(?) LIKE ?", sq.Expr(mdQuery), fmt.Sprintf("%%%s%%", f.Search.Text)))
```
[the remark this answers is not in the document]
mcutalo88: Let me DRY it up a bit.

Yeah I can check for that, I also wanted to support filter / searching on the PB directly as its a JSONB col anyway. Let me spruce this up.

backend/service/topology/querybuilder.go
mcutalo88: Yeah I did this last night with many of the strings inputs, but was unsuccessful in actually performing the drop command. Nevertheless will fix this one up.

backend/service/topology/querybuilder.go
mcutalo88: Ah yeah i can delete this

mcutalo88:
Thanks for the reviews yall, I'm ready for another / final review. Thanks.

backend/service/topology/querybuilder_test.go
```diff
@@ -0,0 +1,337 @@
...
+					Field: "metadata.label",
+					Text:  "cat",
+				},
+				TypeUrl: "type.googleapis.com/clutch.aws.ec2.v1.AutoscalingGroup",
+				Metadata: map[string]string{
+					"label":  "value",
+					"label2": "value2",
+				},
+			},
+			expect: "SELECT * FROM topology_cache WHERE quote_literal(metadata->>'label') LIKE $1 AND resolver_type_url = $2 AND metadata @> $3::jsonb",
```
danielhochman, line 161: the where clause still appears to be vulnerable to injection, it doesn't look like Expr prevents it

backend/service/topology/topology.go
```diff
@@ -93,6 +95,58 @@ func (c *client) GetTopology(ctx context.Context) error {
 	return nil
 }
 
-func (c *client) SearchTopology() error {
-	return nil
+func (c *client) Search(ctx context.Context, req *topologyv1.SearchRequest) ([]*topologyv1.Resource, uint64, error) {
```
danielhochman, line 98: nit: change return type to string of nextPageToken

danielhochman (commented):
lgtm with the exception of some nits and one last possible injection vuln

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,181 @@
...
+func filterQueryBuilder(query sq.SelectBuilder, f *topologyv1.SearchRequest_Filter) (sq.SelectBuilder, error) {
+	if f.Search != nil && len(f.Search.Field) > 0 {
+		searchIdentiferExpr := sq.Expr("id")
+		identifer, err := getFilterSortPrefixIdentifer(f.Search.Field)
+		if err != nil {
+			return sq.SelectBuilder{}, err
+		}
+
+		if identifer == column {
+			searchIdentiferExpr = sq.Expr(strings.TrimPrefix(f.Search.Field, columnIdentifer))
```
danielhochman, line 86: this path is still injectable
```go
input: &topologyv1.SearchRequest_Filter{
    Search: &topologyv1.SearchRequest_Filter_Search{
        Field: "column.id NOT NULL;--",
        Text:  "cat",
    },
},

> SELECT * FROM topology_cache WHERE id NOT NULL;-- LIKE $1
```

echoing my above comment i would do overall validation on search field rather than nested in convertMetadataToQuery on the overall query.

backend/service/topology/querybuilder.go
```diff
@@ -0,0 +1,181 @@
...
+const (
+	column                   = "column"
+	columnIdentifer          = "column."
+	metadata                 = "metadata"
+	metadataIdentifer        = "metadata."
+	queryDefaultLimit uint64 = 100
+	maxResultLimit           = 1000
+)
+
+var metadataQueryValidator = regexp.MustCompile(`^[a-zA-Z0-9->'"]*$`)
```
danielhochman, line 26: i think it would make more sense to check user input components rather than the overall query.

i can still include `'` and `--` in my query which lets me muck with the query (even though without spaces i'm not sure if it's possible to form a valid query)

backend/service/topology/querybuilder_test.go
```diff
@@ -0,0 +1,343 @@
...
+					"label": "value",
+				},
+			},
+			sort: &topologyv1.SearchRequest_Sort{
+				Field:     "metadata.meow.iam.a.cat",
+				Direction: topologyv1.SearchRequest_Sort_ASCENDING,
+			},
+			pageToken: "10",
+			limit:     5,
+			expect:    "SELECT id, data, metadata FROM topology_cache WHERE metadata->'search'->'field' LIKE $1 AND resolver_type_url = $2 AND metadata @> $3::jsonb ORDER BY $4 ASC LIMIT 5 OFFSET 50",
```
danielhochman, line 104: have you tested this query against a real DB? i think that it needs to be `metadata->'search'->>'field'` (double `>>`)

mcutalo88: Ive tested many permutations against a real database yes. I dont think the double `>>` is necessary but I check again.

backend/service/topology/querybuilder.go
mcutalo88: ah yeah true let me do that

backend/service/topology/querybuilder_test.go
mcutalo88: :( yeah `>>` returns text while `>` returns `jsonb`.

mcutalo88: Actually what am I doing, we dont even support this metadata is just a simple map https://github.com/lyft/clutch/blob/topology-paginated-api-impl/api/topology/v1/topology_api.proto#L226-L228

danielhochman: hmm maybe that should be nested like https://github.com/lyft/clutch/blob/topology-paginated-api-impl/api/topology/v1/topology_api.proto#L216

mcutalo88: Nice let me use that

mcutalo88: Nice much better.

Given this query
```json
{
    "limit": 100,
    "page_token": "0",
    "sort": {
        "direction": "ASCENDING",
        "field":"column.id"
    },
    "filter": {
        "search": {
            "field":"metadata.levelone.leveltwo.field",
            "text": "value"
        },
        "typeUrl": "type.googleapis.com/clutch.k8s.v1.Pod"
    }
}
```

Produces this SQL 
```sql
SELECT id, data, metadata FROM topology_cache WHERE metadata->'levelone'->'leveltwo'->>'field' LIKE $1 AND resolver_type_url = $2 ORDER BY $3 ASC LIMIT 100 OFFSET 0
```

which yields this result
```json
{
    "resources": [
        {
            "id": "clutch-main-1234",
            "pb": {
                "@type": "type.googleapis.com/clutch.k8s.v1.Pod",
                "cluster": "",
                "namespace": "",
                "name": "",
                "containers": [],
                "nodeIp": "",
                "podIp": "",
                "state": "UNSPECIFIED",
                "startTime": null,
                "labels": {},
                "annotations": {}
            },
            "metadata": {
                "app": "clutch",
                "environment": "staging",
                "levelone": {
                    "leveltwo": {
                        "field": "value"
                    }
                }
            }
        }
    ],
    "nextPageToken": "1"
}

danielhochman (approved):
looks great, minor comments.

however i also ask that you go through and add docstrings to all of querybuilder.go (purpose of function, maybe an example I/O for each) and comments throughout where appropriate.

PR #894, opened 2021-01-12
authn: service test coverage

danielhochman:
### Description
Adds tests to `clutch.service.authn`.

### Testing Performed
CI

### TODOs
- [x] stronger assertions (claim contents, etc)
- [x] some error condition coverage
- [x] pull OIDC server into a mock for module testing purposes
- [x] put OIDC-specific code in a separate file


backend/service/authn/oidc.go
```diff
@@ -0,0 +1,241 @@
+package authn
```
mcutalo88, line 1: Is this all moved code? Or is there anything you would like specifically reviewed?

danielhochman: yes, with the exception of a couple minor argument changes it's a refactor + new test code

mcutalo88 (approved):
Overall It looks good, I followed through the OIDC / other docs to check for correctness but would still need to do some more reading but Im going to +1.

PR #906, opened 2021-01-13
frontend: Fix metadata-table grid props

scarlettperry:
### Description
Resolving these warnings raised. The MutableRow Grid had props that can only be set if the `container` prop was set.
Screen Shot 2021-01-13 at 6 06 51 PM

Screen Shot 2021-01-13 at 6 07 09 PM

### Testing Performed
Locally

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -77,6 +77,7 @@ const TableCell = styled(MuiTableCell)({
 
 const Grid = styled(MuiGrid)({
   display: "flex",
+  alignItems: "center",
   ".MuiFormControl-root": {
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -143,7 +144,7 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
     <TableRow key={data.id}>
       <KeyCell data={data} />
       <TableCell>
-        <Grid spacing={2} alignItems="center">
```
scarlettperry, line 146: removed spacing altogether b/c it never had any effect

danielhochman (approved):

PR #912, opened 2021-01-14
frontend: autofocus input element in header search bar

scarlettperry:
### Description
A bug bash fix
* PR adds the `autofocus` prop to the textfield component so that it's already focused when the search bar is open.

https://material-ui.com/api/text-field/#textfield-api

### Testing Performed
Locally

Untitled 10


danielhochman (approved):
nice. it's the little things. :+1:

PR #931, opened 2021-01-21
frontend: fix dialog closing behavior

misteroh:
### Description

Fix a bug with the  closing behavior of the Dialog frontend component: Closing the modal via ESC, the X button, or clicking the backdrop got the the modal stuck in an OFF state and unable to appear again.

The component now accepts an optional `onClose` callback. This callback can be used to set the `open` prop passed into `Dialog`. If this callback exists, the component will rely on the `open` prop alone to show/hide. Otherwise, the old functionality is preserved where the component will use its own internal `isOpen` state.  Warning: With the latter use case, closing the dialog with the Escape key, clicking on the backdrop, or using the "X" button will cause the modal to be stuck in the off state.

https://github.com/lyft/clutch/pull/674
https://github.com/lyft/clutch/pull/823

### Testing Performed
Storybook

### DIFF
--- a/frontend/packages/core/src/dialog.tsx
+++ b/frontend/packages/core/src/dialog.tsx
@@ -69,7 +69,7 @@ export interface DialogProps extends Pick<MuiDialogProps, "open"> {
 
 const Dialog = ({ title, children, open: isOpenFromProps, onClose }: DialogProps) => {
   const [isOpenFromState, setIsOpen] = React.useState(isOpenFromProps);
-  const onCloseHandlerExists = typeof onClose === "function";
+  const onCloseHandlerExists = !!onClose;
   const closeCallback = () => {
     if (onCloseHandlerExists) {
       onClose();
--- a/frontend/packages/core/src/dialog.tsx
+++ b/frontend/packages/core/src/dialog.tsx
@@ -57,7 +57,7 @@ const DialogActions = styled(MuiDialogActions)({
 
 export interface DialogContentProps {}
 export interface DialogActionsProps {}
-export type DialogCloseReasons = "closeButtonClick" | "escapeKeyDown" | "backdropClick"
+export type DialogCloseReasons = "closeButtonClick" | "escapeKeyDown" | "backdropClick";
 export interface DialogProps extends Pick<MuiDialogProps, "open"> {
   title: string;
   children:
@@ -69,11 +69,7 @@ export interface DialogProps extends Pick<MuiDialogProps, "open"> {
 }
 
 const Dialog = ({ title, children, open, onClose }: DialogProps) => (
-  <MuiDialog
-    PaperComponent={DialogPaper}
-    open={open}
-    onClose={onClose}
-  >
+  <MuiDialog PaperComponent={DialogPaper} open={open} onClose={onClose}>
     <DialogTitle disableTypography>
       <DialogTitleText>{title}</DialogTitleText>
       <IconButton onClick={e => onClose(e, "closeButtonClick")}>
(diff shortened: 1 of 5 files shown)


frontend/packages/core/src/dialog.tsx
```diff
@@ -64,19 +64,29 @@ export interface DialogProps extends Pick<MuiDialogProps, "open"> {
...
+  onClose?: () => void;
 }
 
-const Dialog = ({ title, children, open }: DialogProps) => {
-  const [isOpen, setIsOpen] = React.useState(open);
-  const onClose = () => {
-    setIsOpen(prevOpenState => !prevOpenState);
+const Dialog = ({ title, children, open: isOpenFromProps, onClose }: DialogProps) => {
+  const [isOpenFromState, setIsOpen] = React.useState(isOpenFromProps);
+  const onCloseHandlerExists = typeof onClose === "function";
```
mcutalo88, line 72: I dont know my typescript very well but is there a const for datatypes we could use over the string?

misteroh: hmm... i guess I did define it in the `DialogProps` interface already so the typeof comparison doesn't have to be done... would `const onCloseHandlerExists = onClose ? true : false;` be better?

mcutalo88: I will defer to you but that also looks good!

misteroh: okay pushed out a new commit!

mcutalo88 (commented):
Nice! Looks good, I'm going to defer final approval to @dschaller as he is the domain expert here.

frontend/packages/core/src/dialog.tsx
```diff
@@ -64,19 +64,29 @@ export interface DialogProps extends Pick<MuiDialogProps, "open"> {
     | React.ReactElement<DialogContentProps>[]
     | React.ReactElement<DialogActionsProps>
     | React.ReactElement<DialogActionsProps>[];
+  onClose?: () => void;
```
dschaller, line 67: actually this completely makes sense since we already require state from the parent component. Let's change this to be required, remove the conditional and update the references.

dschaller:
Awesome! Good catch :)

frontend/packages/core/src/dialog.tsx
misteroh: done! can you please take a look and give me your thoughts on the `DialogCloseReasons` type? The Material Dialog unfortunately doesn't export the close reasons and instead just lists them in the JSdoc comment: 
image

frontend/packages/core/src/dialog.tsx
```diff
@@ -57,42 +57,31 @@ const DialogActions = styled(MuiDialogActions)({
...
+export type DialogCloseReasons = "closeButtonClick" | "escapeKeyDown" | "backdropClick"
 export interface DialogProps extends Pick<MuiDialogProps, "open"> {
   title: string;
   children:
     | React.ReactElement<DialogContentProps>
     | React.ReactElement<DialogContentProps>[]
     | React.ReactElement<DialogActionsProps>
     | React.ReactElement<DialogActionsProps>[];
-  onClose?: () => void;
+  onClose: (event?: object, reason?: DialogCloseReasons) => void;
```
misteroh, line 68: is `object` too generic? MuiDialog uses `{}` but our close button would pass a MouseEvent

misteroh:
/lock

Can't merge until everything using Dialog passes onClose

misteroh:
ready for review @dschaller !

frontend/packages/core/src/dialog.tsx
dschaller: If we can make it more specific that would be great but it's not a blocker

dschaller (approved):

dschaller:
Looks great! Thanks @misteroh

PR #924, opened 2021-01-19
blog: frontend asset passthrough

mcutalo88:
Blog post about the frontend asset passthrough. I will upload the images to githubs CDN and update the links once this has been reviewed.

TODO
- [x] upload image assets to githubs CDN
- [x] remove images from PR and update links
- [x] decide on new title
- [x] update the date before merging

PR #939, opened 2021-01-25
docs: support image directory

mcutalo88:
### Description
While developing a recent blog #924, I wanted to add several images to the blog post. In this PR we simply create a new `images` directory that will be parse by the site generator and moved to the destination directory so these assets my be used by anything on the site.

### Testing Performed

local.

### DIFF
--- a/docs/_website/generator/generate.go
+++ b/docs/_website/generator/generate.go
@@ -148,12 +148,24 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// Get all images files to be moved.
+	images, err := getFiles(docsRoot, ".png")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	files = append(files, images...)
+
 	for _, f := range files {
 		relpath, _ := filepath.Rel(docsRoot, f)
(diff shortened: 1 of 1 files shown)


danielhochman:
I'm not quite following here, were images from the static/ directory not usable in blog posts prior to this change?

Also images should always be in the static/ directory so they are properly handled by the build tools.

Last but not least, I think we should probably start using GitHub's hosting by default for blog post images anyways, there is little benefit to keeping them (or any images for that matter) in the repo itself.

docs/_website/generator/generate.go
```diff
@@ -148,6 +148,14 @@ func main() {
 		log.Fatal(err)
 	}
 
+	// Get all images files to be moved.
+	images, err := getFiles(docsRoot, ".png")
```
danielhochman, line 152: this results in a fatal error when the png is fed to generate then ParseFiles

mcutalo88: Idk its been working on my feature branch with my blog post.

mcutalo88:
I actually did not notice thats where all the images were I was only focused on the generated output. Will close this pr.

PR #948, opened 2021-01-27
serverexperimentation: add e2e test for rtds server

snowp:
### Description
Adds an integration test that runs both the RTDS server and a real Envoy. 
Using this we write a test that validates that Envoy is properly enforcing a
fault based on the RTDS configuration issued by the RTDS server.

The test simply verifies that when an experiment is active, the connected Envoy
will be able to receive and enforce the fault configuration. This should allow us to
validate certain interactions between the generated configuration and Envoy,
and make it easier to verify that ECDS faults are working as intended.

Some refactoring is done to split out the test stores into its own package to
clean up the rtds package and to make the test server used in the existing
RTDS test usable from other tests.


Makefile
```diff
@@ -79,6 +79,10 @@ backend-config-validation:
 yarn-install: yarn-ensure
 	$(YARN) --cwd frontend install --frozen-lockfile 
 
+.PHONY: backend-ensure-envoy
```
snowp, line 82: @mcutalo88 We talked briefly about pulling in a binary for integration testing before the holidays, lmk if you have thoughts on this approach. I still need to figure out where to put this binary to avoid a permission denied, but figured I'd flag this for you in case you had opinions before I spent more time on this

danielhochman:
we'll need a build tag for integration tests so it's still possible to run unit only locally without having to install envoy and such

danielhochman:
i'd also prefer to find a way to run it using docker-compose or the like than requiring local installation of envoy for those who do want to run integration tests locally

snowp:
@danielhochman Sounds reasonable, I'll try see what this will look like using Docker via Go or by using docker-compose

snowp:
@danielhochman I updated this to use docker-compose, lmk what you think of this general approach

backend/module/chaos/serverexperimentation/rtds/envoy/envoy.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+
+		return b
+	})
+
+	return nil
+}
+
+// Helper to convert an input yaml into a typed Protobuf message.
+func unmarshalYaml(m proto.Message, input string) error {
```
danielhochman, line 217: helpers exist for this already in [backend/mock/api/api.go](https://github.com/lyft/clutch/blob/4a8a318e2916a01dd1d6c6b5ff5ab0285a4b324e/backend/mock/api/api.go)

backend/module/chaos/serverexperimentation/rtds/envoy/envoy.go
```diff
@@ -0,0 +1,237 @@
+package envoy
```
danielhochman, line 1: this code belongs in a _test.go or mock file

backend/module/chaos/serverexperimentation/rtds/envoy/envoy.go
```diff
@@ -0,0 +1,237 @@
...
+
+		return b
+	})
+
+	return nil
+}
+
+// AddCluster adds a cluster to the list of static clusters.
+func (e *EnvoyConfig) AddCluster(input string) error {
+	cluster := &cluster.Cluster{}
```
danielhochman, line 173: nit: cluster var collides with cluster package import

backend/module/chaos/serverexperimentation/rtds/envoy/envoy.go
danielhochman: yeah after further review this belongs in some kind of mock or test helper package. this will be really useful for future envoy functionality as well. neat stuff.

backend/module/chaos/serverexperimentation/rtds/integration/Dockerfile-envoy
```diff
@@ -0,0 +1,13 @@
+from golang:1.15.6 as build
```
danielhochman, line 1: nit: i guess compose is permissive but none of the syntax highlighters or linters seem to like lowercase directives
```suggestion
FROM golang:1.15.6 as build
```

backend/module/chaos/serverexperimentation/rtds/integration/Dockerfile
```diff
@@ -0,0 +1,5 @@
+from golang:1.15.6
```
danielhochman, line 1:
```suggestion
FROM golang:1.15.6
```

backend/module/chaos/serverexperimentation/rtds/integration/Dockerfile-envoy
```diff
@@ -0,0 +1,13 @@
...
+
+COPY . /build
+
+RUN cd /build/module/chaos/serverexperimentation/rtds/integration/config && go build -o /generate_config
+RUN /generate_config > /config.json
+
+from envoyproxy/envoy:v1.17.0
```
danielhochman, line 8:
```suggestion
FROM envoyproxy/envoy:v1.17.0
```

danielhochman (commented):
i really like this. the organization needs some work. there's too much extra stuff embedded and nested in the actual module package that would be hard to track if we keep adding other modules with similar tests. and we would need to also get the tests running in CI.

maybe i should just take a whack at moving stuff around to establish a clutch integration testing pattern since there are no other examples of this kind of code yet?

let me know what you think.

donyu:
👍 This is awesome. Do you think it's worth adding some documentation for this in the public docs?

snowp:
@danielhochman I believe I've addressed all your comments, ptal

snowp:
@danielhochman ping

danielhochman (approved):

PR #949, opened 2021-01-27
db: enhance migrate tooling to allow federated migrations

### LINKED ISSUE #947: db: migrations and table management from scaffolded application
**Description**

Currently there's no way to manage tables using Clutch's migration tooling in a private repository that consumes the Clutch core (i.e. a scaffolded app).

As users can create modules in a scaffolded instance of Clutch, users should also be able to manage tables associated with those modules using the core tooling.

It's slightly tricky due to the way that `migrate` manages versions. Most likely we will need to namespace the versioning separately and use a separate set of versions for private instance migrations.

danielhochman:
### Description
Allows federated migrations, in a different directory, on different schemas. See linked issue for more details.

```
$ go run migrate.go -namespace foo -migrationDir migrations -c ../../../examples/postgres-config/clutch-config.yaml 
2021-01-27T13:53:07.885-0600    INFO    migrate/migrate.go:129  using migration table   {"migrationTable": "schema_migrations_foo"}
2021-01-27T13:53:07.887-0600    INFO    migrate/migrate.go:147  using migration directory       {"migrationDir": "migrations"}
2021-01-27T13:53:07.887-0600    INFO    migrate/migrate.go:97   using database  {"hostInfo": "clutch@localhost:5432"}
2021-01-27T13:53:07.887-0600    WARN    migrate/migrate.go:99   migration has the potential to cause irrevocable data loss, verify information above

*** Continue with migration? [y/N]
```

### GitHub Issue

Fixes #947

### DIFF
--- a/backend/cmd/migrate/migrate.go
+++ b/backend/cmd/migrate/migrate.go
@@ -12,7 +12,6 @@ import (
 	"github.com/golang-migrate/migrate/v4"
 	"github.com/golang-migrate/migrate/v4/database/postgres"
 	_ "github.com/golang-migrate/migrate/v4/source/file"
-	"github.com/golang/protobuf/ptypes"
 	_ "github.com/lib/pq"
 	"github.com/uber-go/tally"
 	"go.uber.org/zap"
@@ -24,9 +23,11 @@ import (
 )
 
 type MigrateFlags struct {
-	Force     bool
-	Down      bool
-	BaseFlags *gateway.Flags
+	Force        bool
+	Down         bool
+	Namespace    string
+	MigrationDir string
+	BaseFlags    *gateway.Flags
 }
 
 func (m *MigrateFlags) Link() {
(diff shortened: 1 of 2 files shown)


backend/cmd/migrate/migrate.go
```diff
@@ -187,6 +207,12 @@ func main() {
 	logger, _ := zap.NewDevelopment()
 	logger = logger.WithOptions(zap.AddStacktrace(zap.FatalLevel + 1))
 
+	// Check that both namespace and migration dir are either empty or filled.
+	if (f.Namespace == "") != (f.MigrationDir == "") {
```
mcutalo88, line 211: When `MigrationsTable == postgres.DefaultMigrationsTable` and if `len(f.MigrationDir) > 0` we should probably stop the user from accidentally corrupting their database, no?

Also this check can go elsewhere i just thought of it while reading this line.

mcutalo88: I guess this case should be sufficient

danielhochman: the namespace is appended to `{$DefaultMigrationsTable}_` so it's not possible to accidentally override the default migrations table

mcutalo88: I see now cool.

mcutalo88 (approved):

PR #950, opened 2021-01-27
docs: add component library links

dschaller:
### Description

Adds links to component website in clutch.sh header and on frontend development page.

### DIFF
--- a/docs/_website/src/theme/Navbar/index.js
+++ b/docs/_website/src/theme/Navbar/index.js
@@ -38,11 +38,6 @@ var items = [
     label: 'Community',
   },
   {
-    to: 'https://storybook.clutch.sh',
-    icon: "fe fe-layers",
-    label: 'Storybook',
-  },
-  {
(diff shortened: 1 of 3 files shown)


danielhochman:
can you add the link to the footer as well?

docs/_website/src/theme/Navbar/index.js
```diff
@@ -37,6 +37,11 @@ var items = [
     icon: "fe fe-message-square",
     label: 'Community',
   },
+  {
+    to: 'https://storybook.clutch.sh',
+    icon: "fe fe-layers",
+    label: 'Storybook',
+  },
```
danielhochman, line 44: not sure we need it in the header actually, most people probably won't know what it is or be confused if they click it. better to link to it from docs pages and put it in the footer. wdyt?

dschaller: Sounds good to me

danielhochman (approved):

PR #955, opened 2021-01-28
typo fix in data-layout

misteroh:
`Missing depedency for data layout` => `Missing dependency for data layout`


danielhochman (approved):
thanks!

PR #958, opened 2021-01-29
frontend: fix nil k8s clientset

mcutalo88:
### Description
When going to confrim `resolverInput.clientset`

### Testing Performed

Local.

### GitHub Issue

Reported in [#clutch slack](https://lyftoss.slack.com/archives/C015UJ6LED9/p1611927757004400)


danielhochman:
There's no guarantee the cluster name will match the clientset name, so this approach may not work in all cases.

Another option is to make `GetK8sClientset` operate more permissively, matching on cluster if the clientset isn't found. That still has the potential for issues if there are multiple clientsets on the same cluster with different users and different permissions.

The foolproof fix would be to pass the clientset back in some sort of additional metadata field on the Search call. But I think it would need to per-item, so we would have to wrap `SearchResponse.Results` with some metadata container.

PR #980, opened 2021-02-01
authn: make storage a standalone component

danielhochman:
### Description
Makes authn storage a top-level Clutch component, as opposed to being instantiated and tracked by the authn provider. This will allow other components to read authn tokens, or multiple authn providers to write their tokens using a shared DB connection.

I also updated the doc generator to be able to parse multiple components per package.

### Testing Performed
CI (existing unit tests)


docs/_website/generator/components.go
```diff
@@ -47,7 +51,13 @@ func getClutchComponentFromFile(path string) (*Component, bool) {
 		}
 	}
 
-	name := node.Scope.Lookup("Name")
+	var name *ast.Object
+	for k, v := range node.Scope.Objects {
+		if strings.HasSuffix(k, "Name") && v.Kind == ast.Con {
+			name = v
```
mcutalo88, line 57: Can you break early here?

mcutalo88 (approved):

PR #985, opened 2021-02-03
frontend: Add onClick handler to TableRow and Change max-width of Dialog

misteroh:
### Description

- Added onClick handler to TableRow
- Changed max-width of Dialog Paper to fit-content and added min-width: 600px

Needed for some extra flexibility in styling and functionality

### Testing Performed

Locally

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -1,6 +1,9 @@
 import * as React from "react";
 import styled from "@emotion/styled";
-import type { TableProps as MuiTableProps } from "@material-ui/core";
+import type {
+  TableProps as MuiTableProps,
+  TableRowProps as MuiTableRowProps,
+} from "@material-ui/core";
 import {
   Paper,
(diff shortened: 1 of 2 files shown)


dschaller:
@misteroh can you expand on the use case for this?

misteroh:
@dschaller sure! the Dialog is currently very difficult to style as even putting a container around it and targeting a child .Mui-Dialog (or whatever the class name is) doesn't work because the Dialog is rendered elsewhere in the page with what I assume is a React Portal. With the className prop we can do a styled(Dialog)({...})

the className prop being added to the TableProps interface actually isn't strictly necessary as it currently passes `...props` to the MuiTable anyway, but we might as well for type checking.

The table row on click is needed to trigger an event when clicking on the table row. It's even in the docs for the material-ui Table (https://material-ui.com/components/tables/#data-table), and as currently implemented nothing happens when you click on our Clutch table row which is something I'd like for Windshield

dschaller (approved):

PR #991, opened 2021-02-05
frontend: add className props to Dialog and Table Interface

misteroh:
### Description
See title

### Testing Performed
Locally

Fixes #
- Can now style Dialog


dschaller:
Discussed IRL and we are going to punt on this until we decide to allow global style overrides.

PR #993, opened 2021-02-05
k8s: return error when listing clientsets

danielhochman:
### Description
Since clientsets can be dynamic (per-user, etc), it makes sense to return an error when listing clientsets.

### Testing Performed
Unit and build.


backend/service/k8s/clientset.go
```diff
@@ -138,12 +138,12 @@ type managerImpl struct {
 	clientsets map[string]*ctxClientsetImpl
 }
 
-func (m *managerImpl) Clientsets(ctx context.Context) map[string]ContextClientset {
+func (m *managerImpl) Clientsets(ctx context.Context) (map[string]ContextClientset, error) {
```
scarlettperry, line 141: in this func, not seeing a case when an error is returned

danielhochman: it would only happen internally currently where clientsets are dynamic

rithujohn191 (approved):

PR #1006, opened 2021-02-08
data layout: reset error state on load

dschaller:
### Description

Fix bug where previous error state would be presented when a workflow was retried.

### Testing Performed

manual


frontend/packages/data-layout/src/layout.tsx
```diff
@@ -31,6 +31,10 @@ const useDataLayout = (key: string, opts?: object): DataLayout => {
     throw new Error(`Non-existant data layout key: ${key}`);
   }
 
+  React.useEffect(() => {
```
danielhochman, line 34: add comment

Issue #921, opened 2021-01-18
resolver: autocomplete for queries


micnncim:
**Description**

I want the autocomplete feature in Resolvers.

It's not been implemented yet though `Resolver` interface has a method for it with comment-out:

https://github.com/lyft/clutch/blob/f14f1e551628e1631bafc08c709c9981db444091/backend/resolver/resolver.go#L41

Also, https://clutch.sh/docs/development/backend/#resolvers says:

> In the future, the resolver will handle autocomplete and asynchronous validation of form input for the frontend.

ref: https://github.com/lyft/clutch/blob/f14f1e551628e1631bafc08c709c9981db444091/docs/about/roadmap.md

Complexity [S/M/L]: M?

**Wireframes, Mockups, or Screenshots (if applicable)**

danielhochman:
@mcutalo88 is working on this as we speak! first he's building it out for the AWS workflows, since they currently do straightforward ID lookups. it is powered by the topology cache that he is also building, since AWS API calls can be slow for users with lots of cloud resources.

later we will also add the ability to transform objects in the topology cache into more complicated formats, such as `{cluster}/{podID}` for k8s.

mcutalo88:
Docs issue created https://github.com/lyft/clutch/issues/1143, going to close this out now!

PR #1009, opened 2021-02-08
frontend: autocomplete

mcutalo88:
### Description
Autocomplete component for the frontend. In this PR we amend the `text-field` component to take additional properties to support autocomplete. The autocomplete implementation leverages the Mui's autocomplete component, which we also use for the workflow search bar.

A few notes on behavior:
* A user can always submit a query even if it does not exist in the list of results
* Hitting enter will submit the form as well as the button of course
* Recent results are cached for the lifetime of the workflow, which I think is fine once you navigate away from the workflow that state will be lost.
* If the backend is not configured to handle auto complete the default `text-field` will be used for all workflows.

In a followup PR I would like to add support for labels on the search result, which can provide a bit more metadata about the result. There are some more things to consider before we add metadata to the results with a bit of backend work needed, I felt it was best to leave this out of the initial implementation for now.

### Testing Performed
Local.

### GitHub Issue
This makes progress on issues #921 and #30, a few more follow up PR's to support Kubernetes autocomplete should close out #921 with some additional documentation on how to set it up.

autocompletedemo


frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -55,6 +74,8 @@ const QueryResolver: React.FC<QueryResolverProps> = ({ schemas, submitHandler })
         error={!!error}
         helperText={error?.message || error?.type || ""}
         endAdornment={<SearchIcon />}
+        isAutoCompleteable={isAutoCompleteable}
+        autocompleteCallback={v => autoComplete(inputType, v)}
```
mcutalo88, line 78: Im still looking at how to make optional props so I can remove `isAutoCompleteable`

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -128,6 +170,8 @@ export interface TextFieldProps
     >,
     Pick<MuiInputProps, "readOnly" | "endAdornment"> {
   onReturn?: () => void;
+  isAutoCompleteable?: boolean;
+  autocompleteCallback?: (v: string) => Promise<any>;
```
dschaller, line 174: you can get rid of the boolean here and just use the presence of the callback
```suggestion
  autocompleteCallback?: (v: string) => Promise<any>;
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
     );
   }
 
+  if (isAutoCompleteable) {
```
dschaller, line 211:
```suggestion
  if (autocompleteCallback !== undefined) {
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+          })
+          .catch(err => {
+            helpText = err;
+          });
+      }, 500)
+    ).current;
+
+    // TODO (mcutalo): support option.label in the renderOption
+    return (
+      <Autocomplete
```
dschaller, line 227: nit: I would extract this out to a separate functional component

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+        freeSolo
+        size="small"
+        options={autoCompleteOptions}
+        PopperComponent={renderPopper}
```
dschaller, line 231:
```suggestion
        PopperComponent={Popper}
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+        getOptionLabel={option => (option.id ? option.id : option)}
```
dschaller, line 232:
```suggestion
        getOptionLabel={option => (option?.id ? option.id : option)}
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+        onInputChange={(e, v, r) => autoCompleteDebounce(v)}
```
dschaller, line 233:
```suggestion
        onInputChange={(_, v) => autoCompleteDebounce(v)}
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -92,6 +100,40 @@ const StyledTextField = styled(BaseTextField)({
...
+// popper containing the search result options
+const Popper = styled(MuiPopper)({
+  ".MuiAutocomplete-paper": {
+    boxShadow: "0px 5px 15px rgba(53, 72, 212, 0.2)",
+  },
+  ".MuiAutocomplete-option": {
+    height: "48px",
+    padding: "0px",
+  },
+  ".MuiAutocomplete-option[data-focus='true']": {
```
dschaller, line 112: I'd need to play around with this component but I believe you can use the focus state similar to [here](https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/Input/toggle.tsx#L11)

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+      <Autocomplete
+        freeSolo
+        size="small"
+        options={autoCompleteOptions}
+        PopperComponent={renderPopper}
+        getOptionLabel={option => (option.id ? option.id : option)}
+        onInputChange={(e, v, r) => autoCompleteDebounce(v)}
+        renderOption={option => (
+          <ResultGrid container alignItems="center">
+            <Grid item xs>
```
dschaller, line 236: super nit: can you extract this out into a component?

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+              <ResultLabel>{option.id}</ResultLabel>
+            </Grid>
+          </ResultGrid>
+        )}
+        renderInput={params => (
+          <StyledTextField
+            {...params}
```
dschaller, line 243: reactism
```suggestion
        renderInput={props => (
          <StyledTextField
            {...props}
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+        freeSolo
+        size="small"
+        options={autoCompleteOptions}
+        PopperComponent={renderPopper}
+        getOptionLabel={option => (option.id ? option.id : option)}
+        onInputChange={(e, v, r) => autoCompleteDebounce(v)}
+        renderOption={option => (
+          <ResultGrid container alignItems="center">
+            <Grid item xs>
+              <ResultLabel>{option.id}</ResultLabel>
```
dschaller, line 237: Is the option guaranteed to have an `id`? the `getOptionLabel` prop accounts for the fact it might be missing. How should we handle that here?

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -162,6 +208,55 @@ export const TextField = ({
...
+            </Grid>
+          </ResultGrid>
+        )}
+        renderInput={params => (
+          <StyledTextField
+            {...params}
+            onFocus={onChange}
```
dschaller, line 244: super nit: you could probably extract out the props here into a shared constant since we repeat them below for the standard text field

frontend/packages/core/src/Resolver/input.tsx
dschaller: you got it :)

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -14,18 +14,29 @@ import {
...
 import { TextField } from "../Input/text-field";
+import { client } from "../network";
 
 import type { ChangeEventTarget } from "./hydrator";
 import { convertChangeEvent, hydrateField } from "./hydrator";
 
 const Form = styled.form({});
 
 interface QueryResolverProps {
+  inputType: string;
```
dschaller, line 25: can you add a comment on what this is for?

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -41,6 +52,14 @@ const QueryResolver: React.FC<QueryResolverProps> = ({ schemas, submitHandler })
     setQueryData(convertChangeEvent(event).target.value);
   };
 
+  // If there is more than 1 schema dont enable autocomplete as that is not currently supported
+  let isAutoCompleteable = false;
```
dschaller, line 56:
```suggestion
  let autocompleteEnabled = false;
```

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -41,6 +52,14 @@ const QueryResolver: React.FC<QueryResolverProps> = ({ schemas, submitHandler })
...
+  const autoCompleteableSchemas = schemas.map(schema => schema?.metadata?.search?.autocompleteable);
```
dschaller, line 57:
```suggestion
  const schemasWithAutocomplete = schemas.map(schema => schema?.metadata?.search?.autocompleteable);
```

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -41,6 +52,14 @@ const QueryResolver: React.FC<QueryResolverProps> = ({ schemas, submitHandler })
...
+
+  if (autoCompleteableSchemas.length === 1 && autoCompleteableSchemas[0] === true) {
```
dschaller, line 59: Is it a requirement that the first schema must support autocomplete or can any of them as long as it's just one? If it can be any you can use filter above to get a list of just the schemas with autocomplete enabled and check the length of that to determine if autocomplete should be enabled

dschaller:
Can you add a story for the autocomplete text field variant?

frontend/packages/core/src/Resolver/input.tsx
mcutalo88: Do I need to dynamically assign the `autoComplete` function then? Currently it's always truth right?

frontend/packages/core/src/Resolver/input.tsx
mcutalo88: Currently I have no way of knowing if a specific resource is enabled for autocomplete but rather is the backend configured to support auto complete eg: is the topology service enabled. Which is propagated up through the `autocomplete_enabled` field which will be present on all schemas, this acts as our feature flag. 

But yes I can use the filter as you suggest and just make sure its `>= 1`

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: Yeah the `id` should always be present, with the `label` being optional.  Which I have yet to add label support quite yet, I plan to do that in a follow up PR as there is a bit more backend work needed.

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: I just used all of the styling that was done for the search component https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/AppLayout/search.tsx#L102-L118, I didn't change much with the css besides the border of the result box that was casting a weird shadow.

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: ah wow i didnt know it would just pass down the props for you nice.

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: Either way works, i can use this name also.

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: ok let me parameterize the differences and then we can make it its own component eg: `InputProps`

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: Talked in person and were going to pass on this for now.

frontend/packages/core/src/Input/stories/text-field.stories.tsx
```diff
@@ -50,3 +50,22 @@ WithEndAdornment.args = {
   defaultValue: "Search",
   endAdornment: <SearchIcon />,
 };
+
+const autoComplete = value => {
+  return new Promise((resolve, reject) => {
+    resolve({
+      results: [
+        { id: "clutch", label: "" },
```
dschaller, line 58: super nit: it would be cool to have these be configurable within the story but that's a nice to have not a requirement

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -106,6 +144,23 @@ const IconButton = styled(MuiIconButton)({
   },
 });
 
+interface resultOption {
```
dschaller, line 147:
```suggestion
interface ResultOption {
```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -106,6 +144,23 @@ const IconButton = styled(MuiIconButton)({
...
+  id: string;
+  label: string;
+}
+
+interface ResultProps {
+  option: resultOption;
+}
+
```
dschaller, line 155: It looks like `resultOption` isn't used anywhere outside of result props. You could do something like
```suggestion
interface ResultProps {
  option: {
    id: string;
    label: string;
  };
}

```

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -106,6 +144,23 @@ const IconButton = styled(MuiIconButton)({
...
+const Result: React.FC<ResultProps> = ({ option }) => (
```
dschaller, line 156:
```suggestion
const AutocompleteResult: React.FC<AutocompleteResultProps> = ({ option }) => (
```

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: ah ok nice

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -106,6 +144,23 @@ const IconButton = styled(MuiIconButton)({
   },
 });
 
+interface resultOption {
+  id: string;
+  label: string;
+}
+
+interface ResultProps {
+  option: resultOption;
```
dschaller, line 153: Why have `option` here? It seems like the result can have an id and label props directly

frontend/packages/core/src/Input/stories/text-field.stories.tsx
mcutalo88: oh yeah let me add that

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -14,18 +14,31 @@ import {
...
 
 import type { ChangeEventTarget } from "./hydrator";
 import { convertChangeEvent, hydrateField } from "./hydrator";
 
 const Form = styled.form({});
 
 interface QueryResolverProps {
+  // The inputType is the orignal resource type requested
+  // eg: clutch.aws.ec2.v1.AutoscalingGroup
+  inputType: string;
```
dschaller, line 27: Typescript can pick this up as a docstring if you have it in the format
```suggestion
  /**
   * The inputType is the orignal resource type requested
   * eg: clutch.aws.ec2.v1.AutoscalingGroup
   */
  inputType: string;
```

frontend/packages/core/src/Input/text-field.tsx
mcutalo88: yeah we could do that let me change

dschaller (approved):

PR #1011, opened 2021-02-09
gateway: error log handler

scarlettperry:
### Description
PR adds a new error log handler that can (1) unpack a gRPC error and return the Status or (2) return the original error.

### Testing Performed
Unit tests
Locally

**examples**
1) a conventional error

```
ERROR   k8s/job.go:33   error deleting Job      {"serviceName": "clutch.service.k8s", "error": "jobs.batch \"foo\" not found"}
```

2) a gRPC error
```
ERROR   k8s/job.go:35   error deleting Job      {"serviceName": "clutch.service.k8s", "error": {"code":5,"message":"Job not found"}}
```

3) a gRPC error showing the extra details that had been appended to the Status
```
ERROR   k8s/job.go:36   error deleting Job      {"serviceName": "clutch.service.k8s", "error": {"code":5,"message":"Job not found","details":[{"@type":"type.googleapis.com/google.rpc.ResourceInfo","resourceType":"Job","description":"foo not found"},{"@type":"type.googleapis.com/google.rpc.ResourceInfo","resourceType":"Job","description":"foobar not found"}]}}
```

### DIFF
--- a/backend/gateway/log/log.go
+++ b/backend/gateway/log/log.go
@@ -19,7 +19,7 @@ func ProtoField(key string, m proto.Message) zap.Field {
 }
 
 // NamedError returns a zap field that logs as the embedded JSON representation of the error.
-// If's a gRPC error, it converts it to a Status and returns
+// If it's a gRPC error, it converts it to a Status and returns
 // the unpacked information. Otherwise, it returns the original error.
 func NamedError(key string, err error) zap.Field {
 	s, ok := status.FromError(err)
@@ -27,4 +27,9 @@ func NamedError(key string, err error) zap.Field {
 		return zap.NamedError(key, err)
 	}
 	return ProtoField(key, s.Proto())
+}
+
+// Error calls NamedError with the key "error"
+func Error(err error) zap.Field {
+	return NamedError("error", err)
 }
--- a/backend/gateway/log/log.go
+++ b/backend/gateway/log/log.go
@@ -18,10 +18,10 @@ func ProtoField(key string, m proto.Message) zap.Field {
 	return zap.Any(key, json.RawMessage(b))
 }
 
-// NamedError returns a zap field that logs as the embedded JSON representation of the error.
+// NamedErrorField returns a zap field that logs as the embedded JSON representation of the error.
 // If it's a gRPC error, it converts it to a Status and returns
 // the unpacked information. Otherwise, it returns the original error.
-func NamedError(key string, err error) zap.Field {
+func NamedErrorField(key string, err error) zap.Field {
 	s, ok := status.FromError(err)
 	if !ok {
 		return zap.NamedError(key, err)
@@ -29,7 +29,7 @@ func NamedError(key string, err error) zap.Field {
 	return ProtoField(key, s.Proto())
 }
 
-// Error calls NamedError with the key "error"
-func Error(err error) zap.Field {
-	return NamedError("error", err)
+// ErrorField calls NamedErrorField with the key "error"
+func ErrorField(err error) zap.Field {
+	return NamedErrorField("error", err)
 }
--- a/backend/gateway/log/log.go
+++ b/backend/gateway/log/log.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 
 	"go.uber.org/zap"
-	spb "google.golang.org/genproto/googleapis/rpc/status"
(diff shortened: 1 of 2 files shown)


backend/gateway/log/log.go
```diff
@@ -16,3 +18,23 @@ func ProtoField(key string, m proto.Message) zap.Field {
...
+// representation of a gRPC Status message.
+func GrpcStatusField(key string, err error) zap.Field {
+	var m *spb.Status
+	// the second return val is a bool reporting whether it was able to
+	// convert the input error to a Status. If false, FromError returns a new Status with
+	// codes.Unknown and the original input error message.
+	// https: //github.com/grpc/grpc-go/blob/master/status/status.go#L81
+	s, _ := status.FromError(err)
+
+	m = s.Proto()
```
scarlettperry, line 32: returns Status as spb.Status proto message
https://pkg.go.dev/google.golang.org/grpc@v1.34.0/internal/status#Status.Proto

backend/gateway/log/log.go
```diff
@@ -16,3 +18,23 @@ func ProtoField(key string, m proto.Message) zap.Field {
...
+
+// GrpcStatusField takes in an error and returns a zap field that logs as the embedded JSON
+// representation of a gRPC Status message.
+func GrpcStatusField(key string, err error) zap.Field {
+	var m *spb.Status
+	// the second return val is a bool reporting whether it was able to
+	// convert the input error to a Status. If false, FromError returns a new Status with
+	// codes.Unknown and the original input error message.
+	// https: //github.com/grpc/grpc-go/blob/master/status/status.go#L81
+	s, _ := status.FromError(err)
```
danielhochman, line 30: i'd still return the original error in the event it's not a gRPC error. then we can use this function everywhere without worrying about whether or not it's a status and that changing the error up to appear as if it was a status. and for that matter we can rename this function to NamedError and create an Error helper just like zap has to be used everywhere.
```suggestion
	s, ok := status.FromError(err)
	if !ok {
	   return zap.NamedError(key, err)
	}
```

backend/gateway/log/log.go
```diff
@@ -16,3 +18,23 @@ func ProtoField(key string, m proto.Message) zap.Field {
...
+
+	m = s.Proto()
+
+	b, err := protojson.Marshal(m)
+	if err != nil {
+		return zap.Any(key, m)
+	}
+
+	return zap.Any(key, json.RawMessage(b))
```
danielhochman, line 39: redundant
```suggestion
	return ProtoField(key, s.Proto())
```

danielhochman (commented):
nice!

backend/gateway/log/log.go
scarlettperry: got it, yeah that would be better. will update

backend/gateway/log/log.go
scarlettperry: nice! thanks for catching

backend/gateway/log/log.go
scarlettperry: updated the local testing examples in the PR description

danielhochman (commented):
looks good, though can we also add the convenience `func Error(err error) zap.Field` that just called NamedError with key `"error"`?

scarlettperry:
> looks good, though can we also add the convenience `func Error(err error) zap.Field` that just called NamedError with key `"error"`?

oh! sorry misunderstood. will update

backend/gateway/log/log.go
```diff
@@ -16,3 +17,14 @@ func ProtoField(key string, m proto.Message) zap.Field {
 	}
 	return zap.Any(key, json.RawMessage(b))
 }
+
+// NamedError returns a zap field that logs as the embedded JSON representation of the error.
+// If's a gRPC error, it converts it to a Status and returns
```
danielhochman, line 22:
```suggestion
// If it's a gRPC error, it converts it to a Status and returns
```

danielhochman (approved):
:balloon:

rithujohn191 (approved):
😀

danielhochman:
real quick, i wonder if we should just add `Field` to the end of each method name. let's discuss in standup?

scarlettperry:
> real quick, i wonder if we should just add `Field` to the end of each method name. let's discuss in standup?

ok! will wait to merge

PR #1019, opened 2021-02-10
core: add ErrorDetails proto and use in resolver

danielhochman:
### Description
Introduces the ErrorDetails proto object, that the frontend will know how to render from. Leverages the error wrapping field from the ErrorDetails proto in the resolver.

### Testing Performed
Unit.

### TODOs
- [x] Tests


dschaller (approved):

backend/module/resolver/response.go
```diff
@@ -59,18 +60,40 @@ func (r *response) marshalResults(results *resolver.Results) error {
 	return nil
 }
 
+func allStatusMatch(c codes.Code, sl []*statuspb.Status) bool {
```
mcutalo88, line 63: Lets write a test for this

api/api/v1/error.proto
```diff
@@ -0,0 +1,16 @@
...
+
+import "google/rpc/status.proto";
+
+// Any error information beyond code and status should be included here and added to the error in the status details
+// field. The frontend knows how to render all of the fields in a user-friendly way. If there is extremely verbose error
+// information, consider adding it using a different type, e.g. from the errdetails package. Any details not using this
+// type will still be accessible to the user in a raw format.
+message ErrorDetails {
+  // If there are any underlying errors that were being wrapped, they are presented here.
+  repeated google.rpc.Status wrapped = 1;
```
mcutalo88, line 15: I dont think I can see the full picture, but how is a list of status useful to the FE without mappings to more descriptive error messages? Or is that part coming next?

mcutalo88: Oh i see now https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto#L28-L47

mcutalo88: I was thinking this was just status code at first

danielhochman: i don't follow the question exactly. each status has it's own message and code. this field allows us to render the highlighted region on the frontend:
image

mcutalo88 (approved):

PR #1021, opened 2021-02-10
tools: add vscode workspace configuration

dschaller:
### Description

Adds a VSCode workspace configuration file to support opening Clutch backend and frontend simultaneously in the same vscode window.

### Testing Performed

manual


.vscode/clutch.code-workspace
```diff
@@ -0,0 +1,24 @@
...
+		{
+			"path": "../frontend"
+		},
+		{
+			"path": "../tools"
+		},
+		{
+      "name": "ci",
+      "path": "../.github"
+    }
```
mcutalo88, line 21: nit: fix formatting

danielhochman:
weird, vscode works for me with no config

mcutalo88:
Interesting it has never worked for me on OSX

mcutalo88 (approved):

dschaller:
Certain features of VSCode will work but some do not. @danielhochman can you open up a test file and see if typescript can resolve global types for jest? That didn't work for me on mac

PR #1023, opened 2021-02-11
sourcecontrol: add support for dynamic repository options

dschaller:
### Description

Adds support for the backend to drive owner and visibility options displayed on the frontend. This change also allows for these to be set via configuration to limit what users can set for both owner and visibility of a new repository.

### Testing Performed

n/a


api/sourcecontrol/v1/sourcecontrol.proto
```diff
@@ -9,9 +9,17 @@ import "google/protobuf/any.proto";
 import "validate/validate.proto";
 
 import "api/v1/annotations.proto";
+import "config/module/sourcecontrol/v1/sourcecontrol.proto";
 import "sourcecontrol/github/v1/github.proto";
 
 service SourceControlAPI {
+  rpc GetRepositoryOptions(GetRepositoryOptionsRequest) returns (GetRepositoryOptionsResponse) {
+    option (google.api.http) = {
+      post : "/v1/sourcecontrol/repositoryOptions",
```
danielhochman, line 18: nit: name should match, rpc style
```suggestion
      post : "/v1/sourcecontrol/getRepositoryOptions",
```

api/sourcecontrol/v1/sourcecontrol.proto
```diff
@@ -21,6 +29,20 @@ service SourceControlAPI {
...
+
+// An entity encompasses members and organizations who are both owners of or contributors to source code.
+message Entity {
+  string name = 1;
+  string photo_url = 2;
+}
+
+message GetRepositoryOptionsResponse {
+  repeated Entity available_owners = 1;
+  repeated clutch.config.module.sourcecontrol.v1.Visibility visibility_options = 2;
```
danielhochman, line 43: i don't think we've run into this before, but config objects should not leak over to the RPC side. i think the converse doesn't necessarily need to be true so you could move visibility_options over to the RPC side and then reference it on the config side.

one reason for this is that at some point we want to stop bundling service config protos over to the frontend where they are unused.

api/config/module/sourcecontrol/v1/sourcecontrol.proto
```diff
@@ -0,0 +1,15 @@
...
+
+package clutch.config.module.sourcecontrol.v1;
+
+option go_package = "github.com/lyft/clutch/backend/api/config/module/sourcecontrol/v1;sourcecontrolv1";
+
+import "sourcecontrol/v1/sourcecontrol.proto";
+
+message Config {
+  // Name of the owner for new repositories.
+  repeated string owners = 1;
```
mcutalo88, line 11: Does owners here imply organizations at least for a github context?

dschaller: It can be both organizations as well as a member since you can create accounts for the organizations you belong to but also your own github account.

mcutalo88 (approved):

PR #1025, opened 2021-02-12
Fix incorrect name attached to cluster radio group props

donyu:
The `upstreamClusterType` was no longer being updated when selecting one of the values from the radio group. This fixes the bug, and I have some ideas on how to refactor our React code to follow paradigms to prevent this kind of logic leak.

### Description
When migrating our fields, we didn't update the property name passed into the onChange() prop function in the <RadioGroup> component.

### Testing Performed
Tested locally by verifying state in <ExperimentDetail> before & after.


donyu:
👀 @kathan24

kathan24 (approved):

donyu:
💨

PR #1027, opened 2021-02-12
gateway: add an error interception interface for services

danielhochman:
### Description
Allow services to intercept and rewrite errors in the middleware chain by exposing an `InterceptError(error) error` function on the service.

This will make it so that services can effectively register a function that will look in on an error, see if it can convert it from whatever error type (e.g. k8s error or aws error) to a `status.Status` for rich error information on the client.

The reason it makes sense in an interceptor is so that every single function in a service doesn't have to call some error enrichinment function before the return. It can just blindly return if err != nil and trust that it will eventually be caught by the interceptor before making it to the client as a code 2 with only the message preserved.

### Testing Performed
TODO

### TODOs
- [x] tests
- [x] determine if it makes sense for modules to have the same capability (will defer)


backend/gateway/gateway.go
```diff
@@ -88,6 +89,14 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
 	initScope := scope.SubScope("gateway")
 	initScope.Counter("start").Inc(1)
 
+	// Create the error interceptor so services can register error interceptors if desired.
+	var unaryErrorMiddleware *errorintercept.Middleware
+	if m, err := errorintercept.New(); err != nil {
+		logger.Fatal("could not create error interceptor middleware", zap.Error(err))
+	} else {
+		unaryErrorMiddleware = m.(*errorintercept.Middleware)
+	}
```
mcutalo88, line 98: nit: fatal logs should call `os.Exit(1)` we probably don't need the else unless we change log level.

mcutalo88 (approved):
Looks sweet, just left a small nit.

PR #1038, opened 2021-02-17
frontend: data-layout only merge data of same types on hydration end

dschaller:
### Description

Update data-layout to only merge existing data and new data when the types are the same.

### Testing Performed

added tests


mcutalo88 (approved):
I think this looks good, Im curious to learn more on how we had the mismatch but we can do that async.

PR #1042, opened 2021-02-18
data layout: documentation and cleanup

dschaller:
### Description

Cleanup the data-layout.
- Add docstrings to the layout props and functions
- Remove unused `setError` and `setLoading` functions as they are not necessary to the component
- remove support for additional `opts` prop as it doesn't appear to be used
- fix hydration based on cache flag and existing data

### GitHub Issue

Fixes #


danielhochman (approved):
nice

PR #1052, opened 2021-02-19
sourcecontrol: fetch owners and visibility from server in create repo

dschaller:
### Description

Read possible owners and visibility in source control create repository workflow from the server. The server will also have the ability to control control which owners and visibility are available to users.

This removes the additional step of adding collaborators which wasn't working before. We can add it back in in a subsequent PR when that functionality is working.

sourcecontrol

### Testing Performed

Manual

### GitHub Issue

Fixes #


backend/module/sourcecontrol/sourcecontrol.go
```diff
@@ -31,22 +42,104 @@ func New(*any.Any, *zap.Logger, tally.Scope) (module.Module, error) {
...
-func (m *mod) CreateRepository(ctx context.Context, req *sourcecontrolv1.CreateRepositoryRequest) (*sourcecontrolv1.CreateRepositoryResponse, error) {
-	return m.github.CreateRepository(ctx, req)
+func contains(s []string, str string) bool {
+	for _, v := range s {
+		if v == str {
+			return true
+		}
+	}
+	return false
+}
```
mcutalo88, line 65: Lets add tests for this

backend/module/sourcecontrol/sourcecontrol.go
```diff
@@ -31,22 +42,104 @@ func New(*any.Any, *zap.Logger, tally.Scope) (module.Module, error) {
...
+
+func (m *mod) ManageableOrganizations(ctx context.Context) ([]*sourcecontrolv1.Entity, error) {
+	organizations, err := m.github.ListOrganizations(ctx, "")
```
mcutalo88, line 68: we should just make `""` a constant kinda like what kubernetes has, when you reference all namespaces its `""` but they have a type [metav1.NamespaceAll](https://github.com/kubernetes/apimachinery/blob/da970ee67a0683f95ab94dece488e7adb4eaacdf/pkg/apis/meta/v1/types.go#L288) so its clear to the reader why its an empty string.

backend/module/sourcecontrol/sourcecontrol.go
```diff
@@ -31,22 +42,104 @@ func New(*any.Any, *zap.Logger, tally.Scope) (module.Module, error) {
...
+	writableOrgs := []*sourcecontrolv1.Entity{}
+	if err != nil {
+		return nil, err
+	}
```
mcutalo88, line 72: nit: move error under the `m.github.ListOrganizations` call

backend/module/sourcecontrol/sourcecontrol.go
```diff
@@ -31,22 +42,104 @@ func New(*any.Any, *zap.Logger, tally.Scope) (module.Module, error) {
...
+	configVisibility := m.config.VisibilityOptions
+	var visibilityOptions []sourcecontrolv1.Visibility
+	if configVisibility == nil {
+		visibilityOptions = []sourcecontrolv1.Visibility{
+			sourcecontrolv1.Visibility_PUBLIC,
+			sourcecontrolv1.Visibility_PRIVATE,
+		}
+	} else {
+		visibilityOptions = configVisibility
+	}
```
mcutalo88, line 118:
```suggestion
	visibilityOptions := m.config.VisibilityOptions
	if visibilityOptions == nil {
		visibilityOptions = []sourcecontrolv1.Visibility{
			sourcecontrolv1.Visibility_PUBLIC,
			sourcecontrolv1.Visibility_PRIVATE,
		}
	}
```

backend/module/sourcecontrol/sourcecontrol.go
dschaller: I was trying to avoid having the logic of writable orgs within the list orgs call because you could technically use the github service to see all of a users orgs just for membership purposes. I could create a new function that is just `ListManageableOrganizations` or could have a parameter to control which orgs to return based on permissions. what do you think?

backend/module/sourcecontrol/sourcecontrol.go
```diff
@@ -31,22 +42,101 @@ func New(*any.Any, *zap.Logger, tally.Scope) (module.Module, error) {
...
+	return false
+}
+
+func (m *mod) ManageableOrganizations(ctx context.Context) ([]*sourcecontrolv1.Entity, error) {
+	organizations, err := m.github.ListOrganizations(ctx, github.CurrentUser)
+	writableOrgs := []*sourcecontrolv1.Entity{}
+	if err != nil {
+		return nil, err
+	}
```
scarlettperry, line 72: i think this is what Mike was referring to
```suggestion
	organizations, err := m.github.ListOrganizations(ctx, github.CurrentUser)
	if err != nil {
		return nil, err
	}
	writableOrgs := []*sourcecontrolv1.Entity{}
```

frontend/packages/core/src/Input/select.tsx
```diff
@@ -178,6 +179,7 @@ const StyledSelect = styled(BaseSelect)({
 interface SelectOption {
   label: string;
   value?: string;
+  startAdornment?: React.ReactElement;
```
scarlettperry, line 182: would be nice to have this example on storybook but doesn't have to be done in this PR

frontend/workflows/sourcecontrol/src/create-repository.tsx
```diff
@@ -113,18 +102,25 @@ const Confirm: React.FC<WizardChild> = () => {
...
 
-const CreateRepository: React.FC<WorkflowProps> = ({ heading, options }) => {
+const CreateRepository: React.FC<WorkflowProps> = ({ heading }) => {
   const dataLayout = {
-    extraOptionsData: {},
-    resourceData: { cache: false },
+    repositoryData: {
+      cache: false,
+      hydrator: () => {
+        return client.post("/v1/sourcecontrol/getRepositoryOptions");
```
scarlettperry, line 119: could the github mocks in the backend be updated?  https://github.com/lyft/clutch/blob/main/backend/mock/service/githubmock/githubmock.go, I think it's just missing `GetRepositoryOptions`. Just so that we can easily test the frontend

scarlettperry (approved):
lgtm! just some non-blocking comments

PR #1063, opened 2021-02-22
k8s: adding more fields to cronjob proto

smonero:
Add some fields to the cronjob proto

### Description

Need some extra stuff in the cronjob description - i.e suspend, active, concurrency policy, and starting deadline seconds.
https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

### Testing Performed

unit test

### GitHub Issue

Fixes #


mcutalo88 (commented):
Nice looks fine overall, lets also add these new fields here https://github.com/lyft/clutch/blob/2792e9d58070223e10ce42d11cc43ab1b71da9d8/backend/service/k8s/cronjob.go#L65-L78 and update tests.

api/k8s/v1/k8s.proto
```diff
@@ -643,6 +643,20 @@ message CronJob {
...
+  enum ConcurrencyPolicy {
+    UNSPECIFIED = 0;
+    UNKNOWN = 1;
+    ALLOW = 2;
+    FORBID = 3;
+    REPLACE = 4;
+  }
+  ConcurrencyPolicy concurrency_policy = 9;
+
+  int64 starting_deadline_seconds = 10;
```
danielhochman, line 659: i think this needs a value wrapper since it's optional (null different from 0)

api/k8s/v1/k8s.proto
```diff
@@ -643,6 +643,20 @@ message CronJob {
 
   map<string, string> labels = 5;
   map<string, string> annotations = 6;
+
+  bool suspend = 7;
+  int32 active_jobs = 8;
```
danielhochman, line 648: is this supposed to be repeated or is it a count? maybe `num_active_jobs` would be better in case we embed more information about the active jobs later?

danielhochman (commented):
the more fields the merrier! a couple of comments

api/k8s/v1/k8s.proto
smonero: sounds good.  I'll make it num_active_jobs since its a count for now

danielhochman (approved):
lgtm! comments are discretionary, i just tend to avoid the wrappers if i can since they are annoying esp on the frontend.

PR #1064, opened 2021-02-22
k8s: add create job API call

rithujohn191:
### Description
This PR adds a CreateJob API call to the K8s service which enables a user to create a Kubernetes Job object

### Testing Performed
- Unit tests
- Local testing
```
$ curl -X POST   http://localhost:3000/v1/k8s/createJob -H 'Content-Type: application/json'   -d '{
"namespace": "envoy-staging",
"job_config":  $JOB,
}'
{"job":{"cluster":"kind-clutch-local", "namespace":"envoy-staging", "name":"test-job", "annotations":{}, "labels":{}}
```

```
kubectl get jobs -n envoy-staging
NAME                 COMPLETIONS   DURATION   AGE
job-staging-test-1   0/1           12d        12d
job-staging-test-2   0/1           12d        12d
test-job             0/1                      4m52s
```


api/k8s/v1/k8s.proto
```diff
@@ -801,6 +809,27 @@ message DeleteJobRequest {
...
 }
 
+message JobConfig {
+  google.protobuf.Value value = 1;
+}
+
+message CreateJobRequest {
+  option (clutch.api.v1.id).patterns = {
+    type_url : "clutch.k8s.v1.Job",
+    pattern : "{cluster}/{namespace}/{name}"
```
scarlettperry, line 819:
```suggestion
    pattern : "{cluster}/{namespace}"
```
a name isn't being sent in the request

backend/service/k8s/job.go
```diff
@@ -20,6 +22,31 @@ func (s *svc) DeleteJob(ctx context.Context, clientset, cluster, namespace, name
...
+
+	// convert Job config into a runtime object and then type cast to a
+	// batch job
+	config := []byte(jobConfig.GetStringValue())
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	obj, _, err := decode(config, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	job := obj.(*v1.Job)
```
scarlettperry, line 39: nice! TIL

backend/service/k8s/job_test.go
```diff
@@ -126,6 +127,36 @@ func TestDeleteJob(t *testing.T) {
...
+			clientsets: map[string]*ctxClientsetImpl{"testing-clientset": &ctxClientsetImpl{
+				Interface: cs,
+				namespace: "testing-namespace",
+				cluster:   "testing-cluster",
+			}},
+		},
+	}
+
+	value := structpb.NewStringValue(batchJob)
+	job, err := s.CreateJob(context.Background(), "testing-clientset", "testing-cluster", "testing-namespace", value)
```
scarlettperry, line 155: is there a common error case we can test as well?

api/k8s/v1/k8s.proto
rithujohn191: Thanks for catching this

api/k8s/v1/k8s.proto
```diff
@@ -801,6 +809,27 @@ message DeleteJobRequest {
 message DeleteJobResponse {
 }
 
+message JobConfig {
```
danielhochman, line 812: is this wrapper necessary? seems like we could just use `google.Protobuf.Value job_config` in the request.

api/k8s/v1/k8s.proto
```diff
@@ -801,6 +809,27 @@ message DeleteJobRequest {
...
+  string cluster = 2 [ (validate.rules).string = {min_bytes : 1} ];
+  string namespace = 3 [ (validate.rules).string = {min_bytes : 1} ];
+  JobConfig job_config = 4;
+}
+
+message CreateJobResponse {
+  Job job = 1;
+}
+
+// This message type is used to add support for nullable strings and is an
```
danielhochman, line 832:
```suggestion
```

backend/service/k8s/job_test.go
rithujohn191: Added a case where the job Config passed in is invalid and it returns an error

backend/service/k8s/job.go
danielhochman, line 39: i'd think i'd actually push this conversion code up to the module since it deals with user input transformation. if other code paths want to create a job, they wouldn't need to go back and forth between yaml and the go struct.

danielhochman (commented):
looking good, i like the use of the blob as we discussed

api/k8s/v1/k8s.proto
rithujohn191: Yep we could, I just thought the naming will help with understanding what the value mean, but I could change it and add  a comment if its less confusing

backend/service/k8s/job.go
rithujohn191: Ah ok and then change the function signature of this method to take in the K8s job directly?

danielhochman: yep!

api/k8s/v1/k8s.proto
rithujohn191: Followed the pattern from here for consistency: https://github.com/lyft/clutch/blob/main/api/envoytriage/v1/output.proto#L24. Let me know if you prefer I change it

danielhochman: that was done because we might extract variables from the dump later out into the message in addition to the raw content. but in this case we already have a higher level object.

i agree though it works either way so you can go with your preference.

my other question i guess is `Config` the right name for this? would `Spec` be a more k8s-ish name?

rithujohn191: The Job object has a field within it called Spec, so I didn't use spec cause I thought it would confuse the user that they only have to provide the spec blob

backend/service/k8s/job.go
rithujohn191: Done

backend/module/k8s/job.go
```diff
@@ -21,3 +25,23 @@ func (a *k8sAPI) ListJobs(ctx context.Context, req *k8sapiv1.ListJobsRequest) (*
...
+	}
+	// convert Job config into a runtime object and then type cast to a
+	// batch job
+	config := []byte(req.JobConfig.Value.GetStringValue())
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	obj, _, err := decode(config, nil, nil)
+	if err != nil {
+		return nil, err
+	}
+	job := obj.(*batchv1.Job)
```
danielhochman, line 41: add an `ok` check?

rithujohn191: Thanks for catching that. Added

backend/tools.go
```diff
@@ -11,6 +11,7 @@ import (
 	_ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway"
 	_ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2"
 	_ "github.com/shurcooL/vfsgen"
-	_ "google.golang.org/protobuf/cmd/protoc-gen-go"
 	_ "google.golang.org/grpc/cmd/protoc-gen-go-grpc"
+	_ "google.golang.org/protobuf/cmd/protoc-gen-go"
+	_ "k8s.io/api"
```
danielhochman, line 16:
```suggestion
```

danielhochman: i think this is leftover from the earlier compiling k8s protos attempt

rithujohn191: hmmm...weird that make backend-verify doesn't clean it up. I'll remove it

rithujohn191: Ah we add to this only when we don't want `go mod tidy` to clean it up

danielhochman (approved):
:tada: :ship:

backend/tools.go
```diff
@@ -11,6 +11,6 @@ import (
 	_ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway"
 	_ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2"
 	_ "github.com/shurcooL/vfsgen"
-	_ "google.golang.org/protobuf/cmd/protoc-gen-go"
 	_ "google.golang.org/grpc/cmd/protoc-gen-go-grpc"
+	_ "google.golang.org/protobuf/cmd/protoc-gen-go"
```
rithujohn191, line 15: This is just rearranging it alphabetically

PR #1069, opened 2021-02-23
kubernetes: cache to re-list resources hourly

mcutalo88:
### Description
Performs a full re-list for all watched resources types by the Kubernetes cache. This works in tandem with the LightweightInformer to keep the cache as fresh as possible while also keeping items in cache that may not see many updates. This is similar to upstream's informer's resync while still keeping the benefit to overall memory footprint and actually pulling fresh data from the api servers.

For example some Kubernetes resources dont see frequent enough updates which in-turn does not emit an informer event. If updates to these resources are so infrequent they can be cleaned up by the topology cache which removes stale items.

In order to thread through the TopologyCache TTL to the kubernetes cache I changed the Interface for `CacheableTopology` which could be useful for other services of the topology cache for similar situations like this one.

### Testing Performed
Locally.

### DIFF
--- a/backend/service/k8s/cache.go
+++ b/backend/service/k8s/cache.go
@@ -28,7 +28,7 @@ func (s *svc) CacheEnabled() bool {
 	return true
 }
 
-func (s *svc) StartTopologyCaching(ctx context.Context) (<-chan *topologyv1.UpdateCacheRequest, error) {
+func (s *svc) StartTopologyCaching(ctx context.Context, ttl time.Duration) (<-chan *topologyv1.UpdateCacheRequest, error) {
 	// There should only ever be one instances of all the informers for topology caching
 	// We lock here until the context is closed
 	if !s.topologyInformerLock.TryAcquire(topologyInformerLockId) {
@@ -41,13 +41,13 @@ func (s *svc) StartTopologyCaching(ctx context.Context) (<-chan *topologyv1.Upda
 	}
 	for name, cs := range clientsets {
 		s.log.Info("starting informer for", zap.String("cluster", name))
-		go s.startInformers(ctx, name, cs)
+		go s.startInformers(ctx, name, cs, ttl)
 	}
 
 	return s.topologyObjectChan, nil
 }
 
-func (s *svc) startInformers(ctx context.Context, clusterName string, cs ContextClientset) {
+func (s *svc) startInformers(ctx context.Context, clusterName string, cs ContextClientset, ttl time.Duration) {
 	informerHandlers := cache.ResourceEventHandlerFuncs{
 		AddFunc:    s.informerAddHandler,
 		UpdateFunc: s.informerUpdateHandler,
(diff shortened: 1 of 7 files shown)


backend/service/k8s/cache.go
```diff
@@ -85,6 +91,59 @@ func (s *svc) startInformers(ctx context.Context, cs ContextClientset) {
...
 
+// cacheFullRelist will list all resources and push them to the topology cache for processing.
+// The importance of doing a semi frequent full re-list give us better cache accuracy,
+// while also keeping resources that are infrequently updated from being cleaned up by the cache TTL.
+// This works in tandem with the LightweightInformers above.
+//
+// Notably we intentionally run these in serial, not only can this cause memory pressure but
+// also being mindful of the kubernetes api servers to reduce burst load.
+func (s *svc) cacheFullRelist(ctx context.Context, lwPods, lwDeployments, lwHPA *cache.ListWatch) {
+	ticker := time.NewTicker(time.Hour * 1)
```
danielhochman, line 102: hardcoding this here is not great because if the cache TTL dips below this value, because it's configurable or otherwise modified, then we're going to have the same problem.

we need a way to get the TTL value from cache and then use some lesser value here.

backend/service/k8s/cache.go
```diff
@@ -85,6 +91,59 @@ func (s *svc) startInformers(ctx context.Context, cs ContextClientset) {
...
+	for {
+		// The informers will only ever do a full list once on boot
+		// we will wait the hour before doing another full list again
+		select {
+		case <-ticker.C:
+			pods, err := lwPods.List(metav1.ListOptions{})
+			if err != nil {
+				s.log.Warn("Unable to list pods to populate Kubernetes cache", zap.Error(err))
+				return
```
danielhochman, line 111: why return? if pods list fails might deployments list succeed? also seems like we might want this at error severity, not a warn, given that there's no retry until the next run.

danielhochman (commented):
the joys of cache invalidation :smile:

backend/service/k8s/cache.go
mcutalo88: Ah yeah I had these in parallel before let me remove the returns.

backend/service/k8s/cache.go
mcutalo88: Yes true this is a problem, I think ill just do 50% of the topocache ttl time. If the ttl is 60min then will refresh every 30min, how does that sound?

mcutalo88: Actually the TTL for the cache is not yet honored i created a ticket to follow up with a fix for that and will include this in the todo. 

https://github.com/lyft/clutch/issues/1081

danielhochman: can we go ahead and fix that ticket first? it should be a pretty quick fix

mcutalo88: Yeah I can do that, just threading through the configuration to other services might be interesting unless I change the cacheable topology interface will see what i do on the other pr. Other than that it's straight forward.

backend/service/topology/topology.go
```diff
@@ -79,6 +79,11 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
 		return c, nil
 	}
 
+	// If TTL is not set default to two hours
+	if c.config.Cache.Ttl == nil {
+		c.config.Cache.Ttl = ptypes.DurationProto(time.Hour * 2)
+	}
```
danielhochman, line 85: instead of storing/modifying the config, please add a `ttl time.Duration` member to the struct.

danielhochman (approved):
looks good, please address comments before merge

PR #1070, opened 2021-02-23
experimentation: Envoy ECDS

kathan24:
### Description

Adding Envoy Extension Config Discovery Service in Clutch. In the future, faults will be injected via ECDS vs RTDS. 

### GitHub Issue

Fixes #


api/config/module/chaos/experimentation/xds/v1/xds.proto
```diff
@@ -26,4 +26,10 @@ message Config {
 
   // The interval at which to send heartbeat responses for TTL'd resources.
   google.protobuf.Duration heartbeat_interval = 6;
+
+  // list of clusters for which ECDS is enabled. * means it will be enabled for all
+  repeated string ecds_enabled_clusters = 7;
```
snowp, line 31: an alternative API that avoids the need for special strings would be 

```
message ECDSAllowList {
  repeated enabled_clusters = 1;
}

// Specifies which clusters ECDS is enabled for, causing faults to be communicated over ECDS instead of RTDS where supported. If unset, all ECDS will be used for all clusters.
ECDSAllowList ecds_allow_list = 7;

```

api/config/module/chaos/experimentation/xds/v1/xds.proto
```diff
@@ -26,4 +26,10 @@ message Config {
...
+
+  // A prefix to use to override runtime keys for ECDS
+  string ecds_fault_runtime_prefix = 8;
```
snowp, line 34: I'm not sure if its worth it making this configurable (just hard code to something nobody would ever configure?), but if are we should speak to its purpose a bit more

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,246 @@
...
+	"go.uber.org/zap"
+
+	proto "github.com/golang/protobuf/proto"
+
+	experimentation "github.com/lyft/clutch/backend/api/chaos/experimentation/v1"
+	serverexperimentation "github.com/lyft/clutch/backend/api/chaos/serverexperimentation/v1"
+)
+
+const (
+	HeaderXEnvoyDownstreamServiceCluster = `x-envoy-downstream-service-cluster`
```
snowp, line 25: nit: something like `EnvoyDownstreamServiceClusterHeader` might read better?

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,246 @@
...
+	ExternalFaultFilterConfigName        = `envoy.egress.extension_config.%s`
+	InternalFaultFilterConfigName        = `envoy.extension_config`
+	FaultFilterTypeURL                   = `type.googleapis.com/envoy.extensions.filters.http.fault.v3.HTTPFault`
+
+	DelayPercentRuntime    = "%s.http.delay.percentage"
+	DelayDurationRuntime   = "%s.http.delay.fixed_duration_ms"
+	AbortHttpStatusRuntime = "%s.http.abort.http_status"
+	AbortPercentRuntime    = "%s.http.abort.abort_percent"
```
snowp, line 33: I'm guessing most of these can be private?

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,246 @@
...
+		Denominator: gcpType.FractionalPercent_HUNDRED,
+	},
+}
+
+type ECDSConfig struct {
+	enabledClusters []string
+
+	faultRuntimePrefix string
+
+	requestedResourcesMap sync.Map
```
snowp, line 64: Looking at this I think this might not work because this would be a copy of the map used by the callbacks? An integration test that this works right would be awesome

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,246 @@
...
+			continue
+		}
+
+		isExternalFault, abort, delay, err = createAbortDelayConfig(httpFaultConfig, downstreamCluster)
+		if err != nil {
+			logger.Errorw("Unable to create ECDS filter fault", "config", httpFaultConfig)
+			continue
+		}
+
+		// We can only perform one fault test per cluster with ECDS. Hence we break.
```
snowp, line 94: Maybe log something in this case?

backend/module/chaos/serverexperimentation/xds/ecds_server.go
```diff
@@ -0,0 +1,42 @@
...
+		server: gcpServerV3.NewServer(ctx, config, callbacks),
+	}
+}
+
+func (e ecdsServer) StreamExtensionConfigs(stream extensionconfigservice.ExtensionConfigDiscoveryService_StreamExtensionConfigsServer) error {
+	return e.server.StreamHandler(stream, resource.ExtensionConfigType)
+}
+
+func (e ecdsServer) DeltaExtensionConfigs(extensionconfigservice.ExtensionConfigDiscoveryService_DeltaExtensionConfigsServer) error {
+	panic("implement me")
```
snowp, line 33: Maybe return an error here instead? This means that if we somehow get delta streams hitting clutch it would start crashing

backend/module/chaos/serverexperimentation/xds/xds.go
```diff
@@ -146,9 +157,12 @@ func (s *Server) newScopedStats(subScope string) serverStats {
 func (s *Server) Register(r module.Registrar) error {
 	PeriodicallyRefreshCache(s)
 	// RTDS V3 Server
-	rtdsServer := gcpServerV3.NewServer(s.ctx, s.snapshotCacheV3, &rtdsCallbacks{callbacksBase{s.newScopedStats("v3"),
+	rtdsServer := gcpServerV3.NewServer(s.ctx, s.snapshotCacheV3, &rtdsCallbacks{callbacksBase{s.newScopedStats("rtds"),
```
snowp, line 160: We'll want to update the dashboards after

backend/module/chaos/serverexperimentation/xds/xds.go
```diff
@@ -215,3 +229,47 @@ func (c *rtdsCallbacks) OnFetchRequest(context.Context, *gcpDiscoveryV3.Discover
...
+	return c.onStreamOpen(ctx)
+}
+
+func (c *ecdsCallbacks) OnStreamClosed(streamID int64) {
+	c.logger.Debugw("ECDS onStreamClosed", "streamID", streamID)
+	c.onStreamClosed(streamID)
+}
+
+func (c *ecdsCallbacks) OnStreamRequest(streamID int64, req *gcpDiscoveryV3.DiscoveryRequest) error {
+	c.requestedECDSResources.Store(req.Node.Cluster, req.ResourceNames)
```
snowp, line 255: Tracking this per cluster is a bit tricky, I'm not sure if this works as is. For example, if a service is deploying a change that adds a new ECDS resource, two independent Envoys might be sending requests with different sets of resources, causing this per cluster value to flap.

I'm also not sure if Envoy properly combines all ECDS resources into one request (it doesn't for EDS in certain cases), so we should probably verify that (integration test would be nice here). 

The simplest thing to do is just track all the values seen globally, but with some work we can get this working per cluster, lmk which one you prefer.

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -197,3 +228,13 @@ func getEnforcer(httpFaultConfig *serverexperimentation.HTTPFaultConfig) string
...
+
+func containsString(list []string, str string) bool {
+	for _, element := range list {
+		if element == str {
+			return true
+		}
+	}
+
+	return false
+}
```
snowp, line 240: Instead of scanning each time I would build a `map[string]struct{}` and just do a map lookup

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -105,8 +130,14 @@ func setSnapshot(resource []gcpTypes.ResourceWithTtl, cluster string, snapshotCa
 		return nil
 	}
 
-	logger.Infow("Setting snapshot", "cluster", cluster, "resource", resource)
-	snapshot := gcpCacheV3.NewSnapshotWithTtls(computedVersion, nil, nil, nil, nil, resource, nil)
+	snapshot := gcpCacheV3.Snapshot{}
+	logger.Infow("Setting snapshot", "cluster", cluster, "resource", resource, "isECDSResource", isECDSResource)
+	if isECDSResource {
```
snowp, line 135: maybe just pass through the type name? or even pass `map[string][]gcpTypes.ResourceWithTtl` to make this a bit more generic

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -64,10 +64,21 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
 		if _, exist := clusterFaultMap[cluster]; !exist {
 			logger.Debugw("Removing experiments for cluster", "cluster", cluster)
 
-			// in order to remove fault, we need to set the snapshot with empty runtime resource
-			emptyRuntimeResource := generateRTDSResource([]*experimentation.Experiment{}, rtdsConfig, ttl, logger)
+			var emptyResource []gcpTypes.ResourceWithTtl
+			ecdsClusterEnabled := containsString(ecdsConfig.enabledClusters, cluster) || containsString(ecdsConfig.enabledClusters, "*")
+			if ecdsClusterEnabled {
+				// in order to remove fault, we need to set the snapshot with default ecds config
+				emptyResource = generateEmptyECDSResource(cluster, ecdsConfig, logger)
```
snowp, line 71: Wouldn't we set this for all clusters? Otherwise clusters that have the ECDS filter configured but that have not been enabled in Clutch will be stuck waiting for the init timeout

snowp (changes requested):
Thanks for getting this started!

api/config/module/chaos/experimentation/xds/v1/xds.proto
```diff
@@ -26,4 +26,12 @@ message Config {
...
   // The interval at which to send heartbeat responses for TTL'd resources.
   google.protobuf.Duration heartbeat_interval = 6;
+
+  message ECDSAllowList {
+    repeated string enabled_clusters = 1;
+  }
+
+  // Specifies which clusters ECDS is enabled for, causing faults to be communicated over ECDS instead of RTDS where
+  // supported. If unset, ECDS will be used for all clusters.
+  ECDSAllowList ecds_allow_list = 7;
```
snowp, line 36: I might have suggested this originally but I think we'd want some kind of gating to prevent all configuration to suddenly start using ECDS. Maybe something like

```
message ECDSConfiguration {
  message AllowList {
    repeated string enabled_clusters = 1;
  }

  AllowList allow_list = 1;
}

EcdsConfiguration ecds = 1;
```

With `ecds: {}` meaning enable it for everything, and `ecds: { allow_list: { enabled_clusters: [...] } }` using a list of clusters? Alternatively we can oneof this with a bool always true param, but I'm not sure if that is much cleaner.

The tl;dr is that we don't want someone pulling in this change to switch everything to ECDS if they don't change their configuration..

backend/internal/test/integration/helper/envoytest/envoy.go
```diff
@@ -69,7 +69,7 @@ type EnvoyHandle struct{}
 
 // NewEnvoyHandle creates a new handle for the Envoy under test after waiting for it to initialize.
 func NewEnvoyHandle() (*EnvoyHandle, error) {
-	timeout := time.NewTimer(5 * time.Second)
+	timeout := time.NewTimer(40 * time.Second)
```
snowp, line 72: I don't think we need this long of a timeout?

backend/internal/test/integration/xds/cmd/envoyconfiggen/main.go
```diff
@@ -13,7 +13,7 @@ name: envoy.fault
 typed_config:
   '@type': type.googleapis.com/envoy.extensions.filters.http.fault.v3.HTTPFault
   abort:
-    http_status: 400
+    http_status: 503
```
snowp, line 16: Doesn't this make it hard to tell the difference between an injected fault and a 503 UC?

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,255 @@
...
+	"go.uber.org/zap"
+
+	experimentation "github.com/lyft/clutch/backend/api/chaos/experimentation/v1"
+	serverexperimentation "github.com/lyft/clutch/backend/api/chaos/serverexperimentation/v1"
+)
+
+const (
+	EnvoyDownstreamServiceClusterHeader   = `x-envoy-downstream-service-cluster`
+	FaultFilterConfigNameForExternalFault = `envoy.egress.extension_config.%s`
+	FaultFilterConfigNameForInternalFault = `envoy.extension_config`
```
snowp, line 26: These can all be private I would think? same with a lot of the other variables in this file I believe

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,255 @@
...
+		AbortPercentRuntime:    abortPercentRuntime,
+	}
+
+	var faultFilterName string
+	if isExternalFault {
+		faultFilterName = fmt.Sprintf(FaultFilterConfigNameForExternalFault, upstreamCluster)
+	} else {
+		faultFilterName = FaultFilterConfigNameForInternalFault
+
+		// match downstream cluster with request header else fault will be applied to requests coming from all downstream
```
snowp, line 118: This is a bit hard to parse if you don't already know what's going on, how about something like `We match against the EnvoyDownstreamServiceCluster header to only apply these faults to the specific downstream cluster`?

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	if _, exists := ecdsConfig.ecdsResourceMap.requestedResourcesMap[cluster]; !exists {
+		logger.Debugw("Cluster not found in ECDS resource map", "cluster", cluster)
+		return []gcpTypes.ResourceWithTtl{}
+	}
+
+	ecdsConfig.ecdsResourceMap.mu.Lock()
+	resourceNames := ecdsConfig.ecdsResourceMap.requestedResourcesMap[cluster]
+	ecdsConfig.ecdsResourceMap.mu.Unlock()
```
snowp, line 174: might be cleaner to expose this as functions on the safe resource map so callers don't have to worry about locking?

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,255 @@
...
+		Abort: DefaultAbortFaultConfig,
+
+		// override runtimes so that default runtime is not used.
+		DelayPercentRuntime:    delayPercentRuntime,
+		DelayDurationRuntime:   delayDurationRuntime,
+		AbortHttpStatusRuntime: abortHttpStatusRuntime,
+		AbortPercentRuntime:    abortPercentRuntime,
+	}
+
+	serializedFaultFilter, err := proto.Marshal(faultFilter)
```
snowp, line 161: Should we just store this in a var somewhere? No need to do this every time if the config is identical

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,255 @@
...
+	ecdsResourceMap *SafeEcdsResourceMap
+}
+
+func generateECDSResource(experiments []*experimentation.Experiment, ttl *time.Duration, logger *zap.SugaredLogger) []gcpTypes.ResourceWithTtl {
+	var downstreamCluster string
+	var upstreamCluster string
+	var isExternalFault = false
+
+	abort := DefaultAbortFaultConfig
+	delay := DefaultDelayFaultConfig
```
snowp, line 76: Seems like createAbortDelay already handles this default logic, so I don't think we need these?

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,255 @@
...
+					Seconds: int64(httpFaultConfig.GetLatencyFault().GetLatencyDuration().GetFixedDurationMs()),
+				},
+			},
+			Percentage: &gcpType.FractionalPercent{
+				Numerator:   httpFaultConfig.GetLatencyFault().GetPercentage().GetPercentage(),
+				Denominator: gcpType.FractionalPercent_HUNDRED,
+			},
+		}
+	default:
+		return false, abort, delay, fmt.Errorf("unknown fault type %v", httpFaultConfig)
```
snowp, line 251: I would return nil instead of a default abort/delay in the error cases

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -77,36 +82,59 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
...
 		logger.Debugw("Injecting fault for cluster", "cluster", cluster)
 
-		runtimeResource := generateRTDSResource(experiments, rtdsConfig, ttl, logger)
-		err := setSnapshot(runtimeResource, cluster, snapshotCache, logger)
+		// Enable ECDS if cluster present in the map
+		if _, exists := ecdsConfig.enabledClusters[cluster]; exists {
+			resources[gcpTypes.ExtensionConfig] = generateECDSResource(experiments, ttl, logger)
+
+			// generate empty Runtime config to clear any existing RTDS faults
+			resources[gcpTypes.Runtime] = generateRTDSResource([]*experimentation.Experiment{}, rtdsConfig, ttl, logger)
```
snowp, line 94: Seems like we don't actually implement the logic for enabling this for everything? That's probably okay for now though we'll want to update the proto docs in that case

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -77,36 +82,59 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
...
-		currentSnapshotVersion = currentSnapshot.GetVersion(cluster)
+		currentSnapshotVersion, _ = computeChecksum(currentSnapshot)
+	}
+
+	snapshot := gcpCacheV3.Snapshot{}
+	for resourceType, resources := range resourceMap {
+		computedVersion, err := computeChecksum(resources)
+		if err != nil {
+			continue
+		}
```
snowp, line 120: Could we make this not fail silently in some way? Logs, errors,  etc

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -77,36 +82,59 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
...
-	if err != nil {
-		return err
-	}
-
+func setSnapshot(resourceMap map[gcpTypes.ResponseType][]gcpTypes.ResourceWithTtl, cluster string, snapshotCache gcpCacheV3.SnapshotCache, logger *zap.SugaredLogger) error {
 	currentSnapshotVersion := ""
 	currentSnapshot, err := snapshotCache.GetSnapshot(cluster)
 	if err == nil {
-		currentSnapshotVersion = currentSnapshot.GetVersion(cluster)
+		currentSnapshotVersion, _ = computeChecksum(currentSnapshot)
```
snowp, line 112: Seems unnecessary to recompute the checksum here? Would it be cleaner to compute the checksum of each resource type and set the snapshot if any of them are different?

backend/module/chaos/serverexperimentation/xds/xds_cache_test.go
```diff
@@ -279,3 +291,131 @@ func TestComputeVersionReturnValue(t *testing.T) {
...
+		upstream, downstream, err := getClusterPair(config)
+		assert.NoError(t, err)
+
+		downstreamCluster = downstream
+
+		// Test Internal Fault
+		if upstream == testCluster {
+			testClusterFaults = append(testClusterFaults, experiment)
+
+			// Since ECDS can only perform one experiment per cluster
```
snowp, line 318: This comment seems incomplete, could you elaborate?

backend/module/chaos/serverexperimentation/xds/xds_cache_test.go
```diff
@@ -279,3 +291,131 @@ func TestComputeVersionReturnValue(t *testing.T) {
...
+	testCache := gcpCacheV3.NewSnapshotCache(false, gcpCacheV3.IDHash{}, nil)
+	testCluster := "serviceA"
+	var downstreamCluster string
+	mockExperimentList := mockGenerateFaultData(t)
+
+	var testClusterFaults []*experimentation.Experiment
+	for _, experiment := range mockExperimentList {
+		config := &serverexperimentation.HTTPFaultConfig{}
+		err := ptypes.UnmarshalAny(experiment.GetConfig(), config)
+		if err != nil {
```
snowp, line 305: `assert.NoError`

backend/module/chaos/serverexperimentation/xds/xds_cache_test.go
```diff
@@ -279,3 +291,131 @@ func TestComputeVersionReturnValue(t *testing.T) {
...
+		t.Errorf("Unable to unmarshall typed config")
+	}
+
+	assert.EqualValues(t, 10, httpFaultFilter.GetAbort().GetPercentage().GetNumerator())
+	assert.EqualValues(t, 404, httpFaultFilter.GetAbort().GetHttpStatus())
+	assert.EqualValues(t, 0, httpFaultFilter.GetDelay().GetPercentage().GetNumerator())
+	assert.EqualValues(t, 0, httpFaultFilter.GetDelay().GetFixedDelay().GetSeconds())
+	assert.EqualValues(t, 1, len(httpFaultFilter.GetHeaders()))
+	assert.EqualValues(t, EnvoyDownstreamServiceClusterHeader, httpFaultFilter.GetHeaders()[0].GetName())
+	assert.EqualValues(t, downstreamCluster, httpFaultFilter.GetHeaders()[0].GetExactMatch())
```
snowp, line 357: Might find that proto equality of a fully constructed proto ends up being easier to read + ensures that we're not setting fields that we don't expect to set

backend/module/chaos/serverexperimentation/xds/xds_integration_test.go
```diff
@@ -111,3 +122,40 @@ func awaitExpectedReturnValueForSimpleCall(t *testing.T, e *envoytest.EnvoyHandl
 
 	return nil
 }
+
+func TestEnvclearoyECDSFaults(t *testing.T) {
```
snowp, line 126: test name seems messed up

backend/module/chaos/serverexperimentation/xds/xds_integration_test.go
```diff
@@ -111,3 +122,40 @@ func awaitExpectedReturnValueForSimpleCall(t *testing.T, e *envoytest.EnvoyHandl
...
+
+	experiment := createTestExperiment(t, 404, ts.Storer)
+
+	err = awaitExpectedReturnValueForSimpleCall(t, e, awaitReturnValueParams{
+		timeout:        2 * time.Second,
+		expectedStatus: 404,
+	})
+	assert.NoError(t, err, "did not see faults enabled")
+
+	// Not testing TTL by stopping the server as its currently not supported for ECDS in Upstream Envoy
```
snowp, line 153: This comment seems a bit odd because it assumes the reader was expecting to see a TTL test here which they probably don't, maybe reword a bit? Maybe just turn this into a TODO

api/config/module/chaos/experimentation/xds/v1/xds.proto
snowp: Having read the rest of the code we could also just document that this doesn't have a "allow all" mode at the moment and leave the API as is

snowp (changes requested):

backend/internal/test/integration/helper/envoytest/envoy.go
kathan24: In fact, I had to increase the timeout to 60 seconds. Not sure where the time is used

backend/module/chaos/serverexperimentation/xds/ecds.go
kathan24: if there is an error, would it not be better to send default configs (which will revert the fault)?

backend/module/chaos/serverexperimentation/xds/ecds.go
kathan24: By storing, do you mean a global list of `faultFilters`? I think it might be an overkill

backend/module/chaos/serverexperimentation/xds/ecds.go
kathan24: This is just a safety mechanism incase `experiments` are empty. In that case, we need some default

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,302 @@
...
+				break
+			}
+		}
+
+		if isMissing {
+			missingResources = append(missingResources, newResource)
+		}
+	}
+
+	safeResource.requestedResourcesMap[cluster] = append(safeResource.requestedResourcesMap[cluster], missingResources...)
```
snowp, line 99: Should we just use a map[string]struct[] here? That way we could just unconditionally add all the resources

backend/module/chaos/serverexperimentation/xds/ecds.go
```diff
@@ -0,0 +1,302 @@
...
+	var isExternalFault bool
+	var abort *gcpFilterFault.FaultAbort
+	var delay *gcpFilterCommon.FaultDelay
+
+	switch httpFaultConfig.GetFault().(type) {
+	case *serverexperimentation.HTTPFaultConfig_AbortFault:
+		switch httpFaultConfig.GetFaultTargeting().GetEnforcer().(type) {
+		case *serverexperimentation.FaultTargeting_DownstreamEnforcing:
+			// Abort External Fault
+			isExternalFault = true
```
snowp, line 253: It seems a bit odd for someone not familiar with the lyft use case to see external used here since the API just talks about downstream vs upstream enforcing. Maybe be consistent and call this "isEgressFault" or something?

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -62,12 +62,17 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
...
 	for _, cluster := range snapshotCache.GetStatusKeys() {
 		if _, exist := clusterFaultMap[cluster]; !exist {
+			emptyResources := make(map[gcpTypes.ResponseType][]gcpTypes.ResourceWithTtl)
+
 			logger.Debugw("Removing experiments for cluster", "cluster", cluster)
 
+			// in order to remove fault, we need to set the snapshot with default ecds config
+			emptyResources[gcpTypes.ExtensionConfig] = generateEmptyECDSResource(cluster, ecdsConfig, logger)
+
 			// in order to remove fault, we need to set the snapshot with empty runtime resource
```
snowp, line 72: Combine these two comments into one?

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -77,37 +82,56 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
...
+
 		logger.Debugw("Injecting fault for cluster", "cluster", cluster)
 
-		runtimeResource := generateRTDSResource(experiments, rtdsConfig, ttl, logger)
-		err := setSnapshot(runtimeResource, cluster, snapshotCache, logger)
+		// Enable ECDS if cluster present in the map
+		if _, exists := ecdsConfig.enabledClusters[cluster]; exists {
+			resources[gcpTypes.ExtensionConfig] = generateECDSResource(experiments, cluster, ttl, logger)
+
+			// generate empty Runtime config to clear any existing RTDS faults
```
snowp, line 93: This is also to provide an initial RTDS response to clients booting up right? Maybe add to comment or be less specific

snowp (changes requested):
Thanks this is looking really good! Just a few more comments

snowp (approved):
LGTM, thanks!

danielhochman (approved):

PR #1076, opened 2021-02-25
k8s: extend configmap definition to include data

rithujohn191:
### Description
Extend configmap proto definition to include Data field which is a simple key-value map

### Testing Performed
- unit tests


api/k8s/v1/k8s.proto
```diff
@@ -729,6 +729,10 @@ message ConfigMap {
 
   map<string, string> annotations = 4;
   map<string, string> labels = 5;
+
+  // represents a configmaps data field which is simple
+  // key:value map
+  google.protobuf.Struct data = 6;
```
danielhochman, line 735: is there a reason it's not `map<string, string>`?

also checked the underlying type https://github.com/kubernetes/api/blob/65a259a606b34fbfa8861674531aab27da4f1381/core/v1/types.go#L5895

rithujohn191: The thought here was that when a user needs to provide this value (say when they are creating a configmap), they will have to convert the config from json to a map[string]string. So was trying to avoid the extra conversion step for the user.
Also read that Maps cannot be repeated, optional, or required (https://developers.google.com/protocol-buffers/docs/overview#maps_features) and figured that we might want to keep this field optional

rithujohn191: If we are okay with making the user provide us a map for the data field I could change it to a `map<string, string>`

danielhochman: ah i think i see, the user can provide `foo: {bar: baz}` and the resulting configmap is `{foo.bar: baz}`?

rithujohn191: Yep for example the following are some possible values for the Data field in a configmap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: game-demo
data:
  # property-like keys; each key maps to a simple value
  player_initial_lives: "3"
  ui_properties_file_name: "user-interface.properties"

  # file-like keys
  game.properties: |
    enemy.types=aliens,monsters
    player.maximum-lives=5    
  user-interface.properties: |
    color.good=purple
    color.bad=yellow
    allow.textmode=true
```
So its not truly always just a single string mapping to another string value

danielhochman: ok i think the proper type in this case is actually `map<string, protobuf.Value>` which enforces the root type is always a map.

i don't think you can have 
```yaml
data:
  - foo
  - bar
```

rithujohn191: Ahhh ok you might be right. Let me test that out

rithujohn191: Updated. I think this covers all the cases

backend/service/k8s/configmap.go
```diff
@@ -71,11 +72,21 @@ func protoForConfigMap(cluster string, k8sconfigMap *v1.ConfigMap) *k8sapiv1.Con
...
 		Namespace:   k8sconfigMap.Namespace,
 		Name:        k8sconfigMap.Name,
 		Labels:      k8sconfigMap.Labels,
 		Annotations: k8sconfigMap.Annotations,
 	}
+	// optional fields
+	if k8sconfigMap.Data != nil {
+		dataMap := make(map[string]*structpb.Value, len(k8sconfigMap.Data))
+		for k, v := range k8sconfigMap.Data {
+			dataMap[k], _ = structpb.NewValue(v)
```
rithujohn191, line 87: @danielhochman is it okay to absorb this error here?

danielhochman: in this case, you can call `NewStringValue` which doesn't return error.

though we should document why we're not unpacking the complex types though and just using the flat string. i'm actually not sure it's the best idea. if we want to be able to read, update, write then the read ideally would return the same format that we end up using for the write. maybe i'm missing something?

rithujohn191: Reading up a little, realized that we only need to support the map[string]string type for data, its upto the user to pack other values as string. So will go ahead and change the type also also add BinaryData as a field in the proto. Thanks for being patient with this one

rithujohn191: Updated 👍

danielhochman (approved):
:ship:

PR #1077, opened 2021-02-25
topology: simplify leader elect db connection

mcutalo88:
### Description
Fix a bug where we would leak database connections. We explicitly call `conn.Close()` and remove the logic to manually unlock advisory locks as they are automatically released when the connection is severed.

This makes it simpler to reason about and the connection logic more clear.

### Testing Performed

locally.

### DIFF
--- a/backend/service/topology/cache.go
+++ b/backend/service/topology/cache.go
@@ -43,14 +43,14 @@ func (c *client) acquireTopologyCacheLock(ctx context.Context) {
 		default:
 			c.log.Error("lost connection to database, trying to reconnect...", zap.Error(err))
 		}
+		// Closes the db connection which will also release the advisory lock
+		conn.Close()
 
 		select {
 		case <-ticker.C:
 			continue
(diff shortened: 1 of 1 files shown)


scarlettperry (approved):

backend/service/topology/cache.go
```diff
@@ -43,14 +43,13 @@ func (c *client) acquireTopologyCacheLock(ctx context.Context) {
 		default:
 			c.log.Error("lost connection to database, trying to reconnect...", zap.Error(err))
 		}
+		conn.Close()
```
danielhochman, line 46: add comment saying it releases lock

danielhochman (approved):

backend/service/topology/cache.go
mcutalo88: I actually already have one when I open the conn a 20 lines up or so

mcutalo88: i can add an extra one here also tho

PR #1079, opened 2021-02-25
k8s: full error detail preservation

danielhochman:
### Description
This preserves everything that is included in K8s error information into the details field of the converted grpc.Status.

Due to the zaniness that is K8s protos, we had to create our own analogous type for the K8s status since it panics when serializing to Any due to pointer/value mismatches in the results.

### Testing Performed
Unit


backend/service/k8s/errors_test.go
```diff
@@ -58,3 +61,51 @@ func TestConvertError(t *testing.T) {
...
+
+	err := ConvertError(&k8serrors.StatusError{ErrStatus: ks})
+
+	s, ok := status.FromError(err)
+	assert.True(t, ok)
+	assert.NotNil(t, s)
+	assert.Len(t, s.Details(), 1)
+
+	ret := s.Details()[0].(*k8sv1.Status)
+	assert.EqualValues(t, ks.Message, ret.Message)
```
rithujohn191, line 99: 👏 Thanks for adding thorough testing

PR #1080, opened 2021-02-25
k8s: always return a gRPC status

danielhochman:
### Description
Replaces all uses of `fmt.Errorf` and `errors.New` in the `k8s` service and module with their proper gRPC status and code.

This will ensure accurate metrics and sensible error information is displayed to the user.

### Testing Performed
Manual audit.


rithujohn191 (approved):
💯

PR #1083, opened 2021-02-26
core/topology/aws/k8s: Support proto patterns as valid search input

mcutalo88:
### Description

Support for utilizing proto patterns as valid search input for a resource.

To support this change this PR does the following.
* Update the topology cache to utilize the patterns as the new `id` instead of name, enabling autocomplete.
* All resolver apis have been updated to parse search queries as patterns, if successfully parsed a targeted looked up is performed which should be faster in cases when we would otherwise fanned out.

### Testing Performed
Unit / Local

### TODOs
- [x] revisit instance id search resolver
- [x] revisit k8s clientset code

### DIFF
--- a/backend/gateway/meta/meta.go
+++ b/backend/gateway/meta/meta.go
@@ -155,7 +155,7 @@ func HydratedPatternForProto(pb proto.Message) string {
 //  namespace: mynamespace
 //  name: nameofresource
 // }
-func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]string, error) {
+func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]string, bool, error) {
 	m := pb.ProtoReflect()
 	opts := m.Descriptor().Options().ProtoReflect()
 
@@ -174,7 +174,7 @@ func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]
 			convertedRegex := fmt.Sprintf("^%s$", fieldNameRegexp.ReplaceAllString(pattern.Pattern, "(.*)"))
 			patternRegex, err := regexp.Compile(convertedRegex)
 			if err != nil {
-				return nil, err
+				return nil, false, err
 			}
 
 			// Extract the regex groups, index 0 is always the input string
@@ -187,7 +187,7 @@ func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]
 			}
 		}
 	}
-	return result, nil
+	return result, true, nil
 }
 
 func extractProtoPatternFieldNames(pattern *apiv1.Pattern) []string {
--- a/backend/gateway/meta/meta.go
+++ b/backend/gateway/meta/meta.go
@@ -120,7 +120,7 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
 }
 
 // HydratedPatternForProto takes a proto and returns its pattern populated with values
-func HydratedPatternForProto(pb proto.Message) string {
+func HydratedPatternForProto(pb proto.Message) (string, error) {
 	m := pb.ProtoReflect()
 	opts := m.Descriptor().Options().ProtoReflect()
 
@@ -137,11 +137,10 @@ func HydratedPatternForProto(pb proto.Message) string {
 
 		// At the time of writing there is only support for a single pattern
 		// this list should only have one item to return
-		return populatedPattern[0]
+		return populatedPattern[0], nil
 	}
 
-	// todo: rethink error case
-	return ""
+	return "", fmt.Errorf("the supplied proto does not have a pattern: [%T]", pb)
 }
 
 // ExtractPatternValuesFromString takes a string value and maps the patterns from a proto pattern
--- a/backend/gateway/meta/meta.go
+++ b/backend/gateway/meta/meta.go
@@ -187,6 +187,14 @@ func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]
 			}
 		}
 	}
+
+	// Check that all of the fields have values
+	for _, value := range result {
+		if len(value) == 0 {
+			return result, false, nil
+		}
+	}
+
 	return result, true, nil
 }
 
--- a/backend/gateway/meta/meta.go
+++ b/backend/gateway/meta/meta.go
@@ -168,7 +168,7 @@ func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]
 
 		for _, pattern := range id.Patterns {
 			// The variable names on the pattern
-			variableNames := fieldNameRegexp.FindAllStringSubmatch(pattern.Pattern, -1)
+			patternFields := extractProtoPatternFieldNames(pattern)
 
 			// Convert the pattern into a regex
 			converedRegex := fmt.Sprintf("^%s$", fieldNameRegexp.ReplaceAllString(pattern.Pattern, "(.*)"))
@@ -177,17 +177,26 @@ func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]
 				return nil, err
 			}
 
-			if patternRegex.MatchString(value) {
-				// Extract the regex groups, index 0 is always the input string
-				subStringGroups := patternRegex.FindAllStringSubmatch(value, -1)
-				for i, name := range variableNames {
+			// Extract the regex groups, index 0 is always the input string
+			subStringGroups := patternRegex.FindAllStringSubmatch(value, -1)
+			if subStringGroups != nil {
+				for i, name := range patternFields {
 					// Plus one here because the first value is the input string
-					result[name[1]] = subStringGroups[0][i+1]
+					result[name] = subStringGroups[0][i+1]
 				}
 			}
 		}
 	}
 	return result, nil
+}
+
+func extractProtoPatternFieldNames(pattern *apiv1.Pattern) []string {
+	results := []string{}
+	variableNames := fieldNameRegexp.FindAllStringSubmatch(pattern.Pattern, -1)
+	for _, name := range variableNames {
+		results = append(results, name[1])
+	}
+	return results
 }
 
 func resolveField(pb proto.Message, name string) []*auditv1.Resource {
--- a/backend/gateway/meta/meta.go
+++ b/backend/gateway/meta/meta.go
@@ -2,7 +2,6 @@ package meta
 
 import (
 	"fmt"
-	"log"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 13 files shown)


backend/resolver/k8s/k8s.go
```diff
@@ -174,36 +174,66 @@ func (r *res) Search(ctx context.Context, typeURL, query string, limit uint32) (
...
 		if idPattern.MatchString(query) {
-			for _, name := range clientsets {
+			mappedValues, err := meta.PatternValueMapping(&k8sv1api.Pod{}, query)
+			if err == nil && len(mappedValues["cluster"]) > 0 && len(mappedValues["namespace"]) > 0 && len(mappedValues["name"]) > 0 {
 				handler.Add(1)
-				go func(name string) {
+				go func(cluster, namespace, name string) {
 					defer handler.Done()
-					pod, err := r.svc.DescribePod(ctx, name, "", metav1.NamespaceAll, query)
+					pod, err := r.svc.DescribePod(ctx, cluster, cluster, namespace, name)
```
mcutalo88, line 182: double check the `clientset` here, I dont want to fanout again here if we have the cluster information but this relies on the `clientset` == `cluster`.

backend/gateway/meta/meta.go
```diff
@@ -119,6 +119,77 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
...
+	m := pb.ProtoReflect()
+	opts := m.Descriptor().Options().ProtoReflect()
+
+	populatedPattern := []string{}
+
+	if opts.Has(identifierTypeDescriptor) {
+		v := opts.Get(identifierTypeDescriptor)
+		id := v.Message().Interface().(*apiv1.Identifier)
+
+		for _, pattern := range id.Patterns {
```
mcutalo88, line 133: Check support for more than one pattern, currently I believe its only 1.

mcutalo88:
I still have a few items to circle back on, but this is in a state where it can start receiving reviews.

backend/gateway/meta/meta.go
```diff
@@ -119,6 +119,77 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
...
 }
 
+// ExtractProtoPatternsValues takes a proto and returns its pattern populated with values
+func ExtractProtoPatternsValues(pb proto.Message) string {
+	m := pb.ProtoReflect()
+	opts := m.Descriptor().Options().ProtoReflect()
+
+	populatedPattern := []string{}
+
+	if opts.Has(identifierTypeDescriptor) {
```
danielhochman, line 129: can/should we share any code with `func ResourceNames`?

backend/gateway/meta/meta.go
```diff
@@ -119,6 +119,77 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
 	return nil
 }
 
+// ExtractProtoPatternsValues takes a proto and returns its pattern populated with values
+func ExtractProtoPatternsValues(pb proto.Message) string {
```
danielhochman, line 123: `PopulatedPatternForProto`, `HydratedPatternForProto` or similar

backend/gateway/meta/meta.go
```diff
@@ -119,6 +119,77 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
...
+	// Field and Value result map
+	result := map[string]string{}
+
+	if opts.Has(identifierTypeDescriptor) {
+		v := opts.Get(identifierTypeDescriptor)
+		id := v.Message().Interface().(*apiv1.Identifier)
+
+		for _, pattern := range id.Patterns {
+			// The variable names on the pattern
+			variableNames := fieldNameRegexp.FindAllStringSubmatch(pattern.Pattern, -1)
```
danielhochman, line 171: i would pull this out into a func that returns `[]string` and use everywhere we're doing this

backend/gateway/meta/meta.go
```diff
@@ -119,6 +119,77 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
...
+
+			// Convert the pattern into a regex
+			converedRegex := fieldNameRegexp.ReplaceAllString(pattern.Pattern, "(.*)")
```
danielhochman, line 174: i think you also need a `^` and `$` at the beginning and end of the converted regex to avoid partial match.

e.g. `(.*?)/(.*?)` matching `foo/bar/baz`

danielhochman (commented):
:fish_cake:

backend/resolver/aws/asg.go
```diff
@@ -22,20 +24,37 @@ func (r *res) resolveAutoscalingGroupsForInput(ctx context.Context, input proto.
...
-			groups, err := r.client.DescribeAutoscalingGroups(ctx, region, ids)
-			select {
-			case handler.Channel() <- resolver.NewFanoutResult(groups, err):
-				return
-			case <-handler.Cancelled():
-				return
+	for _, id := range ids {
+		// if the pattern matches then proceeded
+		mappedValues, err := meta.PatternValueMapping(&ec2v1.AutoscalingGroup{}, id)
+		if err == nil && len(mappedValues["region"]) > 0 && len(mappedValues["name"]) > 0 {
```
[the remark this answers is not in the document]
mcutalo88: Ah yeah that makes more sense and I wont have to have these long if statements to check if the value is present.

backend/gateway/meta/meta.go
mcutalo88: Yeah let me see what we can share

backend/gateway/meta/meta.go
mcutalo88: @danielhochman is it safe to assume that as it currently stands you can only have one pattern?

mcutalo88:
Working on a few of the feedbacks here but local testing and refactoring have gone really well and its getting close to its final state, so requesting another review.

backend/gateway/meta/meta.go
mcutalo88: Looking at this more both use cases for this have slightly different needs, i would need to return the raw output to satisfy both. Leaving this func with just a single line.

```
func extractProtoPatternFieldNames(pattern *apiv1.Pattern) [][]string {
	return  fieldNameRegexp.FindAllStringSubmatch(pattern.Pattern, -1)
}
```

`resolverPattern` is the other usage 
https://github.com/lyft/clutch/blob/f5e5ce7ea88be003e99dcc1fffffcabe5401ad6a/backend/gateway/meta/meta.go#L246

danielhochman (commented):
looking good, a couple of questions

backend/resolver/aws/asg.go
```diff
@@ -22,20 +24,40 @@ func (r *res) resolveAutoscalingGroupsForInput(ctx context.Context, input proto.
...
-				return
-			case <-handler.Cancelled():
-				return
+	for _, id := range ids {
+		patternValues, ok, err := meta.ExtractPatternValuesFromString((*ec2v1.AutoscalingGroup)(nil), id)
+		if err != nil {
+			return nil, err
+		}
+
+		if ok && len(patternValues["region"]) > 0 && len(patternValues["name"]) > 0 {
```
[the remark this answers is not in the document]
mcutalo88: yeah missed this one ty.

backend/gateway/meta/meta.go
```diff
@@ -119,6 +119,99 @@ func ResourceNames(pb proto.Message) []*auditv1.Resource {
...
+	if opts.Has(identifierTypeDescriptor) {
+		v := opts.Get(identifierTypeDescriptor)
+		id := v.Message().Interface().(*apiv1.Identifier)
+
+		for _, pattern := range id.Patterns {
+			// The variable names on the pattern
+			patternFields := extractProtoPatternFieldNames(pattern)
+
+			// Convert the pattern into a regex
+			convertedRegex := fmt.Sprintf("^%s$", fieldNameRegexp.ReplaceAllString(pattern.Pattern, "(.*)"))
```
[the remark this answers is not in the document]
mcutalo88: I have a test for this case, it's technically valid but I dont currently think it's a problem. If we get to a place where multiple patterns are valid then this could get interesting.

backend/resolver/aws/instances.go
```diff
@@ -38,19 +40,44 @@ func (r *res) resolveInstancesForInput(ctx context.Context, input proto.Message)
...
+				go func(region string) {
+					defer handler.Done()
+					instances, err := r.client.DescribeInstances(ctx, region, []string{normalizedId})
+					select {
+					case handler.Channel() <- resolver.NewFanoutResult(instances, err):
+						return
+					case <-handler.Cancelled():
+						return
+					}
+				}(region)
```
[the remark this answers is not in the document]
mcutalo88: Yeah true this is a bit gnarly, let me refactor a bit.

backend/gateway/meta/meta.go
mcutalo88: talked in person and we only support one pattern for now, left a comment below as to why we assume there is only one item in the list.

backend/resolver/aws/asg.go
```diff
@@ -22,20 +24,32 @@ func (r *res) resolveAutoscalingGroupsForInput(ctx context.Context, input proto.
...
-			select {
-			case handler.Channel() <- resolver.NewFanoutResult(groups, err):
-				return
-			case <-handler.Cancelled():
-				return
-			}
-		}(region)
-	}
+	for _, id := range ids {
+		patternValues, ok, err := meta.ExtractPatternValuesFromString((*ec2v1.AutoscalingGroup)(nil), id)
```
danielhochman, line 28: looking at this again, this call belongs on :138 in the Search function, as it doesn't apply to the Resolve() path, only the Search() path.

i.e.
```golang
case typeURLAutoscalingGroup:
  patternValues, ok, err := meta.ExtractPatternValuesFromString((*ec2v1.AutoscalingGroup)(nil), id)
  if ok {
    return r.autoscalingGroupResults(ctx, patternValues["region"], []string{patternValues["name"]}, limit)
  }
  return r.autoscalingGroupResults(ctx, resolver.OptionAll, []string{query}, limit)
```

same probably goes for other uses of the extract func

backend/resolver/k8s/k8s.go
```diff
@@ -174,36 +174,66 @@ func (r *res) Search(ctx context.Context, typeURL, query string, limit uint32) (
...
+			for _, clientset := range clientsets {
+				namespace := metav1.NamespaceAll
+				podQuery := query
+				cluster := ""
+
+				if ok {
+					namespace = patternValues["namespace"]
+					podQuery = patternValues["name"]
+					cluster = patternValues["cluster"]
+				}
```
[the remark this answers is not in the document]
mcutalo88: Looking at this again I dont remember why I ever put it in the loop, let me clean it up.

danielhochman (approved):
:ship:

PR #1089, opened 2021-02-28
slacksink: Support custom slack messages in slack config

scarlettperry:
### Description
PR
* updates the slack config to allow users to specify a custom slack message for a `/service/method`
* adds a function that returns the custom slack message from the config if preconditions are met
* adds unit test for ^

A separate PR https://github.com/lyft/clutch/pull/1093 has the transformer to parse the custom text from the config with metadata from the audit event that returns a formatted slack message.

### Testing Performed
Unit test and locally


backend/clutch-config.yaml
```diff
@@ -35,6 +35,19 @@ modules:
   - name: clutch.module.k8s
   - name: clutch.module.kinesis
 services:
+  - name: clutch.service.auditsink.slack
```
scarlettperry, line 38: will remove these changes before merging. showing an example of how the set up will look like in the `clutch-config`

backend/service/auditsink/slack/slack_helper.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	auditv1 "github.com/lyft/clutch/backend/api/audit/v1"
+	configv1 "github.com/lyft/clutch/backend/api/config/service/auditsink/slack/v1"
+	"github.com/lyft/clutch/backend/middleware"
+)
+
+// GetSlackOverrideText returns the custom slack message from the slack override config if the FullMethod matches
+// the serive and method in the audit event. Otherwise ok is false and an empty string is returned.
+func GetSlackOverrideText(override *configv1.Override, event *auditv1.RequestEvent) (string, bool) {
```
scarlettperry, line 11: Following the pattern for how we decide if we write an audit event to the slack sink. For example, we check if the filter specified in the config matches given values in the audit event (https://github.com/lyft/clutch/blob/main/backend/service/auditsink/slack/slack.go#L52 -> https://github.com/lyft/clutch/blob/main/backend/service/auditsink/sink.go#L19).

Similarly here, I check if the FullMethod in the slack override config matches the service and method in the audit event.

backend/service/auditsink/slack/slack_helper.go
```diff
@@ -0,0 +1,28 @@
...
+	if override == nil {
+		// no slack overrides
+		return "", false
+	}
+
+	for _, customSlack := range override.CustomSlackMessages {
+		service, method, ok := middleware.SplitFullMethod(customSlack.FullMethod)
```
scarlettperry, line 18: https://github.com/lyft/clutch/blob/4d0b7faab3f2b5cc4c5b6bb9de84eb52b174d393/backend/middleware/middleware.go#L19

api/config/service/auditsink/slack/v1/slack.proto
```diff
@@ -13,4 +13,17 @@ message SlackConfig {
   string channel = 2 [ (validate.rules).string = {min_bytes : 1} ];
 
   clutch.config.service.audit.v1.Filter filter = 3;
+  Override override = 4;
```
danielhochman, line 16: should this be flattened to `repeated CustomSlackMessage overrides`? do you anticipate adding anything else to Override?

api/config/service/auditsink/slack/v1/slack.proto
```diff
@@ -13,4 +13,17 @@ message SlackConfig {
...
+}
+
+message CustomSlackMessage {
```
danielhochman, line 19: nit: package name repeats
```suggestion
message CustomMessage {
```

api/config/service/auditsink/slack/v1/slack.proto
```diff
@@ -13,4 +13,17 @@ message SlackConfig {
...
+  // The full method in the format of `/SERVICE/METHOD`
+  string full_method = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  // The custom slack message for `/SERVICE/METHOD`
```
danielhochman, line 22: i would give slightly more detail in the doc here about how it's a template. obviously can write much more thorough docs once we settle on the approach for templating so maybe a TODO works as well.

also a link to slack's formatting docs would be nice.

backend/service/auditsink/slack/slack_helper.go
danielhochman, line 11: let's hang this method off of the sink and store the overrides on the sink's struct.

danielhochman, line 11: nit in general try to avoid repeating the package name in its members
```suggestion
func GetOverrideText(override *configv1.Override, event *auditv1.RequestEvent) (string, bool) {
```

danielhochman: also method can be private

danielhochman (commented):
minor refactoring comments, good overall

backend/service/auditsink/slack/slack_helper.go
scarlettperry: wanted to keep it public so that we could use it in the internal implementation

api/config/service/auditsink/slack/v1/slack.proto
scarlettperry: flattened
don't plan on adding anything else to Override

api/config/service/auditsink/slack/v1/slack.proto
scarlettperry: updated

api/config/service/auditsink/slack/v1/slack.proto
scarlettperry: added more details and left a to-do to expand on it (& docs to clutch.sh) after the transformer PR goes in

backend/service/auditsink/slack/slack_helper.go
scarlettperry: added Overrides to the sink's `svc` struct

RE "hang this method off of the sink" - can you clarify a bit more? Jst want to make sure I understand. For example, for the e2e flow, I was planning on calling the func in https://github.com/lyft/clutch/blob/main/backend/service/auditsink/slack/slack.go#L64 like so:
```
// Check if custom slack message provided in the config
	customMessage, ok := GetOverrideMessage(s.overrides, event)
	if ok {
		// call the slack transformer
	}
```

scarlettperry: oh, perhaps you meant changing it to a receiver function:
`func (s *svc) GetOverrideMessage (event *auditv1.RequestEvent)`, though then I wouldn't be able to reuse it in the internal implementation.

But I don't mind writing another func for the Lyft use-case.

danielhochman: what i would do is write a helper func that generates an lookup table (LUT) from the []CustomMessage, e.g. `func(overrides []*CustomMessage) map[string]CustomMessage`. call that during init and store it on the service struct. then the lookup is trivial (and efficient) in anyone who generates an LUT from overrides using the helper.

danielhochman: discussing in other comment thread https://github.com/lyft/clutch/pull/1089#discussion_r586589334

danielhochman: one more thought, if you want to be able to call GetOverride(service, method string) rather than combining the service and method string from event you can have the helper generate a OverrideLookup struct that keeps the map internally and exposes that func, then store the OverrideLookup func on the service struct

scarlettperry: thank you for the feedback! went with the 2nd approach and added new unit tests

backend/service/auditsink/slack/slack.go
```diff
@@ -29,20 +29,28 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
...
+		channel:   config.Channel,
 	}
 	return s, nil
 }
 
+// OverrideLookup stores a map of custom slack messages that are
+// provided in the slack config
+type OverrideLookup struct {
+	messages map[string]*configv1.CustomMessage
+}
```
danielhochman, line 46: let's move this into `slack_helper.go`

backend/service/auditsink/slack/slack.go
```diff
@@ -29,20 +29,28 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
...
+		filter:    config.Filter,
+		overrides: NewOverrideLookup(config.Overrides),
+		scope:     scope,
+		slack:     slack.New(config.Token),
+		channel:   config.Channel,
 	}
 	return s, nil
 }
 
+// OverrideLookup stores a map of custom slack messages that are
```
danielhochman, line 42: nit clarity
```suggestion
// OverrideLookup stores a map of the full method to the custom slack messages that are
```

backend/service/auditsink/slack/slack_helper.go
```diff
@@ -0,0 +1,46 @@
...
+	configv1 "github.com/lyft/clutch/backend/api/config/service/auditsink/slack/v1"
+)
+
+// NewOverrideLookup creates a map from the overrides. The generated OverrideLookup is
+// used to easily retrieve an override by key (/service/method).
+func NewOverrideLookup(overrides []*configv1.CustomMessage) *OverrideLookup {
+	if len(overrides) == 0 {
+		// no overrides
+		return nil
+	}
```
danielhochman, line 15: i would return an empty struct instead of nil. if OverrideLookup gets more complicated it is bug prone to have to remember to nil check in every func.

alternatively you could deal in the value, e.g. return OverrideLookup instead of pointer. i know in Clutch we normally default to using pointers for all structs, but in this case it would be fine (and probably more efficient) to use the value, especially where the members of the struct are reference types as the map is here.

backend/service/auditsink/slack/slack_helper.go
```diff
@@ -0,0 +1,46 @@
...
+
+// GetOverrideMessage uses the OverrideLookup and the audit event's service + method name to
+// check if an override exists for the key service/method. If found, the custom message is returned.
+// Otherwise ok is false.
+func (o *OverrideLookup) GetOverrideMessage(service, method string) (string, bool) {
+	// /service/method
+	// ex. /clutch.k8s.v1.K8sAPI/DescribePod
+	pattern := "/%s/%s"
+
+	if o == nil || len(o.messages) == 0 {
```
danielhochman, line 35: per other comment, 
```suggestion
	if len(o.messages) == 0 {
```

backend/service/auditsink/slack/slack_helper.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+}
+
+// GetOverrideMessage uses the OverrideLookup and the audit event's service + method name to
+// check if an override exists for the key service/method. If found, the custom message is returned.
+// Otherwise ok is false.
+func (o *OverrideLookup) GetOverrideMessage(service, method string) (string, bool) {
+	// /service/method
+	// ex. /clutch.k8s.v1.K8sAPI/DescribePod
+	pattern := "/%s/%s"
```
danielhochman, line 33: nit move below if statement early check

api/config/service/auditsink/slack/v1/slack.proto
```diff
@@ -13,4 +13,23 @@ message SlackConfig {
...
+/*
+By default, the sink provides a formatted message using a subset of info in an audit event. A custom
+message can alo be created for a `/SERVICE/METHOD` using the metadata (API Request/Response) in an
+audit event. The custom message will be appended to the default message for a richer slack audit.
+*/
+message CustomMessage {
+  // The full method in the format of `/SERVICE/METHOD`
+  string full_method = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  /*
+  The slack sink uses the Go text/template pacakge. Example of a custom message for /clutch.k8s.v1.K8sAPI/ResizeHPA:
```
danielhochman, line 30: spelling
```suggestion
  The slack sink uses the Go text/template package. Example of a custom message for /clutch.k8s.v1.K8sAPI/ResizeHPA:
```

danielhochman (commented):
looking good overall, a few comments

api/config/service/auditsink/slack/v1/slack.proto
```diff
@@ -13,4 +13,36 @@ message SlackConfig {
...
+  repeated CustomMessage overrides = 4;
+}
+
+// TODO: (sperry) update docs & add to clutch.sh
+/*
+By default, the sink provides a formatted message using a subset of info in an audit event. A custom
+message can alo be created for a /SERVICE/METHOD using the audit event's metadata (API Request/Response).
+The custom message will be appended to the default message for a richer slack audit.
+*/
+message CustomMessage {
```
scarlettperry, line 26: added more details and examples. in a follow up PR will work on some doc for clutch.sh

backend/service/auditsink/slack/slack.go
scarlettperry: moved

backend/service/auditsink/slack/slack_helper.go
scarlettperry: thanks, yeah agree. in latest commit went with the second approach and also updated `GetOverrideMessage` to have a value receiver since the func does not modify the OverrideLookup map

backend/service/auditsink/slack/slack_helper.go
scarlettperry: updated

PR #1092, opened 2021-03-01
aws: error converter and interceptor

danielhochman:
### Description
Extract the code and relevant message from AWS SDK errors rather than returning the stringified error with all of the information as an unknown error.

**Before:**
- **`Code 2 (UNKNOWN)`**
- `{"code": 2, "message": "operation error S3: GetObject, https response error StatusCode: 403, RequestID: DSFEFFCCWE7, HostID: O5ErieRNfwEFWEJFIWJEFIWJEFWCJJCCJCJCIWJECICJWEOCWECJW3ZHDCrc=, api error InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records."}`

**After:**
- **`Code 7 (PERMISSION_DENIED)`**
- `{"code": 7, "message":"InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records."}`

### Testing Performed
Unit and manual integration.

### DIFF
--- a/backend/service/aws/aws_test.go
+++ b/backend/service/aws/aws_test.go
@@ -6,10 +6,6 @@ import (
 	"fmt"
 	"testing"
 	"time"
-
-	"github.com/aws/smithy-go"
-	"google.golang.org/grpc/status"
-
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/autoscaling"
 	astypes "github.com/aws/aws-sdk-go-v2/service/autoscaling/types"
--- a/backend/service/aws/aws_test.go
+++ b/backend/service/aws/aws_test.go
@@ -7,16 +7,21 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/smithy-go"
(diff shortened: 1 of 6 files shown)


backend/service/aws/aws_test.go
```diff
@@ -343,6 +345,21 @@ func TestDescribeAutoscalingGroupsErrorHandling(t *testing.T) {
 	assert.Error(t, err2)
 }
 
+func TestErrorIntercept(t *testing.T) {
+	c := &client{}
+	{
```
scarlettperry, line 350: do the test scenarios ned to be wrapped in `{`, `}`

scarlettperry (approved):
nice! jst 1 comment

backend/service/aws/aws_test.go
danielhochman: it scopes the vars so you don't accidentally using a value from a previous scenario. i like to use it for really simple tests instead of splitting them out into individual tests and repeating the setup.

scarlettperry: oh nice! good to know

PR #1093, opened 2021-03-01
slacksink: Add transformer to parse custom slack msg and audit event metadata

scarlettperry:
### Description
PR adds logic to parse audit event metadata based on what's requested for the custom slack message. A separate PR https://github.com/lyft/clutch/pull/1089 updates the slack config to allow users to specify this custom slack message.

### Testing Performed
**Examples**:

**(1) Custom slack message for ResizeHPA wants metadata from the API request**
```
"*Min size*: [[.Request.sizing.min]]\n*Max size*: [[.Request.sizing.max]]"
```
slack message [preview](https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%22%60UNKNOWN%60%20performed%20%60ResizeHPA%60%20via%20%60clutch.k8s.v1.K8sAPI%60%20using%20Clutch%20on%20resource(s)%3A%5Cn-%20kind-clutch-local%2Fenvoy-staging%2Fenvoy%20(%60clutch.k8s.v1.HPA%60)%5Cn*Min%20size*%3A%201%5Cn*Max%20size*%3A%202%22%7D)
```
`UNKNOWN` performed `ResizeHPA` via `clutch.k8s.v1.K8sAPI` using Clutch on resource(s):
- kind-clutch-local/envoy-staging/envoy (`clutch.k8s.v1.HPA`)
`Min size` is 2 and `Max size` is 3
```

**(2) Custom slack message for DescribePod wants metadata from the API request AND response**
```
"Pod [[.Request.name]] status is `[[.Response.pod.state]]`"
```
slack message [preview](https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%22Pod%20envoy-958dcf4fc-2mztk%20status%20is%20%60RUNNING%60%22%7D)
```
...<default message>...
Pod envoy-958dcf4fc-2mztk status is `RUNNING`
```
**(3) Custom slack messages where the metadata requested is type Map or List**
Custom slack message for UpdateDeployment uses `slackList` helper to format a message
```
"*Updated labels*:[[slackList .Request.fields.labels]]"
```
slack message [preview](https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%22%60UNKNOWN%60%20performed%20%60UpdateDeployment%60%20via%20%60clutch.k8s.v1.K8sAPI%60%20using%20Clutch%20on%20resource(s)%3A%5Cn-%20kind-clutch-local%2Fenvoy-staging%2Fenvoy%20(%60clutch.k8s.v1.Deployment%60)%5Cn*Updated%20labels*%3A%5Cn-%20bar%3A%20value%5Cn-%20foo%3A%20value%22%7D)
```
`UNKNOWN` performed `UpdateDeployment` via `clutch.k8s.v1.K8sAPI` using Clutch on resource(s):
- kind-clutch-local/envoy-staging/envoy (`clutch.k8s.v1.Deployment`)
*Updated labels*:
- bar: value
- foo: value
```
Custom slack message for UpdatePod uses `slackList` helper and Go Template `range` to format a message
```
 "*Expected Preconditions*:[[range $$k, $$v := .Request.expectedObjectMetaFields.annotations]]
 \n- [[$$k]]: [[range $$i, $$j := $$v]][[$$j]][[end]][[end]]
 \n*Updated Annotations*:[[slackList .Request.objectMetaFields.annotations]]
 \n*Removed Annotations*:[[slackList .Request.removeObjectMetaFields.annotations]]"
```
slack message [preview](https://api.slack.com/docs/messages/builder?msg=%7B%22text%22%3A%22%60UNKNOWN%60%20performed%20%60UpdatePod%60%20via%20%60clutch.k8s.v1.K8sAPI%60%20using%20Clutch%20on%20resource(s)%3A%5Cn-%20kind-clutch-local%2Fenvoy-production%2Fenvoy-7784b58d-txdg2%20(%60clutch.k8s.v1.Pod%60)%5Cn*Expected%20Preconditions*%3A%5Cn-%20foo%3A%20old%20value%5Cn*Updated%20Annotations*%3A%5Cn-%20foo%3A%20new%20value%5Cn-%20bar%3A%20new%20value%5Cn*Removed%20Annotations*%3A%5Cn-%20foo%5Cn-%20bar%22%7D)
```
`UNKNOWN` performed `UpdatePod` via `clutch.k8s.v1.K8sAPI` using Clutch on resource(s):
- kind-clutch-local/envoy-production/envoy-7784b58d-txdg2 (`clutch.k8s.v1.Pod`)
*Expected Preconditions*:
- foo: old value
*Updated Annotations*:
- foo: new value
- bar: new value
*Removed Annotations*:
- foo
- bar
```

### TODOs
- [x] Add unit tests

### DIFF
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -117,7 +117,7 @@ func parseFile(path string, pb proto.Message, template bool) error {
 
 		Solution is to use Clutch-specific templating tokens in the config that are then replaced
 		with the Go Template synax
-		1) $$ in lieu $
+		1) $$ in lieu of $
 		2) [[ ]] in lieu of {{ }}
 	*/
 
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -106,16 +106,17 @@ func parseFile(path string, pb proto.Message, template bool) error {
 		}
 	}
 
-	contents = []byte(processContents(contents))
+	// Interpolate environment variables
+	expandedData := os.ExpandEnv(string(contents))
+
+	contents = []byte(processTemplateToken(expandedData))
 
 	return parseYAML(contents, pb)
 }
 
-func processContents(contents []byte) string {
-	// Interpolate environment variables
-	expandedData := os.ExpandEnv(string(contents))
-	// Replace the Clutch-specific templating token with the Go $ templating token
-	return strings.ReplaceAll(expandedData, "%%", "$")
+// Replace the Clutch-specific templating token with the Go $ template syntax
+func processTemplateToken(data string) string {
+	return strings.ReplaceAll(data, "%%", "$")
 }
 
 func parseYAML(contents []byte, pb proto.Message) error {
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -106,17 +106,40 @@ func parseFile(path string, pb proto.Message, template bool) error {
 		}
 	}
 
-	// Interpolate environment variables
-	expandedData := os.ExpandEnv(string(contents))
+	/*
+		We want to support defining Go templates in the clutch-config that can be executed at
+		a later time.
+		Two issues that needed to be addressed:
+		1) os.ExpandEnv removes the $ and there isn't a way to provide an escape
+		(open upstream issue: https://github.com/golang/go/issues/43482)
+		2) the config itself can be executed as a template, and so the nested templates would also
+		be executed.
 
-	contents = []byte(processTemplateToken(expandedData))
+		Solution is to use Clutch-specific templating tokens in the config that are then replaced
+		with the Go Template synax
+		1) $$ in lieu $
+		2) [[ ]] in lieu of {{ }}
+	*/
+
+	tokenContent := replaceClutchTokens(string(contents))
+
+	// Interpolate environment variables
+	expandedContent := os.ExpandEnv(tokenContent)
+
+	contents = []byte(replaceToken(expandedContent, "@#@", "$"))
 
 	return parseYAML(contents, pb)
 }
 
-// Replace the Clutch-specific templating token with the Go $ template syntax
-func processTemplateToken(data string) string {
-	return strings.ReplaceAll(data, "%%", "$")
+func replaceClutchTokens(data string) string {
+	data = replaceToken(data, "[[", "{{")
+	data = replaceToken(data, "]]", "}}")
+	// this last replace is done so that os.ExpandEnv does not remove the $
+	return replaceToken(data, "$$", "@#@")
+}
+
+func replaceToken(data string, old string, new string) string {
+	return strings.ReplaceAll(data, old, new)
 }
 
 func parseYAML(contents []byte, pb proto.Message) error {
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"text/template"
 	"time"
 
@@ -105,10 +106,16 @@ func parseFile(path string, pb proto.Message, template bool) error {
 		}
 	}
 
-	// Interpolate environment variables.
-	contents = []byte(os.ExpandEnv(string(contents)))
+	contents = []byte(processContents(contents))
 
 	return parseYAML(contents, pb)
+}
+
+func processContents(contents []byte) string {
+	// Interpolate environment variables
+	expandedData := os.ExpandEnv(string(contents))
+	// Replace the Clutch-specific templating token with the Go $ templating token
+	return strings.ReplaceAll(expandedData, "%%", "$")
 }
 
 func parseYAML(contents []byte, pb proto.Message) error {
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -116,30 +116,36 @@ func parseFile(path string, pb proto.Message, template bool) error {
 		be executed.
 
 		Solution is to use Clutch-specific templating tokens in the config that are then replaced
-		with the Go Template synax
+		with the Go Template syntax
 		1) $$ in lieu of $
 		2) [[ ]] in lieu of {{ }}
 	*/
 
-	tokenContent := replaceClutchTokens(string(contents))
+	tokenContent := bulkReplaceTemplateTokens(string(contents))
 
 	// Interpolate environment variables
 	expandedContent := os.ExpandEnv(tokenContent)
 
-	contents = []byte(replaceToken(expandedContent, "@#@", "$"))
+	contents = []byte(replaceVarTemplateToken(expandedContent))
 
 	return parseYAML(contents, pb)
 }
 
-func replaceClutchTokens(data string) string {
-	data = replaceToken(data, "[[", "{{")
-	data = replaceToken(data, "]]", "}}")
-	// this last replace is done so that os.ExpandEnv does not remove the $
-	return replaceToken(data, "$$", "@#@")
+// swaps the Clutch "Action" tokens for the Go Template "Action" tokens
+// swaps the dollar signs with other characters, otherwise os.ExpandEnv
+// would remove the dollar signs
+func bulkReplaceTemplateTokens(data string) string {
+	sanitize := strings.NewReplacer(
+		"[[", "{{",
+		"]]", "}}",
+		"$$", "@#@",
+	)
+	return sanitize.Replace(data)
 }
 
-func replaceToken(data string, old string, new string) string {
-	return strings.ReplaceAll(data, old, new)
+// swaps the Clutch variable token with the Go Template variable token
+func replaceVarTemplateToken(data string) string {
+	return strings.ReplaceAll(data, "@#@", "$")
 }
 
 func parseYAML(contents []byte, pb proto.Message) error {
--- a/backend/service/auditsink/slack/slack.go
+++ b/backend/service/auditsink/slack/slack.go
@@ -26,7 +26,10 @@ import (
 	"github.com/lyft/clutch/backend/service/auditsink"
 )
 
-const Name = "clutch.service.auditsink.slack"
+const (
+	Name        = "clutch.service.auditsink.slack"
+	defaultNone = "None"
+)
 
 func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service, error) {
 	config := &configv1.SlackConfig{}
@@ -144,7 +147,7 @@ func FormatCustomText(message string, event *auditv1.RequestEvent) (string, erro
 
 	// When a value is nil, the Go Template sets the value as "<no value>".
 	// Ex of when we hit this scenario: https://github.com/lyft/clutch/blob/main/api/k8s/v1/k8s.proto#L860
-	sanitized := strings.ReplaceAll(buf.String(), "<no value>", "null")
+	sanitized := strings.ReplaceAll(buf.String(), "<no value>", defaultNone)
 
 	return sanitized, nil
 }
@@ -182,7 +185,7 @@ func getAuditTemplateData(event *auditv1.RequestEvent) (*auditTemplateData, erro
 // for inputs that are type slice/map, returns a formatted slack list
 func slackList(data interface{}) string {
 	if data == nil {
-		return "None"
+		return defaultNone
 	}
 
 	var b strings.Builder
@@ -193,7 +196,7 @@ func slackList(data interface{}) string {
 	case reflect.Slice:
 		s := reflect.ValueOf(data)
 		if s.Len() == 0 {
-			return "None"
+			return defaultNone
 		}
 		// example:
 		// - foo
@@ -205,7 +208,7 @@ func slackList(data interface{}) string {
 	case reflect.Map:
 		m := reflect.ValueOf(data)
 		if m.Len() == 0 {
-			return "None"
+			return defaultNone
 		}
 		// example:
 		// - foo: value
@@ -216,5 +219,7 @@ func slackList(data interface{}) string {
 		}
 		return b.String()
 	}
+	// We don't return an error so that the custom slack message is still returned
+	// in this way, the user will know they didn't pass the correct input type to slackList
 	return "ERR_INPUT_NOT_SLICE_OR_MAP"
 }
--- a/backend/service/auditsink/slack/slack.go
+++ b/backend/service/auditsink/slack/slack.go
@@ -7,7 +7,9 @@ package slack
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"reflect"
 	"text/template"
 
 	"github.com/golang/protobuf/ptypes"
@@ -46,6 +48,13 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service, 
 type auditEventMetadata struct {
 	Request  map[string]interface{}
 	Response map[string]interface{}
+}
+
+// helper functions to format values in the Go template
+var funcMap = template.FuncMap{
+	// for inputs that are type slice/map, returns a formatted slack list OR `N/A` if the slice/map is empty
+	// currently only iterates 1-level deep
+	"slackList": slackList,
 }
 
 type svc struct {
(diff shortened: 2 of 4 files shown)


backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +121,98 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+		return nil, err
+	}
+
+	return &auditEventMetadata{
+		Request:  requestProto,
+		Response: responseProto,
+	}, nil
+}
+
+func slackList(data interface{}, name string) (string, error) {
```
scarlettperry, line 163: This helper function formats fields that maps and lists & iterates over the values.

I've been stuck/experimenting with how to handle the facet that the map and list values can be any type (usually it's `map<string, string>` or `repeated string` but here's an [example](https://github.com/lyft/clutch/blob/main/api/k8s/v1/k8s.proto#L333) of `map<string, NullableString> where NullableString is a message type with a oneof value).

I currently have something like this https://github.com/lyft/clutch/pull/1093/files#diff-12ec22f81216afa6c8c8d39eeacc01f4ea30bf39b0c3da2525a7f72e7c8beb27R207-R212. The slack output would look like:

the template
```
"*Expected Preconditions*: {{slackList .Request.ExpectedObjectMetaFields `annotations`}}"
```
the slack text
```
`UNKNOWN` performed `UpdatePod` via `clutch.k8s.v1.K8sAPI` using Clutch on resource(s):
- kind-clutch-local/envoy-production/envoy-7784b58d-77m6h (`clutch.k8s.v1.Pod`)
*Expected Preconditions*:
- lyft.net/load-balancing-weight: value:"2"
```

scarlettperry: I initially was using the reflect package (snippet below) but felt I wasn't taking advantage of the fact that the input is a proto type
```
func slackList(data interface{}) string {
     if data == nil {
             return ""
     }
     switch reflect.TypeOf(data).Kind() {
     case reflect.Slice:
             s := reflect.ValueOf(data)
             for i := 0; i < s.Len(); i++ {
                     text += fmt.Sprintf(sliceFormat, s.Index(i).Interface())
             }
             return text
     case reflect.Map:
             m := reflect.ValueOf(data)
             for _, k := range m.MapKeys() {
                     v := m.MapIndex(k)
                     text += fmt.Sprintf(mapFormat, k.Interface(), v.Interface())
             }
             return text
     }
     return ""
 }
```

scarlettperry: cc @danielhochman

danielhochman: my first inclination is to try to see if we simplify the serialization code by going to JSON first with protojson and using that as the basis for the context. for example, oneofs are flattened back into a top level field by protojson.

i'd need to experiment with it a bit more locally though, maybe you have thoughts on that approach though?

backend/gateway/config.go
```diff
@@ -105,12 +106,19 @@ func parseFile(path string, pb proto.Message, template bool) error {
...
+	// Interpolate environment variables
+	expandedData := os.ExpandEnv(string(contents))
+
+	contents = []byte(processTemplateToken(expandedData))
 
 	return parseYAML(contents, pb)
 }
 
+// Replace the Clutch-specific templating token with the Go $ template syntax
+func processTemplateToken(data string) string {
```
scarlettperry, line 118: background: we want to support [Go templating syntax](https://golang.org/pkg/text/template/#hdr-Variables) in `clutch-config.yaml`

`os.ExpandEnv` removes the `$` (and it's [not currently possible](https://github.com/golang/go/issues/43482) to escape the `$`), so we'll need to use our own templating token in the config yaml and then replace it with a `$`, _after_ `os.ExpandEnv` has been called.

as for the clutch templating token, I choose `%%` but it can be something else.
example
```
message: {{range %%v, %%k := .Bar}}{{%%k}}:{{%%v}}{{end}}
```

backend/service/auditsink/slack/slack.go
```diff
@@ -38,6 +44,19 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
...
+type auditEventMetadata struct {
+	Request  map[string]interface{}
+	Response map[string]interface{}
+}
+
+// helper functions to format values in the Go template
+var funcMap = template.FuncMap{
+	// for inputs that are type slice/map, returns a formatted slack list OR `N/A` if the slice/map is empty
+	// currently only iterates 1-level deep
+	"slackList": slackList,
```
scarlettperry, line 57: kept this helper b/c I still think it's helpful for message fields that are simple slice/map types (ie `map<string, string>`, `repeated strings`, etc). This way users won't need to write a `range` for these kind of simple fields and helps cut down on long templating syntax in the config.

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +123,90 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, data); err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+// returns the API request/response details saved in an audit event
+func getAuditMetadata(event *auditv1.RequestEvent) (*auditEventMetadata, error) {
```
scarlettperry, line 148: changed this up to go from proto -> JSON -> map[string]interface{}. the map[string]interface{} is easier to work with than proto, thanks for the the suggestion!

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +123,90 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+func slackList(data interface{}) (string, error) {
+	var text string
+	const listItemFormat = "\n- %v"
+	const mapItemFormat = "\n- %v: %v"
+
+	switch reflect.TypeOf(data).Kind() {
+	case reflect.Slice:
+		s := reflect.ValueOf(data)
+		if s.Len() == 0 {
+			return "`N/A`", nil
```
scarlettperry, line 188: for cases when a slice or map is empty (indicating that nothing was updated/removed etc), have been debating on what's better:
"Removed Annotations":
"Removed Annotations": `N/A`

Alternatively in the Go template itself you can write if/else and use `len` to specify unique defaults but trying to cut down on how much templating we do in the config

backend/service/auditsink/slack/slack_test.go
```diff
@@ -51,3 +55,187 @@ func TestFormat(t *testing.T) {
...
+			event: &auditv1.RequestEvent{
+				RequestMetadata:  &auditv1.RequestMetadata{Body: anyK8sUpdateReq},
+				ResponseMetadata: &auditv1.ResponseMetadata{Body: anyK8UpdateResp},
+			},
+			expectedOutput: "*Removed annotations*:\n- foo\n- bar",
+		},
+		// metdata that is a map, map value is a another map
+		// uses the Golang template `range`
+		{
+			text: "*Expected Preconditions*:{{range $key, $val := .Request.expectedObjectMetaFields.annotations}}\n- {{$key}}: {{range $i, $j := $val}}{{$j}}{{end}}{{end}}",
```
scarlettperry, line 124: this is the `range` syntax for https://github.com/lyft/clutch/blob/main/api/k8s/v1/k8s.proto#L333 which is a `map<string, message> type`

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +123,90 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+
+	return &auditEventMetadata{
+		Request:  requestMetadata,
+		Response: responseMetadata,
+	}, nil
+}
+
+// for inputs that are type slice/map, returns a formatted slack list OR `N/A` if the slice/map is empty
+func slackList(data interface{}) (string, error) {
+	var text string
```
danielhochman, line 180: this is a good use case for a string builder or buffer.
- https://www.calhoun.io/concatenating-and-building-strings-in-go/
- https://dave.cheney.net/high-performance-go-workshop/dotgo-paris.html#avoid_string_concatenation

backend/service/auditsink/slack/slack.go
danielhochman: `None` as a single item in a list is a nice option here.

For the template use case, you could add a helper function which replaces any empty list with a slice containing `["None"]`, e.g. `{{ .MyVar | noneIfEmpty }}`

backend/gateway/config.go
danielhochman: great explanation! let's move it to the godoc.

also i think you will also need to choose a replacement token for the curly brackets... since we run go templating on the file itself at lyft (`-template` to the binary)

one other option, at least for `$` is to replace instances of `$$` with some token first, then do the environment variable expansion, then replace them back to single `$`. e.g. `replaceAll('$$', '@#@') -> os.ExpandEnv -> replaceAll('@#@', '$')`

backend/service/auditsink/slack/slack.go
```diff
@@ -38,6 +44,19 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
 	return s, nil
 }
 
+// The struct is used in a Go template and the fields need to be exported
+type auditEventMetadata struct {
```
danielhochman, line 48: nit: let's call it this or `auditTemplateData`
```suggestion
type auditEventTemplateData struct {
```

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +123,90 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+		// example:
+		// foo: value
+		// bar: value
+		for _, k := range m.MapKeys() {
+			v := m.MapIndex(k)
+			text += fmt.Sprintf(mapItemFormat, k.Interface(), v.Interface())
+		}
+		return text, nil
+	}
+	return "", errors.New("input type is not Slice or Map")
```
danielhochman, line 211: i wonder if we should still return a string as the error so the message itself isn't dropped on the floor.

it would be better to get a slack audit event
```
Mike performed Delete on pod123
- Labels: ERR_INPUT_NOT_SLICE_OR_MAP
```

than not seeing the event in slack and getting an error log.

danielhochman (commented):
almost there! it's working!

backend/gateway/config.go
scarlettperry: RE template in a template - thanks for calling that out! wasn't hitting this case https://github.com/lyft/clutch/blob/main/backend/gateway/config.go#L101.

Latest commit changes
1) $$ -> @#A -> $
2) [[ ]] -> {{ }}

Tested locally, added unit tests, and also a comment explaining the issue/solution

backend/service/auditsink/slack/slack.go
scarlettperry: updated to None & wrote a TODO to add more helper funcs

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +114,54 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+// returns the API request/response details saved in an audit event
+func getAuditMetadata(event *auditv1.RequestEvent) (*auditEventMetadata, error) {
+	var requestMetadata map[string]interface{}
+	var responseMetadata map[string]interface{}
```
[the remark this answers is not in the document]
scarlettperry: updated

backend/service/auditsink/slack/slack.go
scarlettperry: thanks for the links! went with `strings.Builder`

backend/service/auditsink/slack/slack.go
scarlettperry: updated to auditTemplateData

backend/service/auditsink/slack/slack.go
scarlettperry: yeah good idea, updated

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +124,96 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+	}
+
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, data); err != nil {
+		return "", err
+	}
+
+	// When a value is nil, the Go Template returns "<no value>" in its place. Replacing this with null instead.
+	// We hit this scenario with https://github.com/lyft/clutch/blob/main/api/k8s/v1/k8s.proto#L860
+	sanitized := strings.ReplaceAll(buf.String(), "<no value>", "null")
```
scarlettperry, line 147: wrote a unit test here https://github.com/lyft/clutch/pull/1093/files#diff-17fcfcb6bbd591d6dcd0e04f6a9ceb6bb8b2d0cba8ee7ecf84f1cf907ae73dccR134

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +124,96 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+
+	return &auditTemplateData{
+		Request:  requestMetadata,
+		Response: responseMetadata,
+	}, nil
+}
+
+// for inputs that are type slice/map, returns a formatted slack list OR "None" if the slice/map is empty
+func slackList(data interface{}) string {
+	if data == nil {
```
scarlettperry, line 184: forgot to add a nil check, otherwise will panic later at `reflect.TypeOf(data).Kind()`. added two test cases for this: https://github.com/lyft/clutch/pull/1093/files#diff-17fcfcb6bbd591d6dcd0e04f6a9ceb6bb8b2d0cba8ee7ecf84f1cf907ae73dccR104, https://github.com/lyft/clutch/pull/1093/files#diff-17fcfcb6bbd591d6dcd0e04f6a9ceb6bb8b2d0cba8ee7ecf84f1cf907ae73dccR198

backend/service/auditsink/slack/slack.go
danielhochman: should we use `None` here as well?
```suggestion
	sanitized := strings.ReplaceAll(buf.String(), "<no value>", "None")
```

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +124,97 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+		// example:
+		// - foo: value
+		// - bar: value
+		for _, k := range m.MapKeys() {
+			v := m.MapIndex(k)
+			fmt.Fprintf(&b, mapItemFormat, k.Interface(), v.Interface())
+		}
+		return b.String()
+	}
+	return "ERR_INPUT_NOT_SLICE_OR_MAP"
```
danielhochman, line 219: leave comment as to why we don't return `error` itself

backend/service/auditsink/slack/slack.go
```diff
@@ -104,3 +124,97 @@ func formatText(username string, event *auditv1.RequestEvent) string {
...
+
+	var b strings.Builder
+	const sliceItemFormat = "\n- %v"
+	const mapItemFormat = "\n- %v: %v"
+
+	switch reflect.TypeOf(data).Kind() {
+	case reflect.Slice:
+		s := reflect.ValueOf(data)
+		if s.Len() == 0 {
+			return "None"
```
danielhochman, line 196: let's make `"None"` a const

backend/service/auditsink/slack/slack_test.go
```diff
@@ -51,3 +55,195 @@ func TestFormat(t *testing.T) {
...
+	anyK8sUpdateReq, _ := anypb.New(k8sUpdateReq)
+	anyK8UpdateResp, _ := anypb.New(&k8sapiv1.UpdatePodResponse{})
+
+	testCases := []struct {
+		text           string
+		event          *auditv1.RequestEvent
+		expectedErr    bool
+		expectedOutput string
+	}{
+		// metdata from the API request
```
danielhochman, line 86:
```suggestion
		// metadata from the API request
```

danielhochman (approved):
just a few minor suggestions, then :ship:. great work!

scarlettperry:
> just a few minor suggestions, then 🚢. great work!

thank you for all the feedback on this!

PR #1098, opened 2021-03-02
db: configurable max idle connections

mcutalo88:
### Description
Configurable max idle connections for postgres. Will keep the default behavior and limit the configuration to a minimum of 2 idle connections. 

### Testing Performed

Locally.


backend/service/db/postgres/setup.go
```diff
@@ -51,6 +51,10 @@ func New(cfg *any.Any, logger *zap.Logger, scope tally.Scope) (service.Service,
 		return nil, err
 	}
 
+	if pgcfg.MaxIdleConnections > 2 {
+		sqlDB.SetMaxIdleConns(int(pgcfg.MaxIdleConnections))
```
mcutalo88, line 55: I want to write some type of test for this but there are not Getter methods to assert this value. I tried to extract the value via the stats that are exposed without luck.

api/config/service/db/postgres/v1/database.proto
```diff
@@ -32,4 +32,6 @@ message Connection {
 // TODO: Expose more database/sql tunables.
 message Config {
   Connection connection = 1;
+  // The default max idle connections is currently 2 and cannot be configured lower
+  int32 max_idle_connections = 2 [ (validate.rules).int32 = {gt : 2} ];
```
danielhochman, line 36: i would make this an optional value. 0 is a valid value. though negative values are the same as zero. so could also do `if 0 default; else val`

danielhochman (approved):

backend/service/db/postgres/setup.go
danielhochman: yeah, we'd have to replace sql.DB with an interface so you could inject a mock. maybe we should at some point.

PR #1099, opened 2021-03-03
frontend: add new network module

dschaller:
### Description

Add new network client with specified error interfaces.

Follow up PRs will include error handling for authentication and functions to properly identify different types of errors as well as the details we care about.

### Testing Performed

unit


frontend/packages/core/src/Network/errors.ts
```diff
@@ -0,0 +1,319 @@
...
+    | DebugInfo
+    | QuotaFailure
+    | ErrorInfo
+    | PreconditionFailure
+    | BadRequest
+    | RequestInfo
+    | ResourceInfo
+    | Help
+    | Details
+  )[];
```
danielhochman, line 246: shouldn't this just be the types that we have renderers for and `any`? it's impossible to maintain a full list of possible details objects, it could be any proto object.

frontend/packages/core/src/Network/errors.ts
```diff
@@ -0,0 +1,319 @@
...
+ *
+ * @param clientError A client error object.
+ */
+const clutchError = (clientError: AxiosError) => {
+  const { data } = clientError?.response;
+
+  const error = {
+    code: data.code,
+    message: data.message,
+    status: grpcCodeToHttpStatus(data.code),
```
danielhochman, line 299: can we rely on the http status code from the response instead? this could lead to a mismatch.

frontend/packages/core/src/Network/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      data: responseData,
+    } as NetworkError;
+  }
+
+  return Promise.reject(err);
+};
+
+const createClient = () => {
+  const axiosClient = axios.create({
+    validateStatus: status => {
```
danielhochman, line 55: nit: add comment

frontend/packages/core/src/Network/index.ts
```diff
@@ -0,0 +1,66 @@
...
+ *  - Server errors (500–599)
+ */
+export interface HttpStatus {
+  /** The status code. */
+  code: number;
+  /** The status message. */
+  text: string;
+}
+
+const errorInterceptor = (error: AxiosError): Promise<ClutchError | NetworkError | ClientError> => {
```
danielhochman, line 24: curious, can we standardize on ClutchError as return type?

e.g. instead of ClientError just make a fake clutch error that says i'm a client error? or containing a details field that lets us know for certain it's a client error?

seems like it would simplify handling as the error bubbles up through the rest of the code.

frontend/packages/core/src/Network/errors.ts
dschaller: 👍  on the any type (I can swap it out for the generic details interface). It is impossible to account for them all and we could most certainly reduce these but I figured having a starting set would provide the most flexibility if other users wanted to utilize these types in their workflows. Ideally we would just consume a googleapis library that already constructs these from the proto definitions but I could not find one that exists.

frontend/packages/core/src/Network/errors.ts
```diff
@@ -0,0 +1,290 @@
+import type { AxiosError } from "axios";
```
mcutalo88, line 1: From reading the spec I now know these error details are already defined in [googleapis protos](https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto), are we unable to reuse them on the frontend? If so why cant we?

dschaller: there is a short bit in the spec about these not being available via a protojs package that's currently published :)

mcutalo88: Ah ok I i must of missed that part :) Thank you.

frontend/packages/core/src/Network/errors.ts
```diff
@@ -0,0 +1,290 @@
...
+    code: data.code,
+    message: data.message,
+    status: {
+      code: clientError.response.status,
+      text: clientError.response.statusText,
+    } as HttpStatus,
+  } as ClutchError;
+
+  if (data?.details !== undefined && data.details.length > 0) {
+    let details = GRPC_DETAIL_TYPES_MAP[data.details["@type"]];
```
danielhochman, line 274: details is always a list, so i'm not clear how `details["@type"]` would work, doesn't that need to be an array index?

dschaller: oh good catch. This should be a a check on the individual details.

dschaller: I removed this all together since I don't think it adds value here.

danielhochman:
what about headers, etc. TODO?

dschaller:
Yeah, I thought I mentioned we would add headers/metadata/etc. in a follow up PR. I want to get the basic plumbed through to components and then add in other bits to make reviewing easier.

danielhochman (approved):

PR #1112, opened 2021-03-08
frontend: add network helpers

dschaller:
### Description

Add type guard helpers for relevant network detail types. At the moment we only care about determining Help and Clutch details.

### Testing Performed

Unit

### DIFF
--- a/frontend/packages/core/src/Network/errors.ts
+++ b/frontend/packages/core/src/Network/errors.ts
@@ -268,6 +268,7 @@ const grpcResponseToError = (clientError: AxiosError): ClutchError => {
   } as ClutchError;
 
   if (data?.details !== undefined && data.details.length > 0) {
+    // reassign the @type prop to make the details TS friendly.
     const details = data.details.map(detail => {
       const filteredDetail = { ...detail, type: detail["@type"] };
       delete filteredDetail["@type"];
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/Network/errors.ts
```diff
@@ -265,4 +268,12 @@ const grpcResponseToError = (clientError: AxiosError): ClutchError => {
   return error;
 };
 
-export default grpcResponseToError;
+const isHelpDetails = (details: ErrorDetails): details is Help => {
+  return (details as Help).links !== undefined;
```
danielhochman, line 272: can you verify the `@type` field instead?

frontend/packages/core/src/Network/errors.ts
```diff
@@ -259,10 +268,23 @@ const grpcResponseToError = (clientError: AxiosError): ClutchError => {
   } as ClutchError;
 
   if (data?.details !== undefined && data.details.length > 0) {
-    error.details = data.details;
+    const details = data.details.map(detail => {
+      const filteredDetail = { ...detail, type: detail["@type"] };
```
danielhochman, line 272: what about
```suggestion
      const filteredDetail = { ...detail, _type: detail["@type"] };
```

frontend/packages/core/src/Network/errors.ts
```diff
@@ -259,10 +268,23 @@ const grpcResponseToError = (clientError: AxiosError): ClutchError => {
   } as ClutchError;
 
   if (data?.details !== undefined && data.details.length > 0) {
-    error.details = data.details;
+    const details = data.details.map(detail => {
```
danielhochman, line 271: add comment

danielhochman (approved):
lgtm

PR #1115, opened 2021-03-09
k8s: populate start time field in pod

rithujohn191:
### Description
Currently the start time field in the pod proto object is not populated, this PR populates the start time field in the pod proto

### Testing Performed
unit test

### DIFF
--- a/backend/service/k8s/pods.go
+++ b/backend/service/k8s/pods.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"strings"
 
-	"github.com/golang/protobuf/ptypes"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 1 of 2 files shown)


backend/service/k8s/pods.go
```diff
@@ -165,25 +167,28 @@ func (s *svc) checkExpectedObjectMetaFields(expectedObjectMetaFields *k8sapiv1.E
...
-	// TODO: There's a mismatch between the serialization of the timestamp here and what's expected
-	// on the frontend.
-	// var launch *timestamp.Timestamp
-	// if converted, err := ptypes.TimestampProto(k8spod.Status.StartTime.Time); err == nil {
-	// 	launch = converted
-	// }
+	var startTimestamp *timestamppb.Timestamp
+	startTime := k8spod.Status.StartTime
+	if startTime != nil {
+		startTimeProto, err := ptypes.TimestampProto(startTime.Time)
```
danielhochman, line 173: ptypes is deprecated, use `timestamppb.New(startTime.Time)`

rithujohn191: Done

danielhochman:
were you able to test that the front-end handles the time correctly?

rithujohn191:
> were you able to test that the front-end handles the time correctly?

Currently this value is not consumed by the frontend in Clutch directly (the only flow that calls describePod is https://clutch.lyft.net/k8s/pod/delete and it does not use the timestamp value)

smonero:
> were you able to test that the front-end handles the time correctly?

I tried it in my list pods PR and it works using {pod.startTime?.toString()}

PR #1118, opened 2021-03-10
k8s: fix ListRequest protos

rithujohn191:
### Description
Add in option annotations for all the List Resource requests


api/k8s/v1/k8s.proto
```diff
@@ -288,6 +288,11 @@ message ListOptions {
 }
 
 message ListPodsRequest {
+  option (clutch.api.v1.id).patterns = {
+    type_url : "clutch.k8s.v1.Pod",
+    pattern : "{cluster}/{namespace}"
+  };
```
danielhochman, line 294: looking at this a second time i think i see why it wasn't added the first time around. there's some nuance here.

this annotation doesn't quite make sense, because it would say it's a single pod. but there's no ID, which is not a valid pod. so it might be confusing to look at pods that were touched in the audit log and have some with no id. i think it might work anyways but it's an open question.

i think what we might actually want here is something like `clutch.k8s.v1.Namespace` since that's what the user is acting on.

danielhochman:
and looking a bit more at some of the underlying audit code i see another bug now which is a race between UnsentEvents and the response being recorded.

api/k8s/v1/k8s.proto
rithujohn191: But the pattern highlights "{cluster}/{namespace}" right? which would specify only the cluster and the namespace and the audit log would also specify the action which is `ListPods`

rithujohn191:
> and looking a bit more at some of the underlying audit code i see another bug now which is a race between UnsentEvents and the response being recorded.
> 
> this query needs to be updated to query for requests that have a response recorded OR that are older than the timeout
> https://github.com/lyft/clutch/blob/c7e192a0e6878d60f50f7f99d5acd08b7bf8ec59/backend/service/audit/storage/sql/sql.go#L124

Will handle this is in a separate PR

rithujohn191:
Updated annotations for List resource calls to `clutch.k8s.v1.Namespace` since that's what the user is acting on.

rithujohn191:
@danielhochman rebased with main. PR is ready for another pass if you have some extra time :)

danielhochman (approved):

PR #1120, opened 2021-03-14
frontend: new error component

dschaller:
### Description

Revamp error handling on frontend. This PR includes:
- new error component
- updating resolver to handle ClutchErrors and utilize the new error component
- updating data-layout to handle ClutchErrors and utilize the new error component
- updating wizard to handle ClutchErrors and utilize the new error component

Screen Shot 2021-03-13 at 9 47 58 PM
Screen Shot 2021-03-13 at 9 50 11 PM
Screen Shot 2021-03-13 at 9 50 44 PM

### Testing Performed

manual & storybook

### TODO
- [ ] Replace request ID [here](https://github.com/lyft/clutch/pull/1120/files#diff-1a3c31a669452ea0a9bd4d70308dcff1a470d4da686735eb6d2889bdfb39fa89R22)
- [ ] Add in request/response headers to network client


frontend/packages/core/src/Feedback/error/details.tsx
```diff
@@ -0,0 +1,174 @@
...
+                    if (isClutchErrorDetails(detail)) {
+                      const renderMax = 2;
+                      const renderItems = detail.wrapped.slice(0, renderMax);
+                      const remainingItems = detail.wrapped.length - renderMax;
+                      return (
+                        <>
+                          {renderItems.map(wrapped => (
+                            <ListItem key={wrapped.message}>
+                              <ErrorDetailText style={{ fontWeight: 500 }}>
+                                {wrapped.message}
```
danielhochman, line 150: wrapped code missing

frontend/packages/core/src/Feedback/error/details.tsx
```diff
@@ -0,0 +1,174 @@
...
+                  {error.details.map(detail => {
+                    // Only render Clutch Error wrapped details errors here
+                    if (isClutchErrorDetails(detail)) {
+                      const renderMax = 2;
+                      const renderItems = detail.wrapped.slice(0, renderMax);
+                      const remainingItems = detail.wrapped.length - renderMax;
+                      return (
+                        <>
+                          {renderItems.map(wrapped => (
+                            <ListItem key={wrapped.message}>
```
danielhochman, line 148: i can see a scenario where the message is not unique. i think a composite key would work here containing index and message, maybe?

frontend/packages/core/src/text.tsx
```diff
@@ -0,0 +1,46 @@
...
+  whiteSpace: "pre-wrap",
+  wordWrap: "break-word",
+  flexDirection: "row-reverse",
+  display: "flex",
+  overflowY: "scroll",
+});
+
+interface CodeProps {
+  children: string;
+  isCopiable?: boolean;
```
danielhochman, line 28: nit: maybe just call this prop copyButton

copiable is a weird word and the prop also sounds like it might be doing clickjacking to prevent copying or something

danielhochman (commented):
a few quick comments

- need additional line spacing for wrapped errors
- you can save the request ID propagation for a later PR. it will make this PR easier to review and we actually don't even have request IDs in Clutch right now. i thought envoy appended them for us (it's not) but we will need an OSS solution anyways.

frontend/packages/core/src/text.tsx
dschaller: 👍 , renamed to `showCopyButton` since `copyButton` sounds like it would expect a react element

frontend/packages/core/src/Feedback/error/details.tsx
```diff
@@ -0,0 +1,186 @@
...
+                            const color =
+                              grpcCodeToHttpCode(wrapped.code) >= 500 ? "#DB3615" : "#daca37";
+                            return (
+                              // eslint-disable-next-line react/no-array-index-key
+                              <ListItem key={`${idx}-${wrapped.message}`}>
+                                <ErrorDetailText style={{ fontWeight: 500 }}>
+                                  <span style={{ color }}>
+                                    {grpcCodeToText(wrapped.code)}&nbsp;
+                                  </span>
+                                  {wrapped.message}
```
danielhochman, line 159: font weight of message does not match design, should not be bold

frontend/packages/core/src/Feedback/error/details.tsx
```diff
@@ -0,0 +1,186 @@
...
+            {hasWrappedErrorDetails && (
+              <div>
+                <ErrorDetailText style={{ color: "#0D1030" }}>
+                  The following errors were encountered:
+                </ErrorDetailText>
+                <ul>
+                  {error.details.map(detail => {
+                    // Only render Clutch Error wrapped details errors here
+                    if (isClutchErrorDetails(detail)) {
+                      const renderMax = 2;
```
danielhochman, line 144: nit: make global const at top of file somewhere. also i would increase it to somewhere between 3 and 5

frontend/packages/core/src/Feedback/error/details.tsx
```diff
@@ -0,0 +1,186 @@
...
+                      const renderItems = detail.wrapped.slice(0, renderMax);
+                      const remainingItems = detail.wrapped.length - renderMax;
+                      return (
+                        <>
+                          {renderItems.map((wrapped, idx) => {
+                            const color =
+                              grpcCodeToHttpCode(wrapped.code) >= 500 ? "#DB3615" : "#daca37";
```
danielhochman, line 151: this yellow is pretty unreadable. i think maybe pills for the status code that use the color as a background would work better?

frontend/packages/core/src/Feedback/error/details.tsx
```diff
@@ -0,0 +1,186 @@
...
+                        </>
+                      );
+                    }
+                    return null;
+                  })}
+                </ul>
+              </div>
+            )}
+            <Grid container justify="flex-end">
+              <DialogButton onClick={() => setDetailsOpen(true)}>More Details</DialogButton>
```
danielhochman, line 176: this is all uppercase which i don't think we have anywhere else in clutch for buttons

danielhochman: also it should be left-aligned

frontend/packages/core/src/text.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+interface CodeProps {
+  children: string;
+  showCopyButton?: boolean;
+}
+
+const Code = ({ children, showCopyButton = true }: CodeProps) => (
+  <Pre>
+    {showCopyButton && (
+      <CopyButtonContainer container justify="flex-end">
```
danielhochman, line 34: ideally the button wouldn't scroll with the view...

frontend/packages/core/src/text.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <Fab variant="round" size="small">
+          <ClipboardButton text={children} />
+        </Fab>
+      </CopyButtonContainer>
+    )}
+    <Grid container justify="flex-start" alignItems="center">
```
danielhochman, line 40: somethings up with the width here it's wrapping well before the edge 
Screenshot from 2021-03-17 14-07-08

danielhochman: probably nowrap would also work better so a single long error doesn't render it unreadable

frontend/packages/core/src/Feedback/error/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  return (
+    <>
+      <ErrorAlert
+        severity="error"
+        title={error.status.text}
+        data-detailed={hasDetails}
+        action={action}
+      >
+        <ErrorSummaryContainer>
```
danielhochman, line 85: line height is off, not enough space between lines

image

frontend/packages/core/src/text.tsx
```diff
@@ -0,0 +1,46 @@
...
+const CopyButtonContainer = styled(Grid)({
+  marginLeft: "7px",
+});
+
+const Pre = styled.pre({
+  border: "1px solid rgba(13, 16, 48, 0.38)",
+  backgroundColor: "rgba(13,16,48,0.12)",
+  borderRadius: "4px",
+  fontSize: "16px",
+  color: "#242b8c",
```
danielhochman, line 16: colors are a bit off to me actually (bg too dark, blue text is odd), did you not like the figma choices? i forget.

frontend/packages/core/src/Feedback/error/dialog.tsx
```diff
@@ -0,0 +1,26 @@
...
+  error: ClutchError;
+  open: boolean;
+  onClose: () => void;
+}
+
+const ErrorDetailsDialog = ({ error, open, onClose }: ErrorDetailsDialogProps) => {
+  const prettyPrintError = JSON.stringify(error, undefined, 2);
+
+  return (
+    <Dialog title="Full Error Details" onClose={onClose} open={open}>
```
danielhochman, line 17: title should be bold

danielhochman (commented):
first pass, works great though!

frontend/packages/core/src/text.tsx
dschaller: yeah the wrapping is off, I can play around with it but nowrap would introduce an x scroll for long errors which I was trying to avoid. Mainly to allow users to not have to scroll around to read a single error but also because the copy button won't scroll with the text.

frontend/packages/core/src/Feedback/error/dialog.tsx
dschaller: this is the Clutch dialog, are you saying it should be bolded everywhere?

frontend/packages/core/src/Feedback/error/details.tsx
dschaller: Potentially. I didn't have this implemented originally since we hadn't finalized colors with Kanchan yet but it was mentioned in the previous review. I can also just make them all red for now.

frontend/packages/core/src/Feedback/error/details.tsx
dschaller: The design has all uppercase and right-aligned

frontend/packages/core/src/text.tsx
dschaller: Not sure this is possible without absolute positioning but since this is within a modal that is much more difficult.

frontend/packages/core/src/text.tsx
dschaller: Yeah I thought it looked slightly better with more contrast than the designs Kanchan had

frontend/packages/core/src/Feedback/error/details.tsx
danielhochman: yeah all red is fine for now :+1:

frontend/packages/core/src/Feedback/error/details.tsx
danielhochman: ah ok fine for now, i can play with some changes after merge that are closer to one of the other designs

frontend/packages/core/src/Feedback/error/dialog.tsx
danielhochman: yeah actually it's weight 500 in the design system too, weight 400 currently in storybook

danielhochman: Screenshot from 2021-03-18 14-56-11

frontend/packages/core/src/text.tsx
danielhochman: i see what you're saying, yeah that's tricky i guess. throw in a TODO for now no need to spend a bunch of time on it

frontend/packages/core/src/Feedback/error/details.tsx
dschaller: I might be missing what you mean here but the message font weight is the same as figma
Screen Shot 2021-03-19 at 10 33 19 AM

danielhochman: i think figma is having trouble figuring out the mixed weight.
Screenshot from 2021-03-19 12-36-53

danielhochman: if we add backtick or single quote monospace rendering eventually to error messages, that looks like weight 500 but the regular message text is weighted normal

danielhochman (approved):
:ship:

PR #1123, opened 2021-03-15
core: statuserr helper package

danielhochman:
### Description
Create a new package containing an exported version of the 'do all codes match' check we were using in the resolver. This will be useful for any Clutch components that do fanout and have to collect multiple error codes.

### Testing Performed
CI.


danielhochman:
@smonero here's a small refactor that will allow us to use this helper code in our private K8s error handling to determine the proper error code

rithujohn191 (approved):

PR #1131, opened 2021-03-17
backend: set go mod directive to version 1.16

micnncim:
### Description

Upgraded Go version to `1.16` in `/backend` and `/tools` as well as in Dockerfile, GitHub Actions, and so on.

### Testing Performed

`make backend-test` has been passed in local.

### GitHub Issue

Fixes #


danielhochman:
Hi and thanks for looking into this!

So I thought that the version directives in `go.mod` files were the minimum version of golang that can be used: https://golang.org/ref/mod#go-mod-file-go. If you are running go 1.16 and compile a project with 1.13 in `go.mod`, it's compiled with 1.16.

There have been no significant changes to the language since 1.13, so if users of Clutch want to compile with 1.13 and I didn't see a reason to prevent them from doing so:
- https://golang.org/doc/go1.13#language
- https://golang.org/doc/go1.14#language
- https://golang.org/doc/go1.15#language
- https://golang.org/doc/go1.16#language

But it turns out that was only true in go 1.11.3 or earlier ([source](https://blog.ksub.org/bytes/2019/09/15/go.mods-go-directive/)). Now it will compile just fine even if the compile version is below the minimum.

However, we still might want to keep this at 1.13 so we would at least know when we start using features that would break users that are compiling with earlier versions.

Ultimately I guess it makes little difference. I will run it by some others on the team to see if they have an opinion and update here tomorrow.

danielhochman (approved):

PR #1133, opened 2021-03-17
k8s: use unbuffered cache channel

mcutalo88:
### Description
Removing the buffered channel for the k8s cache channel, in an effort to reduce memory footprint caused by back pressure from the database. In a follow up PR I'm going to attempt to improve database throughput for topology queries.

Kubernetes informers to my knowledge can not be throttled, we receive events and must do something with them. First lets look at the flow we currently have in place and explain where I think the bottlenecks are and why this change makes sense.

informer -> k8s cache channel (limit 5000) -> processed by the topology service.

I believe by introducing a buffered channel it contributes to overall memory footprint in a drastic way. Once the channel fills up we are unable to process anymore informer events [example here](https://github.com/lyft/clutch/blob/k8s-cache-remove-bufferchan/backend/service/k8s/cache.go#L186) of the blocking call, which leave the full kubernetes objects in memory and puts us in a position where we cant convert them to `lightweightCacheObject` objects which I created to reduce runtime memory footprint by not storing the entire k8s object.

For these reason I would like to remove the buffer channel completely, i rather have lighter weight clutch objects in memory on that channel then kubernetes full events sitting in memory. Additionally depending on the use of the `LightweightInfromer` some use cases drop update events as they are not necessary, by clogging the channel we queue these up in memory.


dschaller:
Seems reasonable to me as we can't be sure of the number of updates we will receive. If we use an unbuffered channel here does that mean we can only read these events synchronously? If that's the case maybe it's not a a concern because we are limited by the topo service.

For the last paragraph are these improvements you would like to make in future PRs?

mcutalo88:
Going to close this out for now and simply increase the size of the buffered channels. In https://github.com/lyft/clutch/pull/1154, I'm supporting batching inserts for the topology cache which greatly improve write throughput. Will reopen this PR if necessary in the future.

PR #1145, opened 2021-03-19
k8s: add namespace api

rithujohn191:
### Description
Add Kubernetes namespace resource API and describeNamespace API call

### Testing Performed
- Unit test
- Local test:
```
curl -X POST   http://localhost:3000/v1/k8s/describeNamespace -H 'Content-Type: application/json'   -d '{ 
"name": "envoy-staging"
}'
{"namespace":{"cluster":"kind-clutch-local", "name":"envoy-staging", "annotations":{}, "labels":{}}}
```


backend/service/k8s/namespace.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(namespaces.Items) == 1 {
+		return ProtoForNamespace(cs.Cluster(), &namespaces.Items[0]), nil
+	} else if len(namespaces.Items) > 1 {
+		return nil, fmt.Errorf("Located multiple Namespaces")
+	}
+	return nil, fmt.Errorf("Unable to locate namespaces")
```
danielhochman, line 31: use rich errors `status.Error`

api/k8s/v1/k8s.proto
```diff
@@ -852,6 +860,38 @@ message CreateJobResponse {
...
+  option (clutch.api.v1.id).patterns = {
+    type_url : "clutch.k8s.v1.Namespace",
+    pattern : "{cluster}/{name}"
+  };
+
+  string clientset = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  string cluster = 2 [ (validate.rules).string = {min_bytes : 1} ];
+  string name = 3 [ (validate.rules).string = {min_bytes : 1} ];
+
+  map<string, string> labels = 4 [ (validate.rules).map.keys.string = {min_bytes : 1} ];
```
danielhochman, line 886: what does this field do on the request?

danielhochman (commented):
looks good after a quick pass, will take one more look to make sure i didn't miss anything after comments are addressed

api/k8s/v1/k8s.proto
rithujohn191: You can provide namespace specific labels to filter by just like you can for any other resource

danielhochman: i figured that was the intent but when i looked it seems like the field is never read or used for that purpose in the implementation.

rithujohn191: Thanks for pointing out that this is a redundant field for a describe call which is trying to get a specific resource and not list based on labels. I'll clean up the other describe calls in a separate PR as all of them have the label field.

danielhochman (approved):

PR #1149, opened 2021-03-19
frontend: reset data layout data on restart

dschaller:
### Description

Clear data layout when restarting workflows.

restartDL

### Testing Performed

manual

### DIFF
--- a/frontend/packages/data-layout/src/manager.tsx
+++ b/frontend/packages/data-layout/src/manager.tsx
@@ -13,6 +13,14 @@ const assign = (key: string, value: object): Thunk<ManagerLayout, Action> => {
   };
 };
 
+const reset = (): Thunk<ManagerLayout, Action> => {
+  return dispatch => {
+    dispatch({
+      type: ManagerAction.RESET,
+    });
+  };
+};
+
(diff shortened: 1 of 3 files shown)


smonero (approved):
tested for my use case and works

frontend/packages/data-layout/src/manager.tsx
```diff
@@ -76,21 +84,24 @@ interface DataManager {
...
 }
 
 const defaultTransform = (data: object): object => data;
 const defaultErrorTransform = (err: Error): string => {
   return err?.response?.displayText ?? err.message;
 };
 
-const useDataLayoutManager = (layouts: ManagerLayout): DataManager => {
-  const initialState = {};
+const initialLayoutStepState = { data: {}, isLoading: true, error: null };
```
danielhochman, line 95: why is isLoading true the default?

dschaller: Ah good point. I think this gets reset in the layout itself as false [here](https://github.com/lyft/clutch/blob/main/frontend/packages/data-layout/src/layout.tsx#L50). We are definitely going to have to come back and clean up the component.

danielhochman (approved):

PR #1153, opened 2021-03-19
FrontEnd Core: Default table vals

smonero:
FrontEnd Core: Default value prop for TableRow

### Description

Add a prop to the TableRow component so that if a value is null, we can replace it with a default. 
I wasn't sure whether it should be a string or `any`, since in the future maybe someone would want boolean or number, but I put it as any for now.

### Testing Performed

local

### GitHub Issue

Fixes #


frontend/packages/core/src/Table/table.tsx
```diff
@@ -79,13 +79,14 @@ export const Table: React.FC<TableProps> = ({ headings, children, ...props }) =>
 
 export interface TableRowProps extends Pick<MuiTableRowProps, "onClick"> {
   children?: React.ReactNode;
+  defaultValue?: any;
```
dschaller, line 82: We should probably have this be more specific to start with and make it broader later if needed. Let's use `React.ReactNode` as the type as this will allow a react element or a string

frontend/packages/core/src/Table/table.tsx
```diff
@@ -79,13 +79,14 @@ export const Table: React.FC<TableProps> = ({ headings, children, ...props }) =>
...
 }
 
-export const TableRow = ({ children = [], onClick }: TableRowProps) => (
+export const TableRow = ({ children = [], onClick, defaultValue }: TableRowProps) => (
```
dschaller, line 85: What's the use case for specified a default value to the table row directly instead of handling it in the child?

smonero: The use case is to prevent having to write out the default value for each child, which results in lots of copying and pasting.  But maybe I'm going about this in the wrong way.

frontend/packages/core/src/Table/table.tsx
```diff
@@ -79,13 +79,14 @@ export const Table: React.FC<TableProps> = ({ headings, children, ...props }) =>
 
 export interface TableRowProps extends Pick<MuiTableRowProps, "onClick"> {
   children?: React.ReactNode;
+  defaultValue?: React.ReactNode;
```
dschaller, line 82: nit: I would change this to something like `defaultCellValue`

frontend/packages/core/src/Table/table.tsx
```diff
@@ -79,13 +79,14 @@ export const Table: React.FC<TableProps> = ({ headings, children, ...props }) =>
...
 }
 
-export const TableRow = ({ children = [], onClick }: TableRowProps) => (
+export const TableRow = ({ children = [], onClick, defaultValue }: TableRowProps) => (
   <StyledTableRow onClick={onClick}>
     {React.Children.map(children, (value, index) => (
       // eslint-disable-next-line react/no-array-index-key
-      <TableCell key={index}>{value}</TableCell>
+      <TableCell key={index}>{value == null ? defaultValue : value}</TableCell>
```
dschaller, line 89:
```suggestion
      <TableCell key={index}>{value === undefined ? defaultValue : value}</TableCell>
```

dschaller:
Looks good! Thanks for explaining

dschaller (approved):

PR #1154, opened 2021-03-19
topology: cache batch inserts

mcutalo88:
### Description
Support batching inserts enabling higher throughputs of inserts for the topology cache. This is configurable and the default is set to 1.

We periodically flush every 30 seconds if the configured batch size has not been reached, to ensure that items to do not get stuck waiting for the batch size to be reached.

### Testing Performed
Unit / Local.


backend/service/topology/cache.go
```diff
@@ -99,64 +100,110 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
-			if err := c.deleteCache(ctx, obj.Resource.Id, obj.Resource.Pb.TypeUrl); err != nil {
-				c.log.Error("Error deleting cache", zap.Error(err))
+
+			if len(batchInsert) >= int(c.config.Cache.BatchInsertSize) {
+				if err := c.setCache(ctx, batchInsert); err != nil {
+					c.log.Error("Error setting cache", zap.Error(err))
+				}
+				batchInsert = []*topologyv1.Resource{}
+			}
+		case <-time.After(time.Second * 30):
```
danielhochman, line 138: make flush interval configurable or at least const

backend/service/topology/cache.go
```diff
@@ -99,64 +100,110 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
+					c.log.Error("Error deleting cache", zap.Error(err))
+				}
+			default:
+				c.log.Warn("UpdateCacheRequest action is not implemented", zap.String("action", obj.Action.String()))
 			}
-		case topologyv1.UpdateCacheRequest_DELETE:
-			if err := c.deleteCache(ctx, obj.Resource.Id, obj.Resource.Pb.TypeUrl); err != nil {
-				c.log.Error("Error deleting cache", zap.Error(err))
+
+			if len(batchInsert) >= int(c.config.Cache.BatchInsertSize) {
```
danielhochman, line 132: nit: store int(c.config.Cache.BatchInsertSize) as a member of the struct rather than reading it and doing a type conversion every time.

backend/service/topology/cache.go
```diff
@@ -99,64 +100,110 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
+					c.log.Error("Error setting cache", zap.Error(err))
+				}
+				batchInsert = []*topologyv1.Resource{}
+			}
+		case <-time.After(time.Second * 30):
+			if len(batchInsert) > 0 {
+				if err := c.setCache(ctx, batchInsert); err != nil {
+					c.log.Error("Error setting cache", zap.Error(err))
+				}
+				batchInsert = []*topologyv1.Resource{}
```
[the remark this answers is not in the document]
danielhochman: same for other instance of this

backend/service/topology/cache.go
```diff
@@ -99,64 +100,110 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
+			c.scope.Tagged(map[string]string{
+				"service": service,
+			}).SubScope("cache").Gauge("object_channel.queue_depth").Update(float64(len(objs)))
+
+			switch obj.Action {
+			case topologyv1.UpdateCacheRequest_CREATE_OR_UPDATE:
+				batchInsert = append(batchInsert, obj.Resource)
+			case topologyv1.UpdateCacheRequest_DELETE:
+				if err := c.deleteCache(ctx, obj.Resource.Id, obj.Resource.Pb.TypeUrl); err != nil {
+					c.log.Error("Error deleting cache", zap.Error(err))
```
danielhochman, line 126: i think there's an obscure bug here, not sure it's worth worrying about for now.

but let's say i get one insert and my batch size is 10. and then i get a delete every 15 seconds. the insert will never occur.

backend/service/topology/cache.go
```diff
@@ -99,64 +100,110 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
-		default:
-			c.log.Warn("UpdateCacheRequest action is not implemented", zap.String("action", obj.Action.String()))
 		}
 	}
 }
 
-func (c *client) setCache(ctx context.Context, obj *topologyv1.Resource) error {
-	const upsertQuery = `
+func (c *client) prepareBulkCacheInsert(obj []*topologyv1.Resource) ([]interface{}, string) {
+	queryParams := make([]string, 0, len(obj))
```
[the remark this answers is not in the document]
mcutalo88: Yeah true I should also look at the string builder that would be better here.

backend/service/topology/cache.go
mcutalo88: I really wish I could use that config as the source of truth and not spill these fields over into the struct.

backend/service/topology/cache.go
```diff
@@ -99,64 +100,110 @@ func (c *client) startTopologyCache(ctx context.Context) {
 	<-ctx.Done()
 }
 
+// Processes all topology cache update requests.
+// This supports insert batching which defaults to 1, and can be configured by settting BatchInsertSize.
+// Notably a flush will happen every 30 seconds, ensuring all items make it into cache even if the
+// configured BatchInsertSize is not met.
 func (c *client) processTopologyObjectChannel(ctx context.Context, objs <-chan *topologyv1.UpdateCacheRequest, service string) {
```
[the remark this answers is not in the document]
mcutalo88: no. ill attempt to craft one.

backend/service/topology/cache.go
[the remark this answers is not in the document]
mcutalo88: I ran 3 benchmarks with different implementations, 1) as is, 2) using the `strings.Builder` and 3) using `bytes.Buffer`.

The results suggest that the current implantation is the fastest. However they were all pretty close

Each test was processing 10000 items ill commit the benchmark to keep around but here is what it looks like.
```
func BenchmarkPrepareBulkCacheInsert(b *testing.B) {
	tenk := generatePrepareBulkCacheInsertInput(10000)
	topology := &client{}

        // Must reset so we do not include the input generation in the benchmark
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		topology.prepareBulkCacheInsert(tenk)
	}
}
```


// As is with the array impl
```
goos: darwin
goarch: amd64
pkg: github.com/lyft/clutch/backend/service/topology
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkPrepareBulkCacheInsert-8           27    39253226 ns/op  11211254 B/op   319765 allocs/op
PASS
ok    github.com/lyft/clutch/backend/service/topology 1.249s
```

// strings.builder 
```
goos: darwin
goarch: amd64
pkg: github.com/lyft/clutch/backend/service/topology
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkPrepareBulkCacheInsert-8           28    41193166 ns/op  12143981 B/op   319796 allocs/op
PASS
ok    github.com/lyft/clutch/backend/service/topology 1.368s
```

// bytes.buffer
```
goos: darwin
goarch: amd64
pkg: github.com/lyft/clutch/backend/service/topology
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkPrepareBulkCacheInsert-8           27    40331501 ns/op  11756060 B/op   319786 allocs/op
PASS
ok    github.com/lyft/clutch/backend/service/topology 1.284s
```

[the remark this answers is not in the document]
danielhochman: nice!

backend/service/topology/cache.go
```diff
@@ -99,64 +100,108 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
 		ctx,
 		upsertQuery,
-		obj.Id,
-		obj.Pb.GetTypeUrl(),
-		dataJson,
-		metadataJson,
+		args...,
 	)
 	if err != nil {
 		c.scope.SubScope("cache").Counter("set.failure").Inc(1)
```
danielhochman, line 200: should this also be len(obj)?

api/config/service/topology/v1/topology.proto
```diff
@@ -15,4 +15,14 @@ message Config {
 message Cache {
   // The time to live (ttl) for an item in cache, the default is 2 hours (7200 seconds).
   google.protobuf.Duration ttl = 1 [ (validate.rules).duration.gt.seconds = 1 ];
+
+  // Set the batch insert size for setCache() operations. By default this is set to 1.
+  // Depending on your database resource configuration and the number of items to cache,
+  // you can tune this value to improve write performance.
+  int32 batch_insert_size = 2;
+
+  // Defaults to 10 seconds.
```
danielhochman, line 24: curious, is it possible to get the default value with the validation rule on the field?

backend/service/topology/cache.go
```diff
@@ -99,64 +100,108 @@ func (c *client) startTopologyCache(ctx context.Context) {
...
+		if err != nil {
+			c.scope.SubScope("cache").Counter("set.failure").Inc(1)
+			return nil, ""
+		}
+
+		// append args in column order
+		queryArgs = append(queryArgs, o.Id)
+		queryArgs = append(queryArgs, o.Pb.GetTypeUrl())
+		queryArgs = append(queryArgs, dataJson)
+		queryArgs = append(queryArgs, metadataJson)
```
danielhochman, line 176:
```suggestion
		queryArgs = append(queryArgs, o.Id, o.Pb.GetTypeUrl(), dataJson, metadataJson)
```

danielhochman (approved):
lgtm after remaining comments

backend/service/topology/cache.go
mcutalo88: ah yeah good catch.

api/config/service/topology/v1/topology.proto
mcutalo88: oh interesting i didn't know the validation had default support, i wouldn't want to force users to use 10seconds as the floor tho. But let me check it out.

danielhochman: sorry, i meant will the validation prevent someone from entering no value, in which case there is no default anyways

mcutalo88: Ah i got ya my bad, yes let me remove the validation.

backend/service/topology/cache.go
mcutalo88: Yeah i see what your saying, im going to leave a todo and circle back.

Issue #1116, opened 2021-03-10
Service based authn/z


snowp:
Provide a way for services (e.g. entities that cannot provide OIDC headers) to participate in authn/authz.

From offline discussions it sounds like we might be able to generate a JWT token that we store in a db for the
service to use, but will defer to @danielhochman.

PR #1157, opened 2021-03-19
oidc: check token store for token revocation

snowp:
### Description
Supports token revocation by checking the token storage to see if the token is still valid.
We expect tokens to be present in the store unless they have been revoked, so
checking for existence here is sufficient for revocation purposes.

No mechanism to actually revoke the token is added, just the logic for failing verification
when the token is revoked. 

### Testing Performed
Added unit test coverage

### GitHub Issue
#1116

### DIFF
--- a/backend/service/authn/oidc.go
+++ b/backend/service/authn/oidc.go
@@ -184,17 +184,6 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
 		return nil, err
 	}
 
-	// If the token doesn't exist in the token storage anymore, it must have been revoked.
-	// Fail verification in this case.
-	// TODO(perf): Cache the lookup result in-memory for min(60, timeToExpiry) to prevent
-	// hitting the DB on each request. This should also cache whether we didn't find a token.
-	if p.tokenStorage != nil {
-		token, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProvider)
-		if token == nil {
-			return nil, err
-		}
-	}
-
 	if err := claims.Valid(); err != nil {
 		return nil, err
 	}
--- a/backend/service/authn/oidc.go
+++ b/backend/service/authn/oidc.go
@@ -175,6 +175,15 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
 		return nil, err
 	}
 
+	// If the token doesn't exist in the token storage anymore, it must have been revoked.
+	// Fail verification in this case.
+	if p.tokenStorage != nil {
+		token, err := p.tokenStorage.Read(ctx, claims.Subject, p.providerAlias)
+		if token == nil {
+			return nil, err
+		}
+	}
+
 	if err := claims.Valid(); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


danielhochman:
we aren't currently storing the clutch issued token, only the provider token that we use as the underlying claim, so this would break all authn.

see https://github.com/lyft/clutch/blob/e90651bfbc3817b260c23d298806de88292b0309/backend/service/authn/oidc.go#L139-L144. no call to Store after creation of the token that's returned to user.

snowp:
@danielhochman I think we look for the provider auth token (not the clutch issued one), so I'm not sure I see how this would break the authn atm? That said this doesn't seem correct, I'm guessing we should store the clutch issued one with some well known provider name (`clutch`?) and make the look up against that token. Would we need to do anything at all with the backing provider token the revocation case then?

danielhochman:
oh right you're calling with providerAlias. revocation using the provider token doesn't make sense i think, though we may want to clear all user tokens out just to be safe when doing revocation.

yeah, let's store the clutch token with a well-known provider name (which we'll need to for refresh to work).

backend/service/authn/oidc.go
```diff
@@ -175,6 +184,15 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
...
 	}
 
+	// If the token doesn't exist in the token storage anymore, it must have been revoked.
+	// Fail verification in this case.
+	if p.tokenStorage != nil {
+		token, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProviderName)
+		if token == nil {
+			return nil, err
+		}
+	}
```
danielhochman, line 194: do you mind deploying this in a fast-follow to avoid everyone getting logged out at the same time? it will also allow us to verify the tokens are being stored correctly in the db upon exchange.

backend/service/authn/oidc.go
```diff
@@ -175,6 +184,15 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
 		return nil, err
 	}
 
+	// If the token doesn't exist in the token storage anymore, it must have been revoked.
+	// Fail verification in this case.
+	if p.tokenStorage != nil {
+		token, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProviderName)
```
danielhochman, line 190: can we add a `TODO(perf/maybe):` somewhere to cache tokens in-memory for `min(60s, timeUntilexpiry)`? once we have service accounts we could have high programmatic call volume which would roundtrip the DB before each call. thoughts on this scenario?

danielhochman (approved):
lgtm overall, a couple of small qs

backend/service/authn/oidc.go
snowp: Yeah I can imagine this being necessary, and we'd probably also want to include tokens whose lookup failed (and indicate this) to prevent a client using a revoked token from having to hit the DB every time. Will add a TODO

backend/service/authn/oidc.go
snowp: Sure, removed this piece of code and updated the test for now, will follow up with another PR.

PR #1162, opened 2021-03-21
k8s: add some fields to HPA and deployment proto protobufs

smonero:
### Description

We need these for our usecase of the dashboard
Added 2 fields and some tests for HPA, deployment

### Testing Performed

unit
local
### GitHub Issue

Fixes #


danielhochman (approved):

api/k8s/v1/k8s.proto
```diff
@@ -419,6 +422,31 @@ message Deployment {
...
+    message DeploymentCondition {
+      enum DeploymentConditionType {
+        UNSPECIFIED = 0;
+        UNKNOWN = 1;
+        AVAILABLE = 2;
+        PROGRESSING = 3;
+        REPLICA_FAILURE = 4;
+      }
+
+      DeploymentConditionType deployment_condition_type = 1;
```
rithujohn191, line 441: Could you add a comment here saying this is mirroring the upstream values here: https://pkg.go.dev/k8s.io/api/apps/v1#DeploymentConditionType

api/k8s/v1/k8s.proto
```diff
@@ -419,6 +422,31 @@ message Deployment {
...
   map<string, string> labels = 4;
   map<string, string> annotations = 5;
+
+  message DeploymentStatus {
+    uint32 replicas = 1;
+    uint32 updated_replicas = 2;
+    uint32 ready_replicas = 3;
+    uint32 available_replicas = 4;
+    uint32 unavailable_replicas = 5;
+    message DeploymentCondition {
```
rithujohn191, line 432: Having Deployment here feels a little redundant since this is in the `DeploymentStatus` message. Maybe replace this with just `Condition` and same comment for the `Condition Type`

backend/service/k8s/cronjob_test.go
```diff
@@ -106,7 +106,7 @@ func TestProtoForCron(t *testing.T) {
 					ConcurrencyPolicy:       v1beta1.AllowConcurrent,
 					Schedule:                "5 4 * * *",
 					Suspend:                 &[]bool{true}[0],
-					StartingDeadlineSeconds: &[]int64{69}[0],
+					StartingDeadlineSeconds: &[]int64{60}[0],
```
rithujohn191, line 109: Is this just fixing a typo?

api/k8s/v1/k8s.proto
```diff
@@ -419,6 +422,31 @@ message Deployment {
...
+      enum DeploymentConditionType {
+        UNSPECIFIED = 0;
+        UNKNOWN = 1;
+        AVAILABLE = 2;
+        PROGRESSING = 3;
+        REPLICA_FAILURE = 4;
+      }
+
+      DeploymentConditionType deployment_condition_type = 1;
+      google.protobuf.BoolValue condition_status = 2;
```
rithujohn191, line 442: This maps to https://pkg.go.dev/k8s.io/api@v0.20.5/core/v1#ConditionStatus right? so looks like it can have three values true, false and unknown, we might want to make this an enum too

rithujohn191: It looks like there might be plans to expand that to add other fields 
```
In the future, we could add other intermediate conditions, e.g. ConditionDegraded.
```
So enum might be our best bet

backend/service/k8s/cronjob_test.go
smonero: Yes, mcutalo said to stay away from "meme" value numbers like 69 and 420

rithujohn191: For this PR its fine, but going forward lets make a separate PR for fixes like these since its not related to the main task at hand

api/k8s/v1/k8s.proto
```diff
@@ -369,6 +369,9 @@ message HPA {
 
   map<string, string> labels = 5;
   map<string, string> annotations = 6;
+
+  google.protobuf.Int32Value current_cpu_utilization_percentage = 7;
```
rithujohn191, line 373: why is the type `google.protobuf.Int32Value` rather than a regular `int32`

smonero: to prevent it being defaulted to 0.  If it just an int32 instead of an Int32Value (which is a pointer), then that means we can't tell the difference between it actually being 0 and it just not being set, which would default to 0.  If its an Int32Value, we can have it be null in the case where its not set, and 0 when it is actually set.

If we can be sure that it will always be set, or if we don't mind it being 0 defaulted, then we can change it to an int32 rather than an Int32Value.

backend/service/k8s/deployment_test.go
```diff
@@ -245,3 +246,67 @@ func TestProtoForDeploymentClusterName(t *testing.T) {
...
 	}
 }
+
+func TestProtoForDeploymentStatus(t *testing.T) {
+	t.Parallel()
+
+	var deploymentTestCases = []struct {
+		id                      string
+		expectedReplicas        uint32
+		expectedUpdatedReplicas uint32
```
rithujohn191, line 256: don't see the fields expectedReplicas & expectedUpdatedReplicas being used in the test cases below

backend/service/k8s/hpa.go
```diff
@@ -3,6 +3,7 @@ package k8s
 import (
 	"context"
 
+	"github.com/golang/protobuf/ptypes/wrappers"
```
rithujohn191, line 6: since ptypes is deprecated not sure if we want to be using ptypes/wrappers

cc @danielhochman

danielhochman: yeah the new package is wrapperspb if you want to swap it out. if not i'll catch it in my cleanup.

rithujohn191 (approved):

PR #1169, opened 2021-03-22
experimentation: Adding stats for active faults

kathan24:
### Description

Emitting stats when faults are injects and removed. This will help us set alarms when faults are on longer than expected. 

### GitHub Issue

Fixes #

### DIFF
--- a/backend/module/chaos/serverexperimentation/xds/xds_cache.go
+++ b/backend/module/chaos/serverexperimentation/xds/xds_cache.go
@@ -71,7 +71,7 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
 			emptyResources[gcpTypes.ExtensionConfig] = generateEmptyECDSResource(cluster, ecdsConfig, logger)
 			emptyResources[gcpTypes.Runtime] = generateRTDSResource([]*experimentation.Experiment{}, rtdsConfig, ttl, logger)
 
-			err := setSnapshot(emptyResources, cluster, snapshotCache, true, logger, scope)
+			err := setSnapshot(emptyResources, cluster, snapshotCache, logger)
 			if err != nil {
 				logger.Errorw("Unable to unset the fault for cluster", "cluster", cluster,
 					"error", err)
@@ -80,7 +80,10 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
 	}
 
 	// Create/Update experiments
+	activeFaults := 0
 	for cluster, experiments := range clusterFaultMap {
+		activeFaults += 1
+
 		resources := make(map[gcpTypes.ResponseType][]gcpTypes.ResourceWithTtl)
 
 		logger.Debugw("Injecting fault for cluster", "cluster", cluster)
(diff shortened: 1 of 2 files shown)


backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -132,6 +133,16 @@ func setSnapshot(resourceMap map[gcpTypes.ResponseType][]gcpTypes.ResourceWithTt
...
 
+	if isSnapshotEmpty {
+		scope.Tagged(map[string]string{
+			"cluster": cluster,
+		}).Counter("active_faults").Inc(-1)
+	} else {
+		scope.Tagged(map[string]string{
+			"cluster": cluster,
+		}).Counter("active_faults").Inc(1)
+	}
```
danielhochman, line 144: from what i can tell, a gauge would work better here than a counter.

snowp: Yeah a gauge seems like a better fit

snowp (changes requested):
PR looks right (besides the counter/gauge comment) but it seems like a imperfect stat to use to monitor that experiments run for too long, because it relies on stats dropping down to zero for some period of time (e.g. imagine one hour long experiments being scheduled every 30 minutes). 

Might be easy enough to use this for now since they would work with current internal usage of EF, but a more robust strategy might be to periodically query the db for the oldest running experiment and emitting a gauge with its duration.

backend/module/chaos/serverexperimentation/xds/xds_cache.go
kathan24: shoot. I had a Guage and then while looking for tagging did copy/paste and missed replacing it. thanks for catching

kathan24:
I agree with you @snowp that if we run frequent experiments we will not able to know if the experiments are finished or not. However, I don't see experiments running that often in the near term. Whenever we do see some traction, we will switch to the DB lookup technique you described.

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -136,11 +136,11 @@ func setSnapshot(resourceMap map[gcpTypes.ResponseType][]gcpTypes.ResourceWithTt
 	if isSnapshotEmpty {
 		scope.Tagged(map[string]string{
 			"cluster": cluster,
-		}).Counter("active_faults").Inc(-1)
+		}).Gauge("active_faults").Update(-1)
```
snowp, line 139: Is this just going to set the gauge to -1? I think you need to track this value somewhere or compute it periodically

snowp (changes requested):

backend/module/chaos/serverexperimentation/xds/xds_cache.go
snowp: Maybe add a test to make sure this works right?

backend/module/chaos/serverexperimentation/xds/xds_cache.go
```diff
@@ -79,7 +80,10 @@ func refreshCache(ctx context.Context, storer experimentstore.Storer, snapshotCa
 	}
 
 	// Create/Update experiments
+	activeFaults := 0
```
danielhochman, line 83: i think this is equal to `len(clusterFaultMap)` instead of maintaining a separate counter

donyu:
👍

donyu (approved):

snowp (approved):

PR #1171, opened 2021-03-24
core: introduce core protos

mcutalo88:
### Description
Creating a new core clutch package that will be a home to components such as high level `Project` definitions for the Service Catalog, high level envoy primitives such as `Cluster` and other abstractions that we consider core to clutch.

In this PR we are adding envoy `Cluster` as the first core type, the first usage of this will be to support autocomplete for the experimentation functionally. 

### Testing Performed
n/a


api/core/v1/envoy.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package clutch.core.v1;
+
+option go_package = "github.com/lyft/clutch/backend/api/core/v1;corev1";
+
+// Represents a single Envoy Cluster
+message Cluster {
```
mcutalo88, line 8: Should I nest this deeper into a `clutch.core.envoy.v1` package? Or keep in the current package and rename to `EnvoyCluster`?

mcutalo88:
cc @Augustyniak

api/core/v1/envoy.proto
Augustyniak: I like the idea of nesting this deeper into `clutch.core.envoy.v1` - I do not have a strong opinion when it comes to this versus renaming to `EnvoyCluster` but I think that adding `Envoy` *somewhere* makes sense to make it clear what kind of cluster we area dealing with especially since this is going to be a part of a generic `Core` package.

//cc @snowp  in case he has any opinion here.

mcutalo88: Yeah I like the former more. Let me update the PR and we can go from there.

Augustyniak (approved):
Looks good to me

danielhochman (approved):

danielhochman:
i think core should probably just have cluster and this would go in an Envoy package but i think it's fine for now. we can figure out the best course as we work through the catalog stuff.

PR #1174, opened 2021-03-24
oidc: check token store when validating tokens

snowp:
Adds the write part of the token revocation logic introduced in #1157

### DIFF
--- a/backend/service/authn/oidc.go
+++ b/backend/service/authn/oidc.go
@@ -184,19 +184,19 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
 		return nil, err
 	}
 
+	if err := claims.Valid(); err != nil {
+		return nil, err
+	}
+
 	// If the token doesn't exist in the token storage anymore, it must have been revoked.
 	// Fail verification in this case.
 	// TODO(perf): Cache the lookup result in-memory for min(60, timeToExpiry) to prevent
 	// hitting the DB on each request. This should also cache whether we didn't find a token.
 	if p.tokenStorage != nil {
-		token, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProvider)
-		if token == nil {
+		_, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProvider)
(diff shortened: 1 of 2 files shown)


backend/service/authn/oidc.go
```diff
@@ -184,6 +184,17 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
...
 
+	// If the token doesn't exist in the token storage anymore, it must have been revoked.
+	// Fail verification in this case.
+	// TODO(perf): Cache the lookup result in-memory for min(60, timeToExpiry) to prevent
+	// hitting the DB on each request. This should also cache whether we didn't find a token.
+	if p.tokenStorage != nil {
+		token, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProvider)
+		if token == nil {
+			return nil, err
+		}
```
danielhochman, line 195: why isn't this err != nil?

backend/service/authn/oidc.go
```diff
@@ -184,6 +184,17 @@ func (p *OIDCProvider) Verify(ctx context.Context, rawToken string) (*Claims, er
 		return nil, err
 	}
 
+	// If the token doesn't exist in the token storage anymore, it must have been revoked.
+	// Fail verification in this case.
+	// TODO(perf): Cache the lookup result in-memory for min(60, timeToExpiry) to prevent
+	// hitting the DB on each request. This should also cache whether we didn't find a token.
+	if p.tokenStorage != nil {
+		token, err := p.tokenStorage.Read(ctx, claims.Subject, clutchProvider)
```
danielhochman, line 192: i would move this after claims validation, if the token is bad then there's no reason to check it in the database.

backend/service/authn/oidc.go
snowp: It definitely should be, I think at some point I thought I should be checking token for nil to see if it was gone, but we should be getting ErrNoRows back so just checking the err should be sufficient and more correct looking.

danielhochman (approved):
looks good but we should hold off on merge until the revert revert is deployed

PR #1175, opened 2021-03-24
frontend: add FormRow component

dschaller:
### Description

Add a `FormRow` component that can be used within the `Form` component to provide a row layout for fields.

Screen Shot 2021-03-24 at 2 28 07 PM

### Testing Performed

manual and storybook


danielhochman (commented):
what about an alternative which is a row component that can nest within form that would allow you to create multi row forms, e.g.
```
[ field ] [ field ]
[ fld ] [fld] [fld]
[      field      ]
```
which would three `<FormRows>` or something.

danielhochman (approved):
update description plz

PR #1179, opened 2021-03-25
frontend: standardize exports and export all input components

misteroh:
### Description

1. Made sure to export all components in core/input folder e.g. `Switch`,`Radio`, and `Checkbox`
2. Standardized export convention a little bit. (e.g. made `TextField` the default export of `text-field`)
(May want to do the same for all other components, but am a bit wary of touching all of them. For now only did the components that 1.) are named the same as their file name and 2.) are the only component exported from that file)

### Testing Performed
Storybook


danielhochman (approved):
lgtm, thanks

PR #1181, opened 2021-03-25
xds: add helper that ensures Envoy is connected to the control plane

snowp:
Adds a test helper that blocks until Envoy is able to re-establish its xDS stream.
This should help de-flake individual test runs and avoid having to account for the
exponential backoff in other test assertions.

### DIFF
--- a/backend/internal/test/integration/helper/envoytest/envoy.go
+++ b/backend/internal/test/integration/helper/envoytest/envoy.go
@@ -140,9 +140,8 @@ func (e *EnvoyHandle) EnsureControlPlaneConnectivity(prefix string) error {
 			if err != nil {
 				continue
 			}
-			defer resp.Body.Close()
-
 			allStatsString, _ := ioutil.ReadAll(resp.Body)
+			resp.Body.Close()
 			splitStats := strings.Split(string(allStatsString), "\n")
 
 			for _, statString := range splitStats {
--- a/backend/internal/test/integration/helper/envoytest/envoy.go
+++ b/backend/internal/test/integration/helper/envoytest/envoy.go
@@ -3,8 +3,10 @@ package envoytest
 import (
 	"bytes"
 	"errors"
+	"io/ioutil"
 	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	bootstrap "github.com/envoyproxy/go-control-plane/envoy/config/bootstrap/v3"
@@ -16,6 +18,9 @@ import (
 
 	apimock "github.com/lyft/clutch/backend/mock/api"
 )
+
+const EcdsStatPrefix = "http.ingress_http.extension_config_discovery.envoy.extension_config"
+const RuntimeStatPrefix = "runtime"
 
 const baseConfig = `
 node:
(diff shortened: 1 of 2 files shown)


backend/internal/test/integration/helper/envoytest/envoy.go
```diff
@@ -106,6 +111,56 @@ func (e *EnvoyHandle) MakeSimpleCall() (int, error) {
...
+		// arbitrary stats.
+		resp, err := client.Do(r)
+		if err != nil {
+			return err
+		}
+
+		allStatsString, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
```
danielhochman, line 144: would ignoring these errors make more sense and letting it attempt until timeout?

snowp: Yea that's a good idea, would be good to swallow bad responses and whatnot to avoid flakes around startup

backend/internal/test/integration/helper/envoytest/envoy.go
```diff
@@ -106,6 +111,50 @@ func (e *EnvoyHandle) MakeSimpleCall() (int, error) {
...
+	select {
+	case <-timeout.C:
+		return errors.New("timed out waiting for control plane connectivity")
+	case <-ticker.C:
+		// TODO(snowp): Have this parse out a generic map of stats values to make it easier to query
+		// arbitrary stats.
+		// We intentionally ignore errors here, as the proxy might be periodically unavailable but we
+		// don't care as long as it recovers within the timeout.
+		resp, _ := client.Do(r)
+		allStatsString, _ := ioutil.ReadAll(resp.Body)
```
danielhochman, line 139: nit: this is extremely pedantic for a test file but you will leak a connection on each loop here (up to 40 connections) since you're not closing the body.
```suggestion
		resp, err := client.Do(r)
		if err != nil {
			continue
		}
		allStatsString, _ := ioutil.ReadAll(resp.Body)
		resp.Body.Close()
```

backend/internal/test/integration/helper/envoytest/envoy.go
```diff
@@ -106,6 +111,55 @@ func (e *EnvoyHandle) MakeSimpleCall() (int, error) {
...
+			// arbitrary stats.
+			// We intentionally ignore errors here, as the proxy might be periodically unavailable but we
+			// don't care as long as it recovers within the timeout.
+			resp, err := client.Do(r)
+			if err != nil {
+				continue
+			}
+			defer resp.Body.Close()
+
+			allStatsString, _ := ioutil.ReadAll(resp.Body)
```
danielhochman, line 145: won't close until the entire function returns, you need to explicitly close during each iteration of the loop
```suggestion
			allStatsString, _ := ioutil.ReadAll(resp.Body)
			resp.Body.Close()
```

danielhochman (approved):
lgtm with last comment

danielhochman:
oops you beat me to it :+1:

PR #1182, opened 2021-03-25
authn: allow empty refresh tokens in database

### LINKED ISSUE #1177: authn: creds with empty refresh token fail encryption
**Description**
When we try to store a cred with an empty refresh token (which the refresh is always empty for Clutch issued tokens currently), the database fails.

https://github.com/lyft/clutch/blob/8ec3e48657aa68223cb9d90b3efe78dd31df2987/backend/service/authn/storage.go#L73-L76

**Expected Behavior**
An empty refresh token stores an empty refresh token in the database.

danielhochman:
### Description
Previously an error would occur if a credential did not have a refresh token, causing exchange or other token read/write operations to fail.

### Testing Performed
Unit

### GitHub Issue

Fixes #1177

### TODOs
- [x] unit

### DIFF
--- a/backend/service/authn/storage.go
+++ b/backend/service/authn/storage.go
@@ -6,11 +6,12 @@ package authn
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/uber-go/tally"
 	"go.uber.org/zap"
 	"golang.org/x/oauth2"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


snowp (approved):
Thanks for fixing this, looks right to me!

Did you uncover this from my previous change breaking things in staging?

danielhochman:
@snowp yep

dschaller (approved):

PR #1190, opened 2021-03-26
experimentation: framework for auto termination

snowp:
Adds implementation of a experiment monitoring framework that supports specifying a number of TerminationCriteria that will
be evaluated periodically against active experiments, giving them a chance to terminate experiments based on arbitrary
conditions.

This framework relies on the termination action being idempotent: multiple workers might be monitoring the same experiment,
but only the first one to trigger will actually mutate the response, as the termination action is a noop if the experiment is no
longer running.

This PR does not introduce any of the configuration setup necessary to actually configure this for a Clutch deployment,
only the logic for polling active experiments and spawning per experiment goroutines.


backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,156 @@
...
+	go func() {
+		trackedExperiments := map[uint64]context.CancelFunc{}
+		ticker := time.NewTicker(m.outerLoopInterval)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				es, err := m.store.GetExperiments(context.Background(), m.enabledConfigTypes, experimentationv1.GetExperimentsRequest_STATUS_RUNNING)
```
Augustyniak, line 68: Did we update the implementation of GetExperiments so that it supports array of config types now? I cannot find this change. Or is the plan to update it in the following PRs?

backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/golang/protobuf/ptypes"
+	"github.com/uber-go/tally"
+	"go.uber.org/zap"
+
+	experimentationv1 "github.com/lyft/clutch/backend/api/chaos/experimentation/v1"
+	"github.com/lyft/clutch/backend/service/chaos/experimentation/experimentstore"
+)
+
+type TerminationCriteria interface {
+	ShouldTerminate(experimentStarted time.Time, config interface{}) error
```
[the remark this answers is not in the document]
Augustyniak: Idea: Maybe we could allow `ShouldTerminate` method to return a termination reason that we would then record in a database. This way we could make these messages more meaningful and context-aware. Not a must tho.

backend/service/chaos/experimentation/terminator/terminator.go
snowp: The current PR is missing some changes for that, I was gonna do it in here

Right now I think we would blow up at startup since the real store doesn't satisfy the required interface

backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -141,6 +142,17 @@ func (s *storer) CancelExperimentRun(ctx context.Context, id uint64) error {
 	return err
 }
 
+func (s *storer) TerminateExperimentRun(ctx context.Context, id uint64, terminationReason string) error {
```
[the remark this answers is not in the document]
snowp: Yeah probably could, let me take a look

backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,156 @@
...
+	gauge tally.Gauge
+	value uint32
+}
+
+func (t *trackingGauge) inc() {
+	t.gauge.Update(float64(atomic.AddUint32(&t.value, 1)))
+}
+
+func (t *trackingGauge) dec() {
+	t.gauge.Update(float64(atomic.AddUint32(&t.value, ^uint32(0))))
```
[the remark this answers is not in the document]
snowp: https://golang.org/pkg/sync/atomic/#AddUint64

backend/service/chaos/experimentation/terminator/terminator.go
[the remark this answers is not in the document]
snowp: Yeah let me clean this up a bit. Right now it returns an `error` that we record, but maybe it would be clearer with (string, error) and not terminate on errors

backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,156 @@
...
+						cancel()
+						delete(trackedExperiments, e)
+					}
+				}
+			}
+		}
+	}()
+}
+
+// monitorNewExperiments iterates over all the provided experiments, spawning a goroutine to montior each experiment that
```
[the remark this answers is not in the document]
snowp: it is for exported functions, but this isn't one so your suggestion seems nicer

there is some ideas around "if you grep for a function name you want the doc string to show up" and support for godoc only for exported functions), but probably not worth it

backend/module/chaos/serverexperimentation/xds/xds_integration_test.go
```diff
@@ -62,6 +64,59 @@ func TestEnvoyFaults(t *testing.T) {
...
+	criteria := &testCriteria{}
+
+	terminator := terminator.NewTestMonitor(
+		ts.Storer,
+		[]string{"type.googleapis.com/clutch.chaos.serverexperimentation.v1.HTTPFaultConfig"},
+		[]terminator.TerminationCriteria{criteria},
+		ts.Logger.Sugar(),
+		ts.Scope)
+
+	// Canclel to ensure that the terminator doesn't leak into other tests.
```
Augustyniak, line 87: typo - Canclel

backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -26,8 +26,8 @@ const Name = "clutch.service.chaos.experimentation.store"
 // Storer stores experiment data
 type Storer interface {
 	CreateExperiment(context.Context, *any.Any, *time.Time, *time.Time) (*experimentation.Experiment, error)
-	CancelExperimentRun(context.Context, uint64) error
-	GetExperiments(ctx context.Context, configType string, status experimentation.GetExperimentsRequest_Status) ([]*experimentation.Experiment, error)
+	CancelExperimentRun(ctx context.Context, id uint64, reason string) error
+	GetExperiments(ctx context.Context, configTypes string, status experimentation.GetExperimentsRequest_Status) ([]*experimentation.Experiment, error)
```
Augustyniak, line 30: rename to `configType` now that it is a single string again

backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	}
+	return activeExperiments
+}
+
+func (m *monitor) monitorSingleExperiment(ctx context.Context, e *experimentationv1.Experiment) {
+	ticker := time.NewTicker(m.perExperimentCheckInterval)
+	terminated := false
+
+	dany := &ptypes.DynamicAny{}
```
Augustyniak, line 121: so this can be just casted to a specific type - i.e., serverexperimentation model - without any conversions?

backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,176 @@
...
+			for _, c := range m.criterias {
+				terminationReason, err := c.ShouldTerminate(e, dany)
+				// TODO(snowp): The logs here might get spammy, rate limit or terminate montioring routine somehow?
+				if err != nil {
+					m.log.Errorw("error while evaluating termination criteria", "id", e.Id, "error", err)
+					continue
+				}
+
+				if terminationReason != "" {
+					err = m.store.CancelExperimentRun(context.Background(), e.Id, terminationReason)
```
Augustyniak, line 148: nice

backend/service/chaos/experimentation/terminator/terminator.go
```diff
@@ -0,0 +1,176 @@
...
+		case <-ticker.C:
+			if terminated {
+				// Note that we don't exit the goroutine here and instead wait for the outer
+				// monitoring loop to detect that this experiment is no longer active, which will
+				// cause this goroutine to be canceled. Exiting here is not helpful since the outer
+				// loop can race and restart this goroutine.
+				continue
+			}
+			for _, c := range m.criterias {
+				terminationReason, err := c.ShouldTerminate(e, dany)
```
Augustyniak, line 140: I know that we have a gave for the number of currently monitored experiments but could we add a counter for successful and failed (with error returned) `ShouldTerminate` calls?

backend/service/chaos/experimentation/terminator/terminator.go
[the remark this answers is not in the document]
Augustyniak: not sure whether you one to follow up - no strong opinions here - but pining you here in case you want to remove it for consistency

Augustyniak (commented):
Looking good. Minor comments here and there and trying to understand DynamicAny better

backend/service/chaos/experimentation/terminator/terminator.go
snowp: Updated to use a proto.Message as per the deprecation warnings on DynamicAny, so yeah this can just be casted to the specific proto

Augustyniak: nice, created an internal ticket for changing the implementation of `ExperimentConfig` so that it uses `DynamicAny`

Augustyniak (approved):
Looking good 👍

Issue #8, opened 2020-07-18
core: slack gateway


danielhochman:
**Description**

Allow `READ` actions to be accessed through a Slackbot interface.

Initially this could be a generic JSON dump of the endpoint data. Later we can have rich transforms that render the data in a more friendly format.

Complexity [S/M/L]: M

**Wireframes, Mockups, or Screenshots (if applicable)**
`@clutchbot aws asg describe myAsg123`
`@clutchbot k8s pod describe <cluster> <mypod>`

PR #1201, opened 2021-03-30
slackbot: api and config protos

scarlettperry:
### Description
PR creates the protos for the API that will receive requests from the [Slack Events API](https://api.slack.com/apis/connections/events-api#the-events-api__receiving-events).
* given that the request/response contain sensitive data (ie tokens, challenges), I've add the `redacted` option to the request/response protos so that the field values don't get saved in the audit DB. Ideally, it'd be nice to specify which exact fields we want to redact, so I opened an [issue](https://github.com/lyft/clutch/issues/1227). I can follow up on that later.

PR also creates a new config for the user to specify the slack bot token and signing secret, which will be used to create a slack client instance via [slack-go](https://pkg.go.dev/github.com/slack-go/slack) and confirm that each request came from Slack.
* we decided that the module will be provider-specific (ie Slack, Discord, etc) and that the service will be generic, which is why I set the module to have the config.

### Testing Performed
Locally
Screen Shot 2021-04-02 at 4 36 38 PM

### GitHub Issue


api/config/module/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,16 @@
...
+
+package clutch.config.module.bot.slackbot.v1;
+
+option go_package = "github.com/lyft/clutch/backend/api/config/module/bot/slackbot/v1;slackbotv1";
+
+import "validate/validate.proto";
+
+message Config {
+  // Bot tokens begin with "xoxb-", https://api.slack.com/authentication/token-types#bot
+  string bot_token = 1 [ (validate.rules).string = {min_bytes : 1} ];
```
dschaller, line 11: Is there a chance we would want to have potentially n bot tokens if somehow we wanted a single clutch instance to talk to multiple slack teams or potentially have many bots?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+// Information on the bot user
+message Bot {
+  // bot id
+  string id = 1;
+  // if the bot is deleted
+  bool deleted = 2;
+  // bot name
+  string name = 3;
+  // timestamp of when the bot app was last updated
+  int64 updated = 4;
```
dschaller, line 30:
```suggestion
  int64 last_updated = 4;
```

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // unique identifier of the installed Slack application
+  string app_id = 5;
+  // image urls
+  map<string, string> icons = 6;
```
dschaller, line 34: It's kind of unclear what icons is scoped to? Is it possible for a bot to have icons?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// The structure of events vary among types. Full list of event types: https://api.slack.com/events.
+// We will be receiving app_mention events (events that mention the bot) or message.im events (a message posted in a DM
+// with the bot).
+message Event {
+  // the type of event
+  string type = 1;
+  // the user who messaged the bot
+  string user = 2;
```
dschaller, line 46: is this the user name or the user id?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  string bot_id = 3;
+  Bot bot_profile = 4;
```
dschaller, line 48: comments? :)

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // the message text
+  string text = 5;
+  // timestamp
+  string ts = 6;
+  // the channel where the event happened
+  string channel = 7;
```
dschaller, line 54: is this the name of the channel or the id?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // for DM messages, this will be "im"
+  string channel_type = 8;
+  // when the event was dispatched
+  string event_ts = 9;
```
dschaller, line 58: What's the difference between `ts` and this?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  string client_msg_id = 10;
```
dschaller, line 59: comment?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // unique identifier of the workspace where the event occurred
+  string team = 11;
```
dschaller, line 61: is this the same as `team_id` above?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // received when a user interacts with a Block Kit component, schema can vary.
+  // https://api.slack.com/reference/block-kit/interactive-components
+  google.protobuf.Value blocks = 12;
+}
+
+// For more details:
+// https://api.slack.com/apis/connections/events-api#the-events-api__receiving-events
+// https://api.slack.com/enterprise/apps/reference#event
+message EventRequest {
+  option (clutch.api.v1.redacted) = true;
```
dschaller, line 71: add a todo with a link to #1227?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  google.protobuf.Value authorizations = 10;
+  // randomly generated string used as part of the URL verification handshake,
+  // https://api.slack.com/apis/connections/events-api#the-events-api__subscribing-to-event-types__events-api-request-urls__request-url-configuration--verification__url-verification-handshake
+  string challenge = 11;
+}
+
+// Respond back to the Slack Events API with the challenge or a 2xx,
+// https://api.slack.com/apis/connections/events-api#the-events-api__responding-to-events
+message EventResponse {
+  option (clutch.api.v1.redacted) = true;
```
dschaller, line 102: add a todo with a link to #1227?

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  google.protobuf.Value blocks = 12;
+}
+
+// For more details:
+// https://api.slack.com/apis/connections/events-api#the-events-api__receiving-events
+// https://api.slack.com/enterprise/apps/reference#event
+message EventRequest {
+  option (clutch.api.v1.redacted) = true;
+  // verification token to validate the event originated from Slack
+  string token = 1;
```
dschaller, line 73:
```suggestion
  string verification_token = 1;
```

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // if where the event occured is an external shared channel
+  bool is_ext_shared_channel = 2;
+  // unique identifier of the workspace where the event occurred
+  string team_id = 3;
+  // unique identifier of the installed Slack application
+  string api_app_id = 4;
```
dschaller, line 79: is this scoped to only the API? I think the app ID is shared everywhere

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  Event event = 5;
+  // indicates which kind of event this is
+  string type = 6;
+  // identifier for this specific event, globally unique across all workspaces
+  string event_id = 7;
+  // epoch timestamp in seconds indicating when this event was dispatched
+  int64 event_time = 8;
+  // an identifier for this specific event, can be used with list of authorizations to obtain a full list of
+  // installations of your app that this event is visible to
+  string event_context = 9;
```
dschaller, line 91: This might be due to the slack design but I would expect these to be on the event object

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,106 @@
...
+  // an installation of the app, object type
+  google.protobuf.Value authorizations = 10;
```
dschaller, line 93: It's a bit unclear what this is

api/bot/slackbot/v1/slackbot.proto
scarlettperry: `updated` is the field name that the Slack API sends to our endpoint so can't rename
example
```
{
        "id": "B061F7JD2",
        "deleted": false,
        "name": "beforebot",
        "updated": 1449272004,
        "app_id": "A161CLERW",
        "user_id": "U012ABCDEF",
        "icons": {
            "image_36": "https://...",
            "image_48": "https://...",
            "image_72": "https://..."
        }
    }
```

api/bot/slackbot/v1/slackbot.proto
scarlettperry: I'll add a comment that makes it clearer but the Slack API sends us something like the snippet below which refers to the display picture of the bot. I'm guessing that Slack saves a given image in a few different sizes
```
"icons": {
            "image_36": "https://...",
            "image_48": "https://...",
            "image_72": "https://..."
        }
```

api/bot/slackbot/v1/slackbot.proto
scarlettperry: user id, will clarify in comment

api/bot/slackbot/v1/slackbot.proto
scarlettperry: yep! will add a comment

api/bot/slackbot/v1/slackbot.proto
scarlettperry: channel id, will clarify in comment

api/bot/slackbot/v1/slackbot.proto
scarlettperry: yep it's the same as the `team_id` in the outer object that the API sends us. in the event object it's named `team`. Will add a comment that it's an identical value as `team_id`

api/bot/slackbot/v1/slackbot.proto
scarlettperry: similar to https://github.com/lyft/clutch/pull/1201/files#r607310647. the API request send to our endpoint has the field `token`

api/bot/slackbot/v1/slackbot.proto
scarlettperry: yeah due to slack API design. the request they send us have this field on the outer layer and not on the event object 
https://api.slack.com/apis/connections/events-api#the-events-api__receiving-events__events-dispatched-as-json

api/bot/slackbot/v1/slackbot.proto
scarlettperry: i'll add a comment

api/bot/slackbot/v1/slackbot.proto
scarlettperry: yep! will add the todo

api/bot/slackbot/v1/slackbot.proto
scarlettperry: in this case, the response just has the challenge field so i think it's fine to redact the entire response

api/bot/slackbot/v1/slackbot.proto
scarlettperry: from what the slack docs mention, event_ts is when the real-time ts of when the event happened whereas ts is when slack sent us the request for this event
full explanation in the slack docs https://api.slack.com/changelog/2016-05-31-more-events-timestamps-in-rtm-api

will clarify in the comments

api/bot/slackbot/v1/slackbot.proto
scarlettperry: the slack API sends us this field to our endpoint, and it's only on the outer layer of the request
example
```
{
    "token": "XXYYZZ",
    "team_id": "TXXXXXXXX",
    "api_app_id": "AXXXXXXXXX",
    "event": {
            "type": "name_of_event",
            "event_ts": "1234567890.123456",
            "user": "UXXXXXXX1",
            ...
    },
```

api/bot/slackbot/v1/slackbot.proto
scarlettperry: This is a weird one - the Slack Events API don't mention that this is a field sent a request to us and it's also not included in their examples. However, when testing locally, I was getting request failures as their API was sending us this field and we weren't supporting it. Turns out it's actually a mistake this this field is part of the request https://github.com/slackapi/python-slack-sdk/issues/736#issuecomment-653115442 and they were supposed to clean that up. I can add a comment to this affect

api/config/module/bot/slackbot/v1/slackbot.proto
scarlettperry: yeah it crossed my mind that it's something we might do in the future. though if we were to support that, then I think the config structure would be different. Perhaps it would be similar to Omnibot's https://lyft.github.io/omnibot/configuration.html#bots, where we'd want the bot app id, bot name, and bot token to know which messages to route to which bot.

I'll add a todo for now and come back to it when I open a PR for the module b/c may want to add more stuff

scarlettperry:
@dschaller, thank you for all the feedback! addressed, lmk if you have any Qs about the comments I left.

api/bot/slackbot/v1/slackbot.proto
```diff
@@ -0,0 +1,113 @@
...
+// We will be receiving app_mention events (events that mention the bot) or message.im events (a message posted in a DM
+// with the bot).
+message Event {
+  // the type of event
+  string type = 1;
+  // the user id of the user who messaged the bot
+  string user = 2;
+  string bot_id = 3;
+  // metadata on the bot
+  Bot bot_profile = 4;
```
mikecutalo, line 49: if we have the `bot_profile` do we need the `bot_id` as a top level key?

scarlettperry: yeah, the `bot_profile` only gets sent when you DM the bot directly (it's an "im" event and the Slack API sends us some extra metadata) so for other event types, we wouldn't get this and can still make use of the `bot_id` top level key

api/config/module/bot/slackbot/v1/slackbot.proto
dschaller: 👍  sounds good. My only concern would be that modifying the config down the road might be more difficult

api/bot/slackbot/v1/slackbot.proto
dschaller: :(

dschaller (approved):

PR #1188, opened 2021-03-26
experimentation: add endpoint that returns existing experiment if needed

Augustyniak:
### Description

1. Change the type of `id` column of an `experiment_run` from `BIGINT` to `varchar(100)`.
1. Change the type of `experiment_config_id` column of an `experiment_run` from `BIGINT` to `varchar(100)`.
1. Change the type of `id` of an `experiment_config` from `BIGINT` to `varchar(100)`.
1. Add primary keys to `experiment_run` and `experiment_config` tables.
1. Allow users of `CreateExperiment` method call to provide an experiment run identifier. Default to random one if it's not provided by a caller.
1. Add `CreateorGetExperiment` endpoint to `experimentation`. The endpoint returns an existing experiment if experiment with an experiment run ID provided by a caller already exists.

### Testing Performed
1. Manual addition of experiments
1. Running migrations command (both up and down) manually.

PR #1202, opened 2021-03-30
experimentation: fix integration tests

Augustyniak:
### Description
 
The race condition happened when https://github.com/lyft/clutch/pull/1188 was merged which resulted in broken integration tests. Fixing them in this PR.


Augustyniak:
Fixed as part of #1205 instead.

PR #1205, opened 2021-03-30
experimentation: reintroduce experimentation endpoint

Augustyniak:
### Description
Reintroducing the `createOrGetExperiment` endpoint that was originally introduced in https://github.com/lyft/clutch/pull/1188 and reverted in https://github.com/lyft/clutch/pull/1204 due to issues with migrations.

Other than reverting the mentioned PR this PR introduced the following changes:
1. Fix one of the unit tests in `xds_cache.test.go file`.  Previous attempts to fix this tests were made in https://github.com/lyft/clutch/pull/1202.
2. Splits problematic migrations into two separate groups of migrations - one for `experiment_run` and one of `experiment_config` tables - maybe this will help us to get rid of locking but it looks like a good practice overall.

### Testing Performed
Running migrations locally.


snowp (approved):
LGTM assuming we can make sure the migrations won't lock up again before we hit production

backend/module/chaos/serverexperimentation/xds/xds_cache_test.go
```diff
@@ -309,7 +309,16 @@ func TestRefreshCache(t *testing.T) {
 	a, err := ptypes.MarshalAny(&config)
 	assert.NoError(t, err)
 
-	_, err = s.CreateExperiment(context.Background(), a, &now, &now)
+	now := time.Date(2011, 0, 0, 0, 0, 0, 0, time.UTC)
+	assert.NoError(t, err)
```
snowp, line 313: This is just checking the err from L309, remove?

PR #1207, opened 2021-03-31
authn: CreateToken API endpoint 

snowp:
Adds an endpoint that allows an operator to issue a long lived token via an API call.


api/authn/v1/authn.proto
```diff
@@ -52,3 +60,20 @@ message CallbackResponse {
...
 }
+
+message CreateTokenRequest {
+  option (clutch.api.v1.redacted) = true;
+
+  // The subject to issue this token for.
+  string subject = 1;
+
+  // The duration until this token should expire.
+  google.protobuf.Duration expiry = 2;
```
danielhochman, line 71: what is default? i think we'll want no expiration as default or at least a possibility.

backend/service/authn/oidc.go
```diff
@@ -133,6 +133,23 @@ func (p *OIDCProvider) Exchange(ctx context.Context, code string) (*oauth2.Token
...
+	return p.signNewToken(ctx, claims)
+}
+
+func (p *OIDCProvider) CreateToken(ctx context.Context, subject string, expiry time.Duration) (*oauth2.Token, error) {
+	claims := &Claims{
+		StandardClaims: &jwt.StandardClaims{
+			ExpiresAt: time.Now().Add(expiry).Unix(),
+			IssuedAt:  time.Now().Unix(),
+			Issuer:    clutchProvider,
+			Subject:   subject,
```
danielhochman, line 145: we should enforce some prefix for the subject so you can't reasonably override a user, and so that you can tell what type of user it is more easily in logs.

for example we have `system:anonymous`, maybe `service:<subject>`?

backend/service/authn/authn.go
```diff
@@ -51,4 +52,5 @@ type Provider interface {
 	Verify(ctx context.Context, rawIDToken string) (*Claims, error)
 	GetAuthCodeURL(ctx context.Context, state string) (string, error)
 	Exchange(ctx context.Context, code string) (token *oauth2.Token, err error)
+	CreateToken(ctx context.Context, subject string, expiry time.Duration) (token *oauth2.Token, err error)
```
danielhochman, line 55: this belongs in a separate Issuer interface

snowp: I gave this a shot but it results in some awkward casts in some places, maybe I'm missing something?

backend/module/authn/authn.go
```diff
@@ -105,3 +115,28 @@ func (a *api) Callback(ctx context.Context, request *authnv1.CallbackRequest) (*
 		AccessToken: token.AccessToken,
 	}, nil
 }
+
+func (a *api) CreateToken(ctx context.Context, request *authnv1.CreateTokenRequest) (*authnv1.CreateTokenResponse, error) {
```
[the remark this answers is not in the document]
snowp: Yeah, I'll add some basic endpoints for the other endpoints as well just for coverage purposes

danielhochman:
the test failure is due to #1208

snowp:
@danielhochman Mind giving this another look?

backend/mock/service/authnmock/grpcmock.go
```diff
@@ -0,0 +1,20 @@
+package authnmock
```
danielhochman, line 1: can you move this to a more generic package? doesn't seem specific to authn

backend/module/authn/authn.go
```diff
@@ -104,3 +109,32 @@ func (a *api) Callback(ctx context.Context, request *authnv1.CallbackRequest) (*
...
+	case authnv1.CreateTokenRequest_SERVICE:
+		prefixedSubject = "service:" + request.Subject
+	default:
+		return nil, status.Error(codes.InvalidArgument, "invalid token type")
+	}
+
+	var expiry *time.Duration
+
+	if request.Expiry != nil {
+		convertedExpiry, err := ptypes.Duration(request.Expiry)
```
danielhochman, line 125: ptypes.Duration is deprecated, use durationpb package

this is enforced on latest main

backend/module/authn/authn.go
```diff
@@ -104,3 +109,32 @@ func (a *api) Callback(ctx context.Context, request *authnv1.CallbackRequest) (*
...
 }
+
+func (a *api) CreateToken(ctx context.Context, request *authnv1.CreateTokenRequest) (*authnv1.CreateTokenResponse, error) {
+	var prefixedSubject string
+	switch request.TokenType {
+	case authnv1.CreateTokenRequest_SERVICE:
+		prefixedSubject = "service:" + request.Subject
+	default:
+		return nil, status.Error(codes.InvalidArgument, "invalid token type")
+	}
```
danielhochman, line 120: let's put this logic in the service, not the module, since that's where helpers live to deal with deciphering and understanding claims, subjects, etc, and we may want a helper that is service aware. plumb the token type one level deeper.

backend/module/authn/authn.go
```diff
@@ -104,3 +106,21 @@ func (a *api) Callback(ctx context.Context, request *authnv1.CallbackRequest) (*
 		AccessToken: token.AccessToken,
 	}, nil
 }
+
+func (a *api) CreateToken(ctx context.Context, request *authnv1.CreateTokenRequest) (*authnv1.CreateTokenResponse, error) {
```
danielhochman, line 110: i think calls to this endpoint need to fail if no storage is configured, otherwise you can create a permanently valid irrevocable token.

danielhochman (commented):
one last comment

backend/module/authn/authn.go
snowp: That's a great point, let me make the change

danielhochman:
This pattern needs to change to exclude the CreateToken endpoint.
https://github.com/lyft/clutch/blob/2c4246b001914671f639958d7b6c82a3ceb2600e/backend/middleware/authn/authn.go#L31

danielhochman:
Same goes for authz middleware actually
https://github.com/lyft/clutch/blob/eec53c69d8c8511707fffae8ea3ff776a2c249ae/backend/middleware/authz/authz.go#L32

danielhochman:
i can draft that

danielhochman:
https://github.com/lyft/clutch/pull/1266

snowp:
Seems like you merged your change, anything else needed for this PR?

danielhochman (approved):

PR #1210, opened 2021-03-31
frontend: use grpc code as status text in errors

dschaller:
### Description

Translate the gRPC code into text for the error components status text.

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Network/errors.ts
+++ b/frontend/packages/core/src/Network/errors.ts
@@ -1,6 +1,7 @@
 import type { clutch as IClutch } from "@clutch-sh/api";
 import type { AxiosError } from "axios";
 
+import { grpcCodeToText } from "./grpc";
 import type { HttpStatus } from "./index";
 
 interface Details {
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Network/errors.ts
```diff
@@ -263,7 +264,7 @@ const grpcResponseToError = (clientError: AxiosError): ClutchError => {
     message: data.message,
     status: {
       code: clientError.response.status,
-      text: clientError.response.statusText,
+      text: grpcCodeToText(data.code),
```
danielhochman, line 267: can data be null?

danielhochman (approved):

frontend/packages/core/src/Network/errors.ts
dschaller: It'll always be set
```tsx
export interface AxiosResponse<T = any>  {
  data: T;
  status: number;
  statusText: string;
  headers: any;
  config: AxiosRequestConfig;
  request?: any;
}
```

dschaller:
CLA bump

PR #1218, opened 2021-04-01
frontend: add query hydration to resolver

dschaller:
### Description

Add query param hydration support to workflows that utilize the resolver.

state

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Resolver/index.tsx
+++ b/frontend/packages/core/src/Resolver/index.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { useSearchParams } from "react-router-dom";
 import styled from "@emotion/styled";
 import _ from "lodash";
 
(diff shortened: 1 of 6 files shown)


frontend/packages/core/src/Resolver/index.tsx
```diff
@@ -91,6 +92,13 @@ const Resolver: React.FC<ResolverProps> = ({
     );
   };
 
+  const [searchParams, ,] = useSearchParams();
```
danielhochman, line 95: same thing right?
```suggestion
  const [searchParams] = useSearchParams();
```

danielhochman (commented):
how about setting it on submit of the search resolver?

frontend/packages/core/src/Resolver/index.tsx
dschaller: ah yupp :)

dschaller:
👍  that should work. We intentionally reset it on restart so there shouldn't be any issues with that

danielhochman (commented):
last request, can we populate the search field with defaultValue as q in case submission fails and its needs to be edited?

danielhochman (approved):
great! 26 lines buys us at least a month or two on real state hydration

dschaller:
just have to remember to clean it up ;)

PR #1219, opened 2021-04-01
wizard: add width prop

danielhochman:
### Description
Adds a width prop to the wizard component.

Before
Screenshot from 2021-04-01 17-37-34

After
Screenshot from 2021-04-01 17-36-54

### Testing Performed
Manual, build.

### DIFF
--- a/frontend/packages/wizard/src/wizard.tsx
+++ b/frontend/packages/wizard/src/wizard.tsx
@@ -32,7 +32,7 @@ interface WizardStepData {
   [index: string]: any;
 }
 
-type ContainerProps = {
+interface ContainerProps {
   width?: "default" | "full";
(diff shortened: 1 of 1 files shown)


dschaller (approved):

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -32,17 +32,26 @@ interface WizardStepData {
   [index: string]: any;
 }
 
-const Container = styled(MuiContainer)({
-  padding: "32px",
-  maxWidth: "800px",
-});
+type ContainerProps = {
```
dschaller, line 35:
```suggestion
interface ContainerProps = {
```

PR #1225, opened 2021-04-02
Experimentation: Decouple redis and server fault experimentation protos

donyu:
### Description
As part of promoting open source usage of server & redis experimentation framework, we want to separate redis vs. server experimentation so that either one could be used without a dependency on the other.

This PR splits the relevant protobufs to decouple these experimentation types.

### Testing Performed
Make relevant changes against frontend and make sure that redis experiments can still be started, viewed, and stopped.


api/chaos/redisexperimentation/v1/redisexperimentation.proto
```diff
@@ -0,0 +1,67 @@
...
+message FaultPercentage {
+  // The numerator of a percentage with a fixed denumerator equal to 100
+  // (i.e. percentage equal to 50 results in 50/100 = 50%)
+  uint32 percentage = 1 [ (validate.rules).uint32 = {gt : 0, lte : 100} ];
+}
+
+// The latency duration to apply as part of a latency fault.
+message FaultLatencyDuration {
+  // The fixed latency duration in milliseconds.
+  uint32 fixed_duration_ms = 1 [ (validate.rules).uint32.gt = 0 ];
```
Augustyniak, line 66: I thought that this was hardcoded for redis faults

api/chaos/redisexperimentation/v1/redisexperimentation.proto
```diff
@@ -0,0 +1,67 @@
...
+  // The targeting of the fault describing what redis requests are being considered for faults.
+  RedisFaultTargeting fault_targeting = 1 [ (validate.rules).message.required = true ];
+
+  oneof fault {
+    option (validate.required) = true;
+
+    // The error fault responds to redis commands with an error.
+    RedisErrorFault error_fault = 2;
+    // The latency fault delays redis commands.
+    LatencyFault latency_fault = 3;
```
Augustyniak, line 20: Do we want to add Redis prefix here or remove it from `RedisErrorFault `? I would go with latter I think

api/chaos/redisexperimentation/v1/redisexperimentation.proto
```diff
@@ -0,0 +1,67 @@
...
+package clutch.chaos.redisexperimentation.v1;
+
+import "validate/validate.proto";
+
+option go_package = "github.com/lyft/clutch/backend/api/chaos/redisexperimentation/v1;redisexperimentationv1";
+
+// The configuration of a Redis fault.
+message RedisFaultConfig {
+  // The targeting of the fault describing what redis requests are being considered for faults.
+  RedisFaultTargeting fault_targeting = 1 [ (validate.rules).message.required = true ];
```
Augustyniak, line 12: Maybe drop Redis prefix now that it's in a separate file?

api/chaos/redisexperimentation/v1/redisexperimentation.proto
donyu: For now, but I have an active JIRA (https://jira.lyft.net/browse/RES-2116) to make customizable. I'd like to construct the protos in a way to support the longer-term vision.

api/chaos/redisexperimentation/v1/redisexperimentation.proto
donyu: Yeah no need for the Redis prefix anymore here. I'll drop.

Augustyniak (approved):

PR #1226, opened 2021-04-02
k8s: update pod proto to extract more status info

rithujohn191:
### Description
Update Pod proto message with status reason and PodConditions

### Testing Performed
-local
- Unit


api/k8s/v1/k8s.proto
```diff
@@ -288,6 +307,14 @@ message Pod {
 
   map<string, string> labels = 9;
   map<string, string> annotations = 10;
+
+  // message indicating details about why the pod is in this state.
+  // https://pkg.go.dev/k8s.io/api/core/v1#PodStatus
+  string state_reason = 11;
```
scarlettperry, line 313: how does this get filled? is it something you'll populate in the internal implementation?

rithujohn191: Yea this is a filed in the Pod Status: https://pkg.go.dev/k8s.io/api/core/v1#PodStatus

api/k8s/v1/k8s.proto
```diff
@@ -258,6 +258,25 @@ message Container {
   int32 restart_count = 5;
 }
 
+message PodCondition {
+  enum Type {
+    TYPE_UNSPECIFIED = 0;
+    CONTAINERS_READY = 1;
+    INITIALIZED = 2;
+    READY = 3;
```
scarlettperry, line 266: a super nit, to match the upstream field names
```suggestion
    POD_INITIALIZED = 2;
    POD_READY = 3;
```

scarlettperry (approved):
lgtm! just 1 nit

api/k8s/v1/k8s.proto
rithujohn191: I was trying to match it to the strings defined in https://github.com/kubernetes/api/blob/v0.20.5/core/v1/types.go#L2520 so it's easier to compare. Does the meaning feel unclear?

rithujohn191: The `type PodConditionType` is actually a `string`

rithujohn191: But I see your point of matching it to the const defined here https://pkg.go.dev/k8s.io/api/core/v1#PodConditionType. Just thought that string comparison would be easier if https://github.com/lyft/clutch/pull/1226/files#diff-22775087d9b453b1fc3ce612415f660ea48a35823bab006fda020221bdaa3faeR96 maps to the exact string value

scarlettperry: oh I see! okay yeah I agree, the current name is good 👍

PR #1250, opened 2021-04-06
slackbot: module

scarlettperry:
### Description
PR adds the building blocks for the module that interfaces with the Slack API
* receives & processes the request from Slack
* sends the users's message to the bot service, where it's mapped to a reply (for now, just `Hello, World!` or a default help message)

### Testing Performed
Locally

**App Mention Event** (ie messaging bot in a slack channel)
drawing

**IM Message Event** (ie messaging bot in a DM)
drawing

### GitHub Issue

### TODOs
Example:
- [x] separate ack'ing the slack request and handling the event 
- [x] unit tests


backend/gateway/mux/mux.go
```diff
@@ -180,6 +180,20 @@ func customResponseForwarder(ctx context.Context, w http.ResponseWriter, resp pr
...
 }
 
+// https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_your_gateway/#mapping-from-http-request-headers-to-grpc-client-metadata
+// adds custom rules for the Slack API headers that would otherwise be dropped as they don't meet the preconditions of the default rules
+func customMatcher(key string) (string, bool) {
+	switch key {
+	case "X-Slack-Signature":
+		return key, true
+	case "X-Slack-Request-Timestamp":
+		return key, true
```
danielhochman, line 190: i would forward all X- headers with the grpcgateway- prefix, no need to make this specific to slack really. in our case we could forward all headers really, but starting with x- is fine for now.

let's also put it in a separate PR.

scarlettperry: ok, will move out of this PR

scarlettperry:
@danielhochman still need to add unit tests but can I get some early feedback? I've kept the bot service really simple for now till i get to the grpc reflection step.

Something I'd like your thoughts on is async processing. Slack docs recommend separating the process of (1) acknowledging receipt of the request with a 2xx response and (2) handling the event. In the PR, I left a comment to denote at which step we can do this (after we've verified it's a request from Slack and after we've checked if it's a URL verification event).

We receive 1 event in each request so some thoughts:
* a simple job queue. create an event channel and start a worker goroutine? 
*  i need to look at the audit service code more closely but perhaps something similar in how it passes an event to the audit sink which does whatever it wants with it

do you have something else in mind?

Lastly, a topic we can have a converstation on later are dropped events. Are some dropped events okay? Do we want to persist in memory or externally somewhere (SQS, Kafka). Just something I've been thinking about.

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,249 @@
...
+	// request headers used to verify a request came from Slack
+	hSignature = "x-Slack-Signature"
+	hTimestamp = "x-Slack-Request-Timestamp"
+	// slack event type for DMs with the bot
+	dmMessage = "im"
+)
+
+func New(cfg *anypb.Any, logger *zap.Logger, scope tally.Scope) (module.Module, error) {
+	config := &slackbotconfigv1.Config{}
+	if cfg != nil {
```
danielhochman, line 38: is a nil config valid?

backend/module/bot/slackbot/slack_helper_test.go
```diff
@@ -0,0 +1,53 @@
+package slackbot
```
danielhochman, line 1: nit filename slackbot_helper_test

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,249 @@
...
+		return err
+	}
+
+	// returns a SecretsVerifier object in exchange for an http.Header object and signing secret
+	sv, err := slack.NewSecretsVerifier(headers, signingSecret)
+	if err != nil {
+		return err
+	}
+
+	body, err := protojson.Marshal(req)
```
danielhochman, line 144: this is almost definitely not going to work. if the body is reordered or changed in anyway the signature won't match, and will definitely change after a roundtrip through the proto layer.

unfortunately, we're probably going to have to do some significant hacking to get at the raw input at this point in the code.

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,249 @@
...
+// the request has two layers - the "outer event" and "inner event". At this step, we are examining the outer event
+// this can be 1 of three types: "url_verification" (handled seperately above), "event_callback", and "app_rate_limited"
+func (m *mod) handleEvent(req *slackbotv1.EventRequest) error {
+	switch req.Type {
+	case slackevents.CallbackEvent:
+		return m.handleCallBackEvent(req.Event)
+	case slackevents.AppRateLimited:
+		// this type of event is sent if our endpoint has received more than 30,000 request events in 60 minutes
+		// https://api.slack.com/apis/connections/events-api#the-events-api__responding-to-events__rate-limiting
+		// TODO: (sperry) do we want to handle this differently
```
danielhochman, line 186: log it

backend/module/bot/slackbot/slackbot_helper.go
```diff
@@ -0,0 +1,51 @@
...
+// this is a sanitization check before matching the command
+func TrimRedundantSpaces(text string) string {
+	text = spaceRegex.ReplaceAllString(text, " ")
+	return strings.TrimSpace(text)
+}
+
+// DefaultHelp returns the default help message is the command isn't matched
+func DefaultHelp() string {
+	return fmt.Sprintf("%s\n%s", HelpIntro, HelpDetails)
+}
```
danielhochman, line 43: this should be in the service, i get it's a proof of concept though, so maybe it's fine for now.

danielhochman:
> a simple job queue. create an event channel and start a worker goroutine
yep this sounds good

> Lastly, a topic we can have a converstation on later but wanted to raise: dropped events. Are some dropped events okay? Do we want to persist in memory or externally somewhere (SQS, Kafka)? Just something I've been thinking about.

for now, just drop them on the floor, not worth any extra complexity as long as we're doing straightforward read-only request/response commands. later on we may use postgres as a queue layer similar to audit if we require durability.

backend/module/bot/slackbot/slackbot.go
danielhochman: just saw the screenshots... i'm perplexed as to how it would work.

here's an issue talking about access to raw request bodies https://github.com/grpc-ecosystem/grpc-gateway/issues/652

scarlettperry: sorry, i forgot to leave a PR todo that I'm still figuring out how to get this to work. locally I have this piece commented out

backend/module/bot/slackbot/slackbot_helper.go
scarlettperry: i put it at the module b/c message formatting could look different among providers (ie Slack vs Discord) and I wanted to the the slack code block formatting for the help message

danielhochman: i thought about it for a bit and i don't think the module would be involved in formatting, i think that should still happen in the service. the module just handles ingest and response (either async or sync depending on which service method it calls), and the service handles everything else.

backend/module/bot/slackbot/slackbot.go
scarlettperry: ah thanks! wasn't exactly sure why this wasn't working. hmm okay going to brainstorm on this. we can also verify the request came from slack by checking the verification token sent in the request, though this flow in the PR is the preferred way.

backend/module/bot/slackbot/slackbot.go
scarlettperry: ah nope! 
will update, thx for catching

backend/module/bot/slackbot/slackbot.go
scarlettperry: i'm going to come back to this in a follow up PR. Some things I've been thinking about
(1) change the API design to use `bytes` as the request body like in the example solution in the issue linked above
(2) verify request through mutual tls https://api.slack.com/authentication/verifying-requests-from-slack#verifying-requests-from-slack-using-signing-secrets__mutual-tls
(3) use mux to create a http handler for our slack events endpoint so that I can have access to the raw request body
(4) still make this work with the current API design but don't know what kind of hacking I need to do yet

for now in this PR, I'll verify the request came from Slack by checking the verification token sent by Slack in the request. Slack still supports this style of verification but this will be a short-term solution since signing secrets or a mutal tls is the more secure way.

backend/module/bot/slackbot/slackbot_helper.go
scarlettperry: got it, shuffled things around. the current flow is that the service will have some provider specific entry points that then all route to the generic command that matches a message string to a clutch api call.

scarlettperry:
PR's getting too big. Going to break out the proto changes and service in separate PRs, and this will just have the module flow.

backend/module/bot/slackbot/slackbot.go
danielhochman: (2) mutual TLS is probably not viable because Clutch won't necessarily be deployed at the edge. if there's a proxy there (e.g. Envoy) doing the ability to verify from Clutch will be lost.

(3) too bespoke, circumvents middleware, etc. application logic should exist whenever possible in a gRPC handler.

backend/module/bot/slackbot/slackbot_helper.go
scarlettperry: convo moved to https://github.com/lyft/clutch/pull/1282#pullrequestreview-633711821

backend/module/bot/slackbot/slackbot.go
scarlettperry: ty for the feedback! i'm going to try option #1 in the next PR.

backend/module/bot/slackbot/slackbot_test.go
```diff
@@ -0,0 +1,138 @@
...
+	assert.Error(t, err)
+}
+
+func TestAppMentionEvent(t *testing.T) {
+	m := modMock(t)
+
+	// setting up slack mocks
+	slacktest.NewTestServer()
+	s := slacktest.NewTestServer()
+	go s.Start()
```
scarlettperry, line 111: following testing patterns in go slack package: https://github.com/slack-go/slack/blob/master/slacktest/handlers_test.go#L23-L32

mikecutalo (commented):
Have you thought about capturing some metrics? Maybe success and failure when handling events?

backend/module/bot/slackbot/slackbot_test.go
danielhochman: it's bizarre they don't bother to call Close(), but maybe i'm missing something. i would add defer s.Close() to each test unless it causes a panic or something

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,169 @@
...
+	// it's standard behavior of the Slack Events API that the bot will receive all message events, including from it's own posts
+	// so we need to filter out these events by checking that the BotId field is empty; otherwise by reacting to them we will create an infinte loop
+	if event.BotId == "" {
+		match := m.bot.MatchCommand(event.Text)
+		err := sendBotReply(m.slack, event.Channel, match)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
danielhochman, line 168: nit try not to nest the happy path, it's more readable. here's a good [blog post](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) on the technique

```suggestion
	if event.BotId != "" {
		return nil
	}
	
	match := m.bot.MatchCommand(event.Text)
	return sendBotReply(m.slack, event.Channel, match)
```

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,169 @@
...
+
+// event type for messages that mention the bot directly
+func (m *mod) handleAppMentionEvent(event *slackbotv1.Event) error {
+	match := m.bot.MatchCommand(event.Text)
+	err := sendBotReply(m.slack, event.Channel, match)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
danielhochman, line 151: nit simplify
```suggestion
	return sendBotReply(m.slack, event.Channel, match)
```

backend/module/bot/slackbot/slackbot_helper.go
```diff
@@ -0,0 +1,13 @@
+package slackbot
+
+import (
+	"github.com/slack-go/slack"
+)
+
+func sendBotReply(client *slack.Client, channel, message string) error {
+	_, _, err := client.PostMessage(channel, slack.MsgOptionText(message, false))
```
danielhochman, line 8: use the context versions of everything, e.g. `PostMessageContext` and thread it through the functions.

will save you a bunch of refactoring trouble later on.

scarlettperry:
> Have you thought about capturing some metrics? Maybe success and failure when handling events?

yep! meant to add a TODO, thanks for the reminder. In a followup PR collect some metrics

backend/module/bot/slackbot/slackbot.go
scarlettperry: ty for catching, updated

backend/module/bot/slackbot/slackbot.go
scarlettperry: ty! fixed

backend/module/bot/slackbot/slackbot_test.go
scarlettperry: yeah thought so too

added `s.Stop()` which is their helper to close the server

backend/module/bot/slackbot/slackbot_helper.go
scarlettperry: when threading the ctx through the slack client API returned `Post \"https://slack.com/api/chat.postMessage\": context canceled"}`. 

I think this is due to the asyc handling of the event / ack'ing the Events API - we return a response to the Events API before we've finished handling the event and so the original ctx gets cancelled . For a sanity check, I changed the flow to first handle the event and then ack the Events API, and I did not get this error.

`PostMessage` uses context.background by default https://github.com/slack-go/slack/blob/v0.8.2/chat.go#L123-L125

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,169 @@
+package slackbot
+
```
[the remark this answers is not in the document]
scarlettperry: added for the module and service https://deploy-preview-1250--clutch-preview.netlify.app/docs/components

backend/module/bot/slackbot/slackbot_helper.go
danielhochman: yep, you should use a background context and not the request context when firing off asynchronous work in the background

scarlettperry: got it, thanks! updated

backend/module/bot/slackbot/slackbot.go
```diff
@@ -0,0 +1,168 @@
...
+// One request, one event. https://api.slack.com/apis/connections/events-api#the-events-api__receiving-events
+// TODO: (sperry) capture success/failure metrics when handling the event
+func (m *mod) Event(ctx context.Context, req *slackbotv1.EventRequest) (*slackbotv1.EventResponse, error) {
+	err := verifySlackRequest(m.verificationToken, req.Token)
+	if err != nil {
+		m.logger.Error("verify slack request error", log.ErrorField(err))
+		return nil, status.Error(codes.Unauthenticated, err.Error())
+	}
+
+	m.logger.Info("received event from Slack API")
```
danielhochman, line 83: delete or add more info

danielhochman (approved):

backend/module/bot/slackbot/slackbot.go
scarlettperry: realized the placement of this log at this step of the flow isnt useful b/c we're not yet sure what kind of an event it is. moved it

scarlettperry:
a few todos in the PR. I'll create issues for some, maybe we'll get contributors :)

PR #1252, opened 2021-04-06
k8s: update pod proto to extract more status info

rithujohn191:
### Description
- Adding InitContainer details to the pod message
- Adding additional state details to the container state

These new fields will help us figure out the exact state of the pod at any given time.

### Testing Performed
- local
- unit

### DIFF
--- a/api/k8s/v1/k8s.proto
+++ b/api/k8s/v1/k8s.proto
@@ -256,6 +256,32 @@ message Container {
   State state = 3;
   bool ready = 4;
   int32 restart_count = 5;
+
+  // ref: https://pkg.go.dev/k8s.io/api/core/v1#ContainerState
+  oneof state_details {
+    StateWaiting state_waiting = 6;
+    StateRunning state_running = 7;
+    StateTerminated state_terminated = 8;
+  }
+}
+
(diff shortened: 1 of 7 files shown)


api/k8s/v1/k8s.proto
```diff
@@ -256,6 +256,32 @@ message Container {
...
+  string message = 2;
+}
+
+// ref: https://pkg.go.dev/k8s.io/api/core/v1#ContainerStateRunning
+message StateRunning {
+  google.protobuf.Timestamp start_time = 1;
+}
+
+// ref: https://pkg.go.dev/k8s.io/api/core/v1#ContainerStateTerminated
+message StateTerminated {
```
danielhochman, line 280: what about the StartedAt FinishedAt fields?

danielhochman (approved):
lgtm overall

smonero (approved):

api/k8s/v1/k8s.proto
rithujohn191: I left these out for now because:
The FinishedAT field in the terminated state refers to "the Time at which the container last terminated" and StartedAT field refers to "Time at which previous execution of the container started" which seemed not as relevant as the other fields. The most common use case would be to have the start time in the running state.

We can always add it on later if we find a use case for this

PR #1253, opened 2021-04-07
k8s meta: changing ApplyListOptions()

smonero:
We need a way to select things differently than we currently do.

### Description

Add a selector string as a supplement to the labels, in addition to a validation check.

### Testing Performed

local (k8s flows), unit

### GitHub Issue

Fixes #


backend/service/k8s/meta.go
```diff
@@ -15,6 +15,11 @@ func ApplyListOptions(listOpts *k8sapiv1.ListOptions) metav1.ListOptions {
 	opts := metav1.ListOptions{}
 	if len(listOpts.Labels) > 0 {
 		opts.LabelSelector = labels.FormatLabels(listOpts.Labels)
+	} else if len(listOpts.SelectorString) > 0 {
+		// use the selector string as a backup, it must be formatted correctly
+		// Example: "!abc"
+		// Another example: "a=b,!c,d!=e"
+		opts.LabelSelector = listOpts.SelectorString
```
mikecutalo, line 22: Is there an input validator we could utilize from upstream?

backend/service/k8s/meta.go
```diff
@@ -15,6 +15,11 @@ func ApplyListOptions(listOpts *k8sapiv1.ListOptions) metav1.ListOptions {
 	opts := metav1.ListOptions{}
 	if len(listOpts.Labels) > 0 {
 		opts.LabelSelector = labels.FormatLabels(listOpts.Labels)
+	} else if len(listOpts.SelectorString) > 0 {
```
mikecutalo, line 18: comment on behavior, as labels takes presidents or even error handle that both fields cant be set simultaneously, unless it's valid to merge them together.

api/k8s/v1/k8s.proto
```diff
@@ -319,6 +319,10 @@ message Pod {
 
 message ListOptions {
   map<string, string> labels = 9;
+  // This selector string will be used in case the labels is empty
```
danielhochman, line 322: use a oneof?

smonero: it says that map fields are not allowed in oneofs, because internally they are repeated fields and those are not allowed.  It seems to not complain if I wrap it in a struct.  Is that the way to go?

backend/service/k8s/meta.go
smonero: I had a new idea: how about we just concat them, as in the selector string will be a supplement rather than a backup replacement

api/k8s/v1/k8s.proto
danielhochman: ah that's right. meh, either way is fine i guess. i'll defer to the other thread where you're deciding on whether or not to concat them.

backend/service/k8s/meta.go
rithujohn191: since set-based selectors are not supported by all resources and since `listOptions` is currently used by all List calls how do we plan on restricting the usage?

backend/service/k8s/meta.go
smonero: https://github.com/kubernetes/apimachinery/blob/dd12c7a65e7f49c4e00b5bd5eb93de2fcbf7831d/pkg/labels/selector.go#L836 this one looks good

backend/service/k8s/meta.go
smonero: I guess the plan is to test each one, and since the string is supplemental it won't change anything for the current uses.  Also, with a validation check for the errors that will help stop problems.

I couldn't find a listing of resources that support it vs those that don't other than those already listed in the docs.

rithujohn191 (approved):

PR #1262, opened 2021-04-07
k8s: add pod status string to pod proto

rithujohn191:
### Description
This PR adds a status string to the pod definition that represents the overall status of the pod calculated by looking at the statuses of its containers. This summary status string is important as it reflects the true status of the pod unlike the `State` field which is just a high level overview of the pod-lifecycle.

This code has been referenced from the upstream dashboard: https://github.com/kubernetes/dashboard/pull/5660

### Testing Performed
-local
-unit


danielhochman (commented):
looks great! can we put together some more test coverage?

also does this match or differ from the lyftkube output (not sure if lyftkube is purely kubectl for the status column).

rithujohn191:
@danielhochman yea this matches lyftkube output directly (which is mostly kubectl for the status column). Will add in more tests

rithujohn191:
added more unit tests to cover different pod states

smonero (approved):

rithujohn191:
@danielhochman added in extra test cases to cover different pod states

scarlettperry (approved):

PR #1264, opened 2021-04-08
mux: forward X- headers

scarlettperry:
### Description
Background:
By default, http request headers are passed to gRPC context via the [DefaultHeaderMatcher](https://github.com/grpc-ecosystem/grpc-gateway/blob/519e7f45f48e3e378e3341d076579f3ce52aa717/runtime/mux.go#L63-L74) rule. If the request headers don't meet certain preconditions, they are dropped.

Solution:
PR adds a custom header rule so that `X-` request headers are forwarded (and prefixed with `grpcgateway`). All other headers will be checked by the DefaultHeaderMatcher rule.

Use case for this change: 
For the slackbot work, we need to [verify](https://api.slack.com/authentication/verifying-requests-from-slack#verifying-requests-from-slack-using-signing-secrets__a-recipe-for-security__step-by-step-walk-through-for-validating-a-request) a request came from Slack by certain `X-` request headers sent in the request.

### Testing Performed
Unit test and locally with https://github.com/lyft/clutch/pull/1250

### DIFF
--- a/backend/gateway/mux/mux.go
+++ b/backend/gateway/mux/mux.go
@@ -28,6 +28,10 @@ import (
 	awsservice "github.com/lyft/clutch/backend/service/aws"
 )
 
+const (
+	xHeader = "X-"
+)
+
 var apiPattern = regexp.MustCompile(`^/v\d+/`)
 
 type assetHandler struct {
@@ -180,6 +184,15 @@ func customResponseForwarder(ctx context.Context, w http.ResponseWriter, resp pr
 	return nil
 }
 
+// forward all X- headers and add the grpcgateway prefix
+func customMatcher(key string) (string, bool) {
+	if strings.HasPrefix(key, xHeader) {
+		return runtime.MetadataPrefix + key, true
+	}
+	// the the default header mapping rule
+	return runtime.DefaultHeaderMatcher(key)
+}
+
 func customErrorHandler(ctx context.Context, mux *runtime.ServeMux, m runtime.Marshaler, w http.ResponseWriter, req *http.Request, err error) {
 	//  TODO(maybe): once we have non-browser clients we probably want to avoid the redirect and directly return the error.
 	if s, ok := status.FromError(err); ok && s.Code() == codes.Unauthenticated {
@@ -219,6 +232,7 @@ func New(unaryInterceptors []grpc.UnaryServerInterceptor, assets http.FileSystem
 				UnmarshalOptions: protojson.UnmarshalOptions{},
 			},
 		),
+		runtime.WithIncomingHeaderMatcher(customMatcher),
 	)
 
 	// If there is a configured asset provider, we check to see if the service is configured before proceeding.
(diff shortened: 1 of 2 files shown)


scarlettperry:
RE [comment](https://github.com/lyft/clutch/pull/1250#discussion_r608814743) - want to double check that we should append the grpcgateway prefex to the `-X` headers?

If so, I noticed something interesting locally in which the context metadata had both `grpcgateway-x-forwarded-for` and `x-forwarded-for`. Same situation with `x-forwarded-host`. This is because if those headers are not found ( which is the case since we're appending a prefix to the header), then it gets added by default here https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/context.go#L121-L134.

Might not be an issue but wanted to raise.

scarlettperry:
side note: curious about https://github.com/lyft/clutch/blob/83c47c8adc4411614c0ed588dcecaf12ec8c7f04/backend/middleware/authn/authn.go#L101

`authorization` is part of the [isPermanentHTTPHeader group](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/context.go#L265:6), which [gets the grpcgateway prefix](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L69), so why isn't the lookup `grpcgateway-authorization`? I ask b/c just want to make sure I'm not missing context that would be important to testing this PR.

backend/gateway/mux/mux.go
```diff
@@ -180,6 +184,15 @@ func customResponseForwarder(ctx context.Context, w http.ResponseWriter, resp pr
 	return nil
 }
 
+// forward all X- headers and add the grpcgateway prefix
+func customMatcher(key string) (string, bool) {
+	if strings.HasPrefix(key, xHeader) {
```
danielhochman, line 189: http headers are not technically case sensitive so you need to match `X-` and `x-`.

or you can call the `textproto.CanonicalMIMEHeaderKey` first like the default matcher so they always end up in the same format with `X-`. i think that's the best option.

also add test case for this scenario.

danielhochman:
re: `authorization` handling it's right above the other code you linked
https://github.com/grpc-ecosystem/grpc-gateway/blob/0a9acf6a7e6399acd4da6a2d24880bab0390f233/runtime/context.go#L102-L105

for `X-Forwarded-For` and `X-Forwarded-Host` is the result just that it will get appended twice, once with once without prefix if passed in by caller? alternatively i guess you could exclude those to avoid confusion.

scarlettperry:
> re: `authorization` handling it's right above the other code you linked
> https://github.com/grpc-ecosystem/grpc-gateway/blob/0a9acf6a7e6399acd4da6a2d24880bab0390f233/runtime/context.go#L102-L105
> 
> for `X-Forwarded-For` and `X-Forwarded-Host` is the result just that it will get appended twice, once with once without prefix if passed in by caller? alternatively i guess you could exclude those to avoid confusion.

ah got it, thanks!

yep will get appended 2x, so list will have `prefix-x-forwarded-for`, `x-forwarded-for`, `prefix-x-forwarded-host`, `x-forwarded-host`. Okay will included these 2 from getting the prefix to avoid confusion.

danielhochman (approved):
lgtm! :ship:

PR #1266, opened 2021-04-08
authnz: explicitly list allowed authn methods

danielhochman:
### Description
Be explicit about the authn methods that are always open and move the list to a common palce.

### Testing Performed
CI


backend/middleware/authn/authn.go
```diff
@@ -25,13 +25,6 @@ import (
 
 const Name = "clutch.middleware.authn"
 
-// List of method patterns that should not be blocked by authn.
-// TODO(maybe): convert this to an API annotation or make configurable on the middleware.
-var allowlist = []string{
-	"/clutch.authn.v1.AuthnAPI/*",
```
danielhochman, line 31: in hindsight this was really dangerous.

scarlettperry (approved):

PR #1270, opened 2021-04-08
frontend: add name, value, and onChange props to Radio

misteroh:
### Testing Performed

Locally

### DIFF
--- a/frontend/packages/core/src/Input/stories/radio.stories.tsx
+++ b/frontend/packages/core/src/Input/stories/radio.stories.tsx
@@ -25,3 +25,10 @@ export const Disabled = Template.bind({});
 Disabled.args = {
   disabled: true,
 };
+
+export const NameValueAndOnChange = Template.bind({});
+NameValueAndOnChange.args = {
+  name: "foo",
+  value: "bar",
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/stories/radio.stories.tsx
```diff
@@ -25,3 +25,10 @@ export const Disabled = Template.bind({});
 Disabled.args = {
   disabled: true,
 };
+
+export const NameValueAndOnChange = Template.bind({});
+NameValueAndOnChange.args = {
+  name: "foo",
+  value: "bar",
+  onChange: e => e.target.value,
```
scarlettperry, line 33: `onChange` can be added as an [Action arg](https://storybook.js.org/docs/react/essentials/actions#action-argtype-annotation). example: https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/Input/stories/checkbox.stories.tsx#L10-L12

misteroh: thanks! just updated

scarlettperry (approved):
lgtm!

PR #1272, opened 2021-04-09
topology cache: deduplicate batch inserts

mikecutalo:
### Description

It was possible to batch insert multiple entries of the same item which is not possible with postgres and results in the error below. With this PR I changed the `batchInserts` to a map which removes the possibility for duplicates while also updating the cache with the latest entry assuming that items coming over the channel are received in chronological order which I think its safe to assume for a single resource.

> {"level":"error","ts":1617985716.6457253,"caller":"topology/cache.go:135","msg":"Error setting cache","serviceName":"clutch.service.topology","error":"pq: ON CONFLICT DO UPDATE command cannot affect row a second time","stacktrace":"github.com/lyft/clutch/backend/service/topology.(*client).processTopologyObjectChannel\n\t/code/clutch-private/clutch/backend/service/topology/cache.go:135"}

This also shows a slight improve in areas during benchmarking but i think overall about the same perf, the previous benchmark can be found [here](https://github.com/lyft/clutch/pull/1154#discussion_r605280940).
```
goos: darwin
goarch: amd64
pkg: github.com/lyft/clutch/backend/service/topology
cpu: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz
BenchmarkPrepareBulkCacheInsert-8   	      33	  32537305 ns/op	12094080 B/op	  309784 allocs/op
PASS
ok  	github.com/lyft/clutch/backend/service/topology	1.138s
```

### Testing Performed
Local / Unit.

### DIFF
--- a/backend/service/topology/cache.go
+++ b/backend/service/topology/cache.go
@@ -105,7 +105,7 @@ func (c *client) startTopologyCache(ctx context.Context) {
 // Notably a flush will happen every 30 seconds, ensuring all items make it into cache even if the
 // configured BatchInsertSize is not met.
 func (c *client) processTopologyObjectChannel(ctx context.Context, objs <-chan *topologyv1.UpdateCacheRequest, service string) {
-	var batchInsert []*topologyv1.Resource
+	var batchInsert = make(map[string]*topologyv1.Resource)
 	queueDepth := c.scope.Tagged(map[string]string{
 		"service": service,
 	}).SubScope("cache").Gauge("object_channel.queue_depth")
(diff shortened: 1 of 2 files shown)


backend/service/topology/cache.go
```diff
@@ -131,25 +131,35 @@ func (c *client) processTopologyObjectChannel(ctx context.Context, objs <-chan *
...
 				}
-				batchInsert = batchInsert[:0]
+				batchInsert = make(map[string]*topologyv1.Resource)
 			}
 		}
 	}
 }
 
+func convertBatchInsertToSlice(batch map[string]*topologyv1.Resource) []*topologyv1.Resource {
+	var batchInsert []*topologyv1.Resource
```
danielhochman, line 156: make with preallocation

backend/service/topology/cache.go
```diff
@@ -105,7 +105,7 @@ func (c *client) startTopologyCache(ctx context.Context) {
 // Notably a flush will happen every 30 seconds, ensuring all items make it into cache even if the
 // configured BatchInsertSize is not met.
 func (c *client) processTopologyObjectChannel(ctx context.Context, objs <-chan *topologyv1.UpdateCacheRequest, service string) {
-	var batchInsert []*topologyv1.Resource
+	var batchInsert = make(map[string]*topologyv1.Resource)
```
danielhochman, line 108: add comment on why you're using a map

danielhochman (approved):

PR #1275, opened 2021-04-09
k8s cache: fix cache panic on shutdown

mikecutalo:
### Description

Previously I was closing `s.topologyObjectChan` for every cluster that I was watching which resulted in the error below. Moved the channel close up so its not invoked more than once.

```
panic: close of closed channel

goroutine 221 [running]:
github.com/lyft/clutch/backend/service/k8s.(*svc).startInformers(0xc000070440, 0x46e0ff8, 0xc000070bc0, 0xc00131d1f0, 0x8, 0x4712d48, 0xc001715170, 0x68c61714000)
	/code/clutch-private/clutch/backend/service/k8s/cache.go:94 +0x868
created by github.com/lyft/clutch/backend/service/k8s.(*svc).StartTopologyCaching
	/code/clutch-private/clutch/backend/service/k8s/cache.go:45 +0x1af
```

### Tests
Locally.

```
INFO	topology/topology.go:107	Caught shutdown signal, shutting down topology caching and releasing advisory lock	{"serviceName": "clutch.service.topology"}
INFO	k8s/cache.go:99	Shutting down the kubernetes cache informers	{"serviceName": "clutch.service.k8s"}
INFO	k8s/cache.go:50	Shutting down the kubernetes topo cache channel	{"serviceName": "clutch.service.k8s"}
````

### DIFF
--- a/backend/service/k8s/cache.go
+++ b/backend/service/k8s/cache.go
@@ -45,6 +45,13 @@ func (s *svc) StartTopologyCaching(ctx context.Context, ttl time.Duration) (<-ch
 		go s.startInformers(ctx, name, cs, ttl)
 	}
 
+	go func() {
+		<-ctx.Done()
+		s.log.Info("Closing the kubernetes topologyObjectChan")
+		close(s.topologyObjectChan)
+		s.topologyInformerLock.Release(topologyInformerLockId)
+	}()
+
 	return s.topologyObjectChan, nil
 }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mikecutalo** [comment]
📜

**Comment by danielhochman** [comment]
the channel is closed for each cluster, since `startInformers` is called once for each cluster. the closes and the lock release need to move to the outer scope in `StartTopologyCaching`

File: backend/service/k8s/cache.go [hunk 1]
```diff
@@ -45,6 +45,13 @@ func (s *svc) StartTopologyCaching(ctx context.Context, ttl time.Duration) (<-ch
 		go s.startInformers(ctx, name, cs, ttl)
 	}
 
+	go func() {
+		<-ctx.Done()
+		s.log.Info("Closing the kubernetes topologyObjectChan")
+		close(s.topologyObjectChan)
```
danielhochman, line 51: this will race with the inner stop channel which could lead to write on a closed channel

mikecutalo: I see, I moved the stop channel up to `StartTopologyCaching` so we can control the shutdowns / channel close all from there.

danielhochman (approved):

PR #1282, opened 2021-04-12
slackbot: Add bot service

scarlettperry:
### Description
PR adds the bot service for the slackbot feature. It's barebones and meant to service the module for testing the e2e flow of receiving a request from Slack and sending the bot's reply to the channel.

The service will be built upon in a following PR, which will add the logic of mapping a slack command to a Clutch API call.

### Testing Performed
Unit tests. Full flow tested here https://github.com/lyft/clutch/pull/1250

### GitHub Issue


backend/service/bot/bot.go
```diff
@@ -0,0 +1,59 @@
...
+
+type svc struct {
+	logger *zap.Logger
+	scope  tally.Scope
+}
+
+type Service interface {
+	// Functions that perform provider-specific sanitization/formatting
+	// They all pass the command to the generic MatchCommand function
+	MatchSlackCommand(command string) string
```
danielhochman, line 32: this is redundant on the interface, no need for a MatchSlack, MatchDiscord, etc. the service can be configured to sanitize appropriately in the call to MatchCommand.

backend/service/bot/bot.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type svc struct {
+	logger      *zap.Logger
+	scope       tally.Scope
+	botProvider botv1.Bot
+}
+
+type Service interface {
+	MatchCommand(command string) string
```
danielhochman, line 53: any ideas for a better name? it sounds like you give it a command and it returns a command. but actually it returns the output of the command

scarlettperry: hmm, `getBotReply`/`getBotResponse` which would match what I have drafted in the module when we get back the match
```
match := m.bot.MatchSlackCommand(event.Text)
err := sendBotReply(m.slack, event.Channel, match)
```

danielhochman: i guess it's fine as-is, maybe update the docstring to be more descriptive and give the return variable here a name

danielhochman (approved):

PR #1283, opened 2021-04-12
frontend: support table row actions

dschaller:
### Description

Add support for table row actions

tableActions

### Testing Performed

manual and storybook


rithujohn191:
Looks great! But I wonder if we should align the dropdown menu to the bottom-left of the button, similar to the screenshot below
Screen Shot 2021-04-12 at 2 27 58 PM

dschaller:
👍  updated

rithujohn191 (approved):

PR #1290, opened 2021-04-13
frontend: skip library checks during TS builds

dschaller:
### Description

Skip library type checks during builds. This is a huge speed gain without any immediate type safety issues as this will still check library types being used within Clutch.

### Testing Performed

CI


.github/workflows/frontend.yml
```diff
@@ -30,6 +30,18 @@ jobs:
...
+          steps.node-modules-cache.outputs.cache-hit == 'true' &&
+          github.ref != 'refs/heads/main'
+        id: build-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            **/dist
+            **/tsconfig.tsbuildinfo
+            !**/node_modules
+          key: ${{ runner.os }}-build-${{ hashFiles('**/yarn.lock') }}
```
danielhochman, line 44: use a more specific key, i would change the `modules` one above too

.github/workflows/frontend.yml
```diff
@@ -30,6 +30,18 @@ jobs:
...
+        if: |
+          steps.node-modules-cache.outputs.cache-hit == 'true' &&
+          github.ref != 'refs/heads/main'
+        id: build-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            **/dist
+            **/tsconfig.tsbuildinfo
+            !**/node_modules
```
danielhochman, line 43: just curious is this actually necessary? are there `node_modules` in dist? can't hurt either way i guess

danielhochman (approved):

.github/workflows/frontend.yml
dschaller: This is to avoid caching dist directories under the node_modules directory

.github/workflows/frontend.yml
dschaller: Updated on https://github.com/lyft/clutch/pull/1288/commits/e44bbc7c550eb8dd9224b129c3550d342ec220b1

PR #1302, opened 2021-04-13
authn: read tokens through the provider for refresh

danielhochman:
### Description
Third-party token reads can now take place through the provider so they can be refreshed if needed.

### Testing Performed
TODO

### TODOs

- [x] Tests
- [x] Try to figure out some cleaner refactoring of the provider impl

### DIFF
--- a/backend/service/authn/authn.go
+++ b/backend/service/authn/authn.go
@@ -69,7 +69,8 @@ type Issuer interface {
 	CreateToken(ctx context.Context, subject string, tokenType authnmodulev1.CreateTokenRequest_TokenType, expiry *time.Duration) (token *oauth2.Token, err error)
 }
 
-type IssuerProvider interface {
+type IssuerProviderReader interface {
 	Issuer
 	Provider
+	TokenReader // Read calls are proxied through the IssuerProvider so the token can be refreshed if needed.
 }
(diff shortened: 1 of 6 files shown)


backend/service/authn/authn.go
```diff
@@ -69,7 +69,21 @@ type Issuer interface {
...
+	Read(ctx context.Context, userID, provider string) (*oauth2.Token, error)
+}
+
+type TokenStorer interface {
+	Store(ctx context.Context, userID, provider string, token *oauth2.Token) error
+}
+
+type Storage interface {
+	TokenReader
+	TokenStorer
```
rithujohn191, line 88: Curious why you didn't leave the Storage interface as is i.e 
```
type Storage interface {
	Store(ctx context.Context, userID, provider string, token *oauth2.Token) error
	Read(ctx context.Context, userID, provider string) (*oauth2.Token, error)
}
```
And have Storage as a field in the Service interface?
Do we see `TokenReader` & `TokenStorer` being used separately else where?

rithujohn191: We implemented a similar storage interface for tokens in Dex: https://github.com/dexidp/dex/blob/master/storage/storage.go#L69

A lot more elaborate and we might not need all the methods but might be helpful for reference

danielhochman: i split it because the service doesn't need to proxy writes, only reads. writes only happen internally within the authn service but reads can occur from other components (e.g. K8s reading the provider token).

rithujohn191 (approved):

PR #1306, opened 2021-04-14
frontend: navigate to origin within wizard if present

dschaller:
### GitHub Issue

Fixes #


frontend/packages/wizard/src/wizard.tsx
```diff
@@ -93,6 +102,7 @@ const Wizard = ({ heading, width = "default", dataLayout, children }: WizardProp
         setGlobalWarnings([]);
         setSearchParams({});
         dispatch(WizardAction.BACK);
+        navigateToOrigin();
```
dschaller, line 105: TODO: Update this to ensure back only navigates if on the second step

PR #1309, opened 2021-04-15
meta: update pattern value extraction to allow empty string

rithujohn191:
### Description
The `ExtractPatternValuesFromString` method does not allow fields to have empty string as a valid value. This PR updates the method to allow pattern values to be an empty string .

This enables us to input a search query with one of the fields as an empty string e.g now for the pod delete flow a user can input `<clientset>//<pod-id>` as a valid query to search for a podID in all namespaces in a particular clientset
### Testing Performed
local
ezgif com-gif-maker (27)

### GitHub Issue

Fixes #

### DIFF
--- a/backend/gateway/meta/meta.go
+++ b/backend/gateway/meta/meta.go
@@ -198,13 +198,6 @@ func ExtractPatternValuesFromString(pb proto.Message, value string) (map[string]
 		return result, false, nil
 	}
 
-	// Check that all of the fields have values
-	for _, value := range result {
-		if len(value) == 0 {
-			return result, false, nil
-		}
-	}
-
 	return result, true, nil
 }
 
--- a/backend/gateway/meta/meta_test.go
+++ b/backend/gateway/meta/meta_test.go
@@ -220,6 +220,16 @@ func TestExtractProtoPatternsValues(t *testing.T) {
 			expect:      "",
 			shouldError: true,
 		},
+		{
+			id: "pod",
+			pb: &k8sapiv1.Pod{
+				Cluster:   "foo",
+				Namespace: "",
+				Name:      "cat",
+			},
+			expect:      "foo//cat",
+			shouldError: false,
+		},
 	}
 
 	for _, tt := range tests {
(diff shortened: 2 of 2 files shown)


mikecutalo:
I think this should be ok, but going to think more about the implications for all and future resolvers before approving.

danielhochman:
i would prefer we try to use some sentinel value going forward instead of an empty string, i.e. `*`

mikecutalo:
Thought more about this, here they are in no particular order.

Pattern delimiters are set by the implementer and are completely custom, I could think of a few scenarios where this change could make a false positive match.

Given the example:
```
pattern: "{field1}-{field2}-{field3}"
```
Given the input of `field1--field2`, this would still be a valid match in this use case. However it might be the literal input the user is looking for,  the promise of the resolver is that we should be able to locate a resource from a verity of different inputs so being explicit here I think is important.

Obviously this was a very crude example but the expressiveness of this feature requires us to think carefully about how we want all resolution to work going forward.

I like @danielhochman suggest of the sentinel value, one thing to keep in mind is that an implementer needs to handle this case and makes sure to fanout.

rithujohn191:
@mikecutalo for the edge case that you pointed out even if we use a sentinel value like `*`, we will be faced with the same problem cause if the user specifies `field1-*-field2` and they want that exact match for the jobID we would not be able to tell the difference. and this problem already exists today with the implementation

I could change the code to use the sentinel value and if the user specifies it, we fallback to the default values of the field i.e https://github.com/lyft/clutch/blob/main/backend/resolver/k8s/k8s.go#L181. The way I see it, introducing this has parity with the code we have today and adds functionality to enable users to specify fanout in namespaces if they want

mikecutalo:
This is true, and in that case I'm having second thoughts about the sentinel value at all.

But that only takes into consideration the k8s resolvers and only for namespace, which happens to be easy to fanout as its supported by the kube api for namespaces. What about AWS? a `*/instanceid` would have to be evaluated also or even for k8s `*/namespace/resource` will still have to implement fanout of clientsets in this case which we have code but will have to be handled for this use case.

Or what about cases were fanout is not possible say searching for an instance id `us-east-1/*`?

The more I think about it the more edge cases arise.

rithujohn191:
Then are we okay with taking a stand that we wont support the fanout edge case at all? i.e the user can either provide in the full pattern i.e <cluster>/<namespace>/<podID> or just the podID and nothing else is valid?

mikecutalo:
I dont think we should support wild card fanout at all, we were talking as a team today in standup and we can meet again next week to review it again. Also revisiting how we got here and why we are thinking about this change just to get the previous context.

Lets revisit next week in standup.

rithujohn191:
closing out this PR in favor of a long term solution to support multiple patterns per proto

PR #1311, opened 2021-04-15
k8s proto: disable audit for listPods

smonero:
k8s proto: disable audit for listPods

### Description

use the disable_audit annotation to disable auditing for list pods

### GitHub Issue

Fixes #


danielhochman:
i don't think we need to completely disable audit for these endpoints. rather we should prevent them from being logged by the slack sink.

https://github.com/lyft/clutch-private/blob/d14929231a1eeda758ef348001e247b40dc81454/backend/clutch-config.yaml#L308-L320

smonero:
got it, I'll close this and open a new one

PR #1312, opened 2021-04-15
k8s cache: fix panic during relist

mikecutalo:
### Description

Previously if a failure to look up a resource occurred, the error would be caught but the corresponding code would still execute. If we catch an error dont execute the corresponding code.

Also I added a bit more information to the error log so we know what cluster its coming from.

```
ERROR: Unable to list deployments to populate Kubernetes cache
...
panic: interface conversion: runtime.Object is nil, not *v1.DeploymentList

goroutine 304 [running]:
github.com/lyft/clutch/backend/service/k8s.(*svc).cacheFullRelist(0xc0000717c0, 0x4737b18, 0xc000071b40, 0xc00164ccf0, 0x13, 0xc0013204e0, 0xc001320558, 0xc0013205d0, 0x68c61714000)
	/code/clutch-private/clutch/backend/service/k8s/cache.go:138 +0x10bb
created by github.com/lyft/clutch/backend/service/k8s.(*svc).startInformers
	/code/clutch-private/clutch/backend/service/k8s/cache.go:100 +0x7b3
```

### DIFF
--- a/backend/service/k8s/cache.go
+++ b/backend/service/k8s/cache.go
@@ -118,45 +118,46 @@ func (s *svc) cacheFullRelist(ctx context.Context, cluster string, lwPods, lwDep
 		case <-ticker.C:
 			pods, err := lwPods.List(metav1.ListOptions{})
 			if err != nil {
-				s.log.Error("Unable to list pods to populate Kubernetes cache", zap.Error(err))
-			}
-
-			podItems := pods.(*corev1.PodList).Items
-			for i := range podItems {
-				pod := &podItems[i]
-				if err := ApplyClusterMetadata(cluster, pod); err != nil {
-					s.log.Error("Unable to apply cluster name to pod object", zap.Error(err))
+				s.log.Error("Unable to list pods to populate Kubernetes cache", zap.String("cluster", cluster), zap.Error(err))
+			} else {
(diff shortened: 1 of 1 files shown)


backend/service/k8s/cache.go
```diff
@@ -118,45 +118,46 @@ func (s *svc) cacheFullRelist(ctx context.Context, cluster string, lwPods, lwDep
...
-				s.log.Error("Unable to list pods to populate Kubernetes cache", zap.Error(err))
-			}
-
-			podItems := pods.(*corev1.PodList).Items
-			for i := range podItems {
-				pod := &podItems[i]
-				if err := ApplyClusterMetadata(cluster, pod); err != nil {
-					s.log.Error("Unable to apply cluster name to pod object", zap.Error(err))
+				s.log.Error("Unable to list pods to populate Kubernetes cache", zap.String("cluster", cluster), zap.Error(err))
+			} else {
```
danielhochman, line 122: i think you can call `break` instead of nesting everything

mikecutalo: One resource might fail, but the next one could succeeded. The API servers latencies are variable by resource.

danielhochman: ah, i see now :+1:

danielhochman (approved):

PR #1317, opened 2021-04-15
authn: change endpoint to post and update docs

Augustyniak:
### Description

The subject is prefixed with `service:` if `token_type` is equal to `service`.

### Testing Performed
unit


api/authn/v1/authn.proto
```diff
@@ -26,7 +26,7 @@ service AuthnAPI {
 
   rpc CreateToken(CreateTokenRequest) returns (CreateTokenResponse) {
     option (google.api.http) = {
-      get : "/v1/authn/createToken"
+      post : "/v1/authn/createToken"
```
danielhochman, line 29: you need body : "*", look at others

danielhochman (approved):

PR #1319, opened 2021-04-16
core: FE chip migration

smonero:
### Description

Made a chip that will be used by the various workflows

### Testing Performed

Nothing is currently using it.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/chip.tsx
+++ b/frontend/packages/core/src/chip.tsx
@@ -1,0 +1,32 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { ChipProps as MuiChipProps } from "@material-ui/core";
+import { Chip as MuiChip } from "@material-ui/core";
+
+type variants = "error" | "unknown" | "pending" | "running" | "succeeded";
+export interface ChipProps extends Pick<MuiChipProps, "label"> {
+  variant: variants;
+}
+
+const CHIP_COLOR_MAP = {
+  error: "#FF8A80",
+  unknown: "#FFCC80",
+  pending: "#FFF59D",
+  running: "#C2C8F2",
+  succeeded: "#69F0AE",
+};
+
+const StyledChip = styled(MuiChip)({
+  height: "20px",
+  fontSize: "12px",
+  lineHeight: "20px",
+  cursor: "inherit",
+  margin: "10px",
+});
+
+const ClutchChip = ({ variant, ...props }: ChipProps) => {
+  const styleOverride = { background: CHIP_COLOR_MAP[variant] };
+  return <StyledChip {...props } style={styleOverride} />;
+};
+
+export default ClutchChip;


frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,31 @@
...
+  succeeded: "#69F0AE",
+};
+
+const StyledChip = styled(MuiChip)({
+  height: "20px",
+  fontSize: "12px",
+  lineHeight: "20px",
+  cursor: "inherit",
+  margin: "10px",
+});
```
dschaller, line 25:
```suggestion
const StyledChip = styled(MuiChip)(
  {
    height: "20px",
    fontSize: "12px",
    lineHeight: "20px",
    cursor: "inherit",
    margin: "10px",
  },
  props => ({
    background: CHIP_COLOR_MAP["data-variant'],
  })
);
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const ClutchChip = ({ variant, ...props }: ChipProps) => {
+  return <StyledChip {...props} style={{ background: CHIP_COLOR_MAP[variant] }} />;
```
dschaller, line 28:
```suggestion
  return <StyledChip {...props} data-variant={variant} />;
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
```
dschaller, line 29:
```suggestion
const ClutchChip = ({ variant, ...props }: ChipProps) => (
  <StyledChip {...props} style={{ background: CHIP_COLOR_MAP[variant] }} />;
);
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const StyledChip = styled(MuiChip)({
+  height: "20px",
+  fontSize: "12px",
+  lineHeight: "20px",
+  cursor: "inherit",
+  margin: "10px",
+});
+
+const ClutchChip = ({ variant, ...props }: ChipProps) => {
```
dschaller, line 27:
```suggestion
const Chip = ({ variant, ...props }: ChipProps) => {
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { ChipProps as MuiChipProps } from "@material-ui/core";
+import { Chip as MuiChip } from "@material-ui/core";
+
+type variants = "error" | "unknown" | "pending" | "running" | "succeeded";
+export interface ChipProps extends Pick<MuiChipProps, "label"> {
+  variant: variants;
```
dschaller, line 8: nit: you can inline the variants here since they aren't being used anywhere else.

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { ChipProps as MuiChipProps } from "@material-ui/core";
+import { Chip as MuiChip } from "@material-ui/core";
+
+type variants = "error" | "unknown" | "pending" | "running" | "succeeded";
```
dschaller, line 6: What is the different between running and succeeded? Adding a docstring might help here

dschaller (commented):
Can you add a story for this as well?

frontend/packages/core/src/chip.tsx
smonero: ill change it to be more like the windshield version of WORST...BEST for now

smonero:
putting on hold until we get design feedback on colors of chips

frontend/packages/core/src/chip.tsx
smonero: Btw when I remove the return storybook doesn't compile.  It says "'Chip' cannot be used as a JSX component.
  Its return type 'void' is not a valid JSX element"

frontend/packages/core/src/chip.tsx
smonero: When I tried doing it like this in storybook it doesn't pass the data-variant.  As in, it will always be the default color rather than the color we want.

frontend/packages/core/src/stories/chip.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from "react";
+import type { Meta } from "@storybook/react";
+
+import type { ChipProps } from "../chip";
+import Chip from "../chip";
+
+export default {
+  title: "Core/Chip",
+  component: Chip,
```
dschaller, line 9: Storybook is going to see the variant as a string because of the typing system but if we add this it'll use a select:
```suggestion
  component: Chip,
  argTypes: {
    variant: {
      control: {
        type: 'select',
        options: [{options...}]
      }
    }
  }
```

If you want to reuse the options you can do something like (noted as suggestions as STORYBOOK TYPES):
```suggestion
  component: Chip,
  argTypes: {
    variant: {
      control: {
        type: 'select',
        options: CHIP_VARIANTS
      }
    }
  }
```

frontend/packages/core/src/stories/chip.stories.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from "react";
+import type { Meta } from "@storybook/react";
+
+import type { ChipProps } from "../chip";
+import Chip from "../chip";
```
dschaller, line 5: STORYBOOK TYPES
```suggestion
import { Chip, CHIP_VARIANTS } from "../chip";
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { ChipProps as MuiChipProps } from "@material-ui/core";
+import { Chip as MuiChip } from "@material-ui/core";
+
+export interface ChipProps extends Pick<MuiChipProps, "label"> {
+  variant: "WORST" | "BAD" | "NEUTRALBAD" | "NEUTRAL" | "NEUTRALGOOD" | "GOOD" | "BEST";
```
dschaller, line 7: STORYBOOK TYPES
```suggestion
const CHIP_VARIANTS = ["WORST", "BAD", "NEUTRALBAD", "NEUTRAL", "NEUTRALGOOD", "GOOD", "BEST"] as const;

export interface ChipProps extends Pick<MuiChipProps, "label"> {
  variant: typeof CHIP_VARIANTS[number];
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,50 @@
...
+    background: CHIP_BACKGROUND_COLOR_MAP[props["data-variant"]],
+    color: CHIP_LABEL_COLOR_MAP[props["data-variant"]],
+  })
+);
+
+const Chip = ({ variant, ...props }: ChipProps) => {
+  return <StyledChip {...props} data-variant={variant} />;
+};
+
+export default Chip;
```
dschaller, line 50: STORYBOOK TYPES
```suggestion
export { Chip, CHIP_VARIANTS };
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  NEUTRALBAD: "#0D1030", // greyish
+  NEUTRAL: "#0D1030", // whiteish
+  NEUTRALGOOD: "#3548D4", // purplish
+  GOOD: "#B09027", // yellow
+  BEST: "#40A05A", // green
+};
+
+const StyledChip = styled(MuiChip)(
+  {
+    height: "20px",
```
dschaller, line 32: I think in figma this is `32px`

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,50 @@
...
+    fontSize: "12px",
+    lineHeight: "20px",
+    cursor: "inherit",
+    margin: "10px",
```
dschaller, line 36: I don't think we should have default margin for these

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  NEUTRALGOOD: "#3548D4", // purplish
+  GOOD: "#B09027", // yellow
+  BEST: "#40A05A", // green
+};
+
+const StyledChip = styled(MuiChip)(
+  {
+    height: "20px",
+    fontSize: "12px",
+    lineHeight: "20px",
```
dschaller, line 34: I think you want this css on the label as well as padding
```suggestion
    ".MuiChip-label": {
      fontSize: "12px",
      lineHeight: "20px",
      padding: "6px 12px",
    }
```

frontend/packages/core/src/chip.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Chip as MuiChip } from "@material-ui/core";
+
+export interface ChipProps extends Pick<MuiChipProps, "label"> {
+  variant: "WORST" | "BAD" | "NEUTRALBAD" | "NEUTRAL" | "NEUTRALGOOD" | "GOOD" | "BEST";
+}
+
+const CHIP_BACKGROUND_COLOR_MAP = {
+  WORST: "#F9EAE7", // red
+  BAD: "#FEF8E8", // orange
+  NEUTRALBAD: "#E2E2E6", // greyish
```
dschaller, line 13: I think we will need to set the border independently of the background. Right now this makes the border more  black than the designs have it
Storybook:
Screen Shot 2021-04-21 at 1 20 07 PM
Figma:
Screen Shot 2021-04-21 at 1 20 13 PM

frontend/packages/core/src/stories/chip.stories.tsx
smonero: it only seems to work if I do `as unknown as Meta` .  Not sure if thats the right way to go about things.  Also, seem to have a new issue with the < Chip {...props} part
err

frontend/packages/core/src/chip.tsx
smonero: Ah, I see there is an alpha difference.

PR #1324, opened 2021-04-17
experimentation: pass experiment config as proto.Message

Augustyniak:
### Description

1. Pass experiment configuration as proto message to limit the number of users of `ExperimentConfig` type that need to unmarshal its `Config` property to a specific type. This new property of `ExperimentConfig` is going to be used by the improved - pluggable - implementation of xds cache.
2. Make storer return all experiments that can be successfully retrieved even if some of them are malformed. Previously, an error was returned in this specific case.

### Testing Performed
Unit Tests


backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -180,25 +182,35 @@ func (s *storer) GetExperiments(ctx context.Context, configType string, status e
...
 		if err != nil {
 			return nil, err
 		}
 
-		err = protojson.Unmarshal([]byte(details), config.Config)
+		config, err := NewExperimentConfig(configId, details)
 		if err != nil {
-			return nil, err
+			s.logger.Errorw("failed to initialize experiment config", "configId", configId, "runId", run.Id)
+			s.configInitializationErrorCount.Inc(1)
```
snowp, line 205: initialize sounds a bit strange since we're reading a value from the database, maybe deserialization ?

snowp (changes requested):

backend/service/chaos/experimentation/experimentstore/storer.go
Augustyniak: fixed

backend/service/chaos/experimentation/experimentstore/storer.go
```diff
@@ -269,7 +269,7 @@ func (s *storer) GetListView(ctx context.Context) ([]*experimentation.ListViewIt
 
 		if err != nil {
 			s.logger.Errorw("failed to initialize experiment config", "configId", configId, "runId", run.Id)
```
snowp, line 271: comment still uses init language

snowp (approved):

PR #1344, opened 2021-04-20
tools: update workflow frontend scaffolding

dschaller:
### Description

Update the frontend workflow scaffolding to use esbuild and enforce a min node version of 14

### Testing Performed

manual


mikecutalo (approved):
When we gonna migrate everything else?

PR #1345, opened 2021-04-20
frontend: fix auth redirect for axios client

tilakraj9560:
### Description
Previously, we used to handle unauthenticated users from the middleware and redirect them to the `/v1/authn/login` url which in turn returns 200. This request was handled by the axios client success interceptor and error was thrown to handle cors policy. Since now we have removed that frontend redirection from the UI. we don't have redirection when apis are called through axios clients.

```
const axiosClient = axios.create({
    // n.b. the client will treat any response code >= 400 as an error and apply the error interceptor.
    validateStatus: status => {
      return status < 400;
    },
  });
  axiosClient.interceptors.response.use(resp => resp, errorInterceptor);

```

Reference: https://github.com/lyft/clutch/pull/325

#### Tested manually.

### DIFF
--- a/frontend/packages/core/src/Network/index.ts
+++ b/frontend/packages/core/src/Network/index.ts
@@ -1,4 +1,4 @@
-import type { AxiosError } from "axios";
+import type { AxiosError, AxiosResponse } from "axios";
 import axios from "axios";
 
 import type { ClutchError } from "./errors";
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Network/index.ts
```diff
@@ -59,15 +79,13 @@ const errorInterceptor = (error: AxiosError): Promise<ClutchError> => {
...
-  });
-  axiosClient.interceptors.response.use(resp => resp, errorInterceptor);
+  const instance = axios.create();
+  instance.interceptors.response.use(
+    response => successInterceptor(response),
+    error => errorInterceptor(error)
+  );
 
-  return axiosClient;
+  return instance;
```
dschaller, line 88: I think you want `createClient` to be:
```jsx
const createClient = () => {
  const axiosClient = axios.create({
    // n.b. the client will treat any response code >= 400 as an error and apply the error interceptor.
    validateStatus: status => {
      return status < 400;
    },
  });
  axiosClient.interceptors.response.use(successInterceptor, errorInterceptor);

  return axiosClient;
};
```

dschaller:
Thank you! Once you update the interceptors from my comment the tests should pass.

dschaller (approved):

PR #1346, opened 2021-04-20
frontend: remove styled components

dschaller:
### Description

Remove styled-components dependency now that we are fully using emotion.

### Testing Performed

CI


scarlettperry (approved):

mikecutalo (changes requested):
fix

PR #1351, opened 2021-04-22
experimentation: introduce pluggable resource generators

Augustyniak:
### Description

This PR is going to be merged intro a feature branch - two more PRs are going to be merged into this feature branch before the whole branch is merged into main.

The PR introduces a pluggable core for registering ECDS and RTDS resource generators so that other system components such as redis or serverexperimentation can provide their own implementation of resource generators to orchestrate chaos experiments with the use of RTDS/ECDS.

Please note that although integration tests compile and run they do not pass in this PR - this is going to be fixed by one of the following PRs.

### Testing Performed
Unit tests

PR #1355, opened 2021-04-23
serverexperimentation: implement ECDS and RTDS generators

Augustyniak:
### Description

This PR builds on top of https://github.com/lyft/clutch/pull/1351/files - which introduces an option for plugin generators to be plugged in into `xds` module - and introduces two first generators (implemented as plugins) for the generation of server faults with the use of RTDS and ECDS.

This PR is currently opened against `generalize-runtime-keys-generation-core` branch. After https://github.com/lyft/clutch/pull/1351/files is merged into a feature branch (`generalize-runtime-keys-generation-feature-branch`) current PR is going to be changed so that its base is `generalize-runtime-keys-generation-feature-branch` too.

### Testing Performed
Unit and integration tests.


Makefile
```diff
@@ -81,7 +81,8 @@ yarn-install: yarn-ensure
 
 .PHONY: backend-integration-test
 backend-integration-test:
-	cd backend/internal/test/integration/xds && docker-compose up --build --abort-on-container-exit
+	docker-compose -f backend/internal/test/integration/serverexperimentation/docker-compose.yaml up --build --abort-on-container-exit
+	docker-compose -f backend/internal/test/integration/xds/docker-compose.yaml up --build --abort-on-container-exit
```
Augustyniak, line 85: maybe there is some clever way to run second docker-compose only if the first one succeeds (all tests pass)?

backend/internal/test/integration/serverexperimentation/cmd/envoyconfiggen/main.go
```diff
@@ -0,0 +1,112 @@
+package main
```
Augustyniak, line 1: a lot of copy-pasting for integrations tests - not a perfect solution but I think it should be just fine for our use case for now.

backend/module/chaos/serverexperimentation/server_experimentation_integration_test.go
```diff
@@ -0,0 +1,233 @@
...
+	xdsconfigv1 "github.com/lyft/clutch/backend/api/config/module/chaos/experimentation/xds/v1"
+	serverexperimentationconfigv1 "github.com/lyft/clutch/backend/api/config/module/chaos/serverexperimentation/v1"
+	"github.com/lyft/clutch/backend/internal/test/integration/helper/envoytest"
+	xdstest "github.com/lyft/clutch/backend/internal/test/integration/helper/testserver"
+	"github.com/lyft/clutch/backend/mock/service/chaos/experimentation/experimentstoremock"
+)
+
+// These tests are intended to be run with docker-compose to in order to set up a running Envoy instance
+// to run assertions against.
+func TestEnvoyRTDSFaults(t *testing.T) {
```
Augustyniak, line 30: This tests is 99% copied from old xds_integration_test file

backend/module/chaos/serverexperimentation/server_experimentation_integration_test.go
```diff
@@ -0,0 +1,233 @@
...
+	ts.Stop()
+
+	err = awaitExpectedReturnValueForSimpleCall(t, e, awaitReturnValueParams{
+		timeout:        10 * time.Second,
+		expectedStatus: 503,
+	})
+	assert.NoError(t, err, "did not see faults reverted")
+}
+
+func TestEnvoyECDSFaults(t *testing.T) {
```
Augustyniak, line 71: This tests is 99% copied from old xds_integration_test file

backend/module/chaos/serverexperimentation/server_experimentation_integration_test.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// Helper class for initializing a test RTDS server.
+type TestModuleServer struct {
+	*xdstest.TestServer
+
+	Storer *experimentstoremock.SimpleStorer
+}
+
+func newServerExperimentationTestModuleServer(seConfig *serverexperimentationconfigv1.Config, xdsConfig *xdsconfigv1.Config) (*TestModuleServer, error) {
```
Augustyniak, line 118: this method is a light modification of `NewModuleTest` method that existed previously (and still exists in `xdstest` package)

backend/module/chaos/serverexperimentation/server_experimentation_integration_test.go
```diff
@@ -0,0 +1,233 @@
+// +build integration_only
```
Augustyniak, line 1: These tests has to be moved to `serverexperimentation` file due to some circular dependency issue that occurred for when these tests stayed in`expermentation` (this is due to the fact that `xds` pieces were moved from `serverexperimentation` to `experimentation` package

Makefile
Augustyniak: fixed

backend/internal/test/integration/serverexperimentation/testrunner.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM golang:1.16.0
+
+COPY . /code
+
+RUN cd /code/module/chaos/serverexperimentation && find . -type f -name "*_test.go" ! -name "*_integration_test.go" -exec rm {} \;
```
snowp, line 5: is this necessary when using the build tag?

backend/module/chaos/serverexperimentation/xds/ecds_faults_generator.go
```diff
@@ -172,69 +114,57 @@ func generateECDSResource(experiments []*experimentstore.Experiment, cluster str
...
+
+	return xds.NewECDSResource(enforcingCluster, config)
 }
 
-func generateEmptyECDSResource(cluster string, ecdsConfig *ECDSConfig, logger *zap.SugaredLogger) []gcpTypes.ResourceWithTtl {
-	// cache serialized
-	if SerializedDefaultFaultFilter == nil {
+func (g *ECDSFaultsGenerator) GenerateDefaultResource(cluster string, resourceName string) (*xds.ECDSResource, error) {
+	if resourceName != faultFilterConfigNameForIngressFault && !strings.HasPrefix(resourceName, fmt.Sprintf(faultFilterConfigNameForEgressFault, "")) {
+		return xds.NewEmptyECDSResource(), nil
```
snowp, line 135: i wonder if it would be nicer if we could just return nil and the generator would do the right thing

backend/module/chaos/serverexperimentation/xds/ecds_faults_generator.go
```diff
@@ -172,69 +114,57 @@ func generateECDSResource(experiments []*experimentstore.Experiment, cluster str
...
+	}
+
+	if g.SerializedDefaultFaultFilter == nil {
```
snowp, line 138: should this be done in some initializer instead? seems odd to have it happen inline here the first time we call this

backend/module/chaos/serverexperimentation/xds/ecds_faults_generator.go
```diff
@@ -58,90 +53,38 @@ var DefaultDelayFaultConfig = &gcpFilterCommon.FaultDelay{
 	},
 }
 
-type SafeEcdsResourceMap struct {
-	mu                    sync.Mutex
-	requestedResourcesMap map[string]map[string]struct{}
+type ECDSFaultsGenerator struct {
+	SerializedDefaultFaultFilter []byte
```
snowp, line 57: no need to export this

snowp (changes requested):

backend/module/chaos/serverexperimentation/xds/ecds_faults_generator.go
Augustyniak: Does returning nil, nil feels idiomatic in Go? Feels error prone (accessing nils -> crashing)

snowp: Pretty common: https://github.com/search?q=%22return+nil%2C+nil%22&type=code

As long as the contract specifies that nil is permitted it seems fine to me. The type signature still allows nil, so without nil handling a person trying to use this might run into nil defererences anyways

Augustyniak: most of these are for cases where a method returns `nil, nil, err` or forks of eth crypto client. Anyway, I can make a change - I think it's a reasonable ask.

backend/module/chaos/serverexperimentation/xds/ecds_faults_generator.go
Augustyniak: good point

snowp (approved):

backend/internal/test/integration/serverexperimentation/testrunner.Dockerfile
Augustyniak: seems to work without them 🤷  removed

PR #1373, opened 2021-04-27
ci: run integration tests in parallel to unit tests

Augustyniak:
### Description
Run integration tests in parallel to unit tests (as a separate CI job)

### Testing Performed
N/A

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           path: |
             ~/go/pkg/mod
-          key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}  
+          key: ${{ runner.os }}-go-mod-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-mod-
       - name: go build cache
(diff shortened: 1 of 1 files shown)


.github/workflows/backend.yml
```diff
@@ -63,6 +63,29 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-go-mod-
+      - name: go test cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/go-build
+          key: ${{ runner.os }}-go-test-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-test-
```
danielhochman, line 88: you can remove this caching since the builds all occur in the container and not where the cache is.

this could also poison the cache with empty items since if this job finishes first and populates the cache there will be no items (since it's all in the container).

.github/workflows/backend.yml
```diff
@@ -63,6 +63,29 @@ jobs:
         env:
           GOFLAGS: "-count=1" # idiomatic way to disable test caching so they always run
       - run: make backend-config-validation
+  integration-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.0
```
danielhochman, line 72: setup-go is also not needed actually (in addition to removing the cache)

.github/workflows/backend.yml
Augustyniak: Oh, - I see, that makes sense (if I understand correctly). I removed test cache and kep mod cache. Let me know if I should get rid of the other one too (I do not think so).

danielhochman: you can remove mod cache too, it is not used. everything is happening in the docker container which does not share the cache at all.

Augustyniak: got it - removed

danielhochman (approved):

PR #1379, opened 2021-04-27
experimentation: speed up integration testing

snowp:
Cuts down the time to run integration tests in several ways:
1) build the Go binaries on the host system, allowing for reuse of cached artifacts (build, go mod, etc.)
2) use smaller Docker containers. We go from needing to pull golang and envoyproxy/envoy to pulling alpine and envoyproxy/envoy-alpine, which results in substantially shorter download times
3) set the grace period of the envoy container to 0, this issues a KILL immediately instead of waiting for up to 10s for envoy to handle a TERM

### DIFF
--- a/backend/internal/test/integration/xds/do_integration_test.sh
+++ b/backend/internal/test/integration/xds/do_integration_test.sh
@@ -2,6 +2,7 @@ target_dir="$(pwd)"
 
 export GOOS=linux
 export GOARCH=amd64
+export CGO_ENABLED=0
 
 pushd ../../../..
 	pushd internal/test/integration/xds/cmd/envoyconfiggen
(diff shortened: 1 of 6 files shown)


snowp:
@danielhochman mind taking a look at this? Should improve the integration test time

backend/internal/test/integration/xds/do_integration_test.sh
```diff
@@ -0,0 +1,24 @@
...
+# than naively building within the container setup since we get to share the host Go
+# caches.
+export GOOS=linux
+export GOARCH=amd64
+export CGO_ENABLED=0
+
+# Place the output binaries in $target_dir for use by the Dockerfiles.
+pushd ../../../..
+	pushd internal/test/integration/xds/cmd/envoyconfiggen
+	  go build -o $target_dir/envoyconfiggen main.go
```
danielhochman, line 17: nit: put these in a build/ dir at least and ignore that dir?

danielhochman (approved):

PR #1386, opened 2021-04-29
authn: fix redirect logic and prepare for silent refresh

danielhochman:
### Description
- Full path preservation during redirects, including query params (by removing reliance on referrer).
- Browser requests are redirected. XHR requests are not, and it's up the XHR client to reauthenticate the user.
- Because XHR requests are not redirected, the network client can handle them, which makes silent refresh possible once refresh tokens exist on the server.

### Testing Performed
Manual, Unit.

### TODOs
- [x] Tests
- [x] Full integration test in staging


backend/gateway/mux/util.go
```diff
@@ -0,0 +1,38 @@
...
+	return false
+}
+
+// This is a hack, but it's the best way to get the Accept header from the response forwarder without significant plumbing
+// complexity and overhead via middleware. Do not invoke this in the normal request flow.
+func copyRequestHeadersFromResponseWriter(w http.ResponseWriter) http.Header {
+	req := reflect.ValueOf(w).Elem().FieldByName("req").Elem().FieldByName("Header")
+
+	h := make(http.Header, req.Len())
+	iter := req.MapRange()
```
dschaller, line 31: Should we ensure that `req` here is a Map to prevent a panic produced by calling `MapRange`?

backend/gateway/mux/util.go
```diff
@@ -0,0 +1,38 @@
...
+	for iter.Next() {
+		k := iter.Key().String()
+		v := iter.Value().Index(0).String()
```
dschaller, line 34: is there a chance that `Index(0)` could fail here if the value is empty?

backend/gateway/mux/mux.go
```diff
@@ -201,24 +204,14 @@ func customHeaderMatcher(key string) (string, bool) {
...
-				runtime.DefaultHTTPErrorHandler(ctx, mux, m, w, req, err)
-				return
-			}
-			if redirectPath != referer.Path {
-				redirectPath = fmt.Sprintf("%s?redirect_url=%s", redirectPath, referer.Path)
-			}
+	if isBrowser(req.Header) { // Redirect if it's the browser (non-XHR).
+		if s, ok := status.FromError(err); ok && s.Code() == codes.Unauthenticated {
+			redirectPath := fmt.Sprintf("/v1/authn/login?redirect_url=%s", url.QueryEscape(req.RequestURI))
+			http.Redirect(w, req, redirectPath, http.StatusFound)
```
dschaller, line 210: This might not be a concern since the response forwarder might prevent this edge case but just in case, does `Redirect` set a default value for the `Accept` header similar to how it sets a default `Content-Type` that might cause an infinite redirect loop?

backend/gateway/mux/mux.go
```diff
@@ -171,9 +171,9 @@ func customResponseForwarder(ctx context.Context, w http.ResponseWriter, resp pr
 		http.SetCookie(w, cookie)
 	}
 
-	if redirects := md.HeaderMD.Get("Location"); len(redirects) > 0 {
-		w.Header().Set("Location", redirects[0])
-
+	// Redirect if it's the browser (non-XHR).
+	redirects := md.HeaderMD.Get("Location")
+	if len(redirects) > 0 && isBrowser(copyRequestHeadersFromResponseWriter(w)) {
```
dschaller, line 176: super nit: I think this might be easier to read as `requestHeadersFromResponseWriter`

backend/gateway/mux/util.go
```diff
@@ -0,0 +1,38 @@
...
+	"net/http"
+	"reflect"
+	"strings"
+)
+
+// All modern browsers send `text/html` as the first MIME type in the Accept header for navigation requests.
+// This is naive and doesn't do any actual content negotiation, just checks if the client accepts text/html which
+// is good enough for our use case of determining whether the request originates from the SPA's XHR network client or
+// browser navigation.
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values
```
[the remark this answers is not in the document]
danielhochman: :slightly_smiling_face: I said "modern browsers", I think it's enough unless we want to record browser compatibility in a more structured way throughout the project. But none of the frontend is going to work on IE8 anyways, since React doesn't support IE8 without some additional polyfills (not even considering the myriad of other frontend libraries we use).

The stylesheet request shouldn't matter, since the browser won't be able to get the main page to then fetch the stylesheet without going through the redirect and authing first. Worst case in the extreme race condition where the user requests the main page successfully, then the token expires before the browser can followup with the request for the CSS, they would just get unstyled content (obviously broken page) and need a refresh to fix it. We could pass more info about the request here to check if the original request was for CSS, because otherwise there's no way to differentiate, but personally I don't think it's worth it. Actually, that race would effect non-Opera browsers too, and it would have to be handled by client code in order to remediate it since redirect or not, the user won't get any CSS from the request.

backend/gateway/mux/util.go
```diff
@@ -0,0 +1,38 @@
+package mux
+
+import (
+	"net/http"
+	"reflect"
+	"strings"
+)
+
+// All modern browsers send `text/html` as the first MIME type in the Accept header for navigation requests.
```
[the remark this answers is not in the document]
danielhochman: The alternative I considered was setting a header for API requests from the network client, but we could have other clients internally which make JSON requests and they might not set the header, and that could get confusing for them and us if we continue to try to differentiate between the browser in certain scenarios.

backend/gateway/mux/util.go
danielhochman: req (i.e. Header) has to be a map, but the call to `FieldByName("req")` could panic if the internals of the go http server change so I've added a check there.

backend/gateway/mux/util.go
danielhochman: Technically no, headers will always have values (even if it's just `""`) but I added a check anyways.

backend/gateway/mux/mux.go
danielhochman: The `Accept` header is a request header, and `Redirect` is setting response headers (`Location` and `Content-Type`). Not sure I follow exactly, but the `Content-Type` header shouldn't influence the follow up request after redirect headers at all.

frontend/packages/core/src/Network/index.ts
```diff
@@ -52,6 +38,13 @@ const errorInterceptor = (error: AxiosError): Promise<ClutchError> => {
     return Promise.reject(clientError);
   }
 
+  // This section handles authentication redirects.
+  if (response?.status === 401) {
+    // TODO: turn this in to silent refresh once refresh tokens are supported.
+    const redirectUrl = window.location.href.replace(window.location.origin, "");
```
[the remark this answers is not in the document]
danielhochman: changed it.

edit: actually it looks like search doesn't include any anchor `#`, but i guess we don't rely on that anyways in Clutch

backend/gateway/mux/mux.go
```diff
@@ -182,8 +182,11 @@ func customResponseForwarder(ctx context.Context, w http.ResponseWriter, resp pr
 			}
 			code = headerCodeOverride
 		}
+
+		w.Header().Set("Location", redirects[0])
```
mikecutalo, line 186: Can this ever be more than 1?

backend/gateway/mux/mux_test.go
```diff
@@ -122,3 +128,108 @@ func TestCustomHeaderMatcher(t *testing.T) {
 		assert.Equal(t, test.expectedBool, ok)
 	}
 }
+
+func TestCustomErrorHandler(t *testing.T) {
```
mikecutalo, line 132: What about a test case that asserts the query params stay intact?

frontend/packages/core/src/Network/tests/index.test.ts
```diff
@@ -63,6 +24,37 @@ describe("error interceptor", () => {
...
+          search: "?foo=bar",
+        },
+        writable: true,
+      });
+
+      errorInterceptor(axiosError);
+    });
+
+    it("redirects to provided url", () => {
+      expect(window.location.href).toBe("/v1/authn/login?redirect_url=%2Fexample%3Ffoo%3Dbar");
```
mikecutalo, line 54: I think this was the test I was looking for, can probably ignore my previous comment.

backend/gateway/mux/mux.go
danielhochman: The effective value for headers is the first one, e.g. https://github.com/golang/go/blob/cd28ca3a4e130c88dd8f2d198050620b11d22ce4/src/net/textproto/header.go#L38

Additional values are ignored. But in our case no, there should never be any additional values anyways.

backend/gateway/mux/mux_test.go
danielhochman: see :157, it checks that the whatever full uri was passed in is preserved

mikecutalo: Ah I see now, ty.

mikecutalo (approved):

PR #1389, opened 2021-04-29
datalayout: make new data take precedence over old data

Augustyniak:
### Description
In case of a conflict at the time of wizard state merge, make new data override the old data.

### Testing Performed
Manual testing (went through some wizard flows and verified that they work)


danielhochman (approved):

Augustyniak:
@dschaller let me know if you are OK with this change or whether I should revert it. I do not need it per se due to the recent workaround that we (or rather you) came up with in our "offline" chat.

PR #1399, opened 2021-05-03
experimentation: Adding `Chaos` in server experimentation UI

kathan24:
### Description

To be more explicit and avoid any confusion with the name Experimentation, this PR adds Chaos before Experimentation in Server experiments.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/serverexperimentation/src/index.tsx
+++ b/frontend/workflows/serverexperimentation/src/index.tsx
@@ -9,7 +9,7 @@ const register = (): WorkflowConfiguration => {
       contactUrl: "mailto:hello@clutch.sh",
     },
     path: "server-experimentation",
-    group: "Experimentation",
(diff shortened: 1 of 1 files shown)


frontend/workflows/serverexperimentation/src/index.tsx
```diff
@@ -9,7 +9,7 @@ const register = (): WorkflowConfiguration => {
       contactUrl: "mailto:hello@clutch.sh",
     },
     path: "server-experimentation",
-    group: "Experimentation",
+    group: "Chaos Experimentation",
```
Augustyniak, line 12: should we add chaos word to `displayName` and `description` fields a few lines below?

Augustyniak (approved):
one suggestion

kathan24:
discussed offline to keep the change as is

PR #1400, opened 2021-05-05
tools: add vscode settings for proto paths

dschaller:
### Description

This allows proto paths to resolve correctly when opening code from the root directory or via workspaces

### Testing Performed

manual

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
   "protoc": {
       "options": [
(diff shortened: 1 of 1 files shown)


mikecutalo (approved):

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "protoc": {
+      "options": [
+          "--proto_path=api"
```
danielhochman, line 4: there are other paths that should be included from the build directory for validate and the google core protos, and protoc itself is also there (`path` option)

PR #1408, opened 2021-05-07
k8s: add unix timestamp to pod

smonero:
### Description

This is useful since FE serialization of timestamps is not enabled due to protobufjs issues.

### Testing Performed

local, unit

### GitHub Issue

Fixes #


api/k8s/v1/k8s.proto
```diff
@@ -351,6 +351,11 @@ message Pod {
   // the status of its containers.
   // ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/printers/internalversion/printers.go#L761
   string status = 14;
+
+  // This is here as a workound since protobufjs currently has serialization
+  // issues for well-known types like google.protobuf.Timestamp
+  // Unix timestamp (seconds since Jan 01 1970)
+  int64 start_time_stamp = 15;
```
danielhochman, line 358: would recommend we standardize on `_millis` for this sort of thing. if we end up doing it elsewhere and need millisecond precision it will be nice not to have mixed seconds and millisecond epochs throughout the api.

e.g. `start_time_millis`

backend/service/k8s/pods.go
```diff
@@ -195,6 +195,14 @@ func podDescription(k8spod *corev1.Pod, cluster string) *k8sapiv1.Pod {
 		InitContainers: makeContainers(k8spod.Status.InitContainerStatuses),
 		Status:         getPodStatus(k8spod),
 	}
+
+	// This is here as a workaround since we can't use StartTime
+	if k8spod.Status.StartTime != nil {
+		// Note .Unix() gives seconds, not milliseconds
+		pod.StartTimeMillis = k8spod.Status.StartTime.Unix() * 1000
```
danielhochman, line 202:
```suggestion
		pod.StartTimeMillis = k8spod.Status.StartTime.UnixNano() / 1e6
```

danielhochman (approved):

PR #1416, opened 2021-05-11
backend: stop signal captures to topo context on shutdown

dschaller:
### Description

Upon shutdown via one of the captured system calls the topocache would continue to capture signals preventing the main thread from exiting.

### Testing Performed

manual


mikecutalo:
Do you think this is why we often have orphaned processes during local development? How many times does the OS send a signal? I assume it has to be more than once, because if it's not then stoping here wont help as will eat the signal.

dschaller:
I think this is part of it. I couldn't term the backend process if I was running the topo cache locally. The context would be cancelled but all future attempts to terminate would do nothing. I think there are a few other issues too but I'm still digging.

PR #1417, opened 2021-05-11
k8s: 'clientset not found' fix

danielhochman:
### Description

When deleting a pod, or performing other operation in Kubernetes, there is a problem with knowing which clientset to use to access the resource.

The resolver fans out across all clientsets in order to located the pod. Clientsets are not an intrisic property of a pod. A pod can be accessed by multiple clientsets, because a clientset is a combination of a user and a cluster. Therefore there can be multiple clientsets that access the same cluster, i.e. 'readonly-staging' and 'admin-staging'.

So in the delete facet workflow for example, the resolver would fan out, locate the clientset, and send the located pod information. Even if the server were able to tell the client which clientset was used to locate the pod, it still may not be the correct clientset, so sending back the clientset is not really a good fix for this issue, as the user may still run into issues when they go to delete the pod.

Therefore, instead of attaching additional metadata to the result, we instead allow the server to find the correct clientset by looking for one that matches the cluster of the pod.

If users have two or more clientsets matching the cluster they will receive an error, since it is impossible to know which one is correct. Fixes for this may include editing down the K8s config used by Clutch so that it only ever uses one clientset to access a cluster, enhancing the existing clientset to probe permissions in order to determine which clientset to use, or developing a custom clientset that uses some other heuristic to determine which clientset to use.

Originally reported here: https://lyftoss.slack.com/archives/C015UJ6LED9/p1611927757004400

### Testing Performed
Manual (using the make target to seed a local k8s) and unit.

### TODOs
- [x] unit test coverage


danielhochman:
> Currently the delete pod flow expects the pattern : "{cluster}/{namespace}/{name}" in the omni search field and in the advanced search expects clientset, name & namespace. If we are not allowing the user to have more than 1 clientset per cluster, should we also change the advanced search to be cluster instead or clientset?

We could make that change in a followup, sure.

dschaller (approved):

dschaller:
@rithujohn191 I thought the advanced search expects cluster as well

rithujohn191:
@dschaller nope clutch OSS does not take in cluster field in the advanced search:
Screen Shot 2021-05-17 at 9 27 56 AM

PR #1004, opened 2021-02-08
frontend: fix data-layout data preservation after hydration

dschaller:
### Description

This fixes a bug where existing data set on a data layout will be overwritten when a hydrate response returns.

### Testing Performed

manually
### GitHub Issue

Fixes #152

Issue #152, opened 2020-07-30
data-layout: displayValue not updating values


dschaller:
**Description**
The `displayValue` function of a data layout caches the initial data available when it's called. If a subsequent update is called the new values are not respected.

**Expected Behavior**
`displayValue` should display values that are added from a call to udpate.

**Actual Behavior**
`displayValue` does not show the updated values.

**Version**
main

PR #1421, opened 2021-05-12
frontend: datalayout - add overwrite flag option to hydrate

smonero:
### Description

It might be useful to be able to force an overwrite of data when hydrating.

### Testing Performed

still need to add tests to state.test.tsx etc

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/data-layout/src/manager.tsx
+++ b/frontend/packages/data-layout/src/manager.tsx
@@ -31,7 +31,7 @@ const update = (key: string, value: object): Thunk<ManagerLayout, Action> => {
   };
 };
 
-const hydrate = (key: string): Thunk<ManagerLayout, Action> => {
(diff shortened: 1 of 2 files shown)


smonero:
closing this as its not really necessary (comment from ticket)

Augustyniak made it so that newData will take precedent over existing data in the 1 layer merge at: https://github.com/lyft/clutch/pull/1389
If we change it so that data is always overwritten, then in the error case of missing dependencies, the data will be overwritten with blank data (this would not be optimal)
If we made a change like this, we would have to change everyone's usage of `useDataLayout()` to have a `false` option for hydrate, or make that the default.  This is because in many layouts people use the `useDataLayout()` function but don't want their data overwritten (see item 2)
relevant prs:

https://github.com/lyft/clutch/pull/1389
https://github.com/lyft/clutch/pull/1038
https://github.com/lyft/clutch/pull/1004
https://github.com/lyft/clutch/issues/152

PR #1422, opened 2021-05-12
data-layout: add flag to enable or disable hydration on useDataLayout()

smonero:
### Description

It is useful to be able to not hydrate by default when calling useDataLayout(k).  

### Testing Performed

Tested for my use case and it works.  This is important if we have tables where we want to switch views on data and don't want to call hydrate everytime we switch tabs.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/data-layout/src/layout.tsx
+++ b/frontend/packages/data-layout/src/layout.tsx
@@ -32,8 +32,8 @@ interface DataLayout {
  * Use a registered data layout.
  * 
  * If a hydrate function has been specified this and the layout's data has not been set and shouldHydrate is true this will 
- * populate it's data on the first invocation. If the layout has a cache key set to true and also has existing data 
- * hydrate will not be invoked.
+ * populate it's data on the first invocation. If the layout has a cache key set to true and also has existing data OR
+ * shouldHydrate is false, hydrate will not be invoked.
 
  * @param key The name of the layout registered with the manager.
  */
--- a/frontend/packages/data-layout/src/layout.tsx
+++ b/frontend/packages/data-layout/src/layout.tsx
@@ -24,22 +24,21 @@ interface DataLayout {
   displayValue: () => any;
   /** Loading state of the layout. This is true when data is being hydrated. */
   isLoading: boolean;
-  /** Error state of the layout. This will be a message containing the error encountered when trying to hydrate the lyaout. */
+  /** Error state of the layout. This will be a message containing the error encountered when trying to hydrate the layout. */
   error: ClutchError;
 }
 
 /**
  * Use a registered data layout.
  * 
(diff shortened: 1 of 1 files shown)


frontend/packages/data-layout/src/layout.tsx
```diff
@@ -24,23 +24,23 @@ interface DataLayout {
...
  * Use a registered data layout.
  * 
- * If a hydrate function has been specified this and the layout's data has not been set this will populate it's data
- * on the first invocation. If the layout has a cache key set to true and also has existing data hydrate will not be invoked.
+ * If a hydrate function has been specified this and the layout's data has not been set and shouldHydrate is true this will 
+ * populate it's data on the first invocation. If the layout has a cache key set to true and also has existing data OR
+ * shouldHydrate is false, hydrate will not be invoked.
 
  * @param key The name of the layout registered with the manager.
+ * @param shouldHydrate An option to not hydrate if we don't want to.
```
danielhochman, line 39: nit: i'd just call it `hydrate`

frontend/packages/data-layout/src/layout.tsx
```diff
@@ -24,23 +24,23 @@ interface DataLayout {
...
  */
-const useDataLayout = (key: string): DataLayout => {
+const useDataLayout = (key: string, shouldHydrate: boolean = true): DataLayout => {
```
danielhochman, line 41: let's stick with an options object rather than adding an individual parameter in this case. it will read clearer on invocation and give us room to add more options later if needed.

`useDataLayout("foo", false);`
vs
`useDataLayout("foo", {hydrate: false}`

https://stackoverflow.com/questions/12826977/multiple-arguments-vs-options-object

frontend/packages/data-layout/src/layout.tsx
```diff
@@ -24,22 +24,23 @@ interface DataLayout {
...
- * on the first invocation. If the layout has a cache key set to true and also has existing data hydrate will not be invoked.
+ * If a hydrate function has been specified this and the layout's data has not been set and hydrate is true this will 
+ * populate it's data on the first invocation. If the layout has a cache key set to true and also has existing data OR
+ * hydrate is false, hydrate will not be invoked.
 
  * @param key The name of the layout registered with the manager.
+ * @param opts An options object to allow for things like disabling hydration by default
  */
-const useDataLayout = (key: string): DataLayout => {
+const useDataLayout = (key: string, opts?: object): DataLayout => {
```
danielhochman, line 41: instead of using `object` let's define the type, i think it looks like

```jsx
interface UseDataLayoutOptions {
  hydrate?: bool;
}
```

danielhochman (approved):
lgtm! pending type definition

PR #1424, opened 2021-05-13
data-layout: hydrateAll()

smonero:
### Description

Add a hydrateAll() func to the manager that allows one to hydrate on all the keys, rather than having to call hydrate() on each individual one

### Testing Performed

still testing
### GitHub Issue

Fixes #


smonero:
actually this is kind of tricky and not really useful so its going to be closed for now

PR #1436, opened 2021-05-18
frontend: allow for configurable popper id

dschaller:
### Description

Move hard-coded ID out of popper and allow it to be passed in as a prop

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/AppLayout/drawer.tsx
+++ b/frontend/packages/core/src/AppLayout/drawer.tsx
@@ -118,7 +118,7 @@ const Group = ({ heading, open = false, updateOpenGroup, closeGroup, children }:
       >
         <Avatar>{heading.charAt(0)}</Avatar>
         <GroupHeading align="center">{heading}</GroupHeading>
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/AppLayout/drawer.tsx
```diff
@@ -118,7 +118,7 @@ const Group = ({ heading, open = false, updateOpenGroup, closeGroup, children }:
       >
         <Avatar>{heading.charAt(0)}</Avatar>
         <GroupHeading align="center">{heading}</GroupHeading>
-        <Popper open={open} onClickAway={closeGroup} anchorRef={anchorRef}>
+        <Popper open={open} onClickAway={closeGroup} anchorRef={anchorRef} id="workflow-options">
```
scarlettperry, line 121: is this going to be updated too?
```suggestion
        <Popper open={open} onClickAway={closeGroup} anchorRef={anchorRef} id={id}>
```

danielhochman: that's passing the ID in to the component that now allows an ID

danielhochman (approved):

PR #1445, opened 2021-05-20
authn: add refresh method to issuer

danielhochman:
### Description
Add the ability for the Issuer to refresh a token it previously issued. Additional complexity comes from needing to ensure the underlying provider token is also valid and kicking the user back through the auth flow (by throwing an err) if it's not.

### Testing Performed
Manual and unit.

### TODOs
- [x] Unit tests.
- [ ] Reduce duplication.


mikecutalo (approved):

danielhochman:
kick Actions, which didn't run for previous commit

PR #1449, opened 2021-05-21
proxy module: protos

mikecutalo:
### Description

The proxy module will expose an `rpc` capable of proxying requests to different systems, enabling quicker integration with other systems when a full Clutch Services is not needed.

Before talking to an external system, a `service` must be defined in the module's configuration, to ensure only requests to configured systems are made.

Example Configuration 
```
  - name: clutch.module.proxy
    typed_config:
      "@type": type.google.com/clutch.config.module.proxy.v1.Config
      services:
        - name: pagerduty
          host: https://api.pagerduty.com
          allowed_requests:
            - path: /incidents
              method: GET
          headers:
            - token: ${PAGERDUTY_TOKEN}
```


api/passthrough/v1/passthrough.proto
```diff
@@ -0,0 +1,38 @@
...
+import "google/protobuf/struct.proto";
+import "google/api/annotations.proto";
+import "validate/validate.proto";
+
+// The PassthroughAPI enabling calling external API's without minimal setup.
+// The services must be configured before calling external API's.
+service PassthroughAPI {
+  rpc GetPassthrough(GetPassthroughRequest) returns (GetPassthroughResponse) {
+    option (google.api.http) = {
+      post : "/v1/passthrough/get"
```
danielhochman, line 17: `get` is confusing given that it's also and HTTP method. i would call it `rpc` or `request` or something. or is this the HTTP method? since i don't see that in the request.

mikecutalo: Yeah I was debating between having the HTTP method passed in or specific RPC for each, passing in the method would simplify / reduce the numbers of `rpc` definitions. I think ill move the method info to the request..

api/config/module/passthrough/v1/passthrough.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message Config {
+  repeated Service services = 1;
+}
+
+message Service {
+  // An identifier to reference the service
+  string name = 1 [ (validate.rules).string.min_len = 1 ];
+  // The host URL to communicate with
+  string host = 2 [ (validate.rules).string.min_len = 1 ];
```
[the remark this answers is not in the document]
mikecutalo: Here is an example config of what I have today, are you proposing that we move host into to the passthrough request? 

```
  - name: clutch.module.passthrough
    typed_config:
      "@type": type.google.com/clutch.config.module.passthrough.v1.Config
      services:
        - name: pagerduty
          host: https://pagerduty.com
          headers:
            - token: ${PAGERDUTY_TOKEN}
```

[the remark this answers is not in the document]
danielhochman: ah i see, whoops. i was reading this as request and not config.

api/config/module/passthrough/v1/passthrough.proto
```diff
@@ -0,0 +1,24 @@
...
+  // Paths that are allowed to be invoked
+  repeated string allowed_paths = 3 [ (validate.rules).repeated = {min_items : 1} ];
+  // HTTP methods that are allowed to be invoked
+  repeated string allowed_methods = 4 [ (validate.rules).repeated = {min_items : 1} ];
+  // Additional headers can be set such as auth
```
danielhochman, line 22: these should probably be combined into a message so I can allowlist the combination of path/method rather than each separately.

e.g. 
DELETE /alerts/escalation_policies
GET /alerts
POST /alerts

api/passthrough/v1/passthrough.proto
```diff
@@ -0,0 +1,40 @@
...
+  string service = 1 [ (validate.rules).string.min_len = 1 ];
+  // The HTTP method that will be used for the request
+  string http_method = 2 [ (validate.rules).string.min_len = 1 ];
+  // The URI path to call
+  string path = 3 [ (validate.rules).string.min_len = 1 ];
+  // The request body
+  google.protobuf.Value request = 4;
+}
+
+message RequestPassthroughResponse {
```
danielhochman, line 35: headers?

api/passthrough/v1/passthrough.proto
```diff
@@ -0,0 +1,40 @@
...
+  // The http response code from the service response
+  int32 http_status = 1;
+  // The full repsonse body
```
danielhochman, line 38: spelling

api/passthrough/v1/passthrough.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
+
+package clutch.passthrough.v1;
```
danielhochman, line 3: should we call it proxy instead of passthrough? just thinking of debugging down the line.

danielhochman (approved):

api/passthrough/v1/passthrough.proto
mikecutalo: True proxy makes more sense

api/passthrough/v1/passthrough.proto
mikecutalo: ah yes let me add

api/config/module/proxy/v1/proxy.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+
+package clutch.config.module.passthrough.v1;
+
+option go_package = "github.com/lyft/clutch/backend/api/config/module/passthrough/v1;passthroughv1";
```
dschaller, line 5: Should this be proxy instead of passthrough?

mikecutalo: Yup I think we raced on my commit

PR #1459, opened 2021-05-25
frontend: feature flags

dschaller:
### Description

Builds off #55 to add frontend support for simple Feature Flags.

This adds support for workflow feature flags (e.g. enabling/disabling a route) as well as component level feature flags.

Breaking the feature flags down:
- Both types of feature flags are fetched on a configurable interval (defaults to every 5 minutes).
- Feature flags are cached in the session storage to prevent numerous feature flags on the same page/the workflow feature flags from overwhelming the server with duplicate requests. The flags are stored with a timestamp which allows us to enforce a configurable TTL (defaults to 1 minute). This cache also allows us to continue serving feature flags if subsequent calls for feature flags fail.
- Workflow feature flags are exposed via a prop on the workflow route.
- Component feature flags can take `FeatureOn` and `FeatureOff` children that control what is shown based on the feature flag value.

Workflow Rollout
featureflag

Component Rollout
ff

### Testing Performed

manual

### GitHub Issue
See #4


frontend/packages/core/src/flags.tsx
```diff
@@ -0,0 +1,38 @@
...
+      return {};
+    });
+};
+
+export interface FeatureFlagProps {
+  flagName: string;
+  children: React.ReactNode;
+}
+
+const FeatureFlag = ({ flagName, children }: FeatureFlagProps) => {
```
dschaller, line 27: Need to rethink this a bit

frontend/packages/core/src/AppProvider/workflow.tsx
```diff
@@ -42,6 +42,12 @@ interface Route {
   requiredConfigProps?: string[];
   /** Is the workflow discoverable via search and drawer navigation. This defaults to false. */
   hideNav?: boolean;
+  /**
+   * The feature flag used to determine if the route should be registered.
+   *
+   * If this is not set the route will always be registered.
+   */
+  featureFlag?: string;
```
danielhochman, line 50: can we derive the flag from existing uniquely identifiable information about the workflow instead of requiring the user to provide one?

this makes it impossible for us to publish a canonical feature flag for a workflow.

frontend/packages/core/src/flags.tsx
```diff
@@ -0,0 +1,115 @@
...
+const FF_CACHE_TTL = +process.env.REACT_APP_FF_CACHE_TTL || 60000;
+
+export interface FeatureFlags {
+  [name: string]: IClutch.featureflag.v1.IFlag;
+}
+
+const featureFlags = (): Promise<FeatureFlags> => {
+  const cachedFlags = JSON.parse(sessionStorage.getItem("featureFlags"));
+  if (cachedFlags) {
+    const elapsedSeconds = new Date(new Date().getTime() - cachedFlags.timestamp).getTime();
```
danielhochman, line 17: i think this is elapsed milliseconds

frontend/packages/core/src/AppProvider/workflow.tsx
dschaller: We could use something like the package name and route key but I was thinking it might be nice to allow users to specify a flag if they want to be able to enable all routes behind a single flag

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -12,44 +13,87 @@ import { Theme } from "./themes";
 import type { ConfiguredRoute, Workflow, WorkflowConfiguration } from "./workflow";
 import ErrorBoundary from "./workflow";
 
+const FF_POLL = +process.env.REACT_APP_FF_POLL || 300000;
```
lyft-tla, line 16: What do you think about having a lower bound?

dschaller: I like the idea but I also think it's reasonable to not have one so that users can lower it if they feel like they want more frequent polling.

frontend/packages/core/src/flags.tsx
```diff
@@ -0,0 +1,115 @@
+import * as React from "react";
+import type { clutch as IClutch } from "@clutch-sh/api";
+
+import type { ClutchError } from "./Network/errors";
+import { client } from "./Network/index";
+
+const FF_POLL = +process.env.REACT_APP_FF_POLL || 300000;
```
mikecutalo, line 7: Is there a place we can put `FF_POLL` so we can have a source of truth? I see we set the same const in `frontend/packages/core/src/AppProvider/index.tsx`.

dschaller: Yeah we could have a constants file for core. Good idea

frontend/packages/core/src/flags.tsx
```diff
@@ -0,0 +1,115 @@
...
+const featureFlags = (): Promise<FeatureFlags> => {
+  const cachedFlags = JSON.parse(sessionStorage.getItem("featureFlags"));
+  if (cachedFlags) {
+    const elapsedTime = new Date(new Date().getTime() - cachedFlags.timestamp).getTime();
+    if (elapsedTime < FF_CACHE_TTL) {
+      return new Promise(resolve => resolve(cachedFlags.flags));
+    }
+  }
+  return client
+    .post("/v1/featureflag/getFlags", {}, { timeout: 1000 })
```
mikecutalo, line 23: does `client` set a default timeout?

dschaller: yeah, we rely on the gateway timeout instead of on the client. This is set here intentionally since we don't want feature flags to become a blocking call

mikecutalo (approved):

PR #1470, opened 2021-05-28
authn: use refresh token if present in /login

danielhochman:
### Description
If a valid refresh token is present the user will be issued a new auth and refresh token when hitting /login. Otherwise the user will continue through the normal auth exchange.

### Testing Performed
Unit, manual.


backend/module/authn/authn.go
```diff
@@ -55,6 +57,33 @@ type api struct {
 }
 
 func (a *api) Login(ctx context.Context, request *authnv1.LoginRequest) (*authnv1.LoginResponse, error) {
+	// Attempt refresh. If refresh fails for any reason continue the regular auth flow.
+	md, _ := metadata.FromIncomingContext(ctx)
+	if v := md.Get("grpcgateway-cookie"); len(v) > 0 {
+		if refreshToken, err := mux.GetCookieValue(v, "refreshToken"); err == nil {
+			if t, err := a.issuer.RefreshToken(ctx, &oauth2.Token{RefreshToken: refreshToken}); err == nil {
```
rithujohn191, line 64: This refresh error does not get surfaced is this on purpose?

rithujohn191: Asking because OIDC has some refresh token policy errors (like `obsolete cause its used twice`, `expiration`,etc ) that might be helpful to surface

danielhochman: yep, the user should fall back to the regular auth flow since their refresh token is invalid. however i can split this out into another method and then log any error for debug purposes. will do that.

rithujohn191: 👍  logging the error would be helpful to identify a scenario where the refresh never happens due so some policy error

rithujohn191 (approved):
Looks good apart from the one comment

backend/gateway/mux/util_test.go
```diff
@@ -51,3 +51,17 @@ func TestRequestHeadersFromrResponseWriterMessed(t *testing.T) {
...
+func TestGetCookieValue(t *testing.T) {
+	cookies := []string{
+		"foo=bar;baz=bang",
+		"baz=bloop",
+	}
+	res, err := GetCookieValue(cookies, "foo")
+	assert.NoError(t, err)
+	assert.Equal(t, "bar", res)
+
+	res, err = GetCookieValue(cookies, "baz")
```
dschaller, line 64: should we add a test that when we ask for something that doesn't exist nothing comes back?

backend/module/authn/authn.go
```diff
@@ -52,9 +55,58 @@ func (m *mod) Register(r module.Registrar) error {
...
+	if err != nil {
+		return nil, nil
+	}
+
+	token, err := a.issuer.RefreshToken(ctx, &oauth2.Token{RefreshToken: refreshToken})
+	if err != nil {
+		return nil, err
+	}
+
+	err = grpc.SendHeader(ctx, metadata.New(map[string]string{
```
dschaller, line 81: should this be `SetHeader` instead of `SendHeader`?

danielhochman: good catch, makes more sense to stick to Set instead of Send even though we're not setting headers elsewhere yet. will change other instances of it as well

PR #1477, opened 2021-06-02
gateway: configurable max response size

danielhochman:
### Description
From default of 4MB. Any requests beyond this size should probably move to a paginated API. In the future this value could be made configurable.


smonero (approved):

backend/gateway/gateway.go
```diff
@@ -202,7 +202,11 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
 
 	// Create a client connection for the registrar to make grpc-gateway's handlers available.
 	// TODO: stand up a private loopback listener for the grpcServer and connect to that instead.
-	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", cfg.Gateway.Listener.GetTcp().Address, cfg.Gateway.Listener.GetTcp().Port), grpc.WithInsecure())
+	opts := []grpc.DialOption{
+		grpc.WithInsecure(),
+		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(5 * 1024 * 1024)),
```
dschaller, line 207: super nit: maybe add a todo inline about reading via config?

danielhochman: i can just go ahead and make it configurable now in that case, will do

PR #1482, opened 2021-06-04
k8s proto: add stateful set fields

rithujohn191:
### Description
This PR adds stateful set status fields and creation timestamp to the proto.

### Testing Performed
- unit
- local


backend/service/k8s/statefulset.go
```diff
@@ -67,12 +67,28 @@ func ProtoForStatefulSet(cluster string, statefulSet *appsv1.StatefulSet) *k8sap
...
+		// Convert Unix Timestamp to milliseconds
+		k8sStateful.CreationTimeMillis = statefulSet.CreationTimestamp.UnixNano() / 1e6
+	}
+
+	return k8sStateful
+}
+
+func ProtoForStatus(status appsv1.StatefulSetStatus) *k8sapiv1.StatefulSet_Status {
+	return &k8sapiv1.StatefulSet_Status{
+		Replicas:        uint32(status.Replicas),
```
smonero, line 89: I assume status will never be `nil`?

smonero (approved):

backend/service/k8s/statefulset.go
rithujohn191: Status is the current status of Pods in this StatefulSet so the status and number of replicas can never be nil similar to deployment status

PR #1484, opened 2021-06-04
project: api protos

mikecutalo:
### Description
Initial project api's.

### Testing Performed
n/a


backend/service/project/project.go
```diff
@@ -0,0 +1 @@
+package project
```
danielhochman, line 1: accidental? i guess it's the same as any other boilerplate

api/project/v1/project.proto
```diff
@@ -0,0 +1,28 @@
...
+    option (clutch.api.v1.action).type = READ;
+  }
+}
+
+message GetProjectsRequest {
+  repeated string names = 1 [ (validate.rules).repeated = {min_items : 1} ];
+}
+
+message GetProjectsResponse {
+  repeated clutch.core.project.v1.Project projects = 1;
```
danielhochman, line 27: we may want to preemptively wrap each result in a Message in case we want to somehow extend the result response without adding fields to project, kind of like how we ended up in a place where attaching metadata in the resolver API would result in a breaking change or a less optimal API. given that it's early we can also just go with this and reconsider before some initial usage of the API.

danielhochman (approved):
lgtm, some quick thoughts

backend/service/project/project.go
mikecutalo: Ah yeah I was working on the implementation ill back this out for now

api/project/v1/project.proto
mikecutalo: Yeah good idea, let me leave a TODO.

PR #1485, opened 2021-06-04
frontend: set lower bound on select default index

dschaller:
### Description

Add a lower bound on default option index.

### Testing Performed

manual and unit

### DIFF
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -201,9 +201,7 @@ const Select = ({
   options,
   onChange,
 }: SelectProps) => {
-  const [defaultIdx] = React.useState(
-    defaultOption < options.length && defaultOption > 0 ? defaultOption : 0
-  );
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/select.tsx
```diff
@@ -201,28 +201,33 @@ const Select = ({
   options,
   onChange,
 }: SelectProps) => {
-  if (options?.length === undefined || options?.length === 0) {
-    return null;
-  }
-
-  const defaultIdx = defaultOption < options.length ? defaultOption : 0;
+  const [defaultIdx] = React.useState(
```
danielhochman, line 204: why is this a hook?

dschaller: To ensure that the following dependent states are properly updated in order

danielhochman: i don't follow, would like to understand for my own personal benefit. what happens then if this is not a hook vs if it is?

dschaller: Ah nevermind, ignore this :)

PR #1499, opened 2021-06-10
audit: fix issue where events wouldn't sink

mikecutalo:
### Description
There could be a case where the number of audit events generated due to number of events and the size of them, becomes too large and the 1 second deadline for the update query fails. Once that threshold is surpassed it can never recover.

This pr includes a number of improvements to prevent this from happening.
* Use a select subquery in the update statement that limits the number of events to 100 to ensure smaller payload size.
* Since we are using a select subquery this operation is not concurrency safe, as we do not skip locked rows. To mitigate this we introduce a simple locking to control concurrency.

### TODOs

- [x] fix update sql query
- [x] revisit todo context
- [x] tests


backend/service/audit/setup.go
```diff
@@ -156,9 +159,22 @@ func (c *client) poll(interval time.Duration) {
...
-		readAndFanout()
+
+		isLocked, err := c.storage.AttemptLock(context.TODO(), lockID)
+		if err != nil {
+			c.logger.Error("Error while attempting to get lock", zap.Error(err))
+		}
+
+		if isLocked {
+			readAndFanout()
+			_, err := c.storage.ReleaseLock(context.TODO(), lockID)
```
mikecutalo, line 174: Wasn't immediately sure what i wanted to do with these context's, but maybe ill create a context with a deadline of 1s?

backend/service/audit/storage/sql/sql.go
```diff
@@ -187,6 +187,28 @@ func (c *client) query(ctx context.Context, query string, args ...interface{}) (
...
+	if err := c.db.QueryRowContext(ctx, "SELECT pg_advisory_unlock($1)", lockID).Scan(&unlock); err != nil {
+		c.logger.Error("Unable to perform an advisory unlock", zap.Error(err))
+		return false, err
+	}
+	return unlock, nil
+}
+
+func (c *client) GetLockID(lockID string) uint32 {
+	return storage.ConvertLockToUint32(lockID)
+}
```
danielhochman, line 210: this function being on client doesn't really seem to add anything, just call the Convert func directly?

backend/service/audit/setup.go
```diff
@@ -156,9 +159,22 @@ func (c *client) poll(interval time.Duration) {
 		}
 	}
 
+	lockID := c.storage.GetLockID(auditEventSinkLockId)
 	ticker := time.NewTicker(interval)
```
danielhochman, line 163: we should jitter the interval to minimize the chance that everything will be trying to take the lock at once

backend/service/audit/setup.go
danielhochman: i would pull the context from readAndFanout up, make it 10 seconds or something, and then just share it between all of these funcs

backend/service/audit/storage/sql/sql.go
mikecutalo: My initial though was that other providers might want to compute locks differently, however the interface still requires you to return a `unit32` so yeah I think it's a moot point. Let me remove.

danielhochman (approved):
:balloon:

PR #1506, opened 2021-06-14
project: refactor project api

mikecutalo:
### Description
Refactoring the project api protos. Removing the oss implementation for now, going to reassess how to make it more generic as we make progress on the feature set.


danielhochman (approved):
lgtm! @scarlettperry please take a look as well to make sure you don't spot anything missing that we would need.

scarlettperry:
> lgtm! @scarlettperry please take a look as well to make sure you don't spot anything missing that we would need.

@mikecutalo a mock API would be super helpful but I'm guessing that's coming in the next PR

mikecutalo:
@scarlettperry yeah there was a mock api, which i deleted in this PR as we are moving the impl to private for now. But yes next PR which is coming up soon will have a mock api.

PR #1465, opened 2021-05-27
project selector

scarlettperry:
### Description
Adds Project Selector for Dash which calls the GetProjects API for user-owned and custom projects.

Co-authored-by: Daniel Hochman <danielhochman@users.noreply.github.com>

### TODOs
- [x] Styling (alot of has been done, some small things left to do that can be handled in another PR)
- [ ] Hidden projects on change (will be handled in separate PR)
- [x] API Integration

PR #1507, opened 2021-06-15
project selector: don't add project that already exists

scarlettperry:
### Description
Previously, for each new custom project added to the list, an API call is made to retrieve that project even if the project already exists in the list. 

PR checks if the project has already been added in the state to avoid making an unnecessary API call.

### Testing Performed
before
before

after
after

### DIFF
--- a/frontend/workflows/projectSelector/src/hello-world.tsx
+++ b/frontend/workflows/projectSelector/src/hello-world.tsx
@@ -84,16 +84,24 @@ const useDispatch = () => {
 
 const selectorReducer = (state: State, action: Action): State => {
   switch (action.type) {
-    case "ADD_PROJECTS":
+    case "ADD_PROJECTS": {
+      // a given custom project may already exist in the group so don't trigger a state update for those duplicates
+      const uniqueCustomProjects = action.payload.projects.filter(
+        (project: string) => !(project in state[action.payload.group])
+      );
+      if (uniqueCustomProjects.length === 0) {
+        return { ...state };
+      }
       return {
         ...state,
         [action.payload.group]: {
           ...state[action.payload.group],
           ...Object.fromEntries(
-            action.payload.projects.map(v => [v, { checked: true, custom: true }])
+            uniqueCustomProjects.map(v => [v, { checked: true, custom: true }])
           ),
         },
       };
+    }
     case "REMOVE_PROJECTS":
       // TODO: also remove any upstreams or downstreams related (only) to the project.
       return {
@@ -316,10 +324,6 @@ const ProjectSelector = () => {
   const handleAdd = () => {
     if (customProject === "") {
       return;
-    }
-    if (customProject in state[Group.PROJECTS]){
-      // TODO: alert/show an "error" that the project is already in the list
-      return setCustomProject("");
     }
     dispatch({
       type: "ADD_PROJECTS",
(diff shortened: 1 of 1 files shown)


frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -84,17 +84,24 @@ const useDispatch = () => {
 
 const selectorReducer = (state: State, action: Action): State => {
   switch (action.type) {
-    case "ADD_PROJECTS":
-      // TODO: don't add if it already exists.
+    case "ADD_PROJECTS": {
```
scarlettperry, line 87: fyi wrapping the case in a `{}` b/c was getting the lint error `Unexpected lexical declaration in case block no-case-declarations` https://eslint.org/docs/rules/no-case-declarations

frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -84,17 +84,24 @@ const useDispatch = () => {
...
+      // a given custom project may already exist in the group so don't trigger a state update for those duplicates
+      const uniqueCustomProjects = action.payload.projects.filter(
+        (project: string) => !(project in state[action.payload.group])
+      );
```
scarlettperry, line 91: initially had this simple check https://github.com/lyft/clutch/pull/1507/commits/5152901f0bbb5c694b4abe48938c7052433ff03b#diff-0bcc1b146aa20a98ff28606102658d758469edad49c9a8f8ab6a9ed8fcf0e244R320-R323 but thought it might be better to move it here, especially if we plan to support bulk adding projects via the text box in the future.

scarlettperry:
on a side/related note: in playing with the other action types for the main projects (not the upstream/downstream groups), we trigger an API call each time the state changes (i.e unchecking a project). We should probably only call the API on load and when adding a new project to the list, so perhaps what we'll need to modify the flow here https://github.com/lyft/clutch/pull/1465/files#diff-0bcc1b146aa20a98ff28606102658d758469edad49c9a8f8ab6a9ed8fcf0e244R286-R320

frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -84,17 +84,24 @@ const useDispatch = () => {
...
+      if (uniqueCustomProjects.length === 0) {
+        return { ...state };
```
danielhochman, line 93: nit: return existing state directly since it was not modified
```suggestion
        return state;
```

frontend/workflows/projectSelector/src/hello-world.tsx
danielhochman: yep, that way if we add some other way to add projects in addition to the text box (e.g. query param) the logic still works. :+1:

frontend/workflows/projectSelector/src/hello-world.tsx
danielhochman: i was planning on doing this anyways otherwise you can get some really weird bugs with variables being in scope but not actually existing between cases. can you go ahead and wrap all of the cases in `{ }` for consistency?

danielhochman (approved):

danielhochman:
> on a side/related note: in playing with the other action types for the main projects (not the upstream/downstream groups), we trigger an API call each time the state changes (ex unchecking a project). We should probably only call the API on load and when adding a new project to the list, so perhaps we'll need to modify the flow here https://github.com/lyft/clutch/pull/1465/files#diff-0bcc1b146aa20a98ff28606102658d758469edad49c9a8f8ab6a9ed8fcf0e244R286-R320

the flow there is correct, it's actually that the fake API doesn't respond with any custom projects that are provided so the application keeps calling the API to try and fetch it. once we integrate with the real API it will work correctly (or we can continue to expand the fake API to return responses containing all projects).

PR #1508, opened 2021-06-15
frontend: add typography component

dschaller:
### Description

Add a new Typography component based on the figma designs.

### Testing Performed

manual and storybook


mikecutalo (approved):
Played around with it a bit in storybook and looks good!

frontend/packages/core/src/typography.tsx
```diff
@@ -0,0 +1,140 @@
...
+    | "body3"
+    | "caption1"
+    | "caption2"
+    | "overline"
+    | "input";
+  children: React.ReactNode;
+}
+
+const Typography = ({ variant, children }: TypographyProps) => {
+  return <StyledTypography data-variant={variant}>{children}</StyledTypography>;
```
danielhochman, line 137: does this need to be a data prop?

dschaller: yes, div doesn't have a variant prop

danielhochman: This seems to work fine.

```typescript
interface MyDivProps {
  colorz: string;
}

const MyStyledDiv = styled.div<MyDivProps>(
  {}, props => ({color: props.colorz})
);

const MyDiv = () => <MyStyledDiv colorz="blue" />
```

danielhochman: or just
```typescript
const MyStyledDiv = styled.div<{colorz: string}>(
  {}, props => ({color: props.colorz})
);

const MyDiv = () => <MyStyledDiv colorz="blue" />
```

dschaller: Oh yeah, I guess I can define the interface for it. I was relying on the existing props

PR #1513, opened 2021-06-15
project selector: apply negative margin to full bleed elements

danielhochman:
### Description
Fix elements to hit edges if needed.


danielhochman:
will pad inner elements instead

PR #1514, opened 2021-06-15
project: upstream / downstream dependencies

mikecutalo:
### Description
This adds project dependencies to the project proto. I felt that a new message was better then adding new fields to the project proto as we can more easily add new dependency types. Although I'm open to suggestions and feedback.


api/core/project/v1/project.proto
```diff
@@ -16,4 +16,13 @@ message Project {
...
   // Allows an organization to populate any information they see fit that does not fit our schema
   map<string, google.protobuf.Value> data = 5;
+  ProjectDependencies dependencies = 6;
+}
+
+// Dependencies could be either upstream or downstream of this project.
+// This is also a place where we can add other dependency types such as datastores.
+message ProjectDependencies {
+  // Projects that are Upstreams and Downstreams of this project, the values are other Project.Name's.
+  repeated string upstreams = 1;
```
danielhochman, line 26: can we give these type_urls for possible future expansion?

mikecutalo: Something like this?

```proto
message Dependency {
  string identifier = 1; 
  string type_url = 2;
}

message ProjectDependencies {  
 repeated Dependency upstreams = 1;
```

danielhochman: yep! `id` is fine.

smonero (approved):

PR #1538, opened 2021-06-23
project selector: project api integration

scarlettperry:
### Description
PR adds support for
* calling the API for user-owned projects & custom projects
* processing the API response results and adding it to the State
* an error state/reducer (simple setup for now) for when the API returns an error. if there's an error, it'll be rendered in the project selector

Mentioned the TODOs in the PR to address in follow up PRs.

### Testing Performed
Locally

To test yourself, can run the real local backend or the mock api (search for the "comments" project for the scenario of calling api for a custom project)

mockdata

simplerrstate

### DIFF
--- a/frontend/workflows/projectSelector/src/hello-world.tsx
+++ b/frontend/workflows/projectSelector/src/hello-world.tsx
@@ -173,7 +173,7 @@ const ProjectSelector = () => {
       dispatch({ type: "HYDRATE_START" });
 
       // TODO: have userId check be server driven
-      const requestParams = {"users": [userId()], "projects": []}
+      const requestParams = { users: [userId()], projects: [] };
 
       const customProjects = [];
       _.forEach(Object.keys(state[Group.PROJECTS]), p => {
@@ -192,11 +192,12 @@ const ProjectSelector = () => {
       */
       client
         .post("/v1/project/getProjects", requestParams)
-        .then(resp => dispatch({ type: "HYDRATE_END", payload: { result: resp.data.results } }))
+        .then(resp =>
+          dispatch({ type: "HYDRATE_END", payload: { result: resp?.data?.results || {} } })
+        )
         .catch((err: ClutchError) => {
           dispatch({ type: "HYDRATE_ERROR", payload: { result: err } });
         });
-
     }
   }, [state[Group.PROJECTS]]);
 
--- a/frontend/workflows/projectSelector/src/hello-world.tsx
+++ b/frontend/workflows/projectSelector/src/hello-world.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
-import { TextField } from "@clutch-sh/core";
+import type { ClutchError } from "@clutch-sh/core";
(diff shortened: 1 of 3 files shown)


frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -161,6 +157,10 @@ const ProjectSelector = () => {
 
     let allPresent = true;
     _.forEach(Object.keys(state[Group.PROJECTS]), p => {
+      /*
+      TODO: b/c of this conditional, if a user adds an upstream/downstream we already have the project data for
+      to the custom project group, allPresent will be true and we wont trigger an api call
+      */
```
mikecutalo, line 163: Whats the todo here?

frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -190,25 +214,31 @@ const ProjectSelector = () => {
...
-            />
+          {/* TODO: add plus icon in the text field */}
+          <StyledProjectTextField
+            disabled={state.loading}
+            placeholder="Add a project"
+            value={customProject}
+            onChange={e => setCustomProject(e.target.value)}
+            onKeyDown={e => e.key === "Enter" && handleAdd()}
+            helperText={state.error?.message}
+            error={hasError}
```
scarlettperry, line 240: @dschaller, updated to use the TextField's component helperText and error fields.

frontend/workflows/projectSelector/src/hello-world.tsx
scarlettperry: updated the todo to mention one approach we can take. one approach is updating the conditional to check if the project in state[project] is also in state[upstream]/state[downstream] but need to think on it a bit more

frontend/workflows/projectSelector/src/hello-world.tsx
scarlettperry: Screen Shot 2021-06-24 at 4 31 01 PM

frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -51,10 +52,15 @@ export interface State {
...
-// TODO: subout with full manifest structure (from proto def)
 interface Project {
+  name: string;
+  tier: string;
+  owners: string[];
+  languages: string[];
+  data: any;
   upstreams: string[];
   downstreams: string[];
 }
```
mikecutalo, line 66: Why cant we use the proto here?

mikecutalo (approved):
Left a comment but other it looks sweet!

frontend/workflows/projectSelector/src/hello-world.tsx
scarlettperry: @mikecutalo latest commit updates to use the Project type and also use the request/response types as well https://github.com/lyft/clutch/pull/1538/commits/cdeb56757aebe8e0304981531dd5446fa9961aad

PR #1556, opened 2021-06-28
project selector: remove exclusive upstreams/downstreams of a project

scarlettperry:
### Description
PR 
* creates a mapping of upstream/downstream -> projects(s) so that we can handle removing upstreams/downstreams that are exclusive to a project when a project is removed from `Group.Project` (and in a next PR, hide upstreams/downstreams that are exclusive to a project when the `Group.Project` project is unchecked)

* updates the `REMOVE_PROJECT` state to remove the upstreams/downstreams that are exclusive to the respective project

### Testing Performed
Locally. For this PR, it's easier to test with real data than mock data to catch the edge cases (i.e. we remove a project `foo`. it's upstream `bar` is also in the downstream list. upstream `bar` is exclusive to `foo` but downstream `bar` is shared by both `foo` and `baz`. in this scenario, we remove `bar` from the upstream group but keep it in the downstream group.)


frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -27,13 +27,65 @@ const selectorReducer = (state: State, action: Action): State => {
...
+        action.payload.projects.forEach(p => {
+          _.forIn(state.upstreamToProject, (v, k) => {
+            // if an upstream is associated to the project
+            if (v[p]) {
+              // if the upstream is exclusive to project
+              if (Object.keys(v).length == 1) {
+                upstreamsToRemove.push(k);
+                // remove as it was only tied to this project and b/c both the project and upstream will be removed from the state
+                // otherwise we'll end up with stale information
+                delete state.upstreamToProject[k];
```
scarlettperry, line 51: scenario: we remove project `foo` and upstream `bar` is tied only to `foo`, which we're going to remove. so we delete `"bar": {"foo": true}` b/c this mapping is stale since we've removed `foo` and `bar` from the state.
```
upstreamToProject: {
  "bar": {"foo": true}
}
```

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -27,13 +27,65 @@ const selectorReducer = (state: State, action: Action): State => {
...
+              } else {
+                // project is not exclusive to the upstream but let's remove as it will be removed from Group.PROJECTS state
+                // otherwise we'll end up with stale information
+                delete v[p];
```
scarlettperry, line 55: scenario: we remove project `foo` and upstream `bar` is tied to both `foo` and `baz`. we're not going to remove the upstream `bar` but we delete {"foo": true} from "bar": {"foo": true, "baz": true} b/c we've removed `foo` from the project state. if we don't delete `foo`, then the length of the object will always be greater than 1, and if the user nexts wants to delete `baz`, the upstream will never be removed. 

```
upstreamToProject: {
  "bar": {"foo": true, "baz": true}
}
```

PR #1557, opened 2021-06-28
project selector: update API call

scarlettperry:
### Description
Always make an API with all the custom projects in the group.Projects state


frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -168,16 +168,12 @@ const ProjectSelector = () => {
...
+        // if the project is custom
+        if (state[Group.PROJECTS][p].custom) {
           requestParams.projects.push(p);
         }
       });
 
-      /*
-      TODO: the API doesn't return an error if a custom project is not found so we should first
-      check if the API returns empty results and process that as an error
-      */
```
scarlettperry, line 180: removing as discussed offline that API will be updated to return an error

danielhochman (approved):

PR #1566, opened 2021-06-29
project selector: helper to update project group state

scarlettperry:
### Description
PR does a refactor to add a helper method to handle duplicate logic (ie. whether or not something should be checked true/false). We'll use this pattern of writing helpers moving forward. 

### Testing Performed
Locally with real API data


frontend/workflows/projectSelector/src/state-store.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { Group } from "./hello-world"
+
+export default class StateStore{
+  [Group.PROJECTS]: GroupState;
+  [Group.UPSTREAM]: GroupState;
+  [Group.DOWNSTREAM]: GroupState;
+
+  projectData: { [projectName: string]: IClutch.core.project.v1.IProject };
+  loading: boolean;
+  error: ClutchError | undefined;
```
scarlettperry, line 13: looking into how we this class can extend the `State` interface we've already defined

frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -72,6 +73,8 @@ export const useReducerState = () => {
   return React.useContext(StateContext);
 };
 
+export const stateStore = new StateStore();
```
danielhochman, line 76: don't think this should be exported, state will come from the hook.

scarlettperry: yea there are a few things I want to discuss in our sync that I'm not too sure about

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -4,6 +4,33 @@ import _ from "lodash";
 import type { Action, State } from "./hello-world";
 import { deriveSwitchStatus, Group } from "./hello-world";
 
+// handle custom ; 4th param will be project state object
+const updateGroupstate = (
+  state: State,
+  group: Group,
+  project: string,
+  isCustom?: boolean
```
scarlettperry, line 12: @danielhochman I know you suggested using the ProjectState object but the interface requires the "checked" arg & since what we want to account for is `custom`, I set a bool that the can be passed in. lmk what you think

```
         // user owned project vs custom project
          if (v.from.users.length > 0) {
            newState = updateGroupstate(newState, Group.PROJECTS, k, false);
          } else if (v.from.selected) {
            newState = updateGroupstate(newState, Group.PROJECTS, k, true);
         }
```

danielhochman: i think we also need to be able to add a project with checked = false though 

e.g. https://github.com/lyft/clutch/pull/1465/files#diff-66c15f7738bc8884141f5f91e2c8444c958af9bd00c50832621795772b0f8e53R122-R128

scarlettperry: this is it for upstreams/downstreams (if they're not already in the group) https://github.com/lyft/clutch/pull/1566/files#diff-66c15f7738bc8884141f5f91e2c8444c958af9bd00c50832621795772b0f8e53R24-R28


Group.Projects are checled true if they're not already in the group https://github.com/lyft/clutch/pull/1566/files#diff-66c15f7738bc8884141f5f91e2c8444c958af9bd00c50832621795772b0f8e53R21-R23

danielhochman: i think it would be clearer to keep that logic in the reducer

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -83,59 +105,30 @@ const selectorReducer = (state: State, action: Action): State => {
...
-          // a user owned project
+          // user owned project vs custom project
           if (v.from.users.length > 0) {
-            // preserve the current checked state if the project already exists in this group
-            if (k in state[Group.PROJECTS]) {
-              newState[Group.PROJECTS][k] = { checked: state[Group.PROJECTS][k].checked };
-            } else {
-              newState[Group.PROJECTS][k] = { checked: true };
-            }
+            newState = updateGroupstate(newState, Group.PROJECTS, k, {checked: true, custom: false});
```
danielhochman, line 115: nit you can omit `custom: false`

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -1,9 +1,31 @@
...
+  state: State,
+  group: Group,
+  project: string,
+  projectState: ProjectState,
+): State => {
+  const newState = { ...state };
+  if (project in newState[group]) {
+    // preserve the checked value if the project is already in the group
+    newState[group][project].checked = state[group][project].checked;
+    if (group === Group.PROJECTS) {
```
danielhochman, line 17: this isn't really needed, as we never pass in a custom true on a non group project

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -1,9 +1,28 @@
...
+const updateGroupstate = (
+  state: State,
+  group: Group,
+  project: string,
+  projectState: ProjectState
+): State => {
+  const newState = { ...state };
+  if (project in newState[group]) {
+    // preserve the checked value if the project is already in the group
+    newState[group][project].checked = state[group][project].checked;
```
danielhochman, line 16: redundant
```suggestion
    newState[group][project].checked = state[group][project].checked;
```

danielhochman (approved):

frontend/workflows/projectSelector/src/selector-reducer.tsx
scarlettperry: the suggested change is the same. was trying to figure out what you meant but couldn't spot the redundancy. going to merge this in to incorporate the helper in `add_projects` and can fix in the other PR!

PR #1571, opened 2021-07-01
project selector: update "add_project" to add upstreams/downstreams if we have projectdata

scarlettperry:
### Description

Background
The flow only makes an API call if we don't have data on a project & on each API request we store data on the project and its upstreams/downstreams. So previously, if you added an upstream/downstream as a custom project, we weren't updating the state with its own upstreams/downstreams.

Solution
PR updates the `add_project` state to check if we have project data on a project and if so, we update the state with its upstreams/downstreams.

### Testing Performed
Locally with real API data


frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -28,22 +28,47 @@ const selectorReducer = (state: State, action: Action): State => {
...
       };
+
+      // if the newly added project is in Group.Projects, check to see if we have
+      // project data for it (b/c we don't need to make an API call) and update the state with
+      // its upstreams/downstreams
+      if (action.payload.group === Group.PROJECTS) {
+        uniqueCustomProjects.forEach(v => {
+          if (v in newState.projectData) {
+            newState.projectData[v].dependencies.upstreams[
+              "type.googleapis.com/clutch.core.project.v1.Project"
```
danielhochman, line 53: move TypeURL to global const in file

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -28,22 +28,47 @@ const selectorReducer = (state: State, action: Action): State => {
...
+
+      newState[action.payload.group] = {
+        ...state[action.payload.group],
+        ...Object.fromEntries(uniqueCustomProjects.map(v => [v, { checked: true, custom: true }])),
       };
+
+      // if the newly added project is in Group.Projects, check to see if we have
+      // project data for it (b/c we don't need to make an API call) and update the state with
+      // its upstreams/downstreams
+      if (action.payload.group === Group.PROJECTS) {
```
danielhochman, line 49: try to keep happy path unindented. this is clearer to read.

```suggestion
      if (action.payload.group !== Group.PROJECTS) {
        return newState;
      }
      // continue with the other logic
```

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -28,22 +28,47 @@ const selectorReducer = (state: State, action: Action): State => {
...
       // a given custom project may already exist in the group so don't trigger a state update for those duplicates
       const uniqueCustomProjects = action.payload.projects.filter(
-        (project: string) => !(project in state[action.payload.group])
+        (project: string) => !(project in newState[action.payload.group])
       );
       if (uniqueCustomProjects.length === 0) {
-        return state;
+        return newState;
       }
-      return {
```
danielhochman, line 38: i would keep this as-is (assign to  newState instead of return), you don't want to return copied state unless the state was modified.

danielhochman: also let's add a comment for this logic now that there's the additional step of adding upstreams and downstreams

danielhochman (commented):
small nits otherwise looking good

scarlettperry:
closing this PR b/c easier than rebasing and having to fix conflicts for so many commits.

PR #1573, opened 2021-07-01
project-selector: lint fixes

scarlettperry:
### Description
Main PR changes are the creation of a new `types` file so that we can avoid the cyclical imports issue (ie interface types in X are imported into Y, and Y is imported in X)

After this & this PR https://github.com/lyft/clutch/pull/1574, project selector is ready to be merged to main branch.

### Testing Performed
Ran through everything again locally.


frontend/packages/core/src/Input/checkbox.tsx
```diff
@@ -15,33 +15,31 @@ const FormControl = styled(MuiFormControl)({
   width: "75%",
 });
 
-const StyledCheckbox = styled(MuiCheckbox)(
```
scarlettperry, line 18: some more lint fixes that didn't get picked up in last PR

frontend/workflows/projectSelector/src/hello-world.tsx
```diff
@@ -140,7 +67,7 @@ const ProjectSelector = () => {
   const [state, dispatch] = React.useReducer(selectorReducer, initialState);
 
   React.useEffect(() => {
-    console.log("effect");
+    console.log("effect"); // eslint-disable-line
```
scarlettperry, line 70: keeping these console logs for now. they're useful for debugging

danielhochman (approved):
+1 i have some other thoughts on organization using an `index.tsx` but it can come as a followup to this

PR #1577, opened 2021-07-02
frontend: add custom react-router hooks

dschaller:
### Description

Add custom react-router hooks to:
  -  preserve utm search params
  - provide an easier interface to track origins when routing between pages
  - provide an easier interface to specify search params (see [this](https://github.com/ReactTraining/react-router/issues/7743) for context)

### Testing Performed

manual and unit

### DIFF
--- a/frontend/packages/core/src/navigation.tsx
+++ b/frontend/packages/core/src/navigation.tsx
@@ -46,6 +46,12 @@ export interface NavigateOptions extends ReactRouterNavigateOptions {
 
 /**
  * Convert custom search param object to string.
+ *
+ * ```ts
+ * const searchObject = new URLSearchParams({ query: "value" });
+ * const searchString = convertSearchParam(searchObject);
+ * console.log(searchString); // "?query=value"
+ * ```
  */
 const convertSearchParam = (params: URLSearchParams) => {
   const p = params.toString();
--- a/frontend/packages/core/src/navigation.tsx
+++ b/frontend/packages/core/src/navigation.tsx
@@ -1,0 +1,171 @@
+import type { URLSearchParamsInit } from "react-router-dom";
+import {
+  createSearchParams,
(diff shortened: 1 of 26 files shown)


smonero (approved):

frontend/packages/core/src/navigation.tsx
```diff
@@ -0,0 +1,171 @@
...
+  // Include current URL in state as origin. Defaults to false.
+  origin?: boolean;
+  // Persist UTM search params after navigation. Defaults to true.
+  utm?: boolean;
+}
+
+/**
+ * Convert custom search param object to string.
+ */
+const convertSearchParam = (params: URLSearchParams) => {
```
mikecutalo, line 50: Can you provide an example in the comments?

frontend/packages/core/src/navigation.tsx
```diff
@@ -0,0 +1,171 @@
...
+      }
+    }
+
+    let navPath = {} as PartialPath;
+    if (typeof to === "string") {
+      navPath.pathname = to;
+    } else {
+      navPath = { pathname: to?.pathname, hash: to?.hash };
+    }
+    const [path, search] = navPath.pathname.split("?");
```
mikecutalo, line 110: Is there any utility that would simplify parsing the URL rather than needing to split?

mikecutalo (approved):

mikecutalo (approved):
Do we also need to update the scaffolder to use this new package?

frontend/packages/core/src/navigation.tsx
dschaller: There isn't an easy way to split this since it's just the path. If I wanted to use a utility I would need to rely on the window location to get the hostname.

mikecutalo (approved):

PR #1576, opened 2021-07-01
project selector: update "add_project" state to check project data cache

scarlettperry:
Same as https://github.com/lyft/clutch/pull/1571 but closed b/c commit history issues after I merged the feature branch into main

### Description

Background
The flow only makes an API call if we don't have data on a project & on each API request we store data on the project and its upstreams/downstreams. So previously, if you added an upstream/downstream as a custom project, we weren't updating the state with its own upstreams/downstreams.

Solution
PR updates the `add_project` state to check if we have project data on a project and if so, we update the state with its upstreams/downstreams.

### Testing Performed
Locally with real API data

PR #1578, opened 2021-07-02
project selector: remove exclusive upstreams/downstreams

scarlettperry:
### Description

PR adds
* helpers to collect the exclusive upstreams/downstreams of a project in the "remove_projects" state
* removes the exclusive upstreams/downstreams of the removed project(s)

### Testing Performed
Tested locally with real API data

### DIFF
--- a/frontend/workflows/projectSelector/src/helpers.tsx
+++ b/frontend/workflows/projectSelector/src/helpers.tsx
@@ -53,15 +53,15 @@ const dependencyToProjects = (state: State, group: Group): DependencyMappings =>
 
   const projects = Object.keys(state[group]);
   projects.forEach(p => {
-    const { upstreams, downstreams } = state.projectData[p]?.dependencies;
-    upstreams[PROJECT_TYPE_URL]?.ids.forEach(u => {
+    const { upstreams, downstreams } = state.projectData[p]?.dependencies || {};
+    upstreams?.[PROJECT_TYPE_URL]?.ids.forEach(u => {
       if (!upstreamMap[u]) {
         upstreamMap[u] = { [p]: true };
       } else {
         upstreamMap[u][p] = true;
       }
     });
-    downstreams[PROJECT_TYPE_URL]?.ids.forEach(d => {
+    downstreams?.[PROJECT_TYPE_URL]?.ids.forEach(d => {
       if (!downstreamMap[d]) {
         downstreamMap[d] = { [p]: true };
       } else {
--- a/frontend/workflows/projectSelector/src/helpers.tsx
+++ b/frontend/workflows/projectSelector/src/helpers.tsx
@@ -1,0 +1,108 @@
+import * as React from "react";
+import _ from "lodash";
+
+import type { Action, Group, ProjectState, State } from "./types";
+
+const PROJECT_TYPE_URL = "type.googleapis.com/clutch.core.project.v1.Project";
+
+interface DependencyMappings {
+  upstreams?: { [dependency: string]: { [project: string]: boolean } };
+  downstreams?: { [dependency: string]: { [project: string]: boolean } };
+}
+
+const StateContext = React.createContext<State | undefined>(undefined);
+const useReducerState = () => {
+  return React.useContext(StateContext);
+};
+
+const DispatchContext = React.createContext<(action: Action) => void | undefined>(undefined);
+const useDispatch = () => {
+  return React.useContext(DispatchContext);
+};
+
+// TODO(perf): call with useMemo().
+const deriveSwitchStatus = (state: State, group: Group): boolean => {
+  return (
+    Object.keys(state[group]).length > 0 &&
+    Object.keys(state[group]).every(key => state[group][key].checked)
+  );
+};
+
+const updateGroupstate = (
+  state: State,
+  group: Group,
+  project: string,
+  projectState: ProjectState
+): State => {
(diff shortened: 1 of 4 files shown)


frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -5,6 +5,14 @@ import _ from "lodash";
 import type { Action, ProjectState, State } from "./types";
 import { Group } from "./types";
 
+const TYPEURL = "type.googleapis.com/clutch.core.project.v1.Project";
```
scarlettperry, line 8: will remove after merging https://github.com/lyft/clutch/pull/1576

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -42,6 +50,56 @@ const updateGroupstate = (
   return newState;
 };
 
+// TODO: make an interface for the return value?
+const exclusiveProjectDependencies = (
+  state: State,
+  group: Group,
+  project: string
+): { upstreams: string[]; downstreams: string[] } => {
```
scarlettperry, line 58: debated whether I should define this in a `Dependency` interface but thought perhaps not needed for now? lmk if you prefer an interface tho

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -97,13 +151,32 @@ const selectorReducer = (state: State, action: Action): State => {
...
         ...state,
         [action.payload.group]: _.omit(state[action.payload.group], action.payload.projects),
       };
+
+      if (action.payload.group !== Group.PROJECTS) {
+        return newState;
+      }
+
+      // for the removed projects in Group.Projects we get their upstreams/dowstreams that are exclusive
+      // (i.e. not shared by other projects that weren't removed)
```
danielhochman, line 165: unfortunately i think there's a bug here.

let's say we have

```
foo => a, b
bar => b, c
baz => x, y
```

and we call `remove(foo, bar)`

the result should be `[x, y]` but i think the current code would result in `[b, x, y]`

we could leave a TODO i guess. this complexity is an unfortunate result of my choice to put in multiple projects.

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -97,13 +151,32 @@ const selectorReducer = (state: State, action: Action): State => {
...
+      const upstreams = [];
+      const downstreams = [];
```
danielhochman, line 167: nit clarity
```suggestion
      const exclusiveUpstreams = [];
      const exclusiveDownstreams = [];
```

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -97,13 +151,32 @@ const selectorReducer = (state: State, action: Action): State => {
...
+      action.payload.projects.forEach(p => {
+        // get the upstreams/downstreams that are exclusive to the removed project(s)
+        const dependenices = exclusiveProjectDependencies(state, Group.PROJECTS, p);
```
danielhochman, line 170: nit spelling
```suggestion
        const dependencies = exclusiveProjectDependencies(state, Group.PROJECTS, p);
```

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -7,6 +7,11 @@ import { Group } from "./types";
 
 const PROJECT_TYPE_URL = "type.googleapis.com/clutch.core.project.v1.Project";
 
+interface DependencyMappings {
+  upstreams?: { [dependency: string]: { [project: string]: boolean } };
+  downstreams?: { [dependency: string]: { [project: string]: boolean } };
+}
```
danielhochman, line 13: i think we should move these functions and interfaces to a helper file, wdyt?

frontend/workflows/projectSelector/src/selector-reducer.tsx
scarlettperry: ooh i see - in the func I check to see if the dependency map is a length of 1 but since I'm only checking it for 1 project at at time, the removed projects could be sharing that same dependency but it wouldn't be removed. i'll leave a TODO and followup

frontend/workflows/projectSelector/src/selector-reducer.tsx
scarlettperry: yep sg!

danielhochman (approved):

frontend/workflows/projectSelector/src/helpers.tsx
```diff
@@ -0,0 +1,108 @@
...
+      } else {
+        downstreamMap[d][p] = true;
+      }
+    });
+  });
+
+  return { upstreams: upstreamMap, downstreams: downstreamMap };
+};
+
+const exclusiveProjectDependencies = (
```
scarlettperry, line 76: @danielhochman  was thinking about this more and maybe we'll need a seperate helper for the hidden actions b/c the latter needs to know what is & isn't unchecked. ie if all the unchecked projects share the same dependency, they can be hidden but if any are still used by the checked projects, they can't.

scarlettperry: both can share the func that creates the mapping tho

PR #1585, opened 2021-07-08
frontend: fix autocomplete when chosen value isn't an option

dschaller:
### Description

Fix issue where autocomplete would clear input text if it didn't match an available option.

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Input/text-field.tsx
+++ b/frontend/packages/core/src/Input/text-field.tsx
@@ -228,6 +228,8 @@ const TextField = ({
     },
   };
 
+  // We maintain a defaultVal to prevent the value from changing from underneath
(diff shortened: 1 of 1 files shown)


mikecutalo (approved):

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -228,7 +228,10 @@ const TextField = ({
     },
   };
 
-  const [autoCompleteOptions, setAutoCompleteOptions] = React.useState([]);
+  const [defaultVal] = React.useState<string>((defaultValue as string) || "");
```
danielhochman, line 231: add comment explaining why this is here

danielhochman (approved):

PR #1588, opened 2021-07-09
project-selector: hide exclusive upstreams/downstreams

scarlettperry:
### Description
PR adds support for hiding exclusive upstreams/downstreams of a user-owned/custom project when the latter are unchecked.

PR flow derives the hidden status based on the checked value (ie `true`/`false`) of the user-owned/custom project. In this way, we don't need to handle the flow in the `actions` as a the hidden status of an upstream/downstream is influenced by various user actions (the toggle, the only button, the checkbox, if a new project is added, etc).

### Testing Performed
locally with real API data

### TODOs
- [x] hide an upstream/downstream if all unchecked projects share it

### Questions:
~1. how do we want to handle the switch status b/c we preserve the checked status of hidden projects. we can simply filter by "if checked but not hidden". the `deriveSwitchStatus` helper is used in other parts tho so we'll need to create a helper for the non-switch status use cases.~ **in latest commit, we pass the derived data to the components so the project count already correctly reflects what's _not_ hidden**

~2. how do we want to alert the user to the number of hidden projects. we currently display total number of projects in the chip; we can simply subtract the number of hidden from that project but should we let the users know that some projects have been hidden?~ **we can revisit after collecting user feedback on the hiding behavior**

### DIFF
--- a/frontend/workflows/projectSelector/src/helpers.tsx
+++ b/frontend/workflows/projectSelector/src/helpers.tsx
@@ -48,6 +48,12 @@ const updateGroupstate = (
   return newState;
 };
 
+const updateHidden = (state: State, group: Group, project: string, hidden: boolean): State => {
+  const newState = { ...state };
+  newState[group][project].hidden = hidden;
+  return newState;
+};
+
 const dependencyToProjects = (state: State, group: Group): DependencyMappings => {
   const upstreamMap = {};
   const downstreamMap = {};
@@ -97,7 +103,14 @@ const exclusiveProjectDependencies = (
   return { upstreams, downstreams };
 };
 
+// deriveHiddenStatus evaluates whether to hide an upstream/downstream based on
+// the checked status of/exclusivity to projects in Group.Projects
 const deriveHiddenStatus = (state: State, group: Group, key: string): boolean => {
+  // if true, don't need to go through the flow below as we only hide an upstream/downstream
+  if (group == Group.PROJECTS) {
+    return false;
+  }
+
   const unchecked = [];
   _.forIn(state[Group.PROJECTS], (v, k) => {
     if (!v.checked) {
@@ -108,6 +121,7 @@ const deriveHiddenStatus = (state: State, group: Group, key: string): boolean =>
 
   // no unchecked projects
   if (unchecked.length === 0) {
+    updateHidden(state, group, key, false);
     return false;
   }
 
@@ -116,19 +130,25 @@ const deriveHiddenStatus = (state: State, group: Group, key: string): boolean =>
 
   const exclusive = [];
   if (group === Group.UPSTREAM) {
-    // TODO: if all unchecked projects share that dependency, hide it
     unchecked.forEach(p => {
       _.forIn(dependencyMapping.upstreams, (v, k) => {
+        // hide an upstream if it's exclusive to the unchecked project (ie. no other checked project(s) share it)
         if (v[p] && Object.keys(v).length === 1) {
+          exclusive.push(k);
+        } else if (unchecked.length > 1 && unchecked.every(p => Object.keys(v).includes(p))) {
+          // hide an upstream if all unchecked projects share it
           exclusive.push(k);
         }
       });
     });
   } else if (group === Group.DOWNSTREAM) {
-    // TODO: if all unchecked projects share that dependency, hide it
     unchecked.forEach(p => {
       _.forIn(dependencyMapping.downstreams, (v, k) => {
+        // hide a downstream if it's exclusive to the unchecked project (ie. no other checked project(s) share it)
         if (v[p] && Object.keys(v).length === 1) {
+          exclusive.push(k);
+        } else if (unchecked.length > 1 && unchecked.every(p => Object.keys(v).includes(p))) {
+          // hide a downstream if all unchecked projects share it
           exclusive.push(k);
         }
       });
@@ -136,9 +156,11 @@ const deriveHiddenStatus = (state: State, group: Group, key: string): boolean =>
   }
 
   if (exclusive.includes(key)) {
+    updateHidden(state, group, key, true);
     return true;
   }
 
+  updateHidden(state, group, key, false);
   return false;
 };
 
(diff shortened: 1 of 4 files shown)


frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -153,51 +156,54 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
...
-                >
-                  Only
-                </StyledOnlyButton>
-                <StyledClearIcon>
-                  {state[group][key].custom && (
-                    <IconButton
+          {Object.keys(state[group]).map(
+            key =>
+              !deriveHiddenStatus(state, group, key) && (
+                <StyledMenuItem key={key}>
```
scarlettperry, line 162: no changes from this point on; the rest are just indentation updates b/c of the addition of line `161`

scarlettperry:
@danielhochman noted the main todo in the PR description and other less urgent ones in the PR but the e2e flow of hiding upstreams/downstreams there

frontend/workflows/projectSelector/src/helpers.tsx
```diff
@@ -96,7 +103,75 @@ const exclusiveProjectDependencies = (
...
+  _.forIn(state[Group.PROJECTS], (v, k) => {
+    if (!v.checked) {
+      // collect the unchecked projects, if any
+      unchecked.push(k);
+    }
+  });
+
+  // no unchecked projects
+  if (unchecked.length === 0) {
+    updateHidden(state, group, key, false);
```
scarlettperry, line 124: this is buggy/doesn't always accurately reflect current state of hidden projects. still thinking about how to properly update the global state

scarlettperry: actually the bug was with how I was handling some edge cases w.r.t hiding an upstream/downstream if all its parent projects are unchecked. will push up the fixes

frontend/workflows/projectSelector/src/helpers.tsx
```diff
@@ -47,6 +48,12 @@ const updateGroupstate = (
   return newState;
 };
 
+const updateHidden = (state: State, group: Group, project: string, hidden: boolean): State => {
+  const newState = { ...state };
+  newState[group][project].hidden = hidden;
```
danielhochman, line 53: if it's derived you'll remove it from the state entirely and only use it when rendering. should not be in both places.

scarlettperry: yeah I did it like this so that we'd have a way to know whether upstream/downstream project data should be passed to cards (ie "if checked but not hidden"). 🤔  perhaps I can just do one - use the `deriveHiddenStatus` to only update the hidden state and then check the `hidden` field to decide whether to render

danielhochman: the effect that will update the parent context is in the component, so it will have access to which projects are hidden or not when it updates the parent context

scarlettperry: updated PR to have the derived computation in the project-selector component. i also tested this on your `dash-data-flow` branch and it worked nicely. in the new useEffect hook to update selected projects, I used the `derivedState` value.

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -114,6 +114,10 @@ const ProjectSelector = ({ children }) => {
     }
   }, [state[Group.PROJECTS]]);
 
+  // computes the final state for rendering across other components
+  // (ie. filters out upstream/downstreams that are "hidden")
+  const derivedState = React.useMemo(() => deriveStateData(state), [state]);
```
scarlettperry, line 119: some perf benefits to using `useMemo` for example, without it we'd be calling this func every time the user types in the text field

scarlettperry:
@danielhochman some more context on the recent commit: 

initially I passed down a list of projects that should be hidden as a prop to respective components (ie project group) so that they could filter it from the original state data. dropped the approach since we'd have to remember to filter out the final state data in each component

next i added a new `useState` hook for holding the the derived state value and initialized it with the original state data. in this way, the original state won't be replaced and we'd pass a copy of the final state (that doesn't contain the hidden upstreams/downstreams) to the components for final render. i also added a new `useEffect` hook to listen to changes in state in order to call the derivedstate func which sets the new value of the derived data. while that worked, seemed unnecessary to have a new useState and useEffect hook

in the current approach, I call the derivedstate func directly in the project selector which returns a copy of the state (without the hidden upstreams/downstreams) and I use that value in the StateContext which gets used for render in the project group component. i also tested this out with your data flow PR (value gets used in the new useEffect hook), and it worked out well.

frontend/workflows/projectSelector/src/helpers.tsx
```diff
@@ -96,7 +97,92 @@ const exclusiveProjectDependencies = (
   return { upstreams, downstreams };
 };
 
+// evaluates whether an upstream/downstream should be hidden based on the checked status of/exclusivity to a project(s)
+const hidden = (
```
danielhochman, line 101: seems like it would be more efficient figure out the hidden projects (or conversely the ones that should be displayed) once for all selected projects rather than per project, lots of nested loops happening.

scarlettperry: thx for raising! updated to remove unnecessary extra looping

frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -140,7 +140,7 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
             onChange={() =>
               dispatch({
                 type: "TOGGLE_ENTIRE_GROUP",
-                payload: { group },
+                payload: { group, projects: Object.keys(state[group]) },
```
scarlettperry, line 143: since we may sometimes render out filtered state data, we should only toggle what's shown in the group list. without it, we toggle the status of hidden objects since we currently toggle everything that is in the original state

danielhochman (approved):
i think it might be more efficient to compute the projects that should be displayed rather than computing the hidden projects. then you only need to loop over checked projects and push upstreams and downstreams to the list rather than looping over projects, then the dependency mappings as well, then removing upstream and downstreams.

however i don't think it's any less readable and the perf difference is probably negligible so ship it! great work!

PR #1595, opened 2021-07-12
experimentation: Query by completed status

bwitt:
### Description
Allow for querying for just completed experiments

### GitHub Issue

Fixes #


bwitt:
Need to add support for this in 
https://github.com/lyft/clutch/blob/main/backend/service/chaos/experimentation/experimentstore/storer.go#L178
also

PR #1599, opened 2021-07-13
frontend: use typography in cards and add card header

dschaller:
### Description

Add `CardHeader` component and use Typography component throughout card

### Testing Performed

manual


danielhochman (changes requested):
Storybook is crashing when viewing the Card component with this change.

danielhochman (approved):

PR #1602, opened 2021-07-13
project selector: add new action type to hide upstreams/downstreams

scarlettperry:
### Description
PR adds support for hiding upstreams/downstreams of a user-owned/custom project when the eye icon is clicked

### Testing Performed
Locally

### TODOs
- [ ] add a new field in ProjectState to designate whether a project's dependencies have been hidden. In this we, we can support the next todo below
- [ ] if a dependency is shared by all other projects that have already hidden their other dependencies, then hide this dependency as well
- [ ] when a new project is added, check if any of its dependencies have been hidden and unhide them

### DIFF
--- a/frontend/workflows/projectSelector/src/helpers.tsx
+++ b/frontend/workflows/projectSelector/src/helpers.tsx
@@ -1,7 +1,8 @@
 import * as React from "react";
 import _ from "lodash";
 
-import type { Action, Group, ProjectState, State } from "./types";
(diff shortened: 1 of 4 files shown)


danielhochman:
let's see if we can mock this up first, i'm not sure this will be any less confusing than the original plan and i'd hate to spend a bunch of time implementing it if it turns out not to be the final design.

Issue #33, opened 2020-07-18
gateway: prometheus exporter


danielhochman:
**Description**

https://github.com/uber-go/tally already has a built-in prometheus reporter, just need to add a config hook to https://github.com/lyft/clutch/blob/master/api/config/gateway/v1/gateway.proto and wire it up in the gateway initialization code with some changes to the mux/handler.

Complexity [S/M/L]: S

**Wireframes, Mockups, or Screenshots (if applicable)**

PR #1605, opened 2021-07-14
backend: Add prometheus stats reporter

jeffherald:
### Description
This PR adds the capability to expose stats through a Prometheus endpoint.

### Testing Performed
- Set `gateway.stats` to use `prometheus_reporter` in `clutch-config.yaml`
- `make backend-dev`
- `curl http://localhost:8080/metrics`
### GitHub Issue

Fixes #
https://github.com/lyft/clutch/issues/33


mikecutalo:
📜

backend/gateway/stats/prometheus_reporter.go
```diff
@@ -0,0 +1,13 @@
+package stats
+
+import (
+	gatewayv1 "github.com/lyft/clutch/backend/api/config/gateway/v1"
+	tally_prom "github.com/uber-go/tally/prometheus"
```
mikecutalo, line 5: nit: I would stay away from using `_` in variable names.

backend/gateway/mux/mux.go
```diff
@@ -229,7 +229,7 @@ func customErrorHandler(ctx context.Context, mux *runtime.ServeMux, m runtime.Ma
 	runtime.DefaultHTTPErrorHandler(ctx, mux, m, w, req, err)
 }
 
-func New(unaryInterceptors []grpc.UnaryServerInterceptor, assets http.FileSystem, gatewayCfg *gatewayv1.GatewayOptions) (*Mux, error) {
+func New(unaryInterceptors []grpc.UnaryServerInterceptor, assets http.FileSystem, gatewayCfg *gatewayv1.GatewayOptions, metricsHandler http.Handler) (*Mux, error) {
```
mikecutalo, line 232: nit: lets leave gateway options at the end of the func signature.

backend/gateway/mux/mux.go
```diff
@@ -275,6 +275,16 @@ func New(unaryInterceptors []grpc.UnaryServerInterceptor, assets http.FileSystem
...
 
+	_, ok := gatewayCfg.Stats.Reporter.(*gatewayv1.Stats_PrometheusReporter_)
+	if ok && metricsHandler != nil {
+		metricsPath := "/metrics"
+		promCfg := gatewayCfg.Stats.GetPrometheusReporter()
+		if promCfg.HandlerPath != "" {
+			metricsPath = promCfg.HandlerPath
+		}
+		httpMux.Handle(metricsPath, metricsHandler)
+	}
```
mikecutalo, line 286: How do you want to handle the error case here? 

If we are expecting a metrics handler to be configured but its the incorrect type will fail silently, I think will want to bail out here and fail to start up the application, instead of finding out at runtime the application is not emitting metrics.

mikecutalo (commented):
Looks awesome! Just left a few comments!

backend/gateway/stats/prometheus_reporter.go
jeffherald: fixed

backend/gateway/mux/mux.go
jeffherald: done

backend/gateway/mux/mux.go
jeffherald: Changed to return an error when metrics handler isn't nil and config isn't prometheus

mikecutalo (approved):
lgtm!

PR #1614, opened 2021-07-16
dash: use a dash context updated by selector

danielhochman:
### Description
Instead of directly exposing the selector's context and all of its internal implementation details, we add a new dash context that will be used to track the selected projects and eventually events for the timeline view.

### Testing Performed
Integration with private card.

### TODOs
- [ ] Test with private card.

### DIFF
--- a/frontend/workflows/projectSelector/src/dash.tsx
+++ b/frontend/workflows/projectSelector/src/dash.tsx
@@ -1,35 +1,8 @@
 import * as React from "react";
+
+import { DashDispatchContext, DashStateContext } from "./dash-hooks";
 import ProjectSelector from "./project-selector";
-import type { DashState } from "./types";
-
-type DashActionKind = "UPDATE_SELECTED";
-
-interface DashAction {
-  type: DashActionKind;
-  payload: DashState;
-}
-
-const StateContext = React.createContext<DashState | undefined>(undefined);
-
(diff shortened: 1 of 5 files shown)


scarlettperry (approved):
looks great! tested it with private card

frontend/workflows/projectSelector/src/dash.tsx
```diff
@@ -0,0 +1,35 @@
...
+};
+
+const Dash = ({ children }) => {
+  const [state, dispatch] = React.useReducer(dashReducer, initialState);
+
+  return (
+    <DashDispatchContext.Provider value={dispatch}>
+      <DashStateContext.Provider value={state}>
+        <ProjectSelector />
+        <div>{children}</div>
```
scarlettperry, line 29: card has grid styling so this div affects that
```suggestion
        {children}
```

danielhochman: hmm i did this so the cards wouldn't be siblings with the selector

scarlettperry: ah ok. we can make style updates to the card

danielhochman: i probably need to flex grow this div, let me try that

PR #1624, opened 2021-07-20
proxymodule: fix request body propagation 

mikecutalo:
### Description
There were two issues here, we did not propagate the request body to the API request and when populating the response headers they were appended instead of using the pre allocated space which result in nil  `google.protobuf.Value`, this would fail some gateway validation and you would get the error below.

```
{
    "code": 2,
    "message": "proto: google.protobuf.Value: none of the oneof fields is set",
    "details": []
}
```

### Testing Performed
Manual


backend/module/proxy/proxy.go
```diff
@@ -89,6 +92,16 @@ func (m *mod) RequestProxy(ctx context.Context, req *proxyv1.RequestProxyRequest
...
 
+	var requestBody io.ReadCloser
+	if req.Request != nil {
+		requestJSON, err := protojson.Marshal(req.Request)
+		if err != nil {
+			return nil, err
+		}
+		buff := bytes.NewBuffer(requestJSON)
+		requestBody = ioutil.NopCloser(buff)
+	}
```
danielhochman, line 103: move this below request construction and then you can directly set request.Body instead of maintaining the local var

danielhochman (approved):
lgtm w/ small nit, if you can add any tests easily that would be good too

PR #1626, opened 2021-07-20
backend: github service: add User object to response

smonero:
### Description

It is useful to have some of the other fields that are returned by the go githubv3 API such as avatarURL and ID

### GitHub Issue

Fixes #

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -364,8 +364,7 @@ func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA str
 type Commit struct {
 	Files   []*githubv3.CommitFile
 	Message string
-	Author  *githubv3.CommitAuthor
-	User    *githubv3.User
+	Author  *githubv3.User
 }
 
 func (s *svc) GetCommit(ctx context.Context, ref *RemoteRef) (*Commit, error) {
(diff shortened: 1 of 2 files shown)


backend/service/github/github.go
```diff
@@ -362,21 +362,32 @@ func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA str
...
-	return &Commit{
+	retCommit := &Commit{
 		Files:   commit.Files,
 		Message: commit.GetCommit().GetMessage(),
 		Author:  commit.GetCommit().GetAuthor(),
-	}, nil
+	}
+
+	if commit.Author != nil {
+		if commit.Author.AvatarURL != nil {
```
dschaller, line 384: I think what we want is probably to either distinguish between the github author vs git author or we should just assume that we want the github author and remove the `GetCommit` call above and return the author object to use later, removing the need to have the avatar and id fields here.

backend/service/github/github.go
```diff
@@ -376,6 +377,7 @@ func (s *svc) GetCommit(ctx context.Context, ref *RemoteRef) (*Commit, error) {
 		Files:   commit.Files,
 		Message: commit.GetCommit().GetMessage(),
 		Author:  commit.GetCommit().GetAuthor(),
+		User:    commit.Author,
```
dschaller, line 380: Should we use `GetAuthor` here?

dschaller (commented):
Looks good, I think the only last thing to add would be to add docstring on the difference between author and user :)

backend/service/github/github.go
```diff
@@ -364,18 +364,22 @@ func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA str
...
 }
 
 func (s *svc) GetCommit(ctx context.Context, ref *RemoteRef) (*Commit, error) {
 	commit, _, err := s.rest.Repositories.GetCommit(ctx, ref.RepoOwner, ref.RepoName, ref.Ref)
 	if err != nil {
 		return nil, err
 	}
+	// Distinction between author and commiter:
+	// The author is the person who originally wrote the code. The committer
+	// is assumed to be the person who committed the code on behalf of the original author.
```
dschaller, line 377: I think we can remove this now that we only have one author field
```suggestion
```

dschaller (approved):
lg once the comment is removed

PR #1635, opened 2021-07-23
backend: github service: add commitRange() from comparison API

smonero:
### Description

We need to get a list of commits via the comparison API.

### Testing Performed

unit

### GitHub Issue

Fixes #

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -28,7 +28,6 @@ import (
 	"google.golang.org/grpc/status"
 
 	githubv1 "github.com/lyft/clutch/backend/api/config/service/github/v1"
-	scgithubv1 "github.com/lyft/clutch/backend/api/sourcecontrol/github/v1"
 	sourcecontrolv1 "github.com/lyft/clutch/backend/api/sourcecontrol/v1"
 	"github.com/lyft/clutch/backend/service"
 )
@@ -79,7 +78,7 @@ type Client interface {
 	CreatePullRequest(ctx context.Context, ref *RemoteRef, base, title, body string) (*PullRequestInfo, error)
 	CreateRepository(ctx context.Context, req *sourcecontrolv1.CreateRepositoryRequest) (*sourcecontrolv1.CreateRepositoryResponse, error)
 	CreateIssueComment(ctx context.Context, ref *RemoteRef, number int, body string) error
-	CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA string) (*scgithubv1.CommitComparison, error)
+	CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA string) (*githubv3.CommitsComparison, error)
 	GetCommit(ctx context.Context, ref *RemoteRef) (*Commit, error)
 	GetRepository(ctx context.Context, ref *RemoteRef) (*Repository, error)
 	GetOrganization(ctx context.Context, organization string) (*githubv3.Organization, error)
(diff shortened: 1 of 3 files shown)


backend/service/github/github.go
```diff
@@ -361,6 +361,15 @@ func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA str
 	}, nil
 }
 
+func (s *svc) ListCommitsViaComparison(ctx context.Context, ref *RemoteRef, compareSHA string) ([]*githubv3.RepositoryCommit, error) {
```
dschaller, line 364: Can you add docstring to this? Also, this might be a nit but I think `CompareCommits` or `CommitRange` might be a better name. What do you think?

backend/service/github/github.go
```diff
@@ -361,6 +361,15 @@ func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA str
...
+	comp, _, err := s.rest.Repositories.CompareCommits(ctx, ref.RepoOwner, ref.RepoName, compareSHA, ref.Ref)
+	if err != nil {
+		return nil, fmt.Errorf("could not get commits from comparison for %s and %s. %+v", ref.Ref, compareSHA, err)
```
dschaller, line 367: Should we include the repo name and owner in this?

backend/service/github/github.go
```diff
@@ -361,6 +361,20 @@ func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA str
 	}, nil
 }
 
+/*
+ * CommitRange() is used to get multiple commits between 2 SHAs. It is useful because we can do one API call and get a bunch
+ * of commits, rather than calling GetCommit() multiple times.
+ * compareSHA is the base (parent) while the ref.ref is the head
+ */
+func (s *svc) CommitRange(ctx context.Context, baseSHA string, ref *RemoteRef) ([]*githubv3.RepositoryCommit, error) {
```
danielhochman, line 369: should we just add this data to CompareCommits?

smonero: idk, should we? I guess the reason why I didn't was because I wanted to make it like other helper functions (GetCommit)  that didn't have protos but rather just the struct definitions in the file.  What should be the standard way moving forward? (protos, or not protos?) 

I wasn't sure if I should take over the `scgithubv1.CommitComparison` proto message.  If I put it in there I would need the github commit proto definiton (or recreate some version of it) which I thought was awkward.

smonero: I realized after digging that its actually used in the envoydeploys module which seems to not be used anymore (was made by smercatoris who has since left).  Should we delete all of that as well?

smonero:
Note this will temporarily break private when merged.

backend/service/github/github.go
```diff
@@ -345,20 +344,15 @@ func (s *svc) GetFile(ctx context.Context, ref *RemoteRef, path string) (*File,
 	return f, nil
 }
 
-func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA string) (*scgithubv1.CommitComparison, error) {
+/*
+ * Rather than calling GetCommit() multiple times, we can use CompareCommits to get a range of commits
+ */
+func (s *svc) CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA string) (*githubv3.CommitsComparison, error) {
```
dschaller, line 350: The only concern I have about changing the return type is that it'll break something but I do think this type makes more sense

smonero: Another option is to just go back to the way I had it before (with a separate function) and just leave a TODO to maybe combine the 2 functions.  I thought about adding to the proto as well - the reason why I didn't do that was I didn't see a good way to import all the fields from the github structs in proto form.

danielhochman (approved):

PR #1645, opened 2021-07-26
frontend core: add workflow card framework

scarlettperry:
### GitHub Issue

Fixes #


frontend/packages/core/src/Cards/landing-card.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from "react";
```
scarlettperry, line 1: no changes made to the landing card. copy & paste of what was in `card.tsx`

frontend/packages/core/src/Cards/workflow-card.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from "react";
```
scarlettperry, line 1: no changes made here other than renaming it to `WorkflowCard`

dschaller: Should we include this in the dash package instead of core?

scarlettperry: moved to projectSelector pkg. 

side note: was going to update pkg name to dash for clarity but realized that would cause conflicts in the internal repo as the pkg there is also called dash

frontend/workflows/projectSelector/src/card.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Grid } from "@material-ui/core";
+
+interface CardProps {
+  avatar?: React.ReactNode;
+  title?: React.ReactNode & string;
+  error?: ClutchError;
+  children: React.ReactNode;
+}
+
+const Card = ({ avatar, title, error, children }: CardProps) => (
```
mikecutalo, line 13: Lets add types to the params

scarlettperry: the param interface is above `CardProps` and we we type it here `{ avatar, title, error, children }: CardProps`

mikecutalo: Ah my bad this is a destructuring, i thought it was the return type lol.

mikecutalo (approved):

PR #1648, opened 2021-07-28
frontend: add tooltip to core design & project selector

scarlettperry:
### Description
PR
* adds tooltip to core design system
* adds tooltip stories
* adds tooltip to project selector component 
* makes small change to typography component to accept a color prop

### Testing Performed
locally and storybook

Screen Shot 2021-07-28 at 5 15 51 PM

Screen Shot 2021-07-28 at 5 28 50 PM

Screen Shot 2021-07-28 at 5 28 35 PM

### TODOs
- [x] styling for distance b/w tooltip and reference element

### DIFF
--- a/frontend/packages/core/src/Feedback/tooltip.tsx
+++ b/frontend/packages/core/src/Feedback/tooltip.tsx
@@ -7,29 +7,25 @@ const BaseTooltip = ({ className, ...props }: MuiTooltipProps) => (
   <MuiTooltip classes={{ tooltip: className }} {...props} />
 );
 
-const StyledTooltip = styled(BaseTooltip)(
-  {
-    backgroundColor: "#0D1030",
-    borderRadius: "6px",
-    "&.MuiTooltip-tooltipPlacementLeft": {
-      margin: "0 2px",
-    },
-    "&.MuiTooltip-tooltipPlacementRight": {
-      margin: "0 2px",
-    },
-    "&.MuiTooltip-tooltipPlacementTop": {
-      margin: "2px 0",
-    },
-    "&.MuiTooltip-tooltipPlacementBottom": {
-      margin: "2px 0",
-    },
+const StyledTooltip = styled(BaseTooltip)((props: { maxWidth: string }) => ({
+  maxWidth: props.maxWidth,
+  backgroundColor: "#0D1030",
+  borderRadius: "6px",
+  "&.MuiTooltip-tooltipPlacementLeft": {
+    margin: "0 2px",
   },
-  props => ({
-    maxWidth: props["data-max-width"],
-  })
-);
+  "&.MuiTooltip-tooltipPlacementRight": {
+    margin: "0 2px",
+  },
+  "&.MuiTooltip-tooltipPlacementTop": {
+    margin: "2px 0",
+  },
+  "&.MuiTooltip-tooltipPlacementBottom": {
+    margin: "2px 0",
+  },
+}));
 
-export interface TooltipProps extends Pick<MuiTooltipProps, "placement"> {
+export interface TooltipProps extends Pick<MuiTooltipProps, "interactive" | "placement"> {
   // tooltip reference element (i.e. icon)
   children: React.ReactElement;
   // material ui default is 300px
@@ -40,7 +36,7 @@ export interface TooltipProps extends Pick<MuiTooltipProps, "placement"> {
 
 const Tooltip = ({ children, maxWidth = "300px", title, ...props }: TooltipProps) => {
   return (
-    <StyledTooltip title={title} data-max-width={maxWidth} {...props}>
+    <StyledTooltip title={title} maxWidth={maxWidth} {...props}>
       {children}
     </StyledTooltip>
   );
--- a/frontend/packages/core/src/Feedback/tooltip.tsx
+++ b/frontend/packages/core/src/Feedback/tooltip.tsx
@@ -7,6 +7,7 @@ const BaseTooltip = ({ className, ...props }: MuiTooltipProps) => (
   <MuiTooltip classes={{ tooltip: className }} {...props} />
 );
 
+// TODO: sync with Design on margins for each possible placement
 const StyledTooltip = styled(BaseTooltip)((props: { maxWidth?: string }) => ({
   maxWidth: props.maxWidth,
   backgroundColor: "#0D1030",
--- a/frontend/packages/core/src/Feedback/tooltip.tsx
+++ b/frontend/packages/core/src/Feedback/tooltip.tsx
@@ -10,10 +10,19 @@ const BaseTooltip = ({ className, ...props }: MuiTooltipProps) => (
 const StyledTooltip = styled(BaseTooltip)(
   {
     backgroundColor: "#0D1030",
-    border: "1px solid rgba(13, 16, 48, 0.1)",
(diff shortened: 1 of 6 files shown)


frontend/packages/core/src/typography.tsx
```diff
@@ -118,25 +118,26 @@ type TextVariant =
...
 
 export interface TypographyProps {
   variant: TextVariant;
   children: React.ReactNode;
+  color?: string;
 }
 
-const Typography = ({ variant, children }: TypographyProps) => {
-  return <StyledTypography variant={variant}>{children}</StyledTypography>;
+const Typography = ({ variant, children, color = "#0D1030" }: TypographyProps) => {
```
scarlettperry, line 135: i could have just inline styled the font color in the import component but I think users would want to be able to change the font color, so I added a color prop

frontend/packages/core/src/Feedback/tooltip.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { TooltipProps as MuiTooltipProps } from "@material-ui/core";
+import { Tooltip as MuiTooltip } from "@material-ui/core";
+
+const BaseTooltip = ({ className, ...props }: MuiTooltipProps) => (
```
scarlettperry, line 6: the tooltip component accepts a popper component, so initially I was styling the tooltip by targeting it through the popper. however, things got complicated when I wanted to make the maxWidth prop-based as  I'd have to pass the prop from tooltip --> popper component. I was getting various errors and it felt messy. 

Found this pattern here https://stackoverflow.com/questions/59934683/style-material-ui-tooltip-using-emotion-styled and we also do something similar for our Select component https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/Input/select.tsx#L61-L78

frontend/packages/core/src/Feedback/tooltip.tsx
```diff
@@ -0,0 +1,54 @@
...
+  children: React.ReactElement;
+  // material ui default is 300px
+  maxWidth?: string;
+  // tooltip text
+  title: React.ReactNode;
+}
+
+const Tooltip = ({ children, maxWidth = "300px", title, ...props }: TooltipProps) => {
+  return (
+    <StyledTooltip title={title} data-max-width={maxWidth} {...props}>
```
scarlettperry, line 43: title prop supports html https://material-ui.com/components/tooltips/#customized-tooltips

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -171,9 +177,42 @@ const ProjectSelector = () => {
...
+                    <Typography variant="body3" color="#E7E7EA">
+                      {downstreamDescription}
+                    </Typography>
+                  </TooltipContainer>
+                </>
+              }
+              placement="right-start"
+              maxWidth="400px"
+            >
+              <InfoOutlinedIcon />
```
danielhochman, line 214: icon is too large

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -171,9 +177,42 @@ const ProjectSelector = () => {
...
+              title={
+                <>
+                  <TooltipContainer>
+                    <Typography variant="subtitle3" color="#FFFFFF">
+                      Projects
+                    </Typography>
+                    <Typography variant="body3" color="#E7E7EA">
+                      {projectDescription}
+                    </Typography>
+                  </TooltipContainer>
```
danielhochman, line 192: Actually I'd probably turn this into a loop

danielhochman (commented):
Looks great! A few comments:
- The tooltip needs to be offset vertically a bit upwards. The top is not be directly inline with the icon in the design.
- The tooltip should not disappear when moving the mouse to it so it's possible to select text, click a link or a button within the tooltip.

frontend/packages/core/src/Feedback/tooltip.tsx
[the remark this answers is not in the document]
scarlettperry: thxs, updated in typography component too

frontend/workflows/projectSelector/src/project-selector.tsx
scarlettperry: from the designs, the 2 icons seem to be the same size so I made them both smaller
Screen Shot 2021-07-30 at 12 56 55 PM

The design specifies 18px but it looks too small IMO
Screen Shot 2021-07-30 at 12 50 46 PM

I brought them down to `fontSize=small` instead which is 20px. LMK what you think
Screen Shot 2021-07-30 at 12 50 54 PM

frontend/workflows/projectSelector/src/project-selector.tsx
scarlettperry: turned into a loop

scarlettperry:
@danielhochman  RE 

> The tooltip needs to be offset vertically a bit upwards. The top is not be directly inline with the icon in the design.

I have margin styling w.r.t to how close a tooltip should be to it's reference element but since the tooltip has 12 placement choices https://material-ui.com/components/tooltips/#positioned-tooltips, I've debated making additional changes to where the tooltip should lie since the Figmas don't cover all the placement options. Alternatively though, we can limit the placmement options to just `top`, `bottom`, `left`, and `right`. Wdyt?

frontend/packages/core/src/Feedback/tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "&.MuiTooltip-tooltipPlacementTop": {
+    margin: "2px 0",
+  },
+  "&.MuiTooltip-tooltipPlacementBottom": {
+    margin: "2px 0",
+  },
+}));
+
+export interface TooltipProps extends Pick<MuiTooltipProps, "interactive" | "placement"> {
```
scarlettperry, line 28: fyi `interactive` prop is for making the tooltip not disappear when moving the mouse to it

danielhochman (approved):
we can do the margin fix in TODO and clarify positioning with kanchan

frontend/packages/core/src/typography.tsx
```diff
@@ -118,25 +118,26 @@ type TextVariant =
...
-    fontSize: `${STYLE_MAP[props?.variant || 0].size}px`,
-    fontWeight: STYLE_MAP[props?.variant || 0].weight,
-    lineHeight: `${STYLE_MAP[props?.variant || 0].lineHeight}px`,
-    ...STYLE_MAP[props?.variant || 0].props,
-  })
-);
+const StyledTypography = styled.div<{ color?: string; variant: TextVariant }>(props => ({
+  color: props.color,
+  fontSize: `${STYLE_MAP[props?.variant || 0].size}px`,
+  fontWeight: STYLE_MAP[props?.variant || 0].weight,
```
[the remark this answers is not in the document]
scarlettperry: yeah the `||` shouldn't be necessary. Will open a sep PR now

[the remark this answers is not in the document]
scarlettperry: oh actually, in looking at the code more closely, maybe it makes sense to keep the `||` case. It's not possible to set the prop to an undefined variant but if someone creates a new one (ie `caption4`) and doesn't define some of these css properties, then it's value is `undefinedpx` (tested in the browser). From local testing, it doesn't break anything but I think @dschaller added it as a safety check

PR #1650, opened 2021-07-28
frontend: move project selector to strict typescript

dschaller:
### Description

Enable strict typescript rules for project selector and make appropriate fixes.

### Testing Performed

manual

### DIFF
--- a/frontend/workflows/projectSelector/src/dash-hooks.tsx
+++ b/frontend/workflows/projectSelector/src/dash-hooks.tsx
@@ -4,8 +4,8 @@ import type { DashAction, DashState } from "./types";
 
 export const DashStateContext = React.createContext<DashState | undefined>(undefined);
 
-export const DashDispatchContext = React.createContext<(action: DashAction) => void | undefined>(
-  () => undefined
+export const DashDispatchContext = React.createContext<((action: DashAction) => void) | undefined>(
+  undefined
 );
 
 type useDashUpdaterReturn = {
@@ -17,7 +17,7 @@ export const useDashUpdater = (): useDashUpdaterReturn => {
 
   return {
     updateSelected: projects => {
-      dispatch({ type: "UPDATE_SELECTED", payload: projects });
+      dispatch && dispatch({ type: "UPDATE_SELECTED", payload: projects });
     },
   };
 };
--- a/frontend/workflows/projectSelector/src/dash-hooks.tsx
+++ b/frontend/workflows/projectSelector/src/dash-hooks.tsx
@@ -22,6 +22,6 @@ export const useDashUpdater = (): useDashUpdaterReturn => {
   };
 };
 
-export const useDashState = (): DashState => {
-  return React.useContext<DashState>(DashStateContext);
+export const useDashState = (): DashState | undefined => {
+  return React.useContext<DashState | undefined>(DashStateContext);
 };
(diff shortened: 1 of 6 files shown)


frontend/workflows/projectSelector/src/dash-hooks.tsx
```diff
@@ -5,7 +5,7 @@ import type { DashAction, DashState } from "./types";
 export const DashStateContext = React.createContext<DashState | undefined>(undefined);
 
 export const DashDispatchContext = React.createContext<(action: DashAction) => void | undefined>(
-  undefined
+  () => undefined
 );
```
danielhochman, line 9:
```suggestion
export const DashDispatchContext = React.createContext<((action: DashAction) => void) | undefined>(
  undefined
);
```

frontend/workflows/projectSelector/src/helpers.tsx
```diff
@@ -16,17 +16,15 @@ const useReducerState = () => {
   return React.useContext(StateContext);
 };
 
-const DispatchContext = React.createContext<(action: Action) => void | undefined>(undefined);
+const DispatchContext = React.createContext<(action: Action) => void | undefined>(() => undefined);
```
danielhochman, line 19:
```suggestion
const DispatchContext = React.createContext<((action: Action) => void) | undefined>(undefined);
```

frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -116,8 +116,9 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
 
   const [collapsed, setCollapsed] = React.useState(false);
 
-  const numProjects = Object.keys(state[group]).length;
-  const checkedProjects = Object.keys(state[group]).filter(k => state[group][k].checked);
+  const groupKeys = Object.keys(state?.[group] || {});
```
danielhochman, line 119: nit

```suggestion
  const groupKeys = Object.keys(state?.[group] ?? {});
```

frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -153,53 +154,51 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
...
-                  name={key}
-                  size="small"
-                  disabled={state.loading}
-                  onChange={() =>
+          {groupKeys.sort().map(key => (
+            <StyledMenuItem key={key}>
+              <Checkbox
+                name={key}
+                size="small"
+                disabled={state?.loading}
```
danielhochman, line 162: i think you can eliminate the added `state?` checks by throwing an error from the hook and never returning an undefined state, which should never happen

danielhochman (commented):
first round

frontend/workflows/projectSelector/src/project-group.tsx
dschaller: If we don't expect state to ever be undefined should we change the initialization of it (and types) to not allow for undefined?

scarlettperry: @dschaller , just for my understanding, do you mean something like this?
```
const StateContext = React.createContext({
  [Group.PROJECTS]: {},
  [Group.UPSTREAM]: {},
  [Group.DOWNSTREAM]: {},
  projectData: {},
  loading: false,
  error: undefined,
} as State);
```
we also initialize it with default values in the project selector component in `useReducer`

scarlettperry (approved):

frontend/workflows/projectSelector/src/project-group.tsx
dschaller: Chatted IRL with @scarlettperry and we are going to punt on removing undefined as a type until later.

mikecutalo (approved):

PR #1652, opened 2021-07-29
layout: independent scrolling

danielhochman:
### Description

Before
Peek 2021-07-29 15-29

After
Peek 2021-07-29 15-32


scarlettperry (approved):
played w it locally & lgtm!

frontend/packages/core/src/AppLayout/index.tsx
```diff
@@ -15,9 +15,11 @@ const AppGrid = styled(MuiGrid)({
 
 const ContentGrid = styled(MuiGrid)({
   flex: 1,
-  overflow: "hidden",
+  maxHeight: "calc(100vh - 64px)",
```
dschaller, line 18: super nit: can we export the height from the header and consume it in both spots?

dschaller (approved):

PR #1654, opened 2021-07-29
frontend: fix width layout of dash content

dschaller:
### Description

Fixes width of Dash content layout. Previously it would overflow given the 100% width.

Screen Shot 2021-07-29 at 2 50 35 PM

### Testing Performed

manual

### DIFF
--- a/frontend/workflows/projectSelector/src/card.tsx
+++ b/frontend/workflows/projectSelector/src/card.tsx
@@ -11,7 +11,7 @@ interface CardProps {
 }
 
 const Card = ({ avatar, title, error, children }: CardProps) => (
-  <Grid item xs={12} sm={7}>
+  <Grid item xs={12} sm={6}>
     <ClutchCard>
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectSelector/src/card.tsx
```diff
@@ -11,7 +11,7 @@ interface CardProps {
 }
 
 const Card = ({ avatar, title, error, children }: CardProps) => (
-  <Grid item xs={12} sm={7}>
+  <Grid item xs={12} sm={6}>
```
mikecutalo, line 14: The max size of the grid is 12 right? Giving each card 6 width to fill up? And below does `flex={1}` for both box's just fill that 6 width up?

dschaller: yeah, we want them to display side by side. The flex below for the second box fills up the container.

scarlettperry (approved):

mikecutalo (approved):

PR #1656, opened 2021-07-30
dash: check for projectData before calling update hook

danielhochman:
### Description
Fixes bug where projectData would not be available the first time a project was selected.

### Testing Performed
Manual.

### DIFF
--- a/frontend/workflows/projectSelector/src/project-selector.tsx
+++ b/frontend/workflows/projectSelector/src/project-selector.tsx
@@ -57,6 +57,20 @@ const StyledProgressContainer = styled.div({
   },
 });
 
+
+// Determines if every project has projectData (i.e. the effect has finished fetching the data)
+const allPresent = (state: State): boolean => {
+  let ret = true;
+  // We could potentially check all groups but it's really not necessary since an upstream cannot be added manually.
+  _.forEach(Object.keys(state[Group.PROJECTS]), p => {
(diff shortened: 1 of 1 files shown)


frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -57,6 +57,19 @@ const StyledProgressContainer = styled.div({
...
 
+// Determines if every project has projectData (i.e. the effect has finished fetching the data)
+const allPresent = (state: State): boolean => {
+  let ret = true;
+  // We could potentially check all groups but it's really not necessary since an upstream cannot be added manually.
+  _.forEach(Object.keys(state[Group.PROJECTS]), p => {
+    if (!(p in state.projectData)) {
+      ret = false;
+    }
+    return ret; // Will stop iteration early if false encountered.
```
mikecutalo, line 68: Wont this only iterate once and then return? Also can we remove the usage of `ret` and just return when necessary?

danielhochman: nope when it returns true iteration continues. can't return from the closure, need an outer scoped var to track result.

mikecutalo: Ah ok i see now thanks.

mikecutalo (approved):

PR #1657, opened 2021-07-30
dynamodb: add UpdateTableCapacity endpoint

angethuy:
### Description

This PR implements the UpdateTableCapacity endpoint for the AWS Dynamodb service. UpdateTableCapacity will perform an update action on a specified table's provisioned capacity, checking against limits for safety.

### Testing Performed

Unit and manual tests.


backend/service/aws/dynamodb.go
```diff
@@ -6,18 +6,23 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 
 	dynamodbv1 "github.com/lyft/clutch/backend/api/aws/dynamodb/v1"
 )
 
+const maxDynamodbReadCapacityUnits = 35000
```
rostams-lyft, line 15: Where these values are coming from?

angethuy: Good call out — they're set as constants in the jenkins' job that I'm recreating, but I should probably pull them out into config variables instead so users can set their own limits per their needs

backend/module/dynamodb/dynamodb.go
```diff
@@ -30,7 +30,12 @@ func (a *dynamodbAPI) DescribeTable(ctx context.Context, req *dynamodbv1.Describ
 }
 
 func (a *dynamodbAPI) UpdateTableCapacity(ctx context.Context, req *dynamodbv1.UpdateTableCapacityRequest) (*dynamodbv1.UpdateTableCapacityResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "not implemented")
+	err := a.client.UpdateTableCapacity(ctx, req.Region, req.TableName, req.TargetTableRcu, req.TargetTableWcu)
+	if err != nil {
+		return nil, err
+	}
+
+	return &dynamodbv1.UpdateTableCapacityResponse{}, nil
```
mikecutalo, line 38: Is the response going to echo back the table changes?

angethuy: Yup it'll echo back the table name and the target capacities

backend/service/aws/aws.go
```diff
@@ -119,13 +131,26 @@ type client struct {
...
+type dynamodbConfig struct {
+	scalingLimits scalingLimits
+}
+
+type scalingLimits struct {
+	maxReadCapacityUnits  int64
+	maxWriteCapacityUnits int64
+	maxScaleFactor        float32
+	enableOverride        bool
+}
```
mikecutalo, line 152: Will want to use our protos as the source of truth for this data structure, lets not redefined them in code here.

backend/module/dynamodb/dynamodb.go
mikecutalo: We still going to do the echo back here? Or will that come in a later PR?

backend/service/aws/dynamodb.go
```diff
@@ -2,22 +2,46 @@ package aws
...
 
 	dynamodbv1 "github.com/lyft/clutch/backend/api/aws/dynamodb/v1"
+	awsv1 "github.com/lyft/clutch/backend/api/config/service/aws/v1"
 )
 
+// defaults for the dynamodb settings config
+const (
+	AwsMaxRCU       = int64(40000)
+	AwsMaxWCU       = int64(40000)
+	SafeScaleFactor = float32(2.0)
```
mikecutalo, line 21: In a follow up PR lets make sure to comment the defaults in the configuration proto for the reader.
https://github.com/lyft/clutch/blob/0a6e4a00e72b8bb3a976f801d02f9550adf447d7/api/config/service/aws/v1/aws.proto#L26-L27

backend/service/aws/dynamodb.go
```diff
@@ -57,3 +86,60 @@ func newProtoForGlobalSecondaryIndex(index types.GlobalSecondaryIndexDescription
...
+		return status.Errorf(codes.FailedPrecondition, fmt.Sprintf("Target write capacity exceeds the scale limit of [%.1f]x current capacity", client.dynamodbCfg.ScalingLimits.MaxScaleFactor))
+	}
+	return nil
+}
+
+func (c *client) UpdateTableCapacity(ctx context.Context, region string, tableName string, targetTableRcu int64, targetTableWcu int64) error {
+	cl, err := c.getRegionalClient(region)
+	if err != nil {
+		return err
+	}
```
mikecutalo, line 121:
```suggestion
	if err != nil {
	        c.log.Error("unable to get regional client", zap.Error(err))
		return err
	}
```

lets make sure to log all system errors like these, same applies for other error cases throughout the pr.

backend/module/dynamodb/dynamodb.go
angethuy: I'll add in a separate PR since I'll also have to add the echo fields to the [dynamodb api proto](https://github.com/lyft/clutch/blob/0a6e4a00e72b8bb3a976f801d02f9550adf447d7/api/aws/dynamodb/v1/dynamodb.proto#L78) too right?

mikecutalo: Ah yeah that sounds good, I assumed they were already on the proto.

mikecutalo (approved):

backend/service/aws/dynamodb.go
angethuy: Will follow up with updated proto file that includes comments and echo fields

backend/service/aws/dynamodb.go
```diff
@@ -57,3 +89,63 @@ func newProtoForGlobalSecondaryIndex(index types.GlobalSecondaryIndexDescription
 		ProvisionedThroughput: currentCapacity,
 	}
 }
+
+func isValidIncrease(client *regionalClient, current *types.ProvisionedThroughputDescription, target types.ProvisionedThroughput) error {
+	// check for targets that are lower than current (can't scale down)
+	if *current.ReadCapacityUnits > *target.ReadCapacityUnits {
+		return status.Errorf(codes.FailedPrecondition, fmt.Sprintf("Target read capacity [%d] is lower than current capacity [%d]", *target.ReadCapacityUnits, *current.ReadCapacityUnits))
```
danielhochman, line 96: nit: since you're using errorf you don't need sprintf, e.g.

```suggestion
		return status.Errorf(codes.FailedPrecondition, "Target read capacity [%d] is lower than current capacity [%d]", *target.ReadCapacityUnits, *current.ReadCapacityUnits)
```

backend/service/aws/dynamodb_test.go
```diff
@@ -110,11 +124,133 @@ func TestDescribeTableWithGsiValid(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			got := c.UpdateTableCapacity(context.Background(), "us-east-1", "test-gsi-table", tt.inputRCU, tt.inputWCU)
+			if got.Error() != tt.want {
+				t.Errorf("\nWant error msg: %s\nGot error msg: %s", tt.want, got)
+			}
+		})
+	}
+}
+
+func TestUpdateTableCapacityWithCustomLimits(t *testing.T) {
```
danielhochman, line 204: do we test the success case?

backend/service/aws/dynamodb.go
```diff
@@ -2,23 +2,50 @@ package aws
...
 
 	dynamodbv1 "github.com/lyft/clutch/backend/api/aws/dynamodb/v1"
+	awsv1 "github.com/lyft/clutch/backend/api/config/service/aws/v1"
 )
 
+// defaults for the dynamodb settings config
+const (
+	AwsMaxRCU       = int64(40000)
+	AwsMaxWCU       = int64(40000)
+	SafeScaleFactor = float32(2.0)
```
danielhochman, line 22: in general it's better to use untyped constants
```suggestion
	AwsMaxRCU       = 40000
	AwsMaxWCU       = 40000
	SafeScaleFactor = 2.0
```

danielhochman (approved):
nits

backend/service/aws/dynamodb_test.go
angethuy: Good catch, I'll add tests for success cases

PR #1665, opened 2021-08-03
api: timeSeries proto

smonero:
### Description

We need a timeseries proto to represent events that will be put on the timeline.

One thing I thought about and I don't have a good answer for is how we would prioritize / visualize overlapping events. I guess we will figure that out in the future.

### Testing Performed

none

### GitHub Issue

Fixes #


api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+
+package clutch.timeline.v1;
+
+option go_package = "github.com/lyft/clutch/backend/api/timeline/v1;timelinev1";
+
+import "google/protobuf/any.proto";
+import "validate/validate.proto";
+
+message TimeseriesData {
```
mikecutalo, line 10: Can you write a comment about what this represents?  It looks like just a single datapoint but some more color here would be helpful to the reader.

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,20 @@
...
+  string project_name = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  // For a range, both start time and end time will be present
+  // For a singular event, only start time will be filled in.
+  int64 start_time_millis = 2 [ (validate.rules).int64.gt = 0 ];
+  int64 end_time_millis = 3;
```
mikecutalo, line 15: What if the event is just a moment in time and not a range, how would handle that situation?

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,20 @@
...
+  int64 event_id = 4;
```
mikecutalo, line 16: What will this be used for?

api/timeline/v1/timeseries.proto
smonero: My plan for that was end_time_millis would be 0

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,22 @@
...
+option go_package = "github.com/lyft/clutch/backend/api/timeline/v1;timelinev1";
+
+import "google/protobuf/any.proto";
+import "validate/validate.proto";
+
+// A timeseries datapoint message is useful for organizing events
+// to be displayed in a timeline view.  Users can transform data
+// into a timeseries format and be able to organize them.
+message TimeseriesData {
+  string project_name = 1 [ (validate.rules).string = {min_bytes : 1} ];
```
dschaller, line 14: should we call this project name? It seems like time series data could be used for things even outside of projects.

api/timeline/v1/timeseries.proto
dschaller: They could also be equal. e.g. a moment is just the start and end at the same time

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,22 @@
...
+  // For a range, both start time and end time will be present
+  // For a singular event, only start time will be filled in.
+  int64 start_time_millis = 2 [ (validate.rules).int64.gt = 0 ];
+  int64 end_time_millis = 3;
+  // The plan here is to have each user of this proto to put their unique data in a proto
```
dschaller, line 19: nit: probably would change this to either be more concrete or to be written as a TODO

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,26 @@
...
+
+// A timeseries Point message is useful for organizing events
+// to be displayed in a timeline view.  Users can transform data
+// into a timeseries format and be able to organize them.
+message Point {
+  oneof event {
+    option (validate.required) = true;
+    TimeRange range = 2;
+    int64 singular_event_timestamp_millis = 3;
+  }
```
danielhochman, line 23: nit: naming
```suggestion
  oneof timestamp {
    option (validate.required) = true;
    TimeRange range = 2;
    int64 millis = 3;
  }
```

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,26 @@
...
+  google.protobuf.Any event_info = 5;
+  string event_description = 6;
```
danielhochman, line 25:
```suggestion
  google.protobuf.Any pb = 4;
  string description = 5;
```

(to more closely match https://github.com/lyft/clutch/blob/498410f57fc1b534d941ec2473e1abf3e58d5609/api/topology/v1/topology_api.proto#L224)

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,26 @@
...
+package clutch.timeseries.v1;
+
+option go_package = "github.com/lyft/clutch/backend/api/timeseries/v1;timeseriesv1";
+
+import "google/protobuf/any.proto";
+import "validate/validate.proto";
+
+message TimeRange {
+  int64 start_time_millis = 1 [ (validate.rules).int64.gt = 0 ];
+  int64 end_time_millis = 2 [ (validate.rules).int64.gt = 0 ];
```
danielhochman, line 12: nit verbosity
```suggestion
  int64 start_millis = 1 [ (validate.rules).int64.gt = 0 ];
  int64 end_millis = 2 [ (validate.rules).int64.gt = 0 ];
```

danielhochman (commented):
some naming nits, saving chars will actually pay big dividends when requesting huge timeseries sets in JSON format

mikecutalo (approved):
Looks good I dont have any additional comments atm.

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,26 @@
...
+// A timeseries Point message is useful for organizing events
+// to be displayed in a timeline view.  Users can transform data
+// into a timeseries format and be able to organize them.
+message Point {
+  oneof timestamp {
+    option (validate.required) = true;
+    TimeRange range = 2;
+    int64 millis = 3;
+  }
+  google.protobuf.Any pb = 5;
```
danielhochman, line 24: `4` is missing

danielhochman: actually so is `1`, renumber plz

api/timeline/v1/timeseries.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
```
danielhochman, line 1: folder name is still timeline, should be timeseries

PR #1667, opened 2021-08-03
dynamodb: add config variables for AWS service

angethuy:
This PR adds client settings to the config proto for the AWS service. It enables setting scaling limits for the Dynamodb scaling workflow.

### Description

Proto changes for AWS service.

### Testing Performed
N/A

### TODOs

- [x] generate api


api/config/service/aws/v1/aws.proto
```diff
@@ -9,9 +9,26 @@ import "validate/validate.proto";
...
+
+message ClientSettings {
+  DynamodbScalingLimits dynamodb_scaling_limits = 1;
+}
+
+message DynamodbScalingLimits {
+  // defaults to AWS default quotas if not set
+  // https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#default-limits-throughput-capacity-modes
+  int64 max_read_capacity_units = 1 [ (validate.rules).int64.gte = 1 ];
+  int64 max_write_capacity_units = 2 [ (validate.rules).int64.gte = 1 ];
```
mikecutalo, line 28: Doesn't the validation mark this as a required field in the config?

api/config/service/aws/v1/aws.proto
```diff
@@ -9,9 +9,26 @@ import "validate/validate.proto";
...
   ClientConfig client_config = 2;
+  ClientSettings client_settings = 3;
 }
 
 message ClientConfig {
   // If not set explicity, retries default to 0
   int32 retries = 1 [ (validate.rules).int32.gte = 0 ];
 }
+
+message ClientSettings {
```
mikecutalo, line 20: Should we just put these fields into the `ClientConfig` if so why not?

api/config/service/aws/v1/aws.proto
angethuy: Good catch, I'll add `ignore_empty` constraint to the validation

api/config/service/aws/v1/aws.proto
angethuy: I originally separated them to keep AWS client config variables like `HTTPClient`, `Region`, `Retryer`, `SocketTimeout`, etc distinct from the custom internal settings/constants used by the workflows. Bad naming on my part! Should I just combine all fields under `ClientConfig` or rename ClientSettings?

mikecutalo: Yeah good question since there are a few configuration options for dynamodb maybe we can put it under its own top level message maybe something like this? What do you think?

```proto
message Config {
  DynamodbConfig dynamodb_config = 3;
}

message DynamodbConfig {
...
}
```

angethuy: Looks good to me! Sets up the config struct nicely for additional datastores configs if needed in the future too

scarlettperry (approved):

mikecutalo (approved):

PR #1670, opened 2021-08-04
frontend: revamp table to be responsive

dschaller:
### Description

Revamp the Table component to be responsive and use grid layout.

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -22,14 +22,14 @@ const StyledPaper = styled(MuiPaper)({
   border: "1px solid #E7E7EA",
 });
 
-const StyledTable = styled(MuiTable)<{ actions: boolean; columnCount: number; compress: boolean }>(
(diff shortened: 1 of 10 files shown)


frontend/packages/core/src/Table/table.tsx
```diff
@@ -4,65 +4,188 @@ import {
...
-export const TableCell = styled(MuiTableCell)(
+const StyledTableHead = styled(MuiTableHead)({
+  display: "contents",
+  backgroundColor: "#D7DAF6",
+});
+
+const StyledTableRow = styled(MuiTableRow)<{ responsive?: boolean }>(
   {
+    ":nth-child(even)": {
+      background: "rgba(13, 16, 48, 0.03)",
```
scarlettperry, line 54: Screen Shot 2021-08-05 at 12 11 34 PM

do you think this should maybe be prop based? in case users don't want a diff color for every other row

dschaller: Maybe, this is just going off of the designs. I chose a random header color as well so we can revisit this when Kanchan comes back with finalized designs

scarlettperry (approved):

smonero (approved):

PR #1679, opened 2021-08-06
backend: base64 decode github app pem

dschaller:
### Description

We cannot read in the private key from environment variables because it ends up being stripped. Instead we base64 encode the key with no wraps and decode it here.

### Testing Performed

manual


scarlettperry (approved):

smonero (approved):

mikecutalo (changes requested):
Update the proto and include a comment and maybe even change the name of the config to 'pem_base64' or something along those lines so it's clear.

mikecutalo (approved):

PR #1686, opened 2021-08-09
dash: throw error from useDashState if context invalid

danielhochman:
### Description
Throw an error if the value is not valid for the given context. 

This will eliminate the need for the caller to continually check for and handle a null context value every time they access the context, which is a wholly invalid state anyways and can't be rectified.

### DIFF
--- a/frontend/workflows/projectSelector/src/dash-hooks.tsx
+++ b/frontend/workflows/projectSelector/src/dash-hooks.tsx
@@ -26,7 +26,7 @@ export const useDashState = (): DashState => {
   const value = React.useContext<DashState | undefined>(DashStateContext);
   if (!value) {
     throw new Error(
-      "useDashState was invoked with no value, check that it is a child of the Dash component"
+      "useDashState was invoked outside of a valid context, check that it is a child of the Dash component"
     );
   }
   return value;
--- a/frontend/workflows/projectSelector/src/dash-hooks.tsx
+++ b/frontend/workflows/projectSelector/src/dash-hooks.tsx
@@ -22,6 +22,10 @@ export const useDashUpdater = (): useDashUpdaterReturn => {
   };
 };
 
-export const useDashState = (): DashState | undefined => {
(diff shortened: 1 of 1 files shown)


frontend/workflows/projectSelector/src/dash-hooks.tsx
```diff
@@ -22,6 +22,12 @@ export const useDashUpdater = (): useDashUpdaterReturn => {
   };
 };
 
-export const useDashState = (): DashState | undefined => {
-  return React.useContext<DashState | undefined>(DashStateContext);
+export const useDashState = (): DashState => {
+  const value = React.useContext<DashState | undefined>(DashStateContext);
```
dschaller, line 26: Should we just prevent this from ever being undefined? e.g. initialize the DashState in the context?

danielhochman: It will only be undefined if you're not within the context. The DashState is initialized by the context since it's the Provider with a value.

dschaller: Right, I was thinking we change `DashStateContext`  to only have type `DashState` and set the initialization of the state on line 5 instead of using undefined.

danielhochman: That would mask the instance where the context is uncontrolled by a provider (which is an invalid state). The context should be undefined if it's not initialized within a provider because the card won't work at all and something is wrong. Furthermore cards should not be updating and using the data of a global context outside of a provider if it were to have a default value. For further elaboration see here: https://www.carlrippon.com/react-context-with-typescript-p4/

This a common pattern used by many other large projects: https://sourcegraph.com/search?q=context:global+createContext.*%3F%5C%28undefined&patternType=regexp

Having a globally default value makes sense for simple immutable constant data (like a theme) but not for this type of complex mutable data IMO.

dschaller (approved):

mikecutalo (approved):

PR #1687, opened 2021-08-09
dynamodb: add table status to updatetable response

angethuy:
### Description

This PR adds status fields to the UpdateTableCapacityResponse and UpdateGSICapacityResponse to capture the status of the table, which will indicate the table is `UPDATING` immediately after a successful request is made. Capacity changes are not immediately reflected after an UpdateTable action. A separate query to check the table status must be made after UpdateTable completes execution, to confirm the changes were applied.

More info about UpdateTable: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_UpdateTable.html

### Testing Performed

N/A


api/aws/dynamodb/v1/dynamodb.proto
```diff
@@ -35,6 +35,18 @@ service DDBAPI {
   }
 }
 
+// https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_TableDescription.html
+enum Status {
+  UNSPECIFIED = 0;
```
danielhochman, line 40: you'll want `UNKNOWN` as well if amazon returns a new status that you aren't able to convert into the enum

danielhochman (commented):
lgtm. i would go ahead and combine impl into this PR since it's relatively small.

backend/service/aws/aws.go
```diff
@@ -115,7 +115,7 @@ type Client interface {
 	S3StreamingGet(ctx context.Context, region string, bucket string, key string) (io.ReadCloser, error)
 
 	DescribeTable(ctx context.Context, region string, tableName string) (*dynamodbv1.Table, error)
-	UpdateTableCapacity(ctx context.Context, region string, tableName string, targetIndexRcu int64, targetIndexWcu int64) error
+	UpdateTableCapacity(ctx context.Context, region string, tableName string, targetTableRcu int64, targetTableWcu int64) (*dynamodbv1.Status, error)
```
danielhochman, line 118: i would not return a pointer for status here which is an int

danielhochman (commented):
looks good, two minor nits

backend/service/aws/aws.go
angethuy: Modified to return the status int instead of pointer, and changed `nil, err` returns to `0, err` (Status_UNSPECIFIED) instead

backend/service/aws/dynamodb.go
```diff
@@ -55,11 +55,14 @@ func (c *client) DescribeTable(ctx context.Context, region string, tableName str
 
 	globalSecondaryIndexes := getGlobalSecondaryIndexes(result.Table.GlobalSecondaryIndexes)
 
+	status := newProtoForTableStatus(result.Table.TableStatus)
```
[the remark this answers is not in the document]
angethuy: Modified to `tableStatus`

danielhochman (approved):
lgtm! nice work!

mikecutalo (approved):
Nice

PR #1688, opened 2021-08-09
frontend: ignore non-existent projects in dash

dschaller:
### Description

Dash: fix lookup of projects that do not exist in response.

fixInvalidProjects

### Testing Performed

manual


danielhochman:
@mikecutalo weren't we talking about handling this server-side?

mikecutalo:
@danielhochman we did talk about returning a partial error, but we were back and forth on the Idea. We can revisit the conversation tho. We can huddle up today and hash it out.

PR #1690, opened 2021-08-10
frontend: persist projects in dash

dschaller:
### Description

Add support to persist projects within dash.

persistProjects

### Testing Performed

manual


frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -61,8 +63,12 @@ const StyledProgressContainer = styled.div({
 // Determines if every project has projectData (i.e. the effect has finished fetching the data)
 const allPresent = (state: State): boolean => {
   let ret = true;
-  // We could potentially check all groups but it's really not necessary since an upstream cannot be added manually.
-  _.forEach(Object.keys(state[Group.PROJECTS]), p => {
+  const allProjects = [
```
danielhochman, line 66: nit: a `Set` would be more appropriate here in order to dedupe. not sure if there's a perf implication either way.

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -80,7 +86,12 @@ const ProjectSelector = () => {
 
   const { updateSelected } = useDashUpdater();
 
-  const [state, dispatch] = React.useReducer(selectorReducer, initialState);
+  const localState = window.localStorage.getItem(STORAGE_KEY);
+  const finalizedState = localState ? JSON.parse(localState) : {};
+  const [state, dispatch] = React.useReducer(
+    selectorReducer,
+    _.merge(initialState, finalizedState)
```
danielhochman, line 93: how will we ensure there aren't backwards compatibility issues if we update the structure of the state or storage? i think we need to validate that.

frontend/workflows/projectSelector/src/project-selector.tsx
dschaller: Makes sense for uniqueness but imo it wasn't a huge concern since it'll just be an extra validation. Performance wise, at the scale we are realistically rendering, neither make a difference given the use case. I'll udpate.

mikecutalo (commented):
How are we expiring project data? If you dont interact with the project selector during your sessions the project data will never update is that correct?

Should we be caching all project data or just what the user has selected and query for project data on load?

dschaller:
We aren't storing project data. This is just the project groups. Project data will always be fetched.

frontend/workflows/projectSelector/src/types.tsx
```diff
@@ -37,17 +37,19 @@ interface BackgroundPayload {
 
 export type Action = BackgroundAction | UserAction;
 
-export interface State {
+export interface ProjectsState {
```
dschaller, line 40: Open to discussing this name. `ProjectsState` could be a bit confusing since it's so similar to `ProjectState`. Another option was `HierarchyState`

frontend/workflows/projectSelector/src/storage.tsx
```diff
@@ -0,0 +1,89 @@
...
+import type { GroupState, ProjectsState, ProjectState, State } from "./types";
+import { Group } from "./types";
+
+const LOCAL_STORAGE_STATE_KEY = "dashState";
+
+/**
+ * Determines if an object is of type @type {ProjectState}.
+ */
+const isProjectState = (state: ProjectState | object): state is ProjectState => {
+  return (state as ProjectState).checked !== undefined;
```
danielhochman, line 12: does `as` actually check at runtime?

i'm having trouble finding any docs about this

dschaller: `as` won't do any checking but it'll cast the object to `ProjectState` so we can further check the fields on the object without typescript complaining.

[This page](https://www.typescriptlang.org/docs/handbook/advanced-types.html) is out of date but this is the original documentation on type guards.

danielhochman: but there are other fields, for example `custom` that are not checked here then, is that right?

dschaller: yes that's correct. Those fields are optional and therefore not expected to always be there.

frontend/workflows/projectSelector/src/types.tsx
```diff
@@ -37,27 +37,70 @@ interface BackgroundPayload {
...
+/**
+ * Determines if an object is of type @type {ProjectState}.
+ */
+const isProjectState = (state: ProjectState | object): state is ProjectState => {
+  const pState = state as ProjectState;
+  const checkedProp = pState?.checked;
+  const hasRequiredProps = checkedProp !== undefined && typeof checkedProp === "boolean";
+
+  const validOptionalProps =
+    (pState?.hidden !== undefined ? typeof pState.hidden === "boolean" : true) &&
```
danielhochman, line 67: delete anything related to hidden, it's not used, it's derived instead of stored on the state

frontend/workflows/projectSelector/src/types.tsx
danielhochman: SelectorState, GlobalProjectState? Something like that?

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -159,6 +160,7 @@ const ProjectSelector = () => {
     });
 
     // Update!
+    storeState(state);
```
scarlettperry, line 163: should this use `derivedState` instead like the other areas in the hook? played with it locally tho and didn't see any issues

scarlettperry: nvm using `state` is correct. otherwise if we use derivedState, we don't preserve the checked status of hidden projects.

scarlettperry (approved):

danielhochman (approved):
:+1:

PR #1691, opened 2021-08-10
core: pagerduty info for projects

mikecutalo:
### Description
Adds Pagerduty information for a given project if applicable, the concepts used track Pagerduty definitions. In this case `service_ids` tracks 1:1 with what Pagerduty calls them.s


scarlettperry (approved):

api/core/project/v1/project.proto
```diff
@@ -29,3 +30,10 @@ message ProjectDependencies {
 message Dependency {
   repeated string ids = 1;
 }
+
+// Pagerduty specific information for a project
+message Pagerduty {
```
danielhochman, line 35: super nit: PagerDuty

api/core/project/v1/project.proto
```diff
@@ -17,6 +17,7 @@ message Project {
   // Allows an organization to populate any information they see fit that does not fit our schema
   map<string, google.protobuf.Value> data = 5;
   ProjectDependencies dependencies = 6;
+  Pagerduty pagerduty = 7;
```
danielhochman, line 20: should we nest this in an oncall block or something? just in case we expand for future non-pagerduty

danielhochman (approved):
lgtm, two minor comments

api/core/project/v1/project.proto
mikecutalo: Or we could do an `Alerts` message with any type of information which could also include oncall? Or maybe the word `Alerts` is too specific.

api/core/project/v1/project.proto
```diff
@@ -17,6 +17,7 @@ message Project {
   // Allows an organization to populate any information they see fit that does not fit our schema
   map<string, google.protobuf.Value> data = 5;
   ProjectDependencies dependencies = 6;
+  Alerts alerts = 7;
```
danielhochman, line 20: Alerts sounds like it's going to list out the Alerts

danielhochman (approved):

PR #1700, opened 2021-08-12
backend: github: keep track of rate limit

smonero:
### Description

Use the header response to keep track of the github rate limit

### Testing Performed

make backend-test

### GitHub Issue

Fixes #

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -81,7 +81,7 @@ type Client interface {
 	CreatePullRequest(ctx context.Context, ref *RemoteRef, base, title, body string) (*PullRequestInfo, error)
 	CreateRepository(ctx context.Context, req *sourcecontrolv1.CreateRepositoryRequest) (*sourcecontrolv1.CreateRepositoryResponse, error)
 	CreateIssueComment(ctx context.Context, ref *RemoteRef, number int, body string) error
-	CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA string) (*githubv3.CommitsComparison, error)
+	CompareCommits(ctx context.Context, ref *RemoteRef, compareSHA string) (*githubv3.CommitsComparison, *githubv3.Response, error)
 	GetCommit(ctx context.Context, ref *RemoteRef) (*Commit, error)
 	GetRepository(ctx context.Context, ref *RemoteRef) (*Repository, error)
 	GetOrganization(ctx context.Context, organization string) (*githubv3.Organization, error)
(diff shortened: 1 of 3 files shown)


mikecutalo:
Will want to set this gauge in OSS for any and all GH request, we dont need to pass it back to the caller if that helps.

backend/service/github/github.go
```diff
@@ -95,7 +95,10 @@ func (s *svc) CreateIssueComment(ctx context.Context, ref *RemoteRef, number int
 	com := &githubv3.IssueComment{
 		Body: strPtr(body),
 	}
-	_, _, err := s.rest.Issues.CreateComment(ctx, ref.RepoOwner, ref.RepoName, number, com)
+	_, response, err := s.rest.Issues.CreateComment(ctx, ref.RepoOwner, ref.RepoName, number, com)
+	if response != nil {
+		s.scope.SubScope("github_client").Gauge("rate_limit_remaining").Update(float64(response.Rate.Remaining))
```
danielhochman, line 100: looks good but move to a method invocation e.g.
```go
emitRateLimit(response)
```

i'm also going to look to see if it's possible to just hook the http client so every method doesn't need to remember to call that

danielhochman:
Ok something like this should work...


```go
type StatsRoundTripper struct {
	Wrapped http.RoundTripper
	scope tally.Scope
}

func (st *StatsRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
	resp, err := st.Wrapped.RoundTrip(req)

	if hdr := resp.Header.Get("X-RateLimit-Remaining"); hdr != "" {
		if v, err := strconv.Atoi(hdr); err == nil {
			st.scope.Gauge("foo").Update(float64(v))
		}
	}
	return resp, err
}

func newService() {
...
    httpClient.Transport = &StatsRoundTripper{Wrapped: httpClient.Transport}
    restClient := githubv3.NewClient(httpClient)
...
}


```

backend/service/github/github.go
```diff
@@ -40,12 +41,28 @@ const CurrentUser = ""
...
+	Wrapped http.RoundTripper
+	scope   tally.Scope
+}
+
+func (st *StatsRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	resp, err := st.Wrapped.RoundTrip(req)
+
+	if hdr := resp.Header.Get("X-RateLimit-Remaining"); hdr != "" {
+		if v, err := strconv.Atoi(hdr); err == nil {
+			st.scope.Gauge("github_rate_limit_remaining").Update(float64(v))
```
mikecutalo, line 54: whats the full stats namespace? I think its going to be `clutch.service.github.github_rate_limit_remaining` ? If so lets drop the extra github, you should be able to test this locally.

danielhochman (approved):

mikecutalo (approved):

PR #1701, opened 2021-08-12
dynamodb: add UpdateGSICapacity endpoint

angethuy:
### Description

This PR implements the UpdateGSICapacity endpoint for the AWS Dynamodb service. UpdateGSICapacity will perform an update action on a specified global secondary index's provisioned capacity, checking against limits for safety.

### Testing Performed
Unit and manual.

### TODOs

N/A

### DIFF
--- a/backend/mock/service/awsmock/awsmock.go
+++ b/backend/mock/service/awsmock/awsmock.go
@@ -121,8 +121,9 @@ func (s *svc) UpdateTableCapacity(ctx context.Context, region string, tableName 
 	return stat, nil
 }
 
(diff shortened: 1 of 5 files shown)


danielhochman:
Is it possible to update GSI and Table capacity simultaneously? I could see that being a useful feature if so, and therefore trying to combine the updates into a single endpoint instead of separate ones.

angethuy:
> Is it possible to update GSI and Table capacity simultaneously? I could see that being a useful feature if so, and therefore trying to combine the updates into a single endpoint instead of separate ones.

I spec'd this out thinking that we had to update them separately, but upon further investigation, we can indeed use `UpdateTable` to update all the capacities simultaneously. Will refactor the two endpoints as suggested

PR #1705, opened 2021-08-17
dash: move loading prop to card header

scarlettperry:
### Description
Previously, the loading component was conditionally rendered in between the card header & children and so it pushed down the child component every time there was loading state.

Before:
before

PR moves the loading component to the card header so that it doesn't push down the child component
After:
after

### Testing Performed
locally with internal cards

### DIFF
--- a/frontend/workflows/projectSelector/src/card.tsx
+++ b/frontend/workflows/projectSelector/src/card.tsx
@@ -6,6 +6,7 @@ import { Grid, LinearProgress } from "@material-ui/core";
 
 const StyledProgressContainer = styled.div({
   height: "4px",
+  background: "rgb(235, 237, 251)",
   ".MuiLinearProgress-root": {
     backgroundColor: "rgb(194, 200, 242)",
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectSelector/src/card.tsx
```diff
@@ -6,6 +6,7 @@ import { Grid, LinearProgress } from "@material-ui/core";
 
 const StyledProgressContainer = styled.div({
   height: "4px",
+  background: "rgb(235, 237, 251)",
```
danielhochman, line 9: i would style the CardHeader fragment with background instead so it doesn't have to match

danielhochman (approved):
lgtm, one suggestion

frontend/workflows/projectSelector/src/card.tsx
scarlettperry: updated

mikecutalo (approved):

PR #1717, opened 2021-08-20
core: revamp card header

scarlettperry:
### Description
PR 
* creates a custom header in lieu of the Material component
* adds a new optional field to the header: `sections` which allows users to add subcontent in the card header

### Testing Performed
locally

examples showing the sections take up the full width

Screen Shot 2021-08-27 at 4 49 30 PM

Screen Shot 2021-08-27 at 4 49 53 PM

example of card header with no sections
Screen Shot 2021-08-27 at 4 50 30 PM


frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+  margin: "0px",
+  width: "100%",
+});
+
+const StyledCardHeaderAvater = styled.div({
+  marginRight: "16px",
+  fontSize: "24px",
+});
+
+const StyledDivider = styled(Divider)({
```
dschaller, line 64: Should we sync with Kanchan on if we want this as a component?

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+const StyledCardHeader = styled(Grid)({
+  padding: "16px",
+  minHeight: "72px",
+  margin: "0px",
+  width: "100%",
+});
+
+const StyledCardHeaderAvater = styled.div({
+  marginRight: "16px",
+  fontSize: "24px",
```
dschaller, line 61: Can we use the typography component for this?

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
-interface CardHeaderProps extends Pick<MuiCardHeaderProps, "avatar" | "title"> {
+const StyledCardHeader = styled(Grid)({
+  padding: "16px",
+  minHeight: "72px",
+  margin: "0px",
+  width: "100%",
+});
+
+const StyledCardHeaderAvater = styled.div({
+  marginRight: "16px",
```
dschaller, line 60: Should we include sizing for the icon being 32x32?

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
 
-interface CardHeaderProps extends Pick<MuiCardHeaderProps, "avatar" | "title"> {
+const StyledCardHeader = styled(Grid)({
+  padding: "16px",
+  minHeight: "72px",
+  margin: "0px",
+  width: "100%",
+});
+
+const StyledCardHeaderAvater = styled.div({
```
dschaller, line 59:
```suggestion
const StyledCardHeaderAvatar = styled.div({
```

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+  marginRight: "16px",
+  fontSize: "24px",
+});
+
+const StyledDivider = styled(Divider)({
+  color: "rgba(13, 16, 48, 0.38)",
+  height: "36px",
```
dschaller, line 66: I think these are 40px in figma

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+  alignSelf: "center",
```
dschaller, line 67:
```suggestion
  alignSelf: "stretch",
```

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+});
+
+const StyledGridItem = styled(Grid)({
+  textAlign: "center",
+});
+
+export interface CardHeaderSections {
```
dschaller, line 74:
```suggestion
export interface CardHeaderSummaryProps {
```

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +52,62 @@ const StyledCardHeaderContainer = styled.div({
...
+  height: "36px",
+});
+
+const StyledGridItem = styled(Grid)({
+  textAlign: "center",
+});
+
+export interface CardHeaderSections {
+  title: React.ReactNode;
+  subheader?: React.ReactNode;
```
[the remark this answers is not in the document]
dschaller: We might want to make this more narrow for now since we don't want any element rendered. e.g. probably want this to only be text for now and it might be good to force consumers to use a Typography component instead of just a string.

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,63 @@ const StyledCardHeaderContainer = styled.div({
...
-        padding: "16px",
-      }}
-      disableTypography
-      avatar={avatar}
-      title={<StyledTypography variant="h3">{title}</StyledTypography>}
-    />
-    {children}
-  </StyledCardHeaderContainer>
-);
+// TODO: do we want to move the sections into the Dash card and keep this core card with just avatar and title
```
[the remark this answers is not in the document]
dschaller: I think it makes sense to keep this here. It's similar to having actions off the header. If anything I think it might be good to restructure the files here and separate out the card parts into various files but that's a larger discussion :)

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+const CardHeader = ({ avatar, children, title, sections }: CardHeaderProps) => {
```
dschaller, line 87: nit: you can default sections to an empty list and remove the length check below

frontend/packages/core/src/card.tsx
```diff
@@ -51,23 +49,65 @@ const StyledCardHeaderContainer = styled.div({
...
+      <StyledCardHeader container wrap="nowrap" alignItems="center" spacing={2}>
+        {/* TODO: use avatar component per design doc */}
+        <StyledCardHeaderAvater>{avatar}</StyledCardHeaderAvater>
+        <Grid item xs>
+          <StyledTypography variant="h4">{title}</StyledTypography>
+        </Grid>
+        {sections?.length > 0 &&
+          sections.map(section => (
+            <>
+              <StyledDivider orientation="vertical" />
```
dschaller, line 99: We might want to only prepend a divider if the title is set

frontend/packages/core/src/card.tsx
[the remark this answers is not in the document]
scarlettperry: I changed the `subheader` to be a string type and use the Typography component.

For `title` - in figma, the title color can vary per section and internally, I'm passing in the `TimeAgo` component for some of the sections, so perhaps we keep this field flexible?

frontend/packages/core/src/card.tsx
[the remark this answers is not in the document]
scarlettperry: yeah good idea. added a TODO

frontend/packages/core/src/card.tsx
scarlettperry: yeah, left a comment for Kanchan in figma

frontend/packages/core/src/card.tsx
scarlettperry: yep added 32x32 to the icon container, left a TODO to use the material ui Avatar component when I implement the figma designs

frontend/packages/core/src/card.tsx
scarlettperry: ah, just noticed the height isn't consistent across the mocks. Left a comment for Kanchan to confirm.

frontend/packages/core/src/card.tsx
scarlettperry: with `stretch`,  this happens when the screen size gets smaller:
Screen Shot 2021-09-01 at 2 10 38 PM

frontend/packages/core/src/card.tsx
scarlettperry: yep, updated

frontend/packages/core/src/card.tsx
```diff
@@ -77,7 +78,7 @@ interface CardHeaderProps {
 const CardHeader = ({ avatar, children, title, summary = [] }: CardHeaderProps) => {
   return (
     <StyledCardHeaderContainer>
-      <StyledCardHeader container wrap="nowrap" alignItems="center" spacing={2}>
```
scarlettperry: opted to set custom padding instead for the grid items b/c by default `spacing` adds padding all over  and we only need a bit of right/left padding so that the items don't "touch" each other when the screen size gets smaller

i.e. to avoid this
Screen Shot 2021-09-01 at 9 42 45 PM

with a bit of padding
Screen Shot 2021-09-01 at 9 44 11 PM

frontend/packages/core/src/card.tsx
scarlettperry: left a Q on figma to confirm icon font size

frontend/packages/core/src/card.tsx
scarlettperry: jst noticed the icon container isn't consistent across the mocks. left a Q on figma to confirm

frontend/packages/core/src/card.tsx
```diff
@@ -36,23 +38,71 @@ const StyledCardHeaderContainer = styled.div({
...
+export interface CardHeaderSummaryProps {
+  title: React.ReactNode;
+  subheader?: string;
+}
+
+interface CardHeaderProps {
+  avatar: React.ReactNode;
   children?: React.ReactNode;
+  summary?: CardHeaderSummaryProps[];
+  title: React.ReactNode;
```
dschaller, line 75: nit:
```suggestion
  title: React.ReactElement<Typography>;
```

frontend/packages/core/src/card.tsx
dschaller: ah I missed that 👍

frontend/packages/core/src/card.tsx
```diff
@@ -36,23 +38,71 @@ const StyledCardHeaderContainer = styled.div({
   background: "#EBEDFB",
 });
 
-interface CardHeaderProps extends Pick<MuiCardHeaderProps, "avatar" | "title"> {
+const StyledCardHeader = styled(Grid)({
+  padding: "16px",
+  minHeight: "72px",
+  ".MuiGrid-item": {
+    padding: "0px 8px 0px 8px",
```
dschaller, line 45: super nit:
```suggestion
    padding: "0px 8px",
```

frontend/packages/core/src/card.tsx
```diff
@@ -36,23 +38,71 @@ const StyledCardHeaderContainer = styled.div({
...
+        <Grid item xs>
+          <Typography variant="h4">{title}</Typography>
+        </Grid>
+        {summary.map(section => (
+          <>
+            <StyledDivider orientation="vertical" />
+            <StyledGridItem item xs>
+              {section.title}
+              {section.subheader && (
+                <Typography variant="body3" color="rgba(13, 16, 48, 0.6)">
```
dschaller, line 94: this might be another discrepancy in the designs but some have this as body4

scarlettperry:
left TODOs for things I'm waiting on from Design that can be addressed in follow up PRs: 
* height of the divider / size of the icon container / size of the icon / size of the subheader (mocks have different sizes)
* a core component for dividers

frontend/packages/core/src/card.tsx
scarlettperry: I tried it but it didn't enforce the type when I used another component. Also tried `React.ReactElement<Typography> | React.ReactElement<Typography>[]`; , a pattern we use in the codebase but that didn't enforce the type either 🤔 still looking into, maybe it's a typescript bug

frontend/packages/core/src/card.tsx
scarlettperry: left a TODO and a comment on figma

frontend/packages/core/src/card.tsx
scarlettperry: did some more local testing with other components that type React children (https://github.com/lyft/clutch/search?q=React.ReactElement), and type declarations like these don't enforce the type of the children

it seems like it isn't currently possible to type-check React children (https://stackoverflow.com/questions/59381291/react-typescript-only-allow-children-of-specific-type, https://github.com/Microsoft/TypeScript/issues/14729)

however, i pushed up the change as I think there's still value in documenting what elements _should_ be passed in

smonero (approved):

mikecutalo (approved):

PR #1720, opened 2021-08-23
core: add configurable padding to CardContent

scarlettperry:
### Description

PR also makes the CardContent padding configurable via the Material UI spacing system https://material-ui.com/system/spacing/#api

PR also wraps the Dash card children in the CardContent component to fix issues where the card children have their own border styling. Example below: the Table component's [Paper ](https://github.com/lyft/clutch/blob/d3632a1426c51211f66b690e871a485e09a6f0c0/frontend/packages/core/src/Table/table.tsx#L21)has border styling and so it looks like the card children area isn't aligned with the card header. (Also, the CardContent component will be useful if we want to add card actions to it, ie "expand to see more").
Screen Shot 2021-08-23 at 11 06 17 AM

### Testing Performed
locally

0px padding on the CardContent
Screen Shot 2021-08-31 at 12 03 42 PM

32px padding (the value we currently use) on the Landing cards which use the CardContent as well

Screen Shot 2021-08-31 at 12 04 02 PM

### DIFF
--- a/frontend/packages/core/package.json
+++ b/frontend/packages/core/package.json
@@ -35,7 +35,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/styles": "^4.11.4",
-    "@material-ui/system": "^4.12.1",
+    "@material-ui/system": "^4.11.4",
     "axios": "^0.21.1",
     "history": "^5.0.0",
     "js-cookie": "^3.0.0",
--- a/frontend/packages/core/package.json
+++ b/frontend/packages/core/package.json
@@ -35,6 +35,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/styles": "^4.11.4",
+    "@material-ui/system": "^4.12.1",
     "axios": "^0.21.1",
     "history": "^5.0.0",
     "js-cookie": "^3.0.0",
--- a/frontend/packages/core/src/card.tsx
+++ b/frontend/packages/core/src/card.tsx
@@ -1,17 +1,14 @@
 import * as React from "react";
 import styled from "@emotion/styled";
-import type {
-  CardContentProps as MuiCardContentProps,
-  CardHeaderProps as MuiCardHeaderProps,
-} from "@material-ui/core";
+import type { CardHeaderProps as MuiCardHeaderProps } from "@material-ui/core";
 import {
   Avatar,
   Card as MuiCard,
   CardActionArea,
   CardActionAreaProps,
-  CardContent as MuiCardContent,
   CardHeader as MuiCardHeader,
 } from "@material-ui/core";
+import { spacing } from "@material-ui/system";
 
 import { StyledTypography } from "./typography";
 
@@ -32,16 +29,6 @@ const StyledCard = styled(MuiCard)({
     backgroundColor: "#D7DAF6",
   },
 });
-
-const StyledCardContent = styled(MuiCardContent)((props: { disablePadding?: boolean }) => ({
-  "> .MuiPaper-root": {
-    border: "0",
-    borderRadius: "0",
-  },
-  "&.MuiCardContent-root": {
-    padding: props.disablePadding ? "0px" : "32px",
-  },
-}));
 
 export interface CardProps {
   children?: React.ReactNode | React.ReactNode[];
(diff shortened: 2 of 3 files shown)


dschaller:
I think adjustable padding seems ok. Maybe we want this to be more configurable other than disable or enable? We should probably sync with Kanchan on this.

scarlettperry:
> I think adjustable padding seems ok. Maybe we want this to be more configurable other than disable or enable? We should probably sync with Kanchan on this.

I'll try out using this spacing api https://material-ui.com/system/spacing/#api so that it'll be configurable

frontend/packages/core/src/card.tsx
```diff
@@ -69,7 +58,28 @@ const CardHeader = ({ avatar, children, title }: CardHeaderProps) => (
...
+// We can add more to this list as use cases arise
+interface SpacingProps {
+  padding?: number;
+  // shorthand for padding
+  p?: number;
+}
+
+const BaseCardContent = styled.div<SpacingProps>`
+  ${spacing}
+`;
```
scarlettperry, line 71: lmk your thoughts on this pattern. saw it here https://dev.to/andrecastelo/prop-forwarding-with-emotion-material-ui-and-material-ui-system-40kd and here https://github.com/resir014/chungking/blob/9a03099f5ddcee79216b927958872d4a49ffe03b/packages/chungking-react/src/layout/inline/Inline.tsx#L17-L19. it was tricky to figure out how the spacing system can be used in any component b/c Material UI docs only show it being used in the `Box` component which by default has `spacing` https://material-ui.com/system/spacing/

frontend/packages/core/src/card.tsx
```diff
@@ -1,17 +1,14 @@
 import * as React from "react";
 import styled from "@emotion/styled";
-import type {
-  CardContentProps as MuiCardContentProps,
-  CardHeaderProps as MuiCardHeaderProps,
-} from "@material-ui/core";
```
scarlettperry, line 6: I opted to use a plain div and not use the CardContent component b/c it has padding bugs w.r.t adding extra padding where it assumes the "card action" area will be. More details in this issue  https://github.com/mui-org/material-ui/issues/9871. Initially, I planned to override the component's last-child bottom padding to "0px" or use the `spacing` padding value but we/users might want the "card action area" to be different from the CardContent padding. The CardContent component itself is very simple so decided a div might be better so that we can control the padding more and we can handle the card action padding separately via the the CardAction component  https://material-ui.com/api/card-actions/ if we end up needing it for Dash. LMK what you think.

frontend/packages/core/src/card.tsx
```diff
@@ -69,7 +58,28 @@ const CardHeader = ({ avatar, children, title }: CardHeaderProps) => (
   </StyledCardHeaderContainer>
 );
 
-interface CardContentProps extends MuiCardContentProps {}
+// Material UI Spacing system supports many props https://material-ui.com/system/spacing/#api
+// We can add more to this list as use cases arise
+interface SpacingProps {
```
dschaller, line 63: Can we pick these from material's type interface?

frontend/packages/core/src/card.tsx
dschaller: I think it's fine especially since the card content renders a div anyway. You covered the only downside I can think of which is is we had card actions but I think we could handle that use case

frontend/packages/core/src/card.tsx
dschaller: This is great!

dschaller (commented):
Looks great! Minor comment to pick the type props from the upstream and then we can merge

frontend/packages/core/src/card.tsx
```diff
@@ -8,19 +8,15 @@ import {
...
 
 const StyledCard = styled(MuiCard)({
   boxShadow: "0px 4px 6px rgba(53, 72, 212, 0.2)",
   border: "1px solid rgba(13, 16, 48, 0.1)",
 
-  ".MuiCardContent-root": {
-    color: "#0D1030",
-    fontSize: "16px",
-  },
-
```
scarlettperry: small cleanup

frontend/packages/core/src/card.tsx
```diff
@@ -8,19 +8,15 @@ import {
   CardActionAreaProps,
   CardHeader as MuiCardHeader,
 } from "@material-ui/core";
+import type { SpacingProps as MuiSpacingProps } from "@material-ui/system";
 import { spacing } from "@material-ui/system";
 
-import { StyledTypography } from "./typography";
```
scarlettperry, line 16: in doing this cleanup https://github.com/lyft/clutch/pull/1720/commits/0258d972f274e671ca154b4b345e6df41a596104#r700359762, double checked the typography for the landing card and saw we should be using the Typography component instead.

frontend/packages/core/package.json
```diff
@@ -35,6 +35,7 @@
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.58",
     "@material-ui/styles": "^4.11.4",
+    "@material-ui/system": "^4.12.1",
```
danielhochman, line 38: i'd use a consistent version spec for all the mui libs, though i guess it wouldn't technically matter. if you need an upgrade we can do it separately.

danielhochman (approved):
nice! one small comment

dschaller (approved):

PR #1722, opened 2021-08-23
housekeeping: react-hook-form v7

dschaller:
### Description

Updating react-hook-form and related packages from beta to stable.

### Testing Performed

manual


scarlettperry:
Played with it locally and noticed that:
1) the validation isn't working on the metadata table. Tested on K8s Resize HPA
2) the Search and Advanced Search form fields incorrectly say fields aren't provided when trying to submit via `keydown` (but works if you submit via button). Tested on K8s ResizeHPA

PR #1738, opened 2021-09-02
dash: selector scrolling and chip wrapping

danielhochman:
### Description

Adds independent scrolling for the project selector. Also,

**Chip wrapping**
Screenshot from 2021-09-01 16-46-01

### GitHub Issue

Fixes #


frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -41,16 +50,19 @@ const StyledMenuItem = styled.div({
...
   padding: "0 12px",
 });
 
-const StyledHeaderColumn = styled.div({
+const StyledHeaderColumn = styled.div((props: { grow?: boolean }) => ({
   display: "flex",
+  minHeight: "38px",
   alignItems: "center",
-});
+  flexGrow: props.grow ? 1 : 0,
```
scarlettperry, line 64: a lil confused by how this works. `grow` is passed to the div but I don't see a case when this can be false

scarlettperry:
not a new issue but can the project titles be trimmed at a certain length to avoid the `only` and `x` button from getting pushed out of view? (but can be done in a follow up PR if it's out scope)

Screen Shot 2021-09-07 at 4 40 34 PM

frontend/workflows/projectSelector/src/project-group.tsx
danielhochman: it's only passed as a bool prop on one instance of `StyledHeaderColumn`, default will be false

scarlettperry: oh! i missed that the div is used in other places 👍

danielhochman:
@scarlettperry i'll try to fix in this PR, good find

scarlettperry (approved):
🥳

lyft-tla (approved):
LGTM

PR #1741, opened 2021-09-03
frontend: refactor app provider

dschaller:
### Description

This is a rather large change to the app provider:
  * Make workflow filters synchronous
  * Allow for overrides of an individual workflow props
  * Renamed types for application types
  * Cleaned up code
  * Added unit tests

### Testing Performed

Unit

### DIFF
--- a/frontend/packages/core/src/AppProvider/filters.tsx
+++ b/frontend/packages/core/src/AppProvider/filters.tsx
@@ -1,0 +1,26 @@
+import { featureFlags } from "../flags";
+
+import type { Workflow } from "./types";
+
+/**
+ * Filter workflow routes using available feature flags.
+ *
+ * @param workflows a list of valid Workflow objects.
+ */
+const featureFlagFilter = (workflows: Workflow[]): Promise<Workflow[]> => {
+  return featureFlags().then(flags =>
+    workflows.filter(workflow => {
+      /* eslint-disable-next-line no-param-reassign */
+      workflow.routes = workflow.routes.filter(route => {
+        const show =
+          route.featureFlag === undefined ||
+          (flags?.[route.featureFlag] !== undefined &&
+            flags[route.featureFlag].booleanValue === true);
+        return show;
+      });
+      return workflow.routes.length !== 0;
+    })
+  );
+};
+
+export default featureFlagFilter;
--- a/frontend/packages/core/src/AppProvider/tests/registrar.test.tsx
+++ b/frontend/packages/core/src/AppProvider/tests/registrar.test.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 
-import { workflowRoutes } from "../registrar";
+import { registeredWorkflows, workflowRoutes } from "../registrar";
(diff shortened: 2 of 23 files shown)


frontend/packages/core/src/AppProvider/filters.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import type { Workflow } from "./types";
+
+/**
+ * Filter workflow routes using available feature flags.
+ *
+ * @param workflows a list of valid Workflow objects.
+ */
+const featureFlagFilter = (workflows: Workflow[]): Promise<Workflow[]> => {
+  return featureFlags().then(flags =>
```
dschaller, line 11: TODO: look into the timeout on these as they are now blocking render

frontend/packages/core/src/AppProvider/tests/registrar.test.tsx
```diff
@@ -34,86 +35,198 @@ const WORKFLOW_CONFIG = {
...
+      expect(workflows[0].displayName).toBe("Amazon Web Services");
+    });
+  });
+
+  it("removes workflow overrides from gateway config", () => {
+    gatewayConfig[WORKFLOW_ID].overrides = { displayName: "Amazon Web Services" };
+    return registeredWorkflows(availableWorkflows, gatewayConfig).then(() => {
+      expect(warn).not.toHaveBeenCalled();
+    });
+  });
```
scarlettperry, line 165: is there a typo in the test description? not sure what is getting "removed"

frontend/packages/core/src/AppProvider/types.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+/** Overrides to apply to a workflow within a gateway. */
+export interface WorkflowOverrides {
+  overrides?: Partial<WorkflowBase>;
+}
+
+/** Configuration specific to the gateway running. */
+export interface GatewayConfig {
+  /** Unique package name to collection of routes */
+  [packageName: string]: RouteConfigs | WorkflowOverrides;
```
scarlettperry, line 42: since `RouteConfigs` and/or `WorkflowOverrides` can be passed, is a `GatewayConfig` without `RouteConfigs` valid? if not, I wonder if that might confuse users and lead to some usage errors

frontend/packages/core/src/AppProvider/tests/registrar.test.tsx
scarlettperry: oh i think i understand now https://github.com/lyft/clutch/pull/1741/files#diff-7157db281f494cb33c18ac118bfd99f9ec25a1645e1e9097b15f07e808a825a8R95

scarlettperry:
also, could an example of a config with overrides be added to the docs?

frontend/packages/core/src/AppProvider/types.tsx
dschaller: it is valid but we would remove the package since there are no routes associated with the package

PR #1744, opened 2021-09-10
chaos: update go-control-plane to latest

danielhochman:
### Description
Had to implement interface changes with latest version of go-control-plane.

### Testing Performed
CI.


backend/module/chaos/experimentation/xds/xds.go
```diff
@@ -244,11 +304,25 @@ func (c *ecdsCallbacks) OnStreamRequest(streamID int64, req *gcpDiscoveryV3.Disc
 	return nil
 }
 
+func (c *ecdsCallbacks) OnStreamDeltaRequest(streamID int64, req *gcpDiscoveryV3.DeltaDiscoveryRequest) error {
+	c.safeECDSResources.setResourcesForCluster(req.Node.Cluster, req.ResourceNamesSubscribe)
```
danielhochman, line 308: this is the only line i'm not sure about. there's also `req.ResourceNamesUnsubscribe`. afaict that could be used for optimization but won't cause any problems if left out based on my reading of the other code, in particular `getResourcesForCluster`.

kathan24:
I think we can just keep the body of the Delta methods empty since we don't use [Delta gRPC xDS](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/operations/dynamic_configuration#delta-grpc-xds).

Clutch config is [here](https://github.com/lyft/envoy-static-config/blob/master/templates/top_level_helper.yaml#L69) and as per the doc [here](https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/xds_api#delta-endpoints), is should be DELTA_GRPC if we use delta which we don't

backend/module/chaos/experimentation/xds/xds.go
kathan24: Afair, the whole logic around `ecdsResourceMap` is used for a fail-safe mechanism. 
Eg, Let's say we have injected a fault on a cluster but at the time of removal, the cluster is disconnected from the server. Whenever the connection re-establishes, it will remove the fault.

kathan24 (approved):

mikecutalo (approved):

PR #1745, opened 2021-09-13
dash cards: update/add breakpoints

scarlettperry:
### Description
Currently, the cards remain side by side until the screen size reaches the `xs` breakpoint. Since card info can be dense (i.e. many columns) we want the cards to take the screen's full width at an earlier breakpoint.
* PR updates the small breakpoint value and adds the medium and large breakpoint so that the cards remain side by side until the screen size becomes less than 1280px. https://material-ui.com/customization/breakpoints/
* Tables are used in the internal cards and realized that it has a default breakpoint at which to compress the table rows. Since cards take up full width now, PR changes the default to `small` so that the table rows take up the full space available and also adds a `compressBreakpoint` prop so that user can change the default value.

### Testing Performed
locally

update

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -96,6 +96,8 @@ const TableContainer = ({ children }: TableContainerProps) => (
 export interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   /** The names of the columns. This must be set (even to empty string) to render the table. */
   columns: string[];
+  /** The breakpoint at which to compress the table rows. By default the small breakpoint is used. */
+  compressBreakpoint?: "xs" | "sm" | "md" | "lg" | "xl";
   /** Hide the header. By default this is false. */
   hideHeader?: boolean;
   /** Add an actions column. By default this is false. */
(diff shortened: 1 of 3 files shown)


frontend/workflows/projectSelector/src/dash.tsx
```diff
@@ -1,7 +1,10 @@
 import * as React from "react";
-import { Box } from "@material-ui/core";
+import { Table, TableRow, Typography } from "@clutch-sh/core";
```
scarlettperry, line 2: fyi nothing in this file will be merged. this is just for PR reviewers in case they want to locally test

scarlettperry:
lmk your thoughts on the approach. we can break into full width at a different break point/define our own custom breakpoints

scarlettperry:
converting to draft b/c might also need to update the core table's compress logic. discussing offline with team.

scarlettperry:
updated PR gif with latest changes

smonero (approved):

lyft-tla (approved):

frontend/packages/core/src/Table/table.tsx
```diff
@@ -96,6 +96,8 @@ const TableContainer = ({ children }: TableContainerProps) => (
 export interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   /** The names of the columns. This must be set (even to empty string) to render the table. */
   columns: string[];
+  /** The breakpoint at which to compress the table rows. By default the small breakpoint is used. */
+  compressBreakpoint?: "xs" | "sm" | "md" | "lg" | "xl";
```
dschaller, line 100: Can we use material's interface for these?
`import type { BreakpointDefaults } from "@material-ui/core/styles/createBreakpoints";`

PR #1746, opened 2021-09-13
dash: fix card container overflow behavior

danielhochman:
### Description

Fix overflow behavior of card container, before it would scroll the entire page including the selector.

*AFTER*
Peek 2021-09-13 13-06 AFTER

### Testing Performed
Manual


scarlettperry (approved):
tested and looks great!

dschaller (approved):

PR #1749, opened 2021-09-17
dynamodb: fix BillingMode to proto method

angethuy:
### Description

This PR revises how the BillingMode field for tables is marshaled into proto format. 

### Testing Performed

Manual testing locally + unit tests

### DIFF
--- a/backend/service/aws/dynamodb.go
+++ b/backend/service/aws/dynamodb.go
@@ -45,11 +45,16 @@ func (c *client) DescribeTable(ctx context.Context, region string, tableName str
 	result, err := getTable(ctx, cl, tableName)
 
 	if err != nil {
-		c.log.Error("unable to find table", zap.Error(err))
+		c.log.Warn("unable to find table on region: "+region, zap.Error(err))
 		return nil, err
 	}
 
-	ret := newProtoForTable(result.Table, region)
+	table, ok := result.Table
+	if !ok {
+		return nil, status.Error(codes.NotFound, "Table description not found")
+	}
+
+	ret := newProtoForTable(table, region)
 
 	return ret, nil
 }
--- a/backend/service/aws/dynamodb.go
+++ b/backend/service/aws/dynamodb.go
@@ -124,12 +124,13 @@ func newProtoForIndexStatus(s types.IndexStatus) dynamodbv1.GlobalSecondaryIndex
 // if a table is PAY_PER_REQUEST (on demand), it will have 0 RCU/WCU provisioned
 func checkBillingMode(t *types.ProvisionedThroughputDescription) types.BillingMode {
 	if (*t.ReadCapacityUnits > 0) || (*t.WriteCapacityUnits > 0) {
-		return "PROVISIONED"
+		return types.BillingModeProvisioned
 	}
 	if (*t.ReadCapacityUnits == 0) || (*t.WriteCapacityUnits == 0) {
-		return "PAY_PER_REQUEST"
+		return types.BillingModePayPerRequest
 	}
-	return "UNKNOWN" // unable to infer what billing mode it is
+
+	return "UNSPECIFIED" // unable to infer what billing mode it is
 }
 
 func newProtoForBillingMode(t *types.TableDescription) dynamodbv1.Table_BillingMode {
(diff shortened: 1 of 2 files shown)


scarlettperry (approved):

backend/service/aws/dynamodb.go
```diff
@@ -118,8 +118,8 @@ func newProtoForIndexStatus(s types.IndexStatus) dynamodbv1.GlobalSecondaryIndex
 	return dynamodbv1.GlobalSecondaryIndex_Status(value)
 }
 
-func newProtoForBillingMode(s types.BillingMode) dynamodbv1.Table_BillingMode {
-	value, ok := dynamodbv1.Table_BillingMode_value[string(s)]
+func newProtoForBillingMode(s *types.BillingModeSummary) dynamodbv1.Table_BillingMode {
+	value, ok := dynamodbv1.Table_BillingMode_value[string(s.BillingMode)]
```
dschaller, line 122: Should this be a more explicit nil check?
something like:
```go
if s == nil {
		return dynamodbv1.Table_BILLING_UNKNOWN
}
value, ok := dynamodbv1.Table_BillingMode_value[string(s.BillingMode)]
```

angethuy: Good call, I'll add the explicit check

backend/service/aws/dynamodb.go
```diff
@@ -118,11 +118,33 @@ func newProtoForIndexStatus(s types.IndexStatus) dynamodbv1.GlobalSecondaryIndex
...
 
-func newProtoForBillingMode(s types.BillingMode) dynamodbv1.Table_BillingMode {
-	value, ok := dynamodbv1.Table_BillingMode_value[string(s)]
+// manually check the billing mode by inferring it from throughput
+// to cover cases where AWS does not return the mode in the table description
+// if a table is PROVISIONED, it will have at least 1 RCU/WCU provisioned
+// if a table is PAY_PER_REQUEST (on demand), it will have 0 RCU/WCU provisioned
+func checkBillingMode(t *types.ProvisionedThroughputDescription) types.BillingMode {
+	if (*t.ReadCapacityUnits > 0) || (*t.WriteCapacityUnits > 0) {
+		return "PROVISIONED"
```
mikecutalo, line 127: Shouldn't we have constants for this already defined in proto? Can we the string value there instead of writing them out here? This applies for all use cases in this pr.

backend/service/aws/dynamodb.go
```diff
@@ -118,11 +118,33 @@ func newProtoForIndexStatus(s types.IndexStatus) dynamodbv1.GlobalSecondaryIndex
 	return dynamodbv1.GlobalSecondaryIndex_Status(value)
 }
 
-func newProtoForBillingMode(s types.BillingMode) dynamodbv1.Table_BillingMode {
-	value, ok := dynamodbv1.Table_BillingMode_value[string(s)]
+// manually check the billing mode by inferring it from throughput
+// to cover cases where AWS does not return the mode in the table description
+// if a table is PROVISIONED, it will have at least 1 RCU/WCU provisioned
+// if a table is PAY_PER_REQUEST (on demand), it will have 0 RCU/WCU provisioned
+func checkBillingMode(t *types.ProvisionedThroughputDescription) types.BillingMode {
```
dschaller, line 125: super nit: since this returns a billing mode should it be named something like `getBillingMode` or `billingModeForThroughput`?

backend/service/aws/dynamodb.go
```diff
@@ -118,11 +118,34 @@ func newProtoForIndexStatus(s types.IndexStatus) dynamodbv1.GlobalSecondaryIndex
...
+// if a table is PAY_PER_REQUEST (on demand), it will have 0 RCU/WCU provisioned
+func getBillingMode(t *types.ProvisionedThroughputDescription) types.BillingMode {
+	if (*t.ReadCapacityUnits > 0) || (*t.WriteCapacityUnits > 0) {
+		return types.BillingModeProvisioned
+	}
+	if (*t.ReadCapacityUnits == 0) || (*t.WriteCapacityUnits == 0) {
+		return types.BillingModePayPerRequest
+	}
+
+	return "UNSPECIFIED" // unable to infer what billing mode it is
```
mikecutalo, line 133: Do we not have a type for this one?

angethuy: This check happens before the json response is marshaled into our proto, so I had to use the AWS type enums here since I'm returning an AWS type. [AWS BillingMode enum](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/dynamodb@v1.5.1/types#BillingMode) only has `provisioned` and `pay_per_request` ): 

In theory this check shouldn't ever return UNSPECIFIED because that would mean we're missing both `BillingMode` and `ProvisionedThroughput` from AWS's DescribeTable response

mikecutalo: Alright lets just move this to a const then

PR #1751, opened 2021-09-20
backend: add user groups to auth token

dschaller:
### Description

Add support for appending user groups to auth token

### Testing Performed

Will include unit


backend/service/authn/oidc.go
```diff
@@ -389,6 +408,10 @@ func NewOIDCProvider(ctx context.Context, config *authnv1.Config, tokenStorage S
 		scopes = defaultScopes
 	}
 
+	if c.GroupsFromUserinfo {
+		scopes = append(scopes, "groups")
+	}
```
danielhochman, line 413: i don't remember if this is still required with the separate request to userinfo

dschaller: yeah it is

PR #1752, opened 2021-09-21
core: update CardHeader styling

scarlettperry:
### Description
PR updates the card header styling based on the new changes from Design. High-level callouts:
* card header min height is now 48px
* there's less padding on the card header
* the avatar icon is smaller
* the vertical divider height is smaller

I also updated the typography component because it was missing the body4 variant which is used in the card header.

### Testing Performed
locally (mock cards are included in the PR if you'd like to test as well)

before
Screen Shot 2021-09-21 at 9 48 31 AM

after

Screen Shot 2021-09-27 at 4 33 57 PM

GIF (card header padding is outdated on this gif. this is just to show the changes to the collapse icon)
chevrons


frontend/packages/core/src/card.tsx
```diff
@@ -39,26 +39,33 @@ const StyledCardHeaderContainer = styled.div({
...
+
 // TODO: use material ui avatar component and implement figma design
-const StyledCardHeaderAvatar = styled.div({
-  width: "32px",
-  height: "32px",
+const StyledHeaderAvatar = styled.div({
+  width: "24px",
+  height: "24px",
   marginRight: "16px",
+  fontSize: "18px",
```
scarlettperry, line 61: fyi didn't use the typography component b/c we don't have a matching variant. (font sizes for icons in the core design system are 18px, 24px, and 36x)

frontend/packages/core/src/typography.tsx
```diff
@@ -62,9 +62,14 @@ const STYLE_MAP = {
...
   body3: {
+    size: "14",
+    weight: REGULAR,
+    lineHeight: "18",
+  },
+  body4: {
     size: "12",
     weight: REGULAR,
-    lineHeight: "20",
+    lineHeight: "16",
```
scarlettperry, line 72: some discrepancies with the core design system. `body3` was using `body4` styling and we were missing the `body4` variant

scarlettperry:
since the card header height is now almost half the height than it was previously, the hover state looks too big IMO (we're using the IconButton's default height and width of 48px). What do you think?

Screen Shot 2021-09-21 at 10 42 47 AM

danielhochman:
Can you reduce the space between the icon and title?

danielhochman:
I'd vote we move to a dual-state icon (`_` for collapse, `+` for grow, or chevrons)

and agreed i would reduce the hover state size by as much as possible

danielhochman (approved):
lgtm, with some small suggestions that could come in a separate PR

smonero (approved):

scarlettperry:
> I'd vote we move to a dual-state icon (`_` for collapse, `+` for grow, or chevrons)
> 
> and agreed i would reduce the hover state size by as much as possible

added dual-state icons (chevrons) and reduced icon button to 32x32 (confirmed w Kanchan)

danielhochman (approved):
very nice!

PR #1753, opened 2021-09-21
tools: update golangci-lint 1.42.1

mikecutalo:
### Description
Updating `golangci-lint` to the latest, this required some minor tweaks but nothing major.

Summary of changes
* Now using `sha256sum` as its provide upstream so we dont need to compute `MD5` our selves when checking the binary
* replaced deprecated`scopelint` with supported `exportloopref`
```
WARN [runner] The linter 'scopelint' is deprecated (since v1.39.0) due to: The repository of the linter has been deprecated by the owner.  Replaced by exportloopref.
```

* Fixed a couple new `gocritic` lints
```
service/chaos/experimentation/experimentstore/experiment_run_details.go:25:16: appendAssign: append result not assigned to the same slice (gocritic)
	properties := append(runProperties, configProperties...)
	              ^
service/chaos/experimentation/experimentstore/list_view.go:25:16: appendAssign: append result not assigned to the same slice (gocritic)
	properties := append(runProperties, configProperties...)
```

### Testing Performed
Locally & CI.

```
➜ clutch git:(main) make backend-lint
tools/golangci-lint.sh run
info: Downloading golangci-lint v1.42.1 to build environment
```

### DIFF
--- a/backend/.golangci.yml
+++ b/backend/.golangci.yml
@@ -3,21 +3,21 @@ linters:
   disable-all: true
   enable:
   - deadcode
-  - gci
   - errcheck
+  - exportloopref
+  - gci
   - gocritic
   - goimports
   - gosec
   - gosimple
   - govet
   - ineffassign
-  - scopelint
   - staticcheck
   - structcheck
   - typecheck
   - unused
-  - whitespace
   - varcheck
+  - whitespace
 linters-settings:
   gocritic:
     disabled-checks:
--- a/backend/cmd/assets/placeholder.go
+++ b/backend/cmd/assets/placeholder.go
@@ -1,4 +1,4 @@
-// +build !withAssets
+//go:build !withAssets
 
 package assets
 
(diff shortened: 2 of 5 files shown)


backend/.golangci.yml
```diff
@@ -3,21 +3,21 @@ linters:
   disable-all: true
   enable:
   - deadcode
-  - gci
   - errcheck
+  - exportloopref
```
danielhochman, line 7: may want to use https://github.com/kyoh86/looppointer instead, it's a stricter version https://github.com/kyoh86/looppointer#sensing-policy

backend/cmd/assets/placeholder.go
```diff
@@ -1,4 +1,4 @@
-// +build !withAssets
+//go:build !withAssets
```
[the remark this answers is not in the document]
danielhochman: this requires devs to be on 1.17 though, is that right? i think i'd rather hang on to the old syntax for a bit if possible to give people a change to upgrade.

backend/.golangci.yml
mikecutalo: This is not a supported linter by upstream golangci-lint, we could use it but would require more work to compile if I'm understanding the private plugin system correctly.

danielhochman: ah carry on then, we can take our chances with the less strict version

mikecutalo:
Going to hold on this while we create a preflight script that checks the development requirements #1639.

mikecutalo:
Now that the preflight script has landed were going to move forward on this.

PR #1757, opened 2021-09-22
frontend: metadata table should not update values if errors are present

dschaller:
### Description

If a metadata table row has validation errors it should not update the source field.

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -140,6 +140,8 @@ interface MutableRowProps extends ImmutableRowProps {
 const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, validation }) => {
   const error = validation.errors?.[data.name];
 
+  // intercept the update callback to prevent updates if there are form errors present
+  // based on the validation.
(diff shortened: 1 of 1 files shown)


danielhochman:
Conditional data binding seems like it could lead to strange bugs. User sees one value but another is actually present in the data. Why not block submission instead?

dschaller:
The problem is that the callback can do anything (in this case mutate the original data) which would allow form submission since the form is re-rendered upon data update

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -140,6 +140,9 @@ interface MutableRowProps extends ImmutableRowProps {
 const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, validation }) => {
   const error = validation.errors?.[data.name];
 
+  const updateCallback = (e: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) =>
+    error ? () => {} : onUpdate(e);
```
danielhochman, line 144: add comment

danielhochman (approved):

mikecutalo (approved):

PR #1759, opened 2021-09-23
frontend: ignore non-existent projects in dash

dschaller:
### Description

Dash: fix lookup of projects that do not exist in response.

notFound

### Testing Performed

manual

### DIFF
--- a/frontend/workflows/projectSelector/src/selector-reducer.tsx
+++ b/frontend/workflows/projectSelector/src/selector-reducer.tsx
@@ -1,4 +1,5 @@
 import type { clutch as IClutch } from "@clutch-sh/api";
+import type { ClutchError } from "@clutch-sh/core";
 import _ from "lodash";
 
 import {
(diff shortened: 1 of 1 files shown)


scarlettperry (approved):
tested locally & lgtm!

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -207,16 +208,39 @@ const selectorReducer = (state: State, action: Action): State => {
...
-      */
-      return { ...state, loading: false, error: action?.payload?.result };
+    case "HYDRATE_ERROR": {
+      /**
+       * TODO: do we want to handle the error state differently? For example, when we render the error on the UI,
+       * it won't disapper unless there's a successful API call or if the user refreshes the page. If a user performs other
+       * actions, such as use the toggle/checkbox/ etc. the error message will be still be on the page
+       */
+      const err = action?.payload?.result as ClutchError;
+      // if (grpcCodeToHttpCode(err?.code) !== 404) {
```
mikecutalo, line 218: remove?
```suggestion
```

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -207,16 +208,39 @@ const selectorReducer = (state: State, action: Action): State => {
...
+      if (err.status.code !== 404) {
+        return { ...state, loading: false, error: action?.payload?.result };
+      }
+      const errorMsg = err.message;
+      const validState = { ...state };
+      let projects = validState[Group.PROJECTS];
+      if (errorMsg) {
+        // this message is guaranteed to follow this format from the server
+        const ERROR_FMT = /unable to find project: (.+)/;
```
mikecutalo, line 227: Why do we need to parse the error message? Currently only one 404 case can be hit, when a project is not found. Is it too much to assume that if we see a 404 then this is the case?

dschaller: We need to parse it to extract just the service name that was searched for in order to remove it from the state

mikecutalo: Ah alright i see now thanks

mikecutalo (approved):

PR #1761, opened 2021-09-24
backend: add support for dotenv

dschaller:
### Description

Add support for dotenv files to the backend. This will always allow local env variable to take precedence (either through the command line or `.env.local`). In development, `.env.development` will be loaded, if present.

### Testing Performed

manual

### DIFF
--- a/backend/.air.toml
+++ b/backend/.air.toml
@@ -9,7 +9,7 @@ tmp_dir = ".air"
 # Just plain old shell command. You could use `make` as well.
 cmd = "go build -o .air/clutch ."
 # Binary file yields from `cmd`.
-bin = ".air/clutch"
+bin = ".air/clutch -dev"
 # Watch these filename extensions.
 include_ext = ["go"]
 # Ignore these filename extensions or directories.
--- a/backend/.air.toml
+++ b/backend/.air.toml
@@ -9,7 +9,7 @@ tmp_dir = ".air"
 # Just plain old shell command. You could use `make` as well.
 cmd = "go build -o .air/clutch ."
 # Binary file yields from `cmd`.
-bin = ".air/clutch -dev"
+bin = ".air/clutch -env .env.local"
 # Watch these filename extensions.
 include_ext = ["go"]
 # Ignore these filename extensions or directories.
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -39,7 +39,7 @@ type Flags struct {
 	ConfigPath string
 	Template   bool
 	Validate   bool
-	EnvFile    envFiles
+	EnvFiles   envFiles
 }
 
 // Link register the struct vars globally for parsing by the flag library.
@@ -47,7 +47,7 @@ func (f *Flags) Link() {
 	flag.StringVar(&f.ConfigPath, "c", "clutch-config.yaml", "path to YAML configuration")
 	flag.BoolVar(&f.Template, "template", false, "executes go templates on the configuration file")
 	flag.BoolVar(&f.Validate, "validate", false, "validates the configuration file and exits")
-	flag.Var(&f.EnvFile, "env", "path to additional .env files to load")
+	flag.Var(&f.EnvFiles, "env", "path to additional .env files to load")
 }
 
 // Parse command line arguments.
--- a/backend/gateway/config.go
+++ b/backend/gateway/config.go
@@ -28,6 +28,7 @@ type Flags struct {
 	ConfigPath string
 	Template   bool
 	Validate   bool
+	DevMode    bool
 }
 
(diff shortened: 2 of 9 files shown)


backend/gateway/gateway.go
```diff
@@ -39,7 +41,27 @@ type ComponentFactory struct {
...
+		tmpLogger := newTmpLogger().With(zap.String("file", filename))
+
+		if filename == ".env.development" && !f.DevMode {
+			continue
+		}
+		p, err := filepath.Abs(filename)
+		if err != nil {
+			tmpLogger.Fatal("parsing .env file failed", zap.Error(err))
+		}
+		// nolint
```
smonero, line 58: should we add a comment that explains why we are nolinting here?

smonero (approved):

backend/gateway/config.go
```diff
@@ -28,13 +28,15 @@ type Flags struct {
...
 	Validate   bool
+	DevMode    bool
 }
 
 // Link register the struct vars globally for parsing by the flag library.
 func (f *Flags) Link() {
 	flag.StringVar(&f.ConfigPath, "c", "clutch-config.yaml", "path to YAML configuration")
 	flag.BoolVar(&f.Template, "template", false, "executes go templates on the configuration file")
 	flag.BoolVar(&f.Validate, "validate", false, "validates the configuration file and exits")
+	flag.BoolVar(&f.DevMode, "dev", false, "run clutch in development mode")
```
danielhochman, line 39: this is confusing to me because dev mode usually implies hot restart and other things, (e.g. our target `make backend-dev`)

should we just thread through the specific filename of the env file? that would work better for keeping random env files in examples directories and such too.

danielhochman (commented):
nice! this will make local development much easier, especially for random scenarios

backend/.air.toml
```diff
@@ -9,7 +9,7 @@ tmp_dir = ".air"
 # Just plain old shell command. You could use `make` as well.
 cmd = "go build -o .air/clutch ."
 # Binary file yields from `cmd`.
-bin = ".air/clutch"
+bin = ".air/clutch -env .env.local"
```
mikecutalo, line 12: Since the user will need to create the `.env.local` manually, wont this fail as the file will not exist?

dschaller: Nope, we intentionally ignore errors when the .env files don't exist as it shouldn't block users since these should be optional

mikecutalo: ok cool i though `filepath.Abs` would check for the file existence but I suppose it does not.

backend/gateway/gateway.go
```diff
@@ -39,7 +41,27 @@ type ComponentFactory struct {
...
+		tmpLogger := newTmpLogger().With(zap.String("file", filename))
+
+		p, err := filepath.Abs(filename)
+		if err != nil {
+			tmpLogger.Fatal("parsing .env file failed", zap.Error(err))
+		}
+		// Ignore lint below as it is ok to to ignore dotenv loads as not all env files are guaranteed
+		// to be present.
+		// nolint
+		godotenv.Load(p)
```
mikecutalo, line 59: I think instead of failing silently we should log a warning at the very least that we weren't able to load said env files, what do you think?

dschaller: Yeah, I originally had that when testing locally but it's kind of noisy when starting the app if you don't have the files. I was trying to follow the react-scripts approach where they don't emit anything.

mikecutalo: Talking offline we decided to log on successful loads of env files, that why at a glance its easy to know whats been loaded.

backend/gateway/gateway.go
```diff
@@ -39,7 +41,30 @@ type ComponentFactory struct {
...
+		p, err := filepath.Abs(filename)
+		if err != nil {
+			tmpLogger.Fatal("parsing .env file failed", zap.Error(err))
+		}
+		// Ignore lint below as it is ok to to ignore dotenv loads as not all env files are guaranteed
+		// to be present.
+		// nolint
+		err = godotenv.Load(p)
+		if err == nil {
+			tmpLogger.Info("successfully loaded environment variables")
```
mikecutalo, line 61: add which file was successful in this log as there can be many

dschaller: It's part of the [logger](https://github.com/lyft/clutch/pull/1761/files/c14ffdfbf3e33a1f644c3fda4328302e37faf62d#diff-df82dea03f5bfbf3e5b85c91924864bbc4c9d331eef1b39a6d00f2fae71cbc76R50)

mikecutalo (approved):

PR #1762, opened 2021-09-24
backend: add ability to inherit configs

### LINKED ISSUE #321: Compose configuration files
**Description**
Add the ability to compose the application config from multiple config files.

e.g. `go run main.go -c backend/clutch-config.yaml -c examples/experimentation/clutch-config.yaml`

This would allow for isolated configurations and reduced duplication.

Complexity [S/M/L]: S

dschaller:
### Description

Adds support to extend backend configurations.

For example:
`clutch-config.yaml`
```yaml
extends: clutch-config-base.yaml
gateway:
  listener:
    tcp:
      port: 8081
```

`clutch-config-base.yaml`
```yaml
gateway:
  logger:
    pretty: true
    level: DEBUG
  accesslog:
    # log http 5xx errors by default
    status_code_filters:
      # Unknown - http.StatusInternalServerError
      - equals: 2
      # DeadlineExceeded - http.StatusInternalServerError
      - equals: 4
      # Unimplemented - http.StatusNotImplemented
      - equals: 12
      # Unavailable - http.StatusServiceUnavailable
      - equals: 14
  stats:
    flush_interval: 1s
    log_reporter: {}
  timeouts:
    default: 15s
  middleware:
    - name: clutch.middleware.stats
    - name: clutch.middleware.validate
  listener:
    tcp:
      address: 0.0.0.0
      port: 8080
      secure: false
modules:
  - name: clutch.module.assets
  - name: clutch.module.dynamodb
  - name: clutch.module.healthcheck
```

### Testing Performed

manual

### GitHub Issue

Fixes #321


api/config/gateway/v1/gateway.proto
```diff
@@ -16,6 +16,8 @@ message Config {
   repeated Service services = 2;
   repeated Resolver resolvers = 3;
   repeated Module modules = 4;
+
+  string extends = 5;
```
danielhochman, line 20: i think i would make this repeated so we can aggregate multiple configs easily rather than have to chain configs.

danielhochman (commented):
awesome, one major suggestion

smonero:
should we guard against the case where configs have circular dependencies or is that an adversarial edge case?

dschaller:
This should do that here: https://github.com/lyft/clutch/pull/1762/files#diff-f134efd6fc1e49d005f8487c5c6591dbbf1880450d754149b7dac3b39aaa3c6bR75-R78 :)

dschaller:
@smonero ah yes good catch. I actually should have this solved with the changes from Daniel suggestion to have configs be composable.

smonero (approved):

PR #1764, opened 2021-09-27
tools: preflight version check script

mikecutalo:
### Description
This script checks that your environment has the basic set of requirements to develop Clutch.
We do not stop checks on failure, instead we do a full check of all required dependencies and give the user a summary of the results.

The script can take two params `frontend` or `backend`, that will run a subset of checks for the domain. By default if nothing is specified all checks are run.

### Testing Performed
Locally.

My local versions
```bash
➜  go version
go version go1.17.1 darwin/amd64

➜  node --version
v14.5.0

➜  yarn --version
1.22.11
```

Successful check
```bash
➜  ./tools/prefight-checks.sh
➜  echo $?
0
```

Increase the min requirements to test the failure condition.
```bash
➜ ./tools/prefight-checks.sh
golang version must be >= 2.17.0
node version must be >= 15.0.0
yarn version must be >= 2.22.11

Please refer to the development requirments https://clutch.sh/docs/getting-started/local-build/#requirements
➜ echo $?
1
```

### GitHub Issue
Fixes #

Development requirement checks #1639


mikecutalo:
Thinking about this some more, I want to add two flags `--backend` && `--frontend` so we can run the preflight script for before all make commands.

tools/preflight-checks.sh
```diff
@@ -0,0 +1,106 @@
...
+    elif [ "$1" == "frontend" ]; then
+      frontend
+    else
+      backend
+      frontend
+    fi
+  fi
+
+  if [ "$did_checks_pass" = false ] ; then
+    printf "\nPlease refer to the development requirments https://clutch.sh/docs/getting-started/local-build/#requirements"
```
smonero, line 101:
```suggestion
    printf "\nPlease refer to the development requirements https://clutch.sh/docs/getting-started/local-build/#requirements"
```

tools/preflight-checks.sh
```diff
@@ -0,0 +1,106 @@
...
+    current_version=$(yarn --version)
+    if ! is_version_ok $MIN_YARN_VERSION "$current_version:1"; then
+      echo "yarn version must be >= $MIN_YARN_VERSION"
+      did_checks_pass=false
+    fi
+  fi
+}
+
+main() {
+  # always check OS level requirments
```
smonero, line 86:
```suggestion
  # always check OS level requirements
```

tools/preflight-checks.sh
```diff
@@ -0,0 +1,106 @@
...
+  current_version=$2
+  if [ "$(printf '%s\n' "$required_version" "$current_version" | sort -V | head -n1)" = "$required_version" ]; then 
+    return 0
+  else
+    return 1
+  fi
+}
+
+os() {
+  # If were on OSX lets check for brew and coreutils as they are requriments
```
smonero, line 26:
```suggestion
  # If we're on OSX lets check for brew and coreutils as they are requirements
```

Makefile
```diff
@@ -181,3 +181,14 @@ dev-k8s-up:
...
+.PHONY: preflight-checks-frontend
+preflight-checks-frontend:
+	@tools/preflight-checks.sh frontend
+
+.PHONY: preflight-checks-backend
+preflight-checks-backend:
+	@tools/preflight-checks.sh backend
+
+.PHONY: preflight-checks-all
+preflight-checks-all:
```
dschaller, line 193: nit
```suggestion
preflight-checks:
```

tools/preflight-checks.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Will be set to false if any of the steps fail
+did_checks_pass=true
+
+# Minimum versions
+MIN_GO_VERSION="1.17"
+MIN_NODE_VERSION="14.0.0"
```
dschaller, line 9: This might be too much but it would be nice to parse these from a centralized place (e.g. package.json also has these values and the yarn install script)

tools/preflight-checks.sh
```diff
@@ -0,0 +1,104 @@
...
+  current_version=$2
+  if [ "$(printf '%s\n' "$required_version" "$current_version" | sort -V | head -n1)" = "$required_version" ]; then 
+    return 0
+  else
+    return 1
+  fi
+}
+
+os() {
+  # If we're on OSX lets check for brew and coreutils as they are requirements
```
dschaller, line 26: Should we update the docs to mention running this script to see if your machine is compatible?

tools/preflight-checks.sh
```diff
@@ -0,0 +1,104 @@
...
+    # check brew is installed
+    if command -v brew --version &> /dev/null; then
+      # check if coreutils is installed
+      if ! brew ls --versions coreutils > /dev/null; then
+        echo "coreutils is not installed, this is a required dependency."
+        echo "install by running [brew install coreutils]"
+        did_checks_pass=false
+      fi
+    else
+        echo "brew is not installed, unable to verify coreutils dependency."
```
dschaller, line 38: maybe point users to the homebrew docs to install or include the curl command?

tools/preflight-checks.sh
```diff
@@ -0,0 +1,104 @@
...
+        did_checks_pass=false
+    fi
+  fi
+}
+
+backend() {
+  if ! command -v go -v &> /dev/null; then
+    echo "golang is not installed, this is a required dependency."
```
dschaller, line 46: This could just be a path issue. Maybe mention that it's either not installed or not in the current path?

tools/preflight-checks.sh
mikecutalo: Yeah sure

tools/preflight-checks.sh
mikecutalo: yeah we could do this in a follow up and require all scripts to use the versions defined in a central place

dschaller (approved):

smonero (approved):

PR #1766, opened 2021-09-30
core: add "see more" action to CardContent

scarlettperry:
### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/card.tsx
+++ b/frontend/packages/core/src/card.tsx
@@ -8,9 +8,12 @@ import {
   Divider,
   Grid,
 } from "@material-ui/core";
+import KeyboardArrowDownIcon from "@material-ui/icons/KeyboardArrowDown";
+import KeyboardArrowUpIcon from "@material-ui/icons/KeyboardArrowUp";
 import type { SpacingProps as MuiSpacingProps } from "@material-ui/system";
 import { spacing } from "@material-ui/system";
 
+import { IconButton } from "./button";
 import { Typography, TypographyProps } from "./typography";
 
 // TODO: seperate out the different card parts into various files
@@ -126,20 +129,68 @@ const BaseCardContent = styled.div<SpacingProps>`
   ${spacing}
 `;
 
-const StyledCardContent = styled(BaseCardContent)({
+const StyledCardContentContainer = styled.div({
   "> .MuiPaper-root": {
     border: "0",
     borderRadius: "0",
   },
+  overflow: "hidden",
+});
+
+const StyledExpandButton = styled(IconButton)({
+  width: "32px",
+  height: "32px",
+  color: "#3548D4",
 });
 
 interface CardContentProps extends SpacingProps {
   children?: React.ReactNode | React.ReactNode[];
+  showExpand?: boolean;
 }
 
-const CardContent = ({ children, ...props }: CardContentProps) => (
-  <StyledCardContent {...props}>{children}</StyledCardContent>
-);
+// TODO: make card action component with spacing
+// TODO: make max height a prop
+// TODO: set maxheight to inherit
+// TODO: how to make the bottom card action area more flexible
+const CardContent = ({ children, showExpand = false, ...props }: CardContentProps) => {
+  const ref = React.useRef(null);
+  const [shouldShowExpand, setShouldShowExpand] = React.useState(false);
+  const [expanded, setExpanded] = React.useState(true);
+
+  const maxHeight = 400;
+  const MAX_POSSIBLE_HEIGHT = 1000;
+
+  React.useEffect(() => {
+    if (ref.current.scrollHeight > maxHeight) {
+      setShouldShowExpand(true);
+      setExpanded(false);
+    }
+  }, [maxHeight]);
+
+  return (
+    <BaseCardContent {...props} ref={ref}>
+      <StyledCardContentContainer style={{ maxHeight: expanded ? MAX_POSSIBLE_HEIGHT : maxHeight }}>
+        {children}
+      </StyledCardContentContainer>
+      {showExpand && shouldShowExpand && (
+        <CardActionArea style={{ padding: "8px" }} onClick={() => setExpanded(!expanded)}>
+          <Grid container alignItems="center" spacing={1}>
+            <Grid item>
+              <Typography variant="body4" color="#3548D4">
+                See More
+              </Typography>
+            </Grid>
+            <Grid item>
+              <StyledExpandButton variant="neutral">
+                {expanded ? <KeyboardArrowUpIcon /> : <KeyboardArrowDownIcon />}
+              </StyledExpandButton>
+            </Grid>
+          </Grid>
+        </CardActionArea>
+      )}
+    </BaseCardContent>
+  );
+};
 
 const StyledLandingCard = styled(Card)({
   border: "none",
--- a/frontend/workflows/projectSelector/src/card.tsx
+++ b/frontend/workflows/projectSelector/src/card.tsx
@@ -56,7 +56,9 @@ const Card = ({ avatar, children, error, isLoading, summary, title }: CardProps)
           </StyledProgressContainer>
         </CardHeader>
         {expanded && (
-          <CardContent padding={0}>{error ? <Error subject={error} /> : children}</CardContent>
+          <CardContent padding={0} showExpand>
+            {error ? <Error subject={error} /> : children}
+          </CardContent>
         )}
       </ClutchCard>
     </Grid>
--- a/frontend/workflows/projectSelector/src/dash.tsx
+++ b/frontend/workflows/projectSelector/src/dash.tsx
@@ -1,8 +1,10 @@
 import * as React from "react";
+import { Table, TableRow, Typography } from "@clutch-sh/core";
 import styled from "@emotion/styled";
-import { Box } from "@material-ui/core";
+import { Box, Grid as MuiGrid } from "@material-ui/core";
 import _ from "lodash";
 
+import Card from "./card";
 import { DashDispatchContext, DashStateContext } from "./dash-hooks";
 import ProjectSelector from "./project-selector";
 import type { DashAction, DashState } from "./types";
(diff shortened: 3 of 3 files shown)


frontend/workflows/projectSelector/src/dash.tsx
```diff
@@ -40,7 +46,232 @@ const Dash = ({ children }) => {
       <DashDispatchContext.Provider value={dispatch}>
         <DashStateContext.Provider value={state}>
           <ProjectSelector />
-          <CardContainer>{children}</CardContainer>
+          <CardContainer>
```
scarlettperry, line 49: as usual, nothing in this file will be merged. just for PR reviewers if they'd like to test as well :)

frontend/packages/core/src/card.tsx
```diff
@@ -11,21 +11,14 @@ import {
...
 
 const StyledCard = styled(MuiCard)({
   boxShadow: "0px 4px 6px rgba(53, 72, 212, 0.2)",
   border: "1px solid rgba(13, 16, 48, 0.1)",
-
-  ".MuiCardActionArea-root:hover": {
-    backgroundColor: "#F5F6FD",
-  },
-
-  ".MuiCardActionArea-root:active": {
```
scarlettperry, line 26: decided to style the CardActionArea itself since we're resuing these hover/active states in the "see more" action.

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +119,90 @@ const BaseCardContent = styled.div<SpacingProps>`
...
   "> .MuiPaper-root": {
     border: "0",
     borderRadius: "0",
   },
+  overflow: "hidden",
 });
 
+const BaseCardActionArea = styled(CardActionArea)<SpacingProps>`
+  ${spacing}
+`;
```
scarlettperry, line 132: similar to `CardContent`, spacing on the `CardActionArea` should be configurable and we resuse this component for the landing card

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +119,90 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+
+const StyledExpandButton = styled(IconButton)({
+  width: "32px",
+  height: "32px",
+  color: "#3548D4",
+});
+
+interface CardContentCollapsibleProps {
+  title: string;
+  dualStateIcon?: { true: React.ReactElement; false: React.ReactElement };
```
scarlettperry, line 152: lmk if your thoughts on this. initially also had the option to just pass a single icon but since the designs are dual-state thought maybe we'd want to be opinionated 
```
{
  icon?: React.ReactElement
  dualStateIcon?: ...
}
```

frontend/workflows/projectSelector/src/card.tsx
```diff
@@ -56,7 +56,20 @@ const Card = ({ avatar, children, error, isLoading, summary, title }: CardProps)
...
+            padding={0}
+            collapsible
+            collapseAction={{
+              title: "See More",
+              dualStateIcon: {
+                true: <KeyboardArrowUpIcon />,
+                false: <KeyboardArrowDownIcon />,
+              },
+            }}
+            maxHeight={400}
```
scarlettperry, line 69: The notes on figma say that at the 6th table row we should collapse the card but not all cards will have tables so I think it should be based on at X height we collapse the card. I think 400px looks good, but can get recommendations from Design

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +119,90 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+    }
+  }, [defaultMaxHeight]);
+
+  return (
+    <BaseCardContent {...props} ref={ref}>
+      <StyledCardContentContainer style={{ maxHeight: expanded ? "none" : defaultMaxHeight }}>
+        {children}
+      </StyledCardContentContainer>
+      {collapsible && showExpand && (
+        <StyledCardActionArea padding={1} onClick={() => setExpanded(!expanded)}>
```
scarlettperry, line 188: This isn't on the figma design but I think the entire "see more" area should be a button instead of require the users to have to explicitly click on the IconButton. Lmk what you think

smonero:
awesome! when it is expanded do you think the text should change to "See Less" instead of "See More" ?

scarlettperry:
> awesome! when it is expanded do you think the text should change to "See Less" instead of "See More" ?

thanks for catching that! i'll update

scarlettperry:
latest commit
* adds dualStateTitle
* centers the "see more"
* adds doc strings
* adds default dualStateTitle and dualStateIcon since I think these are pretty standard

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+  color: "#3548D4",
+});
+
+interface CardContentCollapsibleProps {
+  /** The text to show in the collapse action area when the card content is collapsed/not collapsed.
+   * By default, this is "See Less" for true and "See More" for false.
+   */
+  dualStateTitle: { true: string; false: string };
+  /** The icon to show in the collapse action area when the card content is collapsed/not collapsed.
+   * By default, this is a ArrowUp icon for ture and a ArrowDown icon for false.
```
smonero, line 158:
```suggestion
   * By default, this is an ArrowUp icon for true and an ArrowDown icon for false.
```

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+   */
+  dualStateIcon: { true: React.ReactElement; false: React.ReactElement };
+}
+
 interface CardContentProps extends SpacingProps {
   children?: React.ReactNode | React.ReactNode[];
+  /** Make the card content collapse at a set max height. The default is false. */
+  collapsible?: boolean;
+  /** The props for the collapse action if collapisble is enabled */
```
smonero, line 167:
```suggestion
  /** The props for the collapse action if collapsible is enabled */
```

smonero (approved):

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+  collapseAction = {
+    dualStateTitle: { true: "See Less", false: "See More" },
+    dualStateIcon: { true: <KeyboardArrowUpIcon />, false: <KeyboardArrowDownIcon /> },
+  },
+  maxHeight,
+  ...props
+}: CardContentProps) => {
+  const ref = React.useRef(null);
+  const [showExpand, setShowExpand] = React.useState(false);
+  const [expanded, setExpanded] = React.useState(true);
```
dschaller, line 185:
```suggestion
  const [showExpand, setShowExpand] = React.useState<boolean>(false);
  const [expanded, setExpanded] = React.useState<boolean>(true);
```

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+
+  const defaultMaxHeight = maxHeight || "none";
```
dschaller, line 187: Should we make this the default value in the props?

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+  React.useEffect(() => {
+    if (ref.current.scrollHeight > defaultMaxHeight) {
+      setShowExpand(true);
+      setExpanded(false);
+    }
+  }, [defaultMaxHeight]);
+
+  return (
+    <BaseCardContent {...props} ref={ref}>
+      <StyledCardContentContainer style={{ maxHeight: expanded ? "none" : defaultMaxHeight }}>
```
dschaller, line 198: nit: should we be adding these styles to the styled component above?

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+        {children}
+      </StyledCardContentContainer>
+      {collapsible && showExpand && (
+        <StyledCardActionArea padding={1} onClick={() => setExpanded(!expanded)}>
+          <Grid container alignItems="center" spacing={1} justify="center">
+            <Grid item>
+              <Typography variant="body4" color="#3548D4">
```
dschaller, line 205: This is minor but it's a bit odd that the icon weight is heavier than the text. Also, should we disable the hover over on the icon since the whole row highlights also?

dschaller:
Should we make the see more row a bit shorter in height?

scarlettperry:
> Should we make the see more row a bit shorter in height?

yeah agree it was a bit big.

i removed the padding & now the height is 32 (prev was 48)

Screen Shot 2021-10-05 at 11 32 31 AM

if we want it a lil bigger, I can make the padding 6px and the height will be 40
Screen Shot 2021-10-05 at 11 33 26 AM
 
i think no-padding looks good but lmk if you like the 4px version

frontend/packages/core/src/card.tsx
```diff
@@ -126,20 +121,104 @@ const BaseCardContent = styled.div<SpacingProps>`
...
+  width: "32px",
+  height: "32px",
+  color: "#3548D4",
+});
+
+interface CardContentCollapsibleProps {
+  /** The text to show in the collapse action area when the card content is collapsed/not collapsed.
+   * By default, this is "See Less" for true and "See More" for false.
+   */
+  dualStateTitle: { true: string; false: string };
```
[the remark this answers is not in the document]
scarlettperry: this is better, thanks!

frontend/packages/core/src/card.tsx
scarlettperry: updated to set maxHeight's default to `none`

frontend/packages/core/src/card.tsx
scarlettperry: oh yeah true, updated!

frontend/packages/core/src/card.tsx
scarlettperry: > This is minor but it's a bit odd that the icon weight is heavier than the text

I tried changing the font-weight but that didn't do anything. If I make the icon smaller (currently it's 24px) to 20px but I think that then it looks too small in proportion to the text

Screen Shot 2021-10-05 at 12 07 39 PM

dschaller:
I like it without padding 🎉

PR #1772, opened 2021-10-05
backend: add graceful shutdown handling

### LINKED ISSUE #12: gateway: graceful shutdown

dschaller:
### Description

Add graceful shutdown handling to server.

```bash
2021-10-05T15:28:54.310-0700	INFO	gateway/gateway.go:322	listening	{"tcp": {"addr": "0.0.0.0:8080"}}
2021-10-05T15:28:55.300-0700	DEBUG	stats/debug_reporter.go:41	counter	{"name": "clutch.gateway.start", "value": 1, "tags": {}, "type": "counter"}
```

```bash
 > kill -15 9531
```

```bash
2021-10-05T15:29:00.648-0700	DEBUG	gateway/gateway.go:367	server shutdown gracefully
```

### Testing Performed

manual

### GitHub Issue

Fixes #12

### DIFF
--- a/backend/gateway/gateway.go
+++ b/backend/gateway/gateway.go
@@ -359,6 +359,8 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
 
 	<-sc
 
+	signal.Stop(sc)
+
 	// Shutdown timeout should be max request timeout (with 1s buffer).
 	ctxShutDown, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()


backend/gateway/gateway.go
```diff
@@ -75,11 +78,7 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
 	if err != nil {
 		newTmpLogger().Fatal("could not instantiate logger", zap.Error(err))
 	}
-	defer func() {
-		if err := logger.Sync(); err != nil {
-			panic(err)
-		}
-	}()
+	defer logger.Sync()
```
mikecutalo, line 81: How are you handling the error case here now that it's gone?

dschaller: see https://github.com/uber-go/zap/issues/880

backend/gateway/gateway.go
```diff
@@ -340,5 +341,30 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
...
+	sc := make(chan os.Signal, 1)
+	signal.Notify(
+		sc,
+		os.Interrupt,
+		syscall.SIGTERM,
+		syscall.SIGQUIT,
+	)
+
+	go func() {
+		if err = srv.ListenAndServe(); err != http.ErrServerClosed {
```
mikecutalo, line 354: Add a comment here just to explain why we can ignore `http.ErrServerClosed`

mikecutalo:
We talked offline and the last thing that we need is a `signal.Stop(sc)` to stop listening for more signals as we only need to catch one.

backend/gateway/gateway.go
```diff
@@ -340,5 +341,33 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
...
+
+	// Shutdown timeout should be max request timeout (with 1s buffer).
+	ctxShutDown, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	if err = srv.Shutdown(ctxShutDown); err != nil {
+		logger.Fatal("server shutdown failed", zap.Error(err))
+	}
+
+	logger.Debug("server shutdown gracefully")
```
scarlettperry, line 372: should we make this an info log instead? was thinking that'd it be useful to see when testing the change on staging. but i guess the fatal log above would be enough

scarlettperry (approved):

backend/gateway/gateway.go
dschaller: Yeah I wasn't sure if this would be too noisy. I suppose we implicitly log that the server started through the tcp bindings

scarlettperry: sounds good!

PR #1799, opened 2021-10-21
proxy module: add helper for adding ignored headers to Request struct

scarlettperry:
#### Background
Noticed that when we added the host to the header map, the host header wasn't getting sent in the request. Came across https://github.com/golang/go/issues/29865 which mentions that the host added in the header map gets ignored and should instead be added to the `Host` field in the request struct. https://github.com/golang/go/issues/29865

PR adds a new helper to check if the user's header list contains an ignored header. If it does, it get adds to the designated field on the Request struct. For now, PR adds this check for `host` with a TODO to add the other fields that get ignored.

### Testing Performed
Locally with internal repo. Tested the the new service we want to call and the existing pagerduty set up.

### TODOs
- [x] Write tests
- [x] Test internally

### DIFF
--- a/backend/module/proxy/proxy.go
+++ b/backend/module/proxy/proxy.go
@@ -107,7 +107,7 @@ func (m *mod) RequestProxy(ctx context.Context, req *proxyv1.RequestProxyRequest
 		Header: headers,
 	}
 
-	request = addExcludedHeaders(request)
+	addExcludedHeaders(request)
 
 	if req.Request != nil {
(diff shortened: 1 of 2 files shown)


mikecutalo:
Yeah good question, we talked about this internally a bit. I didn't particularly like having a magic string that influenced behavior, or have to do a specific string match which would be straight forward but still a magic match. In my opinion instead of doing it that way, having this configuration be more explicit is more straight forward for end users. 

Although if you feel strongly in a different direction, Im open to talk about it.

PR #1839, opened 2021-11-01
add null type to PopperProps

smonero:
### Description

This is useful to be more like the material-ui popper

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/popper.tsx
+++ b/frontend/packages/core/src/popper.tsx
@@ -89,7 +89,7 @@ export interface PopperProps
   extends Pick<ClickAwayListenerProps, "onClickAway">,
     Pick<MuiPopperProps, "placement"> {
   open: boolean;
-  anchorRef: React.MutableRefObject<HTMLElement>;
+  anchorRef: React.MutableRefObject<HTMLElement> | null;
   children?: React.ReactElement<PopperItemProps> | React.ReactElement<PopperItemProps>[];
   id?: string;
 }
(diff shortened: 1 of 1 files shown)


scarlettperry (approved):

frontend/packages/core/src/popper.tsx
```diff
@@ -89,7 +89,7 @@ export interface PopperProps
   extends Pick<ClickAwayListenerProps, "onClickAway">,
     Pick<MuiPopperProps, "placement"> {
   open: boolean;
-  anchorRef: React.MutableRefObject<HTMLElement>;
+  anchorRef: React.MutableRefObject<HTMLElement> | null;
```
danielhochman, line 92:
```suggestion
  anchorRef?: React.MutableRefObject<HTMLElement>;
```

smonero: https://stackoverflow.com/questions/57243477/can-an-optional-parameter-be-null-in-typescript

Item 5 shows that this approach of using optional to allow for null does not work (it does for `undefined` but not `null`)

danielhochman: ah true, can we have both then? i think both values make sense i.e. `anchorRef?: React.MutableRefObject<HTMLElement> | null`

danielhochman (approved):

frontend/packages/core/src/popper.tsx
smonero: we can

PR #1847, opened 2021-11-03
add MutableRefObject<null> type to Popper prop anchorRef

smonero:
### Description

My previous PR https://github.com/lyft/clutch/pull/1839/files added just the null type, but didn't add the case of where null can be passed into `MutableRefObject` . So this one fixes that.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/popper.tsx
+++ b/frontend/packages/core/src/popper.tsx
@@ -89,7 +89,7 @@ export interface PopperProps
   extends Pick<ClickAwayListenerProps, "onClickAway">,
     Pick<MuiPopperProps, "placement"> {
   open: boolean;
-  anchorRef?: React.MutableRefObject<HTMLElement> | React.MutableRefObject<null> | null;
+  anchorRef: React.MutableRefObject<HTMLElement | null>;
   children?: React.ReactElement<PopperItemProps> | React.ReactElement<PopperItemProps>[];
   id?: string;
 }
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/popper.tsx
```diff
@@ -89,7 +89,7 @@ export interface PopperProps
   extends Pick<ClickAwayListenerProps, "onClickAway">,
     Pick<MuiPopperProps, "placement"> {
   open: boolean;
-  anchorRef?: React.MutableRefObject<HTMLElement> | null;
+  anchorRef?: React.MutableRefObject<HTMLElement> | React.MutableRefObject<null> | null;
```
scarlettperry, line 92: i think this should work
```suggestion
  anchorRef?: React.MutableRefObject<HTMLElement | null> | null;
```

frontend/packages/core/src/popper.tsx
```diff
@@ -89,7 +89,7 @@ export interface PopperProps
   extends Pick<ClickAwayListenerProps, "onClickAway">,
     Pick<MuiPopperProps, "placement"> {
   open: boolean;
-  anchorRef?: React.MutableRefObject<HTMLElement> | null;
+  anchorRef?: React.MutableRefObject<HTMLElement | null> | null;
```
scarlettperry, line 92: wdyt about removing this now since the PR was the correct fix?
```suggestion
  anchorRef?: React.MutableRefObject<HTMLElement | null>;
```

smonero: ok. also how about we remove the question mark as well since we don't need the `undefined` type

scarlettperry: in looking at material ui popper docs, the anchorRef is an optional prop so I think it makes sense to keep the `?`

danielhochman (approved):

frontend/packages/core/src/popper.tsx
smonero: got it

scarlettperry (approved):

PR #1854, opened 2021-11-05
[AWS] multi account config support

mikecutalo:
### Description
This adds configuration support for multiple AWS accounts. This is a backwards compatible change, which was by design as I did not want to completely re-work the configuration.

### Testing Performed
n/a

### GitHub Issue
#1829


api/config/service/aws/v1/aws.proto
```diff
@@ -10,6 +10,11 @@ message Config {
   repeated string regions = 1 [ (validate.rules).repeated = {min_items : 1} ];
   ClientConfig client_config = 2;
   DynamodbConfig dynamodb_config = 3;
+
+  // The current account alias, if this is not set the default will be "default"
+  string account_alias = 4;
```
dschaller, line 15: Is this the primary account?

mikecutalo: Yeah this is the primary account, maybe we rename to include that? `primary_account_alias`?

api/config/service/aws/v1/aws.proto
```diff
@@ -10,6 +10,11 @@ message Config {
   repeated string regions = 1 [ (validate.rules).repeated = {min_items : 1} ];
   ClientConfig client_config = 2;
   DynamodbConfig dynamodb_config = 3;
+
+  // The current account alias, if this is not set the default will be "default"
```
danielhochman, line 14: this is for display purposes? it's a little unclear what it's for.

also accounts have aliases in the ~/.aws/config file which could get confusing...

i wonder if the account will actually already have an assigned alias from the AWS SDK

danielhochman: okay by definition it's the "default" config because we call `LoadDefaultConfig` https://github.com/lyft/clutch/blob/ba8aa49da66f20671cc73e2ee73f84bf380f9465/backend/service/aws/aws.go#L69

and not `LoadSharedConfigProfile`.

so if i had an `~/.aws/config` that looked something like
```toml
[default]
key=123

[staging]
key=456
```

i might set primary_account_alias to `staging` thinking it will use the `staging` profile cred. we don't support this currently but i think we might want to consider how we would incorporate it into the config and not get it confused with this alias that we're currently adding.

mikecutalo: `primary_account_alias` is not just for display purposes, it will also be used when trying to resolve a resource to indicated which account / region config we should be using if were not fanning out. EG: if `primary_account_alias: production` then the resolver patter for a resource would look like so `production/us-east-1/my-asg`.

I think I see what you are saying, if you wanted to specify which`~/.aws/config` profile you want clutch to use by "default" vs what is the account alias for the account your currently running.

If thats the case I would propose something like this
```
primary_account_alias_display_name  # name that will be used when looking for a resource
use_account_profile_from_aws_config # which would use a profile from the aws config
```

mikecutalo: Updated, let me know what you think of the names. I also added the aws config profile to avoid confusion and put a TODO on it.

danielhochman (approved):

api/config/service/aws/v1/aws.proto
```diff
@@ -10,6 +10,20 @@ message Config {
...
+  // The current account alias display name, if this is not set the default will be "default"
+  // The account alias display name will be used when resolving resources.
+  // EG: if this is set to "production" a resource would be referenced like so
+  // "production/us-east-1/my-asg"
+  string primary_account_alias_display_name = 4;
+
+  // If you are using an aws configuration file, this overrides the default profile that is loaded.
+  // TODO: This is currently not implemented, but was created to prevent confusion between
+  // primary_account_alias_display_name
+  string use_account_profile_from_aws_config = 5;
```
dschaller, line 23: talked online: suggested `aws_config_profile_name`

dschaller (approved):

Issue #1797, opened 2021-10-19
core: implement a feedback/NPS system 


scarlettperry:
**Description**
Implement a feedback component on both the header and last step of the Wizard, where users can rate their experience using Clutch/a specific workflow. Backend will persist the feedback submission.

Complexity [S/M/L]: M

**Wireframes, Mockups, or Screenshots (if applicable)**

PR #1859, opened 2021-11-08
feedback: add GetSurveys module

scarlettperry:
### Description
PR adds the API implementation for `GetSurveys` which returns the origin-specific survey details. PR also makes some small related proto updates.

### Testing Performed
locally

example responses

request for either "HEADER" or the numerical value `1`
```
{
    "originSurvey": {
        "HEADER": {
            "prompt": "Rate your experience using Clutch",
            "freeformPrompt": "What's on your mind?",
            "ratingOptions": {
                "one": "bad",
                "two": "ok",
                "three": "great"
            }
        }
    }
}
```

request for 2 origins
```
{
    "originSurvey": {
        "HEADER": {
            "prompt": "Rate your experience using Clutch",
            "freeformPrompt": "What's on your mind?",
            "ratingOptions": {
                "one": "bad",
                "two": "ok",
                "three": "great"
            }
        },
        "WIZARD": {
            "ratingOptions": {
                "one": "bad",
                "two": "ok",
                "three": "great"
            }
        }
    }
}
```

### GitHub Issue

### TODOs
- [x] Add tests


api/config/module/feedback/v1/feedback.proto
```diff
@@ -20,14 +20,16 @@ message Survey {
   RatingOptions rating_options = 3 [ (validate.rules).message = {required : true} ];
 }
 
-message SurveyOrigin {
-  oneof type {
-    option (validate.required) = true;
+enum Origin {
```
scarlettperry, line 23: wanted to use the same enum defined here https://github.com/lyft/clutch/blob/main/api/feedback/v1/feedback.proto#L21-L26 but wasn't able to import it so I tried the method here https://stackoverflow.com/questions/8553147/is-it-possible-to-use-external-enums-outside-message-definitions-with-protocol which also didn't work. Example of what I tried for the latter approach: I had this import defined at the top of the file `import "/feedback/v1/enums.proto";` and then imported the enum like so `clutch.feedback.v1.<enum name>` but got undefined errors. am I missing something? Also, is it good practice to try to reuse the enum?

api/config/module/feedback/v1/feedback.proto
```diff
@@ -20,14 +20,16 @@ message Survey {
...
+  // currently these are the supported placements for feedback
+  ORIGIN_UNSPECIFIED = 0;
+  HEADER = 1;
+  WIZARD = 2;
+}
```
scarlettperry, line 28: wanted to use the same enum defined here https://github.com/lyft/clutch/blob/main/api/feedback/v1/feedback.proto#L21-L26 but wasn't able to import it so I tried the method here https://stackoverflow.com/questions/8553147/is-it-possible-to-use-external-enums-outside-message-definitions-with-protocol which also didn't work. Example of what I tried for the latter approach: I had this import defined at the top of the file `import "/feedback/v1/enums.proto";` and then imported the enum like so `clutch.feedback.v1.<enum name>` but got undefined errors. am I missing something? Also, is it good practice to try to reuse the enum?

api/feedback/v1/feedback.proto
```diff
@@ -31,8 +31,7 @@ message GetSurveysRequest {
     min_items : 1,
     items : {enum : {defined_only : true, not_in : 0}}
   } ];
-  // future: have rules on whether a user should see the feedback survey
-  string user = 2;
```
scarlettperry, line 35: some proto cleanup. decided to remove the fields we aren't going to use in the V1

api/config/module/feedback/v1/feedback.proto
```diff
@@ -20,14 +20,16 @@ message Survey {
...
 
-    // currently these are the supported origins for feedback
-    Survey header = 1 [ (validate.rules).message = {required : true} ];
-    Survey wizard = 2 [ (validate.rules).message = {required : true} ];
```
scarlettperry, line 29: changed it up a bit b/c in creating the survey lookup map, realized with this pattern i'd either have to do case switches on the key type, which would mean remembering to update `NewSurveyLookup` when a new placement is added. I could also have done a reflect.MapKeys() instead but with the new approach felt cleaner

backend/module/feedback/feedback.go
```diff
@@ -0,0 +1,116 @@
...
+	results := make(map[string]*feedbackv1.Survey)
+
+	// this scenario shouldn't happen as the feedback config fields cannot be empty
+	if len(m.surveyMap.surveys) == 0 {
+		return nil, status.Errorf(codes.NotFound, "survey questions were not found")
+	}
+
+	for _, origin := range req.Origins {
+		v, ok := m.surveyMap.GetConfigSurveys(origin)
+		if ok {
```
danielhochman, line 65: nit
```
if !ok {
  continue
}
... rest of code
```

api/config/module/feedback/v1/feedback.proto
danielhochman: this should work:
```proto
import "feedback/v1/feedback.proto";

message SurveyOrigin {
  clutch.feedback.v1.Origin origin = 1 [ (validate.rules).enum = {defined_only : true} ];
  Survey survey = 2 [ (validate.rules).message = {required : true} ];
}
```

danielhochman (commented):
lgtm at first glance, quick comments and can give a final pass

api/config/module/feedback/v1/feedback.proto
scarlettperry: this is what I had tried originally but in trying it again, I realized that the field validation on the import was causing the issue. i removed the validation and now it works. confirmed tho even without the validation, the config parsing will value if we pass in an undefined enum.

this was the output of the pb.validate.go file and the errors
````
	if _, ok := feedbackv1.Origin_name[int32(m.GetOrigin())]; !ok {
		err := SurveyOriginValidationError{
			field:  "Origin",
			reason: "value must be one of the defined enum values",
		}
		if !all {
			return err
		}
		errors = append(errors, err)
	}
````

For this line `if _, ok := feedbackv1.Origin_name[int32(m.GetOrigin())]; !ok {` the errors were `assignment mismatch: 2 variables but 1 value` and `undefined: feedbackv1`

didn't come across a related issue when doing a quick search so perhaps something to do with our proto compile script https://github.com/lyft/clutch/blob/main/tools/compile-protos.sh.

Comment (danielhochman) [reply api/config/module/feedback/v1/feedback.proto]: if feedbackv1 is undefined that means the go_package directive is probably incorrect in one of the proto files

Comment (scarlettperry) [reply api/config/module/feedback/v1/feedback.proto]: go_package imports for both follow our usual pattern - [api](https://github.com/lyft/clutch/blob/main/api/feedback/v1/feedback.proto#L5) and [config](https://github.com/lyft/clutch/blob/main/api/config/module/feedback/v1/feedback.proto#L5) but looks like the issue above stemmed from both having the same package name `feedbackv1`. renamed the config one, added back the validation on the fields, and thinks work now.

File: backend/module/feedback/feedback.go [hunk 6]
```diff
@@ -0,0 +1,117 @@
...
+	feedbackv1.RegisterFeedbackAPIServer(r.GRPCServer(), m)
+	return r.RegisterJSONGateway(feedbackv1.RegisterFeedbackAPIHandler)
+}
+
+func (m *mod) GetSurveys(tx context.Context, req *feedbackv1.GetSurveysRequest) (*feedbackv1.GetSurveysResponse, error) {
+	results := make(map[string]*feedbackv1.Survey)
+
+	// this scenario shouldn't happen as the feedback config fields cannot be empty
+	if len(m.surveyMap.surveys) == 0 {
+		return nil, status.Errorf(codes.NotFound, "survey questions were not found")
```
mikecutalo, line 60: nit: add error log here and below

mikecutalo (approved):
Looking awesome, left a small comment but will approve.

backend/module/feedback/feedback.go
scarlettperry: ah thanks! forgot about error logs, will update.

danielhochman (approved):

PR #1863, opened 2021-11-09
IAM Client + SimulateCustomPolicy Simulator

faris:
IAM client introduction + adding SimulateCustomPolicy API's to Clutch

Description
Adding iam client (that allows users to programmatically test aws policies) as well as adding a wrapper around SimulateCustomPolicy

Testing Performed
Unit tests + Manual Tests (consuming API's in another clutch instance)


backend/service/aws/iam.go
```diff
@@ -0,0 +1,16 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/iam"
+)
+
+func (c *client) SimulateCustomPolicy(ctx context.Context, region string, customPolicySimulatorParams *iam.SimulateCustomPolicyInput) (*iam.SimulateCustomPolicyOutput, error) {
+	rc, err := c.getRegionalClient(region)
```
ayaxpeleo, line 10: just curious, `region` is not used in the AWS IAM client, afaik. But I see this is a general field for clutch clients, is it just ignored internally?

faris: It is used internally, when configuring a client with iam.NewFromConfig(regionCfg). https://github.com/lyft/clutch/blob/main/backend/service/aws/aws.go#L94-L100

We set the client to use a region. When we consume the API we can call it like this: ```c.awsClient.SimulateCustomPolicy(...,reigon,...)``` and it will select the client with the reigon already configured. 

See: 
https://github.com/lyft/clutch/blob/main/backend/service/aws/aws.go#L71
https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/

The reason this is done is because some services are only available in one location, and the fact that it helps a bit with latency when doing AWS operations.

ayaxpeleo: The python library ([boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#client)) obviates the region field, so I wondered if this would do the same.

ayaxpeleo (approved):

backend/service/aws/iam_test.go
```diff
@@ -0,0 +1,92 @@
...
+		ResourceArns:    []string{"arn:aws:s3:::a/*"},
+		ResourcePolicy:  aws.String(`{}`),
+	}
+
+	output1, err1 := c.SimulateCustomPolicy(context.Background(), "us-east-1", simulationInput)
+	assert.Nil(t, output1)
+	assert.Error(t, err1)
+
+	// Test unknown region
+	output2, err2 := c.SimulateCustomPolicy(context.Background(), "us-east-1", simulationInput)
```
angethuy, line 82: Since this is meant to test an unknown region, is the region argument here supposed to be something other than `us-east-1`? Looks like output1 and output2 are currently the same?

faris: Good catch, updated :)

angethuy (approved):
lgtm!

faris:
👀 #platform-tools

danielhochman (approved):

PR #1866, opened 2021-11-09
frontend: reduce autocomplete debouce to 250ms

mikecutalo:
### Description
There are times where I felt auto complete was not firing fast enough to provide a good user experience. Reducing this value by half seems like a good place to start.

### Testing Performed
n/a


mikecutalo:
@dschaller let me know what you think, we have talked about this before but 50% reduction feels like a good start.

dschaller (approved):

dschaller:
this seems like a good start. I think we should probably not go below 200ms though

PR #1868, opened 2021-11-09
Revert "housekeeping: github.com/bufbuild/buf to v0.56.0 (#1833)"

dschaller:
### Description

Reverts https://github.com/lyft/clutch/pull/1833 as it introduces a bug that inserts spaces for protoc-gen-validate rules.

See https://github.com/lyft/clutch-private/pull/1171/files#diff-124de01f1295e85152bb74d818c5cdd8c37fa10c725fc0a33f2ef28e42aff914R74-R75 as an example.

### Testing Performed

manual


danielhochman:
buf doesn't do whitespace linting afaik, that's handled by clang-format. did you test the revert and it worked?

danielhochman:
also if it is buf it seems like we could configure it to fix the problem rather than reverting it.

dschaller:
>buf doesn't do whitespace linting afaik, that's handled by clang-format. did you test the revert and it worked?

I tested this locally and it worked. I'll push up a PR off my local branch to confirm via CI.

dschaller:
Ah I was testing the wrong command. Reverting this does not fix the lint error. :'(

PR #1872, opened 2021-11-12
proto: add new field to Point

scarlettperry:
### Description
PR updates the `Point` proto to have a `href` field so that timeline event marks can link out to external sites.


smonero (approved):

frontend/workflows/projectSelector/src/types.tsx
```diff
@@ -117,12 +117,16 @@ export interface TimeData {
   [eventsKey: string]: EventData;
 }
 
+export interface TimeseriesPoint extends IClutch.timeseries.v1.IPoint {
```
danielhochman, line 120: why not add this field to the proto?

scarlettperry: we can, will update. i guess i saw the proto serving more for backend transformations of the API response

scarlettperry: updated

danielhochman (approved):

PR #1873, opened 2021-11-12
aws: multi account support

mikecutalo:
### Description
This is the last PR in the small series for multi AWS account support. Here we make the final cutover for all AWS resources to support the account dimension. 

### Testing Performed
Unit / Local.

Tested
* Topologycache
* Autocomplete
* Reboot Instance
* Term Instance
* Reize Asg
* Kinesis

### GitHub Issue
#1829


api/aws/ec2/v1/ec2.proto
```diff
@@ -67,7 +68,7 @@ message ResizeAutoscalingGroupResponse {
 message AutoscalingGroup {
   option (clutch.api.v1.id).patterns = {
     type_url : "clutch.aws.ec2.v1.AutoscalingGroup",
-    pattern : "{region}/{name}"
+    pattern : "{account}/{region}/{name}"
```
danielhochman, line 71: the one issue i think i've found is that a query like us-east-1/foo-bar will no longer limit the query to us-east-1

danielhochman (approved):
looks great overall! nice work.

think i may have found one issue in the resolver logic that could be problematic where there are collisions on a resource name between multiple regions but not accounts.

would not affect us but worth thinking about.

api/aws/ec2/v1/ec2.proto
mikecutalo: 🤔 let me see what i can do

api/aws/dynamodb/v1/dynamodb.proto
```diff
@@ -167,13 +171,15 @@ message DescribeTableResponse {
...
-    pattern : "{region}/{table_name}"
+    pattern : "{account}/{region}/{table_name}"
   };
+
   string table_name = 1 [ (validate.rules).string = {min_bytes : 1} ];
   string region = 2 [ (validate.rules).string = {min_bytes : 1} ];
   Throughput table_throughput = 3;
   repeated IndexUpdateAction gsi_updates = 4;
   bool ignore_maximums = 5;
+  string account = 7 [ (validate.rules).string = {min_bytes : 1} ];
```
dschaller, line 182:
```suggestion
  string account = 6 [ (validate.rules).string = {min_bytes : 1} ];
```

api/aws/ec2/v1/ec2.proto
```diff
@@ -52,11 +52,12 @@ message AutoscalingGroupSize {
 message ResizeAutoscalingGroupRequest {
   option (clutch.api.v1.id).patterns = {
     type_url : "clutch.aws.ec2.v1.AutoscalingGroup",
-    pattern : "{region}/{name}"
+    pattern : "{account}/{region}/{name}"
   };
 
   string name = 1 [ (validate.rules).string = {min_bytes : 1} ];
   string region = 2 [ (validate.rules).string = {min_bytes : 1} ];
+  string account = 4 [ (validate.rules).string = {min_bytes : 1} ];
```
dschaller, line 60:
```suggestion
  string account = 3 [ (validate.rules).string = {min_bytes : 1} ];
```

api/aws/ec2/v1/ec2.proto
```diff
@@ -192,4 +197,6 @@ message Instance {
   string availability_zone = 7;
 
   map<string, string> tags = 8;
+
```
dschaller, line 200:
```suggestion
```

backend/mock/service/awsmock/awsmock.go
```diff
@@ -25,30 +25,32 @@ import (
 
 type svc struct{}
 
-func (s *svc) DescribeKinesisStream(ctx context.Context, region string, streamName string) (*kinesisv1.Stream, error) {
+func (s *svc) DescribeKinesisStream(ctx context.Context, account, region, streamName string) (*kinesisv1.Stream, error) {
 	ret := &kinesisv1.Stream{
 		StreamName:        streamName,
 		Region:            region,
+		Account:           "defaul",
```
dschaller, line 32:
```suggestion
		Account:           "default",
```

dschaller (approved):

api/aws/ec2/v1/ec2.proto
mikecutalo: I would think in a case like this I would ideally like to let the fanout to finish instead of short circuiting once it finds a single result to try and capture if there was more than one resource found with the same name and propagate up an error if so, instead of returning the first one we find.

I think this would be a more fundamental change as I would think this should apply to any resolver

PR #1874, opened 2021-11-14
feedback: add SubmitFeedback implementation

scarlettperry:
### Description
PR adds the API implementation to persist a feedback submission to the database as well as the schema for the feedback table. PR also makes small related proto changes.

### Testing Performed
locally w/ pgadmin

example db entry:
Screen Shot 2021-11-18 at 8 21 55 PM

### GitHub Issue

### TODOs
- [x] database schema
- [x] tests


backend/cmd/migrate/migrations/000013_create_feedback_table.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE feedback (
+  -- client_id: unique id generated by the client and used to identify the feedback submission
+  client_id text PRIMARY KEY,
```
scarlettperry, line 3: debated b/w unique & primary key. decided on the latter since it's an identifier of the submission. also debated b/w text and varchar. decided on text since we're not going to set a length requirement on the value

api/feedback/v1/feedback.proto
```diff
@@ -94,14 +92,8 @@ message SubmitFeedbackRequest {
...
 
 message SubmitFeedbackResponse {
 }
-
-// proto used by the Feedback service
-message Submission {
-  // timestamp will be populated by the server
-  google.protobuf.Timestamp submitted_at = 1;
-  Feedback feedback = 2;
-}
```
scarlettperry, line 107: decided to just create a struct in the feedback service

api/feedback/v1/feedback.proto
```diff
@@ -94,14 +92,8 @@ message SubmitFeedbackRequest {
   // TODO: remove if we decide to record feedback only when a user clicks the submit button
   string id = 1 [ (validate.rules).string = {min_bytes : 1} ];
   Feedback feedback = 2 [ (validate.rules).message = {required : true} ];
+  FeedbackMetadata metadata = 3 [ (validate.rules).message = {required : true} ];
```
scarlettperry, line 95: changed it to not be nested for convenience reasons since the feedback and metadata are stored in diff columns.

backend/cmd/migrate/migrations/000013_create_feedback_table.up.sql
```diff
@@ -0,0 +1,17 @@
...
+
+  -- details: json blob of the feedback details
+  details JSONB,
+
+  -- metadata: json blob of info providing more context on the feedback details, such as:
+  -- the survey questions, the origin, whether the feedback was formally submitted, etc.
+  metadata JSONB
+);
+
+CREATE INDEX IF NOT EXISTS sort_submissions ON feedback (submitted_at);
```
scarlettperry, line 15: created index since time based queries will be common

backend/cmd/migrate/migrations/000013_create_feedback_table.up.sql
```diff
@@ -0,0 +1,17 @@
...
+CREATE INDEX IF NOT EXISTS details_json ON feedback USING GIN (details jsonb_path_ops);
+CREATE INDEX IF NOT EXISTS metadata_json ON feedback USING GIN (metadata jsonb_path_ops);
```
scarlettperry, line 17: created these indexes using GIN since we'll be doing queries on the keys of the json objects

scarlettperry:
cc @jdslaugh - tagging just as an fyi

backend/mock/service/feedbackmock/feedbackmock.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/lyft/clutch/backend/service/feedback"
+)
+
+type svc struct{}
+
+func (s svc) SubmitFeedback(ctx context.Context, id string, feedback *feedbackv1.Feedback, metadata *feedbackv1.FeedbackMetadata) error {
+	return nil
+}
+
+func New() feedback.Service {
```
danielhochman, line 16: do we want to register this in the mock server for testing?

backend/service/feedback/feedback.go
```diff
@@ -0,0 +1,104 @@
...
+	if id == "" {
+		return nil, status.Error(codes.InvalidArgument, "client id was empty")
+	}
+
+	if feedback == nil || metadata == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "feedback: %v or metadata: %v provided was nil", feedback, metadata)
+	}
+
+	// rules defined in the proto require that userId, urlPath, and rating not be empty
+	err := feedback.Validate()
```
danielhochman, line 70: this is done automatically in middleware, though i guess it can't hurt to call it again here

backend/service/feedback/feedback.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
```
danielhochman, line 71:
```suggestion
	if err := feedback.Validate(); err != nil {
```

backend/service/feedback/feedback.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
+	}
+
+	// rules defined in the proto require specific enums for the origin and that the survey not be nil
+	err = metadata.Validate()
+	if err != nil {
```
danielhochman, line 77:
```suggestion
	if err := metadata.Validate(); err != nil {
```

backend/service/feedback/feedback.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// the main question that was asked in the feedback component
+	if metadata.Survey.Prompt == "" {
+		return nil, status.Error(codes.InvalidArgument, "metadata survey prompt was empty")
+	}
+
+	if metadata.Survey.RatingOptions == nil {
+		return nil, status.Error(codes.InvalidArgument, "metadata rating options was nil")
+	}
```
danielhochman, line 88: are these not covered in validation?

backend/cmd/migrate/migrations/000013_create_feedback_table.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE feedback (
+  -- client_id: unique id generated by the client and used to identify the feedback submission
+  client_id text PRIMARY KEY,
+
+  submitted_at TIMESTAMP WITH TIME ZONE,
+
```
danielhochman, line 6: i would include the username as an actual field

backend/cmd/migrate/migrations/000013_create_feedback_table.up.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE feedback (
```
danielhochman, line 1: can we normalize the feedback to a rating out of 100 and store that in addition to the raw value?

api/feedback/v1/feedback.proto
```diff
@@ -94,14 +92,8 @@ message SubmitFeedbackRequest {
   // TODO: remove if we decide to record feedback only when a user clicks the submit button
   string id = 1 [ (validate.rules).string = {min_bytes : 1} ];
```
danielhochman, line 93: i'd actually bump the length requirement up here, maybe even use exactly and be prescriptive about how ids should be generated on the client

backend/service/feedback/feedback.go
```diff
@@ -0,0 +1,104 @@
...
+	// rules defined in the proto require that userId, urlPath, and rating not be empty
+	err := feedback.Validate()
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
+	}
+
+	// rules defined in the proto require specific enums for the origin and that the survey not be nil
+	err = metadata.Validate()
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
```
danielhochman, line 78:
```suggestion
		return nil, status.Error(codes.InvalidArgument, err.Error())
```

danielhochman (commented):
looks good overall! minor nits

great job on all the random stuff like WITH TIMEZONE, protojson, etc. lots of attention to detail!

backend/service/feedback/feedback.go
scarlettperry: ah true, decided to remove

backend/service/feedback/feedback.go
scarlettperry: no b/c I'm reusing the `Survey` message that we send in the GetSurveys response https://github.com/lyft/clutch/blob/467205b07853605e1a967d4812dc6a1313c95569/api/feedback/v1/feedback.proto#L53-L66. i haven't seen validation to response messages but lmk if it's fine to add it

my reason for re-using the message was so that whatever gets sent in the GetSurveys response, will be metadata in the SubmitFeedback request.

backend/cmd/migrate/migrations/000013_create_feedback_table.up.sql
scarlettperry: to double check this is what you had in mind -- create a rating enum (1-3 for now) & then have the request send a rating obj like `rating: {score: 3, label: "great"}`?

api/feedback/v1/feedback.proto
scarlettperry: good idea, thx! set the len to exactly 36 chars a a uuid is 32 chars with 4 hyphens.
https://www.npmjs.com/package/uuid, https://www.ietf.org/rfc/rfc4122.txt

should i set this length on the db field as well?

cc @jdslaugh

backend/service/feedback/feedback.go
scarlettperry: kept some checks though, like length of client id and user_id

backend/mock/service/feedbackmock/feedbackmock.go
scarlettperry: registered it but will follow up in another to add error scenarios.

scarlettperry:
@danielhochman addressed PR feedback. PR got larger with the last commit b/c of the proto changes - lmk if I should open a PR with just those changes. 

Latest changes added the normalization of ratings to a score out of a 100. With that, I also created a new RatingLabels proto which includes the raw text selection that gets shown to the user (i.e. bad/ok/great); the keys correspond to the enum emoji rating scale. I synced with Josh on these changes.

example `GetSurveys` API response
```
{
    "originSurvey": {
        "WIZARD": {
            "ratingLabels": {
                "emoji": {
                    "sad": "bad",
                    "neutral": "ok",
                    "happy": "great"
                }
            }
        }
    }
}
```

example `SubmitFeedback` API request
```
{
    "id": "fdbe0b66-a45e-45a2-8d58-1f467c920520",
    "userId": "foo@example.com",
    "feedback": {
        "url_path": "/k8s/deletePod",
        "rating_label": "ok",
        "rating_scale": {
            "emoji": "NEUTRAL"
        }
    },
    "metadata": {
        "origin": 2,
        "survey": {
            "prompt": "Rate your experience using the workflow",
            "freeform_prompt": "How can we improve this feature?",
            "ratingLabels": {
                "emoji": {
                    "sad": "bad",
                    "neutral": "ok",
                    "happy": "great"
                }
            }
        },
        "userSubmitted": false
    }
}
```

example db entries
Screen Shot 2021-11-18 at 8 21 55 PM

backend/service/feedback/feedback.go
```diff
@@ -0,0 +1,132 @@
...
+		if !ok {
+			return -1, fmt.Errorf("unsupported rating: %v", rating)
+		}
+		return v, nil
+	default:
+		return -1, fmt.Errorf("unsupported rating scale: %v", scale)
+	}
+}
+
+func (s *svc) SubmitFeedback(ctx context.Context, id string, userId string, feedback *feedbackv1.Feedback, metadata *feedbackv1.FeedbackMetadata) error {
```
mikecutalo, line 126: I could see the `uuid` being an optional param if we're going to support upserting, if we decided to implement feedback checkpointing. But until then I dont think its best for the caller to have to generate the `uuid`. 

If we do support feedback checkpointing, Submit feedback should return a `uuid` upon first submission and if one is provided then we would do the upsert.

But for now, why have the caller be required to generate a `uuid`?

scarlettperry: this was a suggestion/decision made in a previous PR https://github.com/lyft/clutch/pull/1828 
the advantage of this being generated by the client is that this approach doesn't require coordination from the server. also it's required, because for the V1, we are sending a feedback request each time the user selects an emoji and updating the feedback at the id with the latest submission

mikecutalo: I see your upsert query below, we can stick to the original decision.

mikecutalo (approved):

PR #1880, opened 2021-11-15
core: add emojis and global icon variant sizes

scarlettperry:
### Description
PR adds the emojis that will be used in the feedback system to a new assets dir in the core package. 

PR also adds a global style map based on the icon sizes from the core design system (link in slack) as well as a new size 48px (material also has this size  in the icon system https://developers.google.com/fonts/docs/material_icons) so that we can reuse it as we add other assets.

### Testing Performed
Storybook

emojis

### GitHub Issue


jdslaugh (approved):
I like it.

Two nits:

Should they be named `HappyEmoji` instead of `HappyIcon` since that's what they're labeled as?

And would renaming the files to `emojis/happy` vs `emojis/happy-emoji` make more sense? It seems redundant currently but overall its minor.

Thanks for putting this together!

frontend/packages/core/src/Assests/emojis/happy-emoji.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from "react";
+
+import type { SVGProps } from "../global";
+import { StyledSVG } from "../global";
+
+const HappyIcon = ({ size }: SVGProps) => (
+  <StyledSVG size={size} viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg">
```
dschaller, line 7: can we pass through the remaining props?
```suggestion
const HappyIcon = ({ size, ...props }: SVGProps) => (
  <StyledSVG size={size} viewBox="0 0 48 48" fill="none" xmlns="http://www.w3.org/2000/svg" {...props}>
```

frontend/packages/core/src/Assests/global.ts
```diff
@@ -0,0 +1,33 @@
...
+export const VARIANTS = Object.keys(STYLE_MAP);
+
+export type IconSizeVariant = "xsmall" | "small" | "medium" | "large";
+
+export const StyledSVG = styled.svg<{ size?: IconSizeVariant }>(props => ({
+  width: `${STYLE_MAP[props.size]?.width || STYLE_MAP.small.width}px`,
+  height: `${STYLE_MAP[props.size]?.height || STYLE_MAP.small.height}px`,
+}));
+
+export interface SVGProps {
```
dschaller, line 31: would it make sense to extends reacts SVGProps?

frontend/packages/core/src/Assests/global.ts
```diff
@@ -0,0 +1,33 @@
+import styled from "@emotion/styled";
+
+export const STYLE_MAP = {
+  xsmall: {
+    width: 18,
```
dschaller, line 5: really minor nit: can we extract these out to constants to reduce the duplication?

dschaller:
+1 to @jdslaugh's comment above

scarlettperry:
hmm git isn't picking up my folder renaming changes

frontend/packages/core/src/Assests/global.ts
scarlettperry: updated, lmk if this isn't what you had in mind.

scarlettperry: just pushed up another change to reduce more code duplication. the style map now just has size instead of width/height since both values are the same.

dschaller (approved):

PR #1885, opened 2021-11-17
core: add new color palette to button

scarlettperry:
### Description
PR adds the new color palette `tertiary` that Design added to the core design system (will also be used in the Wizard feedback component). Figma link in slack thread.

### Testing Performed
Storybook

tertiary

### DIFF
--- a/frontend/packages/core/src/button.tsx
+++ b/frontend/packages/core/src/button.tsx
@@ -61,6 +61,18 @@ const COLORS = {
       disabled: "#FFFFFF",
     },
   },
+  tertiary: {
+    background: {
+      primary: "transparent",
+      hover: "#F5F6FD",
+      active: "#D7DAF6",
+      disabled: "transparent",
+    },
+    font: {
+      primary: "#3548D4",
+      disabled: "#0D1030",
+    },
+  },
 } as { [key: string]: ButtonPalette };
 
 const colorCss = (palette: ButtonPalette) => {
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/button.tsx
```diff
@@ -102,7 +114,7 @@ const StyledBorderButton = styled(StyledButton)({
 });
 
 /** Provides feedback to the user in regards to the action of the button. */
-type ButtonVariant = "neutral" | "primary" | "danger" | "destructive";
+type ButtonVariant = "neutral" | "primary" | "danger" | "destructive" | "tertiary";
```
danielhochman, line 117: wouldn't it be secondary?

scarlettperry: secondary is our "neutral" color palette

frontend/packages/core/src/stories/button.stories.tsx
```diff
@@ -25,10 +25,16 @@ Destructive.args = {
   variant: "destructive",
 };
 
-export const Danger = Template.bind({});
-Danger.args = {
-  text: "Rotate",
-  variant: "danger",
```
scarlettperry, line 31: removed b/c we already have the "danger" variant example right above

frontend/packages/core/src/stories/button.stories.tsx
```diff
@@ -25,10 +25,16 @@ Destructive.args = {
...
+export const Neutral = Template.bind({});
+Neutral.args = {
+  text: "Back",
+  variant: "neutral",
```
scarlettperry, line 31: we were missing this example in storybook

frontend/packages/core/src/button.tsx
danielhochman: i would update the design system or the code to match (would prefer the former). no reason to have tertiary without secondary in the impl.

scarlettperry: okay I'll update this new palette name to be "secondary" and leave a comment for Design to rename "secondary" --> "neutral" in the figma file

danielhochman: sgtm!

scarlettperry: updated & left the comments for Design

danielhochman (approved):

jdslaugh (approved):
Thanks!

PR #1899, opened 2021-11-30
backend: add ListPullRequestsWithCommit into Github APIs

mengmichael1:
### Description
Adding the ListPullRequestsWithCommit API into the backend module's available Github APIs. 

https://docs.github.com/en/rest/reference/repos#list-pull-requests-associated-with-a-commit

`GET /repos/{owner}/{repo}/commits/{commit_sha}/pulls`

As part of Lyft's internal extensions for Clutch, we are building a new dashboard to track k8s resources deployed for integrated testing purposes. One of the columns we'd like to surface for our users is the Github PR's branch name, hyperlinked to the Github PR's URL. This allows engineers to have a single-pane-of-glass view over their test deployments of pre-merged Github PR's. The only metadata baked into the k8s artifact is the SHA of the PR, so this API will enable us to translate that into more helpful information (HTML URL, branch name) when engineers are doing their routine maintenance/management.  

### Testing Performed
Added unit tests. First time contributing so seeking advice on what other testing techniques can be used for this type of change :)

### GitHub Issue
n/a

### TODOs
n/a at the time of this posting


backend/mock/service/githubmock/githubmock.go
```diff
@@ -63,6 +63,10 @@ func (s svc) ListOrganizations(ctx context.Context, user string) ([]*githubv3.Or
 	}, nil
 }
 
+func (s svc) ListPullRequestsWithCommit(ctx context.Context, ref *github.RemoteRef, sha string) ([]*github.PullRequestInfo, error) {
+	panic("implement me")
```
dschaller, line 67: can you add a stub for the response? The data from the tests should be ok to use here

backend/service/github/github.go
```diff
@@ -236,6 +242,25 @@ func (s *svc) CreatePullRequest(ctx context.Context, ref *RemoteRef, base, title
 	}, nil
 }
 
+func (s *svc) ListPullRequestsWithCommit(ctx context.Context, ref *RemoteRef, sha string) ([]*PullRequestInfo, error) {
```
danielhochman, line 245: i would expand the signature a bit so this could be used to query for more than just open PRs in the future without a breaking signature change.

fake edit: according to https://docs.github.com/en/rest/reference/repos#list-pull-requests-associated-with-a-commit it doesn't accept `state` as a parameter though, so maybe the comment below is incorrect.

mengmichael1: Good point. Upon looking again, seems like Github has some contradicting information. 

Here (https://github.com/google/go-github/blob/ff33a554ef2c24c1710238b69942fe51e70d89da/github/pulls.go#L121-L123) PullRequestListOptions, which is the opts part of `ListPullRequestsWithCommit(ctx context.Context, owner, repo, sha string, opts *githubv3.PullRequestListOptions)` stats that Default is "open". 

However, the method comment here https://github.com/google/go-github/blob/master/github/pulls.go#L168-L194 seems to imply that the default is open and closed. 

I'll see if I can find out more.

mengmichael1: Following up with go-github here https://github.com/google/go-github/pull/2228

backend/mock/service/githubmock/githubmock.go
mengmichael1: done

mengmichael1:
Addressed both comments, ready for another pass :)

mengmichael1:
I'm out for the weekend, back next week, but would really appreciate a review whenever you get the chance @dschaller @danielhochman :)

backend/service/github/github.go
```diff
@@ -216,6 +218,10 @@ func boolPtr(b bool) *bool {
 	return &b
 }
 
+func intPtr(i int) *int {
+	return &i
+}
```
dschaller, line 223: It looks like this is only used by the tests. Can you move it into that file instead?

backend/service/github/github.go
dschaller: I think Daniel was suggesting supporting either `PullRequestListOptions` in the signature so that callers could specify this.

backend/service/github/github.go
mengmichael1: done!

backend/service/github/github.go
mengmichael1: ahh I see, I jumped the gun too quickly 

done -- added in `state string` into the signature so callers the githubv3 `PullRequestListOptions` struct is abstracted away for callers. wdyt?

backend/service/github/github_test.go
```diff
@@ -672,3 +677,110 @@ func TestGetRepository(t *testing.T) {
...
+	}
+
+	m.actualNumber = 1347
+	m.actualHTMLURL = fmt.Sprintf("https://github.com/%s/%s/pull/%s", owner, repo, strconv.Itoa(m.actualNumber))
+	m.actualBranchName = "my-branch"
+
+	return []*githubv3.PullRequest{
+		{
+			Number:  intPtr(m.actualNumber),
+			State:   strPtr("open"),
```
matthewgrossman, line 707: should this be the `opts.State`?

matthewgrossman: def a nit, just to make the mock more representative of what they passed in

mengmichael1: Hmm I believe it should just be State. This is a mock of https://github.com/lyft/clutch/blob/4bd65211d507baf6d2c95a6fc73ca8a0c74e9446/backend/service/github/iface.go#L37 which returns `[]*githubv3.PullRequest` where `PullRequest` is defined as https://github.com/google/go-github/blob/1d9884e7ac8ca3c6d3d469af28af227fc773a246/github/pulls.go#L30

mengmichael1: You can think of this mock as the mock for what the Github API returns in https://github.com/lyft/clutch/blob/b4651bc598b000cfbf7dea033b3151b4c7d06180/backend/service/github/github.go#L246

matthewgrossman: sorry, I meant this:
```suggestion
			State:   strPtr(opts.State),
```

matthewgrossman: I guess what I'm saying is this mock is hardcoded to return an "open" PR as opposed to something that the user can control, which is what the other args seem to do

backend/service/github/github.go
dschaller: imo the PR list options object is gives the most flexibility for users of the service without having to change / amend the function signature.

backend/service/github/github_test.go
mengmichael1: ah I see what you mean, good catch. updated

backend/service/github/github.go
mengmichael1: Ah ok fair point. Done -- d36be40

mengmichael1:
ptal @dschaller @danielhochman

dschaller (approved):

dschaller:
lgtm - You'll need another +1 from Daniel

danielhochman (approved):

PR #1900, opened 2021-11-30
feedback: update schema of rating labels

scarlettperry:
### Description
A suggestion was raised that `GetSurveys` API response should return the rating labels in an ordered list so that the frontend can assume that the order of the rating labels aligns with the order in which to show the rating icons on the FE, as opposed to the FE hard coding the icons in a certain order.

Previously the rating labels were returned in a map.

### Testing Performed
locally

example response
```
{
    "originSurvey": {
        "WIZARD": {
            "ratingLabels": [
                {
                    "emoji": "SAD",
                    "label": "bad"
                },
                {
                    "emoji": "NEUTRAL",
                    "label": "ok"
                },
                {
                    "emoji": "HAPPY",
                    "label": "great"
                }
            ]
        }
    }
}
```

example feedback config
```
 - name: clutch.module.feedback
    typed_config:
      "@type": types.google.com/clutch.config.module.feedback.v1.Config
      origins:
        - origin: WIZARD
          survey:
            prompt: Rate your experience using this Clutch workflow
            freeform_prompt: What could be better about this workflow?
            rating_labels:
              - emoji: SAD
                label: bad
              - emoji: NEUTRAL
                label: ok
              - emoji: HAPPY
                label: great
```

### GitHub Issue


backend/module/feedback/feedback.go
```diff
@@ -54,13 +54,13 @@ func New(cfg *any.Any, log *zap.Logger, scope tally.Scope) (module.Module, error
 
 // SurveyLookup stores a map of the origin to survey mapping that are
 // provided in the feedback module config
-type SurveyLookup struct {
+type surveyLookup struct {
```
scarlettperry, line 57: just a nit. i think initially I had planned to have this in a helper file so I had it exported

backend/module/feedback/feedback.go
```diff
@@ -97,37 +97,39 @@ func (m *mod) GetSurveys(tx context.Context, req *feedbackv1.GetSurveysRequest)
 	// this scenario shouldn't happen as the request and config have to provide defined enum types
 	// TODO: if multiple orgins are requested at once and only some surveys are found, return partial error
 	if len(results) == 0 {
-		m.logger.Error("origins must be one of HEADER or WIZARD",
+		msg := "surveys not found for the requested origin(s)"
```
scarlettperry, line 100: another nit. original error msg was a lil misleading as they could have passed a valid origin but not have defined its survey questions in the config

api/config/module/feedback/v1/feedback.proto
```diff
@@ -10,7 +10,7 @@ import "feedback/v1/feedback.proto";
 message Survey {
   string prompt = 1 [ (validate.rules).string = {min_bytes : 1} ];
   string freeform_prompt = 2;
-  clutch.feedback.v1.RatingLabels rating_labels = 3 [ (validate.rules).message = {required : true} ];
+  repeated clutch.feedback.v1.RatingLabel rating_labels = 3 [ (validate.rules).repeated = {min_items : 2} ];
```
scarlettperry, line 13: currently we have a 3-point scale but keeping this general to future cases where we support additional scales. but i think we can assume that the rating scale should at least have 2 items.

api/feedback/v1/feedback.proto
```diff
@@ -36,13 +36,6 @@ enum Origin {
 // we can expand to add more emoji options (i.e. for a 4-point scale) and
 // new rating scales (i.e 2-point thumbs up/down scale).
 
-message EmojiRatingLabels {
-  // the corresponding option to show to the user (i.e bad/ok/great)
-  string sad = 1;
-  string neutral = 2;
-  string happy = 3;
-}
-
```
scarlettperry, line 45: can get rid of this. i think the new structure is cleaner.

mikecutalo (approved):

danielhochman (approved):

PR #1903, opened 2021-12-01
feedback: add wizard feedback prop to workflow configuration

scarlettperry:
### Description
PR adds a new interface to workflows' configuration so that users can choose to enable the Wizard feedback placement via the `clutch.config.js`. Reasons for not globally enabling the feedback placement is that feedback placement requires a backend setup and users may not want to opt-in their workflow for feedback.

### Testing Performed
locally

example of prop being adding in the `clutch.config.js`

```
  "@clutch-sh/kinesis": {
    updateShardCount: {
      componentProps: {
        resolverType: "clutch.aws.kinesis.v1.Stream",
        enableFeedback: true,
      },
    },
  },
  "@clutch-sh/dynamodb": {
    updateCapacity: {
      componentProps: {
        resolverType: "clutch.aws.dynamodb.v1.Table",
        enableOverride: true,
        enableFeedback: true,
      },
    },
  },
```

I commented out all the other components to just show the workflow's last step.
Screen Shot 2021-12-01 at 9 39 14 AM

### GitHub Issue

### DIFF
--- a/frontend/workflows/ec2/src/index.tsx
+++ b/frontend/workflows/ec2/src/index.tsx
@@ -13,7 +13,11 @@ interface ConfirmConfigProps {
   notes?: NoteConfig[];
 }
 
-export interface WorkflowProps extends BaseWorkflowProps, ResolverConfigProps, ConfirmConfigProps {}
+interface WizardFeedbackProps {
+  enableFeedback?: boolean;
+}
+
+export interface WorkflowProps extends BaseWorkflowProps, ResolverConfigProps, ConfirmConfigProps, WizardFeedbackProps {}
 export interface ResolverChild extends WizardChild, ResolverConfigProps {}
 export interface ConfirmChild extends WizardChild, ConfirmConfigProps {}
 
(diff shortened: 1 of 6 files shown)


frontend/workflows/ec2/src/index.tsx
```diff
@@ -13,7 +13,15 @@ interface ConfirmConfigProps {
   notes?: NoteConfig[];
 }
 
-export interface WorkflowProps extends BaseWorkflowProps, ResolverConfigProps, ConfirmConfigProps {}
+interface WizardFeedbackProps {
```
jdslaugh, line 16: Is it possible to export and extend this interface from @clutch-sh/core instead of redefining it in each workflow separately?

scarlettperry: ah good idea, thx for the suggestion!

jdslaugh (approved):
Thanks!

mikecutalo (approved):

PR #1906, opened 2021-12-02
k8s: add protos for node

ashleykasim:
### Description
Add protos for [Node](https://pkg.go.dev/k8s.io/api/core/v1#Node) to enable building a [cordon node](https://kubernetes.io/docs/concepts/architecture/nodes/#manual-node-administration) workflow for clutch. `kubectl cordon node` is implemented as `PATCH /api/v1/nodes/$nodename {"spec": "unschedulable": true}` so the clutch proto should follow this convention. The proto currently only supports modifying schedulability, but in the future can be expanded to support modifying other supported fields such as taints, labels, and annotations.


api/k8s/v1/k8s.proto
```diff
@@ -1141,3 +1157,45 @@ message RemoveObjectMetaFields {
...
 }
+
+message Node {
+  option (clutch.api.v1.id).patterns = {
+    type_url : "clutch.k8s.v1.Node",
+    pattern : "{cluster}/{name}"
+  };
+  string name = 1;
+  string cluster = 2;
+  bool unschedulable = 3;
```
dschaller, line 1168: Would it make sense for this be it's own message for Taint specific values?

ashleykasim: cordon's not a taint, it is a separate field on the NodeSpec

dschaller: ah I see it now, thanks!

mikecutalo (approved):

dschaller (approved):

PR #1907, opened 2021-12-02
k8s: implement backend api's for node

ashleykasim:
### Description

Follow up to https://github.com/lyft/clutch/pull/1906

Implement api's for DescribeNode & UpdateNode. Currently supports updating NodeSpec.Unschedulable field only (cordon node).

Updating schedulablity on a node that is already in the desired schedulabiltiy state (i.e. cordoning a node that is already cordoned) should succeed as this is the desired result for the end user.

### Testing Performed
Tested via curl against a remote kube cluster:

DescribeNode:

```
$ curl --location --request POST 'http://localhost:8080/v1/k8s/describeNode' \
--header 'Content-Type: application/json' \
--data-raw '{
    "clientset": "testcontext",
    "cluster": "test",
    "name": "test-node-name"
}'
{"node":{"name":"test-node-name","cluster":"test","unschedulable":false}}%
```

UpdateNode:

```
$ curl --location --request POST 'http://localhost:8080/v1/k8s/updateNode' \
--header 'Content-Type: application/json' \
--data-raw '{
    "clientset": "test-context",
    "cluster": "test",
    "name": "test-node-name",
    "unschedulable": true
}'
{}%

# this correctly updates the node status to SchedulingDisabled (cordoned)

$ kubectl get node test-node-name
NAME                          STATUS                     ROLES    AGE   VERSION
test-node-name  Ready,SchedulingDisabled   <none>   51d   v1.21.1
```

### DIFF
--- a/backend/module/k8s/node.go
+++ b/backend/module/k8s/node.go
@@ -2,15 +2,22 @@ package k8s
 
 import (
 	"context"
-	"errors"
 
 	k8sapiv1 "github.com/lyft/clutch/backend/api/k8s/v1"
 )
 
 func (a *k8sAPI) DescribeNode(ctx context.Context, req *k8sapiv1.DescribeNodeRequest) (*k8sapiv1.DescribeNodeResponse, error) {
-	return nil, errors.New("not implemented")
+	node, err := a.k8s.DescribeNode(ctx, req.Clientset, req.Cluster, req.Name)
+	if err != nil {
+		return nil, err
+	}
+	return &k8sapiv1.DescribeNodeResponse{Node: node}, nil
 }
 
 func (a *k8sAPI) UpdateNode(ctx context.Context, req *k8sapiv1.UpdateNodeRequest) (*k8sapiv1.UpdateNodeResponse, error) {
-	return nil, errors.New("not implemented")
+	err := a.k8s.UpdateNode(ctx, req.Clientset, req.Cluster, req.Name, req.Unschedulable)
+	if err != nil {
+		return nil, err
+	}
+	return &k8sapiv1.UpdateNodeResponse{}, nil
 }
--- a/backend/service/k8s/node.go
+++ b/backend/service/k8s/node.go
@@ -1,0 +1,70 @@
+package k8s
+
+import (
+	"context"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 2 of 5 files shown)


backend/module/k8s/node.go
```diff
@@ -2,15 +2,22 @@ package k8s
...
+	return &k8sapiv1.DescribeNodeResponse{Node: node}, nil
 }
 
 func (a *k8sAPI) UpdateNode(ctx context.Context, req *k8sapiv1.UpdateNodeRequest) (*k8sapiv1.UpdateNodeResponse, error) {
-	return nil, errors.New("not implemented")
+	err := a.k8s.UpdateNode(ctx, req.Clientset, req.Cluster, req.Name, req.Unschedulable)
+	if err != nil {
+		return nil, err
+	}
+	return &k8sapiv1.UpdateNodeResponse{}, nil
```
dschaller, line 22: minor nit: should we return the updated node object here?

backend/service/k8s/node.go
```diff
@@ -0,0 +1,70 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(nodes.Items) == 1 {
+		return ProtoForNode(cs.Cluster(), &nodes.Items[0]), nil
+	} else if len(nodes.Items) > 1 {
+		return nil, status.Error(codes.FailedPrecondition, "located multiple nodes")
```
dschaller, line 32: nit: might be nice to include the name in this response message

backend/module/k8s/node.go
ashleykasim: we don't return the updated object for other k8s types, ex. pod https://github.com/lyft/clutch/blob/main/api/k8s/v1/k8s.proto#L458-L459, I implemented it this way to match the pattern. Either way is fine for me, I suspect I won't need the updated node object for the cordon node workflow though.

dschaller: Good point, for consistency let's keep as empty and I'll mention it in our next team meeting.

dschaller (approved):

mikecutalo (approved):

backend/service/k8s/node.go
ashleykasim: Similar to the previous comment, this matches the message for other types, ex. pods https://github.com/lyft/clutch/blob/main/backend/service/k8s/pods.go#L32. I kept it the same for uniformity. Tbh this condition should basically never be true as kubernetes names are unique per object types (but may be reused - either way only 1 will ever exist at a time) https://github.com/lyft/clutch/blob/main/backend/service/k8s/pods.go#L32

PR #1909, opened 2021-12-03
k8s: resolver for node

ashleykasim:
### Description
Add a resolver for node. Follow up to https://github.com/lyft/clutch/pull/1907 & https://github.com/lyft/clutch/pull/1906

PR #1910, opened 2021-12-03
k8s: cordon/uncordon node workflow

ashleykasim:
### Description
Add workflow for cordon/uncordon node. Follow up to #1906, #1907 , & #1909.

Workflow demo:

https://user-images.githubusercontent.com/2972218/144673825-b2b4f246-6d9a-4f16-91d7-bf34fa2e8c67.mov

### DIFF
--- a/frontend/workflows/k8s/src/cordon-node.tsx
+++ b/frontend/workflows/k8s/src/cordon-node.tsx
@@ -1,0 +1,107 @@
+import React from "react";
+import type { clutch as IClutch } from "@clutch-sh/api";
+import {
+  Button,
+  ButtonGroup,
+  client,
+  Confirmation,
+  MetadataTable,
+  Resolver,
+  Switch,
+  useWizardContext,
+} from "@clutch-sh/core";
+import { useDataLayout } from "@clutch-sh/data-layout";
+import type { WizardChild } from "@clutch-sh/wizard";
+import { Wizard, WizardStep } from "@clutch-sh/wizard";
+
+import type { ConfirmChild, ResolverChild, WorkflowProps } from ".";
+
+const NodeIdentifier: React.FC<ResolverChild> = ({ resolverType }) => {
+  const { onSubmit } = useWizardContext();
+  const resolvedResourceData = useDataLayout("resourceData");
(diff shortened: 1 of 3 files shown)


frontend/workflows/k8s/src/cordon-node.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const node = useDataLayout("resourceData").displayValue() as IClutch.k8s.v1.Node;
+  const updateData = useDataLayout("updateData");
+  return (
+    <WizardStep error={updateData.error} isLoading={updateData.isLoading}>
+      <Confirmation action="Update" />
+      <MetadataTable
+        data={[
+          { name: "Name", value: node.name },
+          { name: "Cluster", value: node.cluster },
+          { name: "Unschedulable", value: String(node.unschedulable) },
```
danielhochman, line 78: ideally this would be a select, radio, or checkbox instead of a string value for bool

danielhochman (approved):
one nit, otherwise nice and clean code!

frontend/workflows/k8s/src/cordon-node.tsx
```diff
@@ -0,0 +1,112 @@
...
+          { name: "Name", value: node.name },
+          { name: "Cluster", value: node.cluster },
+          {
+            name: "Unschedulable",
+            value: node.unschedulable,
+            input: {
+              type: "string",
+              key: "unschedulable",
+              validation: bool(),
+            },
```
dschaller, line 56: I think it might be better to use a toggle here for the state. Something like
```tsx
value: (<Switch checked={resourceData.unschedulable} onChange={update} />),
```

frontend/workflows/k8s/src/cordon-node.tsx
ashleykasim: 👍 will change it to a toggle

frontend/workflows/k8s/src/cordon-node.tsx
ashleykasim: 👍 will change it to a toggle

dschaller (approved):

PR #1924, opened 2021-12-13
feedback: save search params as metadata in feedback submission

scarlettperry:
### Description
PR adds a `urlSearchParam` field for saving the URL query params (if present) as metadata in the feedback submission. This field could be useful in scenarios where a user submits a low score wherein we can check `urlSearchParam` to see if a specific use case wasn't handled well.

### Testing Performed
locally

```
{
....
    "metadata": {
        "origin": 2,
        "userSubmitted": false,
        ...
        "urlSearchParams": "?q=foo%2Ffoo"
    }
}
```

### GitHub Issue


jdslaugh (approved):
Thanks for adding this!

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -158,6 +158,7 @@ const NPSFeedback = (opts: FeedbackOptions) => {
           origin: IClutch.feedback.v1.Origin[opts.origin],
           userSubmitted: hasSubmit,
           survey,
+          ...(window.location.search && { urlSearchParams: window.location.search }),
```
danielhochman, line 161: super nit: this seems like additional complexity for little benefit. what is the benefit over just doing
```suggestion
          urlSearchParams: window.location.search,
```

danielhochman (approved):

frontend/packages/core/src/NPS/feedback.tsx
scarlettperry: i had used this approach for a scenario where the object key being present mattered but yeah in this case, no real benefit. will update!

PR #1941, opened 2022-01-04
FE: add `NotePanel` to resolver along with notes in props

smonero:
### Description

It is useful to have notes so that during the resolver we can give more info to the users of the workflow

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Resolver/index.tsx
+++ b/frontend/packages/core/src/Resolver/index.tsx
@@ -5,6 +5,7 @@ import _ from "lodash";
 import { AccordionGroup } from "../accordion";
 import { useWizardContext } from "../Contexts";
 import { Error } from "../Feedback";
+import type { NoteConfig } from "../Feedback";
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Resolver/index.tsx
```diff
@@ -132,6 +138,7 @@ const Resolver: React.FC<ResolverProps> = ({
               <SchemaResolver key={schema.typeUrl} schema={schema} submitHandler={submitHandler} />
             ))}
           </AccordionGroup>
+          <NotePanel notes={notes} />
```
scarlettperry, line 141: should this be placed above Query Resolver? Saw the screenshot in the internal PR & the notes get placed at the bottom which may be missed by the user.

smonero: yeah I wasn't sure where to put it. I'll move it to the top

scarlettperry (approved):

jdslaugh (approved):

PR #1944, opened 2022-01-05
frontend: filter out transient props

dschaller:
### Description

Add custom styled function that supports filtering out transient properties.

As an added bonus this allows Clutch centralize its versioning of emotion.

See https://github.com/emotion-js/emotion/issues/2193 for additional context.

### Testing Performed

manual & CI


frontend/packages/core/src/NPS/emojiRatings.tsx
```diff
@@ -31,15 +31,15 @@ const EmojiRatings = ({ ratings = [], setRating }: EmojiRatingsProps) => {
   const [selectedRating, selectRating] = useState<Rating>(null);
 
   const StyledIconButton = styled(IconButton)<{
-    selected: boolean;
+    $selected: boolean;
```
dschaller, line 34: @jdslaugh is it ok to filter this out from the resulting element? I know there are tests that asserted this was set.

jdslaugh: Yeah its fine, it was a double check when verifying the opacity but I can just verify the styles themselves

jdslaugh (approved):
This looks really cool!

danielhochman (approved):

smonero (approved):

PR #1963, opened 2022-01-19
frontend: extending select to add grouping option

jdslaugh:
### Description
For anytime NPS component, needed to extend the select component to more clearly separate workflows. Added a grouping option.

#### Without Grouping (Storybook)
Screen Shot 2022-01-19 at 11 59 57 AM

#### With Grouping (Storybook)
Screen Shot 2022-01-19 at 11 59 52 AM

#### In Action (NPS Header)
Screen Shot 2022-01-19 at 11 50 17 AM

### Testing Performed
Manual

### DIFF
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -179,6 +179,8 @@ const StyledSelect = styled(BaseSelect)({
 
   ".MuiListSubheader-root": {
     color: "#939495",
+    cursor: "default",
+    "pointer-events": "none", // disables the select from closing on clicking the subheader
   },
 });
 
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -11,6 +11,7 @@ import {
 } from "@material-ui/core";
 import ErrorIcon from "@material-ui/icons/Error";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
+import { flatten } from "lodash";
 
 const StyledFormControl = styled(MuiFormControl)({
   "label + .MuiInput-formControl": {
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/Input/select.tsx
```diff
@@ -204,41 +213,63 @@ const Select = ({
   const defaultIdx = defaultOption < options.length && defaultOption > 0 ? defaultOption : 0;
   const [selectedIdx, setSelectedIdx] = React.useState(defaultIdx);
 
+  // Flattens all options and sub grouped options for easier retrieval
+  const flatOptions: SelectOption[] = options
+    .map(option => (option.group ? option.group.map(groupOption => groupOption) : option))
+    .reduce((pre: any, cur: any) => pre.concat(cur), []);
```
scarlettperry, line 219: should the type of `pre` and `cur` be `BaseSelectOptions`?

frontend/packages/core/src/Input/select.tsx
```diff
@@ -204,41 +213,63 @@ const Select = ({
...
-    setSelectedIdx(optionValues.indexOf(value));
+    // handle if selecting a header option
+    if (!value) {
+      return;
+    }
+    setSelectedIdx(flatOptions.findIndex(opt => opt?.value === value || opt?.label === value));
     onChange && onChange(value);
   };
 
   if (options.length === 0) {
```
scarlettperry, line 237: should `flatOptions` be used here as well?

scarlettperry (commented):
nice!

frontend/packages/core/src/Input/select.tsx
jdslaugh: Good call out, I've updated it. It technically wouldn't matter but its good to be consistent!

frontend/packages/core/src/Input/select.tsx
jdslaugh: Typings for that block were getting to be a pain and it was starting to look a little cluttered, so refactored it and just utilized flatten from lodash instead and added typings throughout. Good call though (pretty sure it was an any because of the typing issue and planned on returning later to fix it and forgot :) )

scarlettperry (approved):

danielhochman (approved):
super nit: can you do `cursor: default` on the group

Peek 2022-01-20 11-41
?

jdslaugh:
> super nit: can you do `cursor: default` on the group
> 
> Peek 2022-01-20 11-41 ?

Absolutely, Im surprised it wasn't like that be default. Updated!

PR #1964, opened 2022-01-19
frontend: NPS Header Component

jdslaugh:
### Description
- Adding NPS Header Component
- Added 'Large' sizing for IconButton component
- Added testContext file for easier unit testing

Relies on: #1963

Edited: Updated screenshots based on feedback

#### On Popper Open
Screen Shot 2022-01-21 at 3 57 21 PM

#### After Emoji Selection
Screen Shot 2022-01-21 at 3 57 32 PM

#### Feedback Types (Generated from Workflows)
Screen Shot 2022-01-21 at 3 57 42 PM

#### No Error and Input Typing
Screen Shot 2022-01-21 at 3 57 52 PM

#### Input Error Validation
Screen Shot 2022-01-21 at 3 58 10 PM

#### Post Successful Submission
Screen Shot 2022-01-19 at 11 52 39 AM

### Testing Performed
Manual and Unit

### Github Issue


frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -185,21 +204,42 @@ const NPSFeedback = (opts: FeedbackOptions) => {
...
+          <Typography variant={wizardOrigin ? "subtitle3" : "subtitle2"}>
+            {survey.prompt}
+          </Typography>
         </MuiGrid>
-        <MuiGrid item xs={6} style={{ display: "flex", justifyContent: "space-around" }}>
-          <EmojiRatings ratings={survey.ratingLabels} setRating={setSelected} />
+        <MuiGrid
+          item
+          xs={wizardOrigin ? 6 : 12}
+          style={{ margin: "8px 0px", display: "flex", justifyContent: "space-around" }}
```
scarlettperry, line 217: out of curiosity, is the `margin` needed? noticed that it makes the Wizard placement height look larger than before

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -185,21 +204,42 @@ const NPSFeedback = (opts: FeedbackOptions) => {
...
+                  options={options.feedbackTypes}
+                  onChange={setFeedbackType}
+                />
+              </MuiGrid>
+            )}
             <MuiGrid item xs={12}>
               <StyledTextField
                 multiline
                 fullWidth
+                $origin={origin}
```
scarlettperry, line 242: out of curiosity, what does the `$` do?

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -185,21 +204,42 @@ const NPSFeedback = (opts: FeedbackOptions) => {
...
+        <MuiGrid
+          item
+          xs={wizardOrigin ? 6 : 12}
+          style={{ margin: "8px 0px", display: "flex", justifyContent: "space-around" }}
+        >
+          <EmojiRatings
+            ratings={survey.ratingLabels}
+            setRating={setSelectedEmoji}
+            placement={wizardOrigin ? "top" : "bottom"}
+            size={wizardOrigin ? "small" : "large"}
```
scarlettperry, line 223: i know the figma spec is the large size, but does it maybe look too big? maybe we should use the `medium` variant (32x32)? but could just be me

scarlettperry (commented):
nice!!

frontend/packages/core/src/NPS/feedback.tsx
jdslaugh: This is part of the change added via #1944. It makes sure the transient property doesn't get out to the dom.

frontend/packages/core/src/NPS/feedback.tsx
jdslaugh: Yeah its a bit large, I'll adjust it to use medium


Screen Shot 2022-01-21 at 3 53 26 PM

frontend/packages/core/src/NPS/header.tsx
```diff
@@ -0,0 +1,134 @@
...
+}));
+
+const Popper = styled(MuiPopper)({
+  padding: "0 12px",
+  marginLeft: "12px",
+  zIndex: 1201,
+});
+
+const Paper = styled(MuiPaper)({
+  width: "420px",
```
scarlettperry, line 31: i know this is the figma spec but I think at this width it gives the appearance of too much white space. was playing around with diff widths and I think `350px` looks better and it also doesn't break existing styling (screenshot below). lmk what you think

Screen Shot 2022-01-24 at 9 24 40 AM

frontend/packages/core/src/NPS/emojiRatings.tsx
```diff
@@ -27,7 +32,13 @@ const getKey = (map, val): string => Object.keys(map).find(key => map[key] === v
  * @param setRating function which will return the given selected rating
  * @returns rendered EmojiRatings component
  */
-const EmojiRatings = ({ ratings = [], setRating }: EmojiRatingsProps) => {
+const EmojiRatings = ({
+  ratings = [],
+  setRating,
+  placement = "top",
+  size = "small",
```
scarlettperry, line 39: nit, can we rename to `iconSize` since we have `emojiSize`?

frontend/packages/core/src/NPS/emojiRatings.tsx
```diff
@@ -27,7 +32,13 @@ const getKey = (map, val): string => Object.keys(map).find(key => map[key] === v
...
+  emojiSize = "medium",
```
scarlettperry, line 40: nit, since we're using the medium emoji size now, can we make the button size smaller as well? i think it's getting the large hover space from the button size

Screen Shot 2022-01-24 at 9 52 58 AM

scarlettperry (commented):
some last comments related to size.

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -82,15 +94,17 @@ const FeedbackAlert = () => {
...
  */
-const NPSFeedback = (opts: FeedbackOptions) => {
+const NPSFeedback = ({ origin = "HEADER", ...options }: FeedbackOptions) => {
   const [hasSubmit, setHasSubmit] = useState<boolean>(false);
-  const [selected, setSelected] = useState<Rating>(null);
+  const [selectedEmoji, setSelectedEmoji] = useState<Rating>(null);
   const [freeformFeedback, setFreeformFeedback] = useState<string>("");
   const [error, setError] = useState<boolean>(false);
   const [survey, setSurvey] = useState<IClutch.feedback.v1.ISurvey>({});
+  const [feedbackType, setFeedbackType] = useState<string>(null);
```
scarlettperry, line 103: something I thought of just now - i'm wondering if there isn't a need to have the `feedbackType` field for the Header & instead standardize the use of `urlPath` for all feedback placement types. Initially I thought there would be value in having both urlPath and feedbackType as data points for the Anytime placement but since we primarily care what workflow received feedback, maybe it's unnecessary to spread the same data across different fields. lmk what you think!

so maybe we can modify like so:
```
const feedbackUrlPath = wizardOrigin ? window.location.pathname : feedbackType
...
},
   urlPath: feedbackUrlPath,
},
```

jdslaugh: Only issue with this would be if we want to provide feedback outside of workflows (ie General). Since we dont have a true URL for these it would be filled in under urlPath which might seem odd (urlPath: General).

We could possibly swap to just 'feedbackType' and have that be where the url path dumps into.

frontend/packages/core/src/NPS/header.tsx
jdslaugh: Adjusted to 350

frontend/packages/core/src/NPS/emojiRatings.tsx
jdslaugh: Yeah thats a good call, adjust

frontend/packages/core/src/NPS/emojiRatings.tsx
jdslaugh: Yeah, I had to override the padding for that size since the default 12px padding is just too much for that size

frontend/packages/core/src/NPS/feedback.tsx
scarlettperry: ah yeah true, keeping `feedbackType` would be better if we make this change.

i'll ask the team what they think in standup today. maybe others will think we should keep the 2 fields as planned

jdslaugh: Sounds good, we could designate '/' for General as well, only might be an issue in future extensions.

scarlettperry: agree that for future extensions `feedbackType` is more flexible than `urlPath`. let's go with that.

I'll update the mode queries to make sure we're not losing old feedback results where `urlPath` is used

jdslaugh: Sounds good

scarlettperry (approved):
RE https://github.com/lyft/clutch/pull/1964#discussion_r790970537, realized the backend currently requires the urlPath field, so opened https://github.com/lyft/clutch/pull/1984 to address the change.

frontend/packages/core/src/NPS/feedback.tsx
jdslaugh: Merged your changes in!

frontend/packages/core/src/NPS/emojiRatings.tsx
```diff
@@ -27,11 +32,18 @@ const getKey = (map, val): string => Object.keys(map).find(key => map[key] === v
  * @param setRating function which will return the given selected rating
  * @returns rendered EmojiRatings component
  */
-const EmojiRatings = ({ ratings = [], setRating }: EmojiRatingsProps) => {
+const EmojiRatings = ({
+  ratings = [],
+  setRating,
+  placement = "top",
+  buttonSize = "small",
+  emojiSize = "medium",
```
dschaller, line 40: Why do we have both button and emoji size?

frontend/packages/core/src/NPS/emojiRatings.tsx
```diff
@@ -27,11 +32,18 @@ const getKey = (map, val): string => Object.keys(map).find(key => map[key] === v
...
+}: EmojiRatingsProps) => {
   const [selectedRating, selectRating] = useState<Rating>(null);
 
   const StyledIconButton = styled(IconButton)<{
     $selected: boolean;
+    size: string;
```
dschaller, line 46: Just confirming, do you want `size` to be forwarded to the `IconButton` component?

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -43,18 +47,26 @@ export const defaults: IClutch.feedback.v1.ISurvey = {
   ],
 };
 
-const StyledButton = styled(Button)({
-  fontSize: "14px",
-  padding: "0 8px",
-  height: "18px",
-});
+const StyledButton = styled(Button)<{ $origin: Origins }>({}, props =>
+  props.$origin === "WIZARD"
```
dschaller, line 51: nit: maybe extract this into constant

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -9,17 +8,22 @@ import { v4 as uuid } from "uuid";
...
 import { client } from "../Network";
 import type { ClutchError } from "../Network/errors";
+import styled from "../styled";
 import { Typography } from "../typography";
 
 import EmojiRatings, { Rating } from "./emojiRatings";
 
 /** Interfaces */
 
+type Origins = "WIZARD" | "HEADER";
```
dschaller, line 22: nit: Since this isn't repeated it might be easier to read if it is `Origin`

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -82,15 +94,17 @@ const FeedbackAlert = () => {
  * @param opts Available feedback options
  * @returns NPSFeedback component
  */
-const NPSFeedback = (opts: FeedbackOptions) => {
+const NPSFeedback = ({ origin = "HEADER", ...options }: FeedbackOptions) => {
```
dschaller, line 97: nit:
```suggestion
const NPSFeedback = ({ origin = "HEADER", ...props }: FeedbackOptions) => {
```

frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -82,15 +94,17 @@ const FeedbackAlert = () => {
...
   const [hasSubmit, setHasSubmit] = useState<boolean>(false);
-  const [selected, setSelected] = useState<Rating>(null);
+  const [selectedEmoji, setSelectedEmoji] = useState<Rating>(null);
```
dschaller, line 99: nit: 
```suggestion
  const [selectedRating, setSelectedRating] = useState<Rating>(null);
```

frontend/packages/core/src/NPS/header.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const Popper = styled(MuiPopper)({
+  padding: "0 12px",
+  marginLeft: "12px",
+  zIndex: 1201,
+});
+
+const Paper = styled(MuiPaper)({
+  width: "350px",
+  border: "1px solid #E7E7EA",
```
dschaller, line 32: I'm not seeing a border in figma but I might be missing it

frontend/packages/core/src/NPS/header.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (event.target.localName === "body") {
+      return;
+    }
+    if (anchorRef.current && anchorRef.current.contains(event.target)) {
+      return;
+    }
+    setOpen(false);
+  };
+
+  const FeedbackIconProps = {
```
dschaller, line 93: just curious, why not inline these?

frontend/packages/core/src/NPS/emojiRatings.tsx
jdslaugh: For extendability, and since they're no longer mirroring each other I wanted to separate them into their own props so we could override them as necessary.

frontend/packages/core/src/NPS/emojiRatings.tsx
jdslaugh: Yep, the IconButton needs the size but then the IconButton will handle filtering out the transient prop any further.

frontend/packages/core/src/NPS/feedback.tsx
jdslaugh: Good idea, I'll pull it out so I can remove the props

frontend/packages/core/src/NPS/header.tsx
jdslaugh: Leftover from copying from the user popout. Can't see a border anyways, so removing.

frontend/packages/core/src/NPS/header.tsx
jdslaugh: Not sure why I left it that way, good call though, inlining and removing

frontend/packages/core/src/NPS/feedback.tsx
jdslaugh: Good call, adjusting

frontend/packages/core/src/NPS/feedback.tsx
jdslaugh: Can do

frontend/packages/core/src/NPS/emojiRatings.tsx
dschaller: :+1: , just thinking about it though would we ever want to allow these to be that different stylistically?

dschaller (approved):

PR #1965, opened 2022-01-20
FE: Add a central file for organizing colors and files to build the foundation for theming

smonero:
### Description

Setting up a file that other files will be able to reference for colors

The idea here is that this is one step in organizing colors for Clutch. Another step would be replacing various places where we use colors with references to this file instead.

I met with Derek, and he did a lot of work. We originally thought about removing exports because this area is still under construction. I also thought about just adding TODOs for the theming because it seems that no external contributors are interested in working on the theming currently.

### Testing Performed

none

### GitHub Issue

Fixes #


frontend/packages/core/src/Assets/colors.tsx
```diff
@@ -0,0 +1,195 @@
+// COLORS FROM FIGMA
+// NOTE THAT THOSE WITH GRADIENTS ONLY HAVE THEIR BASES
+
+// LIGHT MODE COLORS ////////////////////////////////////////////////////////
+
+// NEUTRAL
+export const CLUTCH_LIGHTMODE_NEUTRAL_5 = "#F8F8F9"; // PRIMARY BUTTON DISABLED / STROKE TERTIARY
```
jdslaugh, line 7: Thanks for adding this!

For the palette system I'm thinking it might work to have it defined a little differently.

Maybe
`const LIGHTMODE: CLUTCHPALETTE = {
NEUTRAL: {...}
};`

This will allow us to type everything so we can extend it easily later and then when we want to swap before palettes it's an easy reference change.

What do you think?

smonero: good idea

frontend/packages/core/src/Assets/colors.tsx
```diff
@@ -0,0 +1,209 @@
+// COLORS FROM FIGMA
+// NOTE THAT THOSE WITH GRADIENTS ONLY HAVE THEIR BASES
+
+interface CLUTCH_PALETTE {
```
smonero, line 4: wasn't sure if this should be capitalized like this

jdslaugh: We could rename it to `ClutchPalette` as well, might look a little cleaner.

One thing to make it a bit easier could be to split out the types from this file, so it doesn't get cluttered, and then type every key as well.

```
export interface ClutchPalette {
   NEUTRAL: NeutralColors;
   BLUE: BlueColors;
}

interface NeutralColors {
   5: string;
   10: string;
   ...
}
```

This way we have every possible value defined and if we want to add any new palettes it will require every single color be present before it will allow it.

frontend/packages/core/src/Assets/colors.tsx
```diff
@@ -0,0 +1,197 @@
+// COLORS FROM FIGMA
+// NOTE THAT THOSE WITH GRADIENTS ONLY HAVE THEIR BASES
+import type { ClutchPalette } from "./colortypes";
+
+export const LIGHT_MODE: ClutchPalette = {
+  NEUTRAL: {
+    5: "#F8F8F9", // PRIMARY BUTTON DISABLED / STROKE TERTIARY
```
jdslaugh, line 7: Looks great! One nit, could we move these line comments to the interfaces so they're referenced if someone creates a new palette?

jdslaugh (approved):

dschaller:
Should we put this in `AppLayout` instead of `Assets` since it'll be injected there?

frontend/packages/core/src/Assets/colortypes.tsx
```diff
@@ -0,0 +1,113 @@
...
+  15: string;
+  20: string;
+  25: string;
+  30: string; // STROKE SECONDARY / TEXT TERTIARY
+  40: string;
+  50: string; // TEXT SECONDARY
+  60: string;
+  70: string;
+  80: string;
+  90: string;
```
dschaller, line 14: I know the figma palette differs from materials but we should try to rectify that. e.g. material already has an interface for Color that we should try to use here.

smonero: I guess I'm a bit confused on a plan of action here. Are you saying we should be using the material colors like this example? 
`import { purple, red } from '@mui/material/colors';

const primary = red[500]; // #f44336
const accent = purple['A200']; // #e040fb
const accent = purple.A200; // #e040fb (alternative method)
`
Or are you saying we should be putting these values into a material theme / palette, like this?

`
import { createTheme } from '@mui/material/styles';
import blue from '@mui/material/colors/blue';

const theme = createTheme({
  palette: {
    primary: blue,
  },
});
`
I found this site https://bareynol.github.io/mui-theme-creator/ which I guess could help with this. 

The idea behind this PR was to have a central file so that in other places in Clutch instead of hardcoding the colors like `x = "#F12345"` we could just have `x =  pallete.Neutral.10 . A step after this would be actually putting the colors into a palette like above.

dschaller: yeah, the idea of using a palette throughout makes sense but I think you'll run into issues since material's palette expects colors to have a certain interface in order to access the various shades. e.g. 
```tsx
export interface Color {
  50: string;
  100: string;
  200: string;
  300: string;
  400: string;
  500: string;
  600: string;
  700: string;
  800: string;
  900: string;
  A100: string;
  A200: string;
  A400: string;
  A700: string;
}
```

frontend/packages/core/src/AppLayout/colortypes.tsx
```diff
@@ -0,0 +1,47 @@
+import type { Color } from "@material-ui/core";
+
+export interface StrokeColors {
+  primary: string;
+  secondary: string;
+  tertiary: string;
+  inverse: string;
+}
+
+export interface StatesColors {
```
dschaller, line 10: Would it make sense to have something like 
```tsx
interface StateColor {
  active: string;
  default: string;
}
```
and then below something like 
```tsx
...
export interface ClutchColorChoices {
  Stroke: StrokeColors;
  Background: BackgroundColors;
  States: {
    hover: StateColor;
    focused: StateColor;
    pressed: StateColor;
    selected: StateColor;
  };
```

frontend/packages/core/src/AppLayout/colortypes.tsx
```diff
@@ -0,0 +1,47 @@
...
+  Green: Color;
+  Amber: Color;
+  Red: Color;
+}
+// TODO: choose a better name for this
+export interface ClutchColorChoices {
+  Stroke: StrokeColors;
+  Background: BackgroundColors;
+  States: StatesColors;
+  Typography: TypographyColors; // Typography and Icons
```
dschaller, line 46: minor nit: these keys should be lowercase

frontend/packages/core/src/AppLayout/colors.tsx
```diff
@@ -0,0 +1,233 @@
...
+    A200: "",
+    A400: "",
+    A700: "",
+  },
+};
+
+// TODO: come up with a better name
+// The reason why I separated them from above is the above is JUST colors
+export const LIGHT_MODE_CHOICES: ClutchColorChoices = {
+  Stroke: {
```
dschaller, line 91: Shouldn't these leverage the color palette so we only need to update one place?

smonero: oh yeah now I see. For some reason I didn't think of that

smonero: hm now that I think about this more I think we need to get Kanchan to make some changes. As in, have colors like black / grey / white etc then put those in a palette, then we can reuse them in the color choices object

jdslaugh: Just found this but looks like MUI supports an alpha functionality to allow for easily changing opacities of theme colors: https://mui.com/guides/migration-v4/#styles

```
import { alpha } from '@mui/material/styles';

 const classes = makeStyles(theme => ({
      backgroundColor: alpha(theme.palette.primary.main, theme.palette.action.selectedOpacity),
 }));
```

frontend/packages/core/src/AppLayout/colors.tsx
```diff
@@ -0,0 +1,256 @@
...
+    A100: "",
+    A200: "",
+    A400: "",
+    A700: "",
+  },
+};
+
+export const DARK_MODE_CHOICES: ClutchColorChoices = {
+  typography: {
+    primary: `${DARK_MODE.neutral[900]}E6`,
```
smonero, line 216: @dschaller just saw the comment about using alpha() in mui5. Should I keep the alpha as a separate field here? Like:
primary: color
alpha: .9

etc. ?

smonero: did some research, didn't really come to a good conclusion though. I wonder if we would be better off organizing our colors like this? https://mui.com/customization/default-theme/?expand-path=$.palette (also they use rgba(x,y,z,o) instead of just putting everything in hex). Or having the palette follow the format like here https://mui.com/customization/palette/

frontend/packages/core/src/AppLayout/colors.tsx
```diff
@@ -0,0 +1,256 @@
...
+    50: "#232542",
+    100: "#30324E",
+    200: "#3E4059",
+    300: "#55566D",
+    400: "#77788A",
+    500: "#8D8E9E",
+    600: "#A4A5B1",
+    700: "#D2D2D8",
+    800: "#E8E8EB",
+    900: "#F8F8F9",
```
dschaller, line 142: It looks like this should be `#FFFFFF`

smonero:
based off the discussion on Tues I updated the states with `alphas`. let me know if you like it better the old way, as I wasn't 100% sure if thats what I should do

smonero:
removing on hold since derek is back

dschaller (commented):
Looks good once we remove the typography

dschaller (approved):

PR #1973, opened 2022-01-21
dynamodb: fix scaling limits

angethuy:
### Description
This PR fixes unintended behavior in the DynamoDB service's logic for enforcing AWS capacity provisioning limits. When scaling up a table that is already provisioned above the max capacity (40000 by default), requests are always erroneously blocked because the requested capacity exceeds the limit.

The fix is to ignore that max limit for tables that are already provisioned above the maximum, and only enforce the scale-factor limit instead (200% of current capacity by default) unless the ignore flag is passed.

### Testing Performed

Unit tests


danielhochman (approved):
great job on this PR! thorough tests and clear comments.

PR #1984, opened 2022-01-24
feedback: remove `urlPath` field from API and replace with `feedbackType`

scarlettperry:
### Description
Addresses https://github.com/lyft/clutch/pull/1964#discussion_r790970537

Initially, the same data was going to be spread across 2 different fields depending on the feedback placement (ie. wizard placement would populate `urlPath` and the header placement would populate `feedbackType`). We decided it would be cleaner and best practice to use one field to collect the workflow url.
* We decided to use `feedbackType` instead of `urlPath` because in the future we might collect feedback on features that won't necessarily have an associated URL.

### Testing Performed
locally end-to-end and viewed DB submissions in pgAdmin.

(highlighted submission is the recent change)
Screen Shot 2022-01-24 at 6 11 18 PM


api/feedback/v1/feedback.proto
```diff
@@ -93,29 +93,24 @@ message FeedbackMetadata {
   // extra info on the feedback (i.e. the survey question, the rating options, the feedback origin, etc.)
   Origin origin = 1 [ (validate.rules).enum = {defined_only : true not_in : 0} ];
   Survey survey = 2 [ (validate.rules).message = {required : true} ];
-  // TODO: remove if we decide to record feedback only when a user clicks the submit button
```
scarlettperry, line 96: small cleanup as we're using the `user_submitted field`

api/feedback/v1/feedback.proto
```diff
@@ -93,29 +93,24 @@ message FeedbackMetadata {
...
   string freeform_response = 4;
-  // (optional) some UI components (i.e the header) will have a dropdown menu for choosing the type of
-  // feedback to submit (i.e. "General", "K8s Delete Pod")
-  string feedback_type = 5;
 }
 
 message SubmitFeedbackRequest {
   // client-genereated unique feedback id, which we will also use to update the feedback (essentially replace with the
   // latest)
-  // TODO: remove if we decide to record feedback only when a user clicks the submit button
```
scarlettperry, line 118: small cleanup as we're using the `id` field

api/feedback/v1/feedback.proto
```diff
@@ -93,29 +93,24 @@ message FeedbackMetadata {
...
   Survey survey = 2 [ (validate.rules).message = {required : true} ];
-  // TODO: remove if we decide to record feedback only when a user clicks the submit button
   bool user_submitted = 3;
   string url_search_params = 4;
 }
 
 message Feedback {
-  // url path of where the feedback was submitted
-  string url_path = 1 [ (validate.rules).string = {min_bytes : 1} ];
+  // url path or workflow name that received feedback
```
jdslaugh: nit: Since 'General' and possibly others wont be a url path or a workflow name, should we adjust this text?

Maybe:
'url path, workflow name or area of feedback'

jdslaugh (approved):
Looks good from a FE and general perspective. Thanks for adjusting this!

PR #1989, opened 2022-01-26
frontend: Remove feature flag setup from NPS wizard placement

scarlettperry:
### Description
We've fully launched the feature internally and monitored it for a few weeks to make sure it's feature ready, so PR removes the feature flag setup in the Wizard.

### Testing Performed
manually

### DIFF
--- a/frontend/packages/app/src/clutch.config.js
+++ b/frontend/packages/app/src/clutch.config.js
@@ -49,7 +49,6 @@ module.exports = {
     deletePod: {
       componentProps: {
         resolverType: "clutch.k8s.v1.Pod",
-        enableFeedback: true,
       },
     },
     resizeHPA: {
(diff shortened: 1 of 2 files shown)


danielhochman:
isn't this still potentially useful for others who are using the wizard component outside of our internal deployment?

scarlettperry:
> isn't this still potentially useful for others who are using the wizard component outside of our internal deployment?

i don't think so since the NPS wizard isn't globally enabled by default. they'd have to add the `enableFeedback` prop per workflow in the clutch js config (docs coming soon!). i thought it'd be a good idea to remove the feature flag by default since it's feature ready but can leave it if that's the preference.

danielhochman (approved):

scarlettperry:
synced with @danielhochman offline. we decided we should leave the feature flag AND globally enable the NPS wizard on all workflows if the feature flag for it is enabled. Same goes for the NPS header placement.

PR #1915, opened 2021-12-06
feedback: add enableFeedback prop to workflow components

scarlettperry:
### Description
Related PR https://github.com/lyft/clutch/pull/1903

### GitHub Issue
https://github.com/lyft/clutch/issues/1797

### TODOs
- [x] https://github.com/lyft/clutch/pull/1917

PR #1917, opened 2021-12-07
feedback: NPS Collect (Wizard)

jdslaugh:
### Description
Initial Implementation of NPS Collector for the Wizard Component.

Currently enabled in only 1 workflow for verification, locked behind a feature flag.

Other work
- Minor adjustments to touched index files to re-export instead of import -> export
- Added emoji selector with a switch to return based on a given type
- Modification of Feedback/tooltip property to remove a console warning
- Added index files for Input
- Extended iconButton to take a size property

##### Wizard Placement
Screen Shot 2021-12-06 at 3 45 24 PM

##### After Emoji Selection
Screen Shot 2021-12-06 at 3 45 33 PM

##### Text Field Validation
Screen Shot 2021-12-06 at 3 46 27 PM

##### Successful Validation
Screen Shot 2021-12-06 at 3 46 36 PM

##### After Submission (via Button)
Screen Shot 2021-12-06 at 3 46 41 PM

### Testing Performed
Manual testing with feature flags

### TODO
- [ ] Modify the survey to fetch on load (similar to feature flags) versus on component load

PR #1990, opened 2022-01-28
feedback: globally enable NPS wizard

scarlettperry:
### Description
Follow up to: https://github.com/lyft/clutch/pull/1989#issuecomment-1023369994. We decided to globally enable the NPS Wizard placement as opposed to having it enabled per workflow given that users will most likely either want to enable it 100% or not at all. To enable it, users will need to turn the feature flag on.

Will follow up with docs for clutch.sh

### Testing Performed
manually tested all workflows

PR undoes the changes in
1) https://github.com/lyft/clutch/pull/1903
2) https://github.com/lyft/clutch/pull/1915

### GitHub Issue

### DIFF
--- a/frontend/packages/wizard/src/wizard.tsx
+++ b/frontend/packages/wizard/src/wizard.tsx
@@ -136,28 +136,26 @@ const Wizard = ({ heading, width = "default", dataLayout, children }: WizardProp
           </WizardContext.Provider>
         </DataLayoutContext.Provider>
         <Grid container justify="center">
-          {((state.activeStep === lastStepIndex && !isLoading) || hasError) && (
+          {((state.activeStep === lastStepIndex && !isLoading && isMultistep) || hasError) && (
             <>
               <SimpleFeatureFlag feature="npsWizard">
                 <FeatureOn>
                   <NPSWizard />
                 </FeatureOn>
               </SimpleFeatureFlag>
-              {(isMultistep || hasError) && (
-                <ButtonGroup>
-                  <Button
-                    text="Start Over"
-                    onClick={() => {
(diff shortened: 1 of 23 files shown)


frontend/packages/wizard/src/wizard.tsx
```diff
@@ -136,28 +136,26 @@ const Wizard = ({ heading, width = "default", dataLayout, children }: WizardProp
           </WizardContext.Provider>
         </DataLayoutContext.Provider>
         <Grid container justify="center">
-          {((state.activeStep === lastStepIndex && !isLoading) || hasError) && (
```
scarlettperry, line 152: noticed that we removed the `isMultistep` check in https://github.com/lyft/clutch/pull/1917 which was previously there(https://github.com/lyft/clutch/blame/917d40a9001d1aee8e28692d74a2e01f80383ccc/frontend/packages/wizard/src/wizard.tsx). Now that we're globally enabling the feedback, going to keep it to the original logic in case there were unknowns unknowns that didn't break before.

scarlettperry: cc @jdslaugh in case there's context I'm missing for the change

jdslaugh: Yeah, so the context for that is based on isMultistep. In this change we will only show feedback if we're a multistep wizard workflow and on the last step. So if there's a single step wizard (not sure if there is or will be), we'll never see it even if it's complete.

That was the reason for the adjustment in logic and moving the start over button under hasError/isMultistep

scarlettperry: ah got it! will undo my changes

PR #1998, opened 2022-02-01
frontend: nps - defaulting displayName

jdslaugh:
### Description
If a workflow route does not have a display name it would be blank, in this case we will default to the workflows display name.

### Testing Performed
Manual


scarlettperry (approved):
thank you!

smonero (approved):

PR #2004, opened 2022-02-02
frontend: NPS Sizing Adjustments

jdslaugh:
### Description
- Fixed combinations of workflows
- Adjusting spacing around label
- Adjusted select styling to max out to 50% of viewport

### DIFF
--- a/frontend/packages/core/src/NPS/header.tsx
+++ b/frontend/packages/core/src/NPS/header.tsx
@@ -65,7 +65,7 @@ export const generateFeedbackTypes = (workflows: Workflow[]): SelectOption[] => 
     typeMap[group].push(
       ...routes.map(route => ({
         label: route.displayName || displayName,
-        value: `/${path}/${route.path}`,
+        value: `/${path}/${route.path}`.replace(/\/\/+/g, "/"),
       }))
     );
   });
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/NPS/header.tsx
```diff
@@ -53,19 +53,29 @@ const StyledFeedbackIcon = styled(IconButton)<{ $open: boolean }>(
...
+    const { group, path, routes, displayName } = workflow;
+
+    if (!typeMap[group]) {
+      typeMap[group] = [];
+    }
+
+    typeMap[group].push(
+      ...routes.map(route => ({
+        label: route.displayName || displayName,
+        value: `/${path}/${route.path}`,
```
scarlettperry, line 68: noticed that the `route.path` for some workflows (like Dash) already included the `/` so the feedbackType ended up being `dash//`

came up w. this but feel free to change the logic if it's too hacky 😅 
```suggestion
        value: route.path.charAt(0) === "/" ? `${path}${route.path}` : `${path}/${route.path}`,
```

jdslaugh: Alright, modified it to replace all multiple occurrences with a single slash, so this should cover most cases.

```
`/${path}/${route.path}`.replace(/\/\/+/g, "/"),
```

smonero (approved):

PR #2006, opened 2022-02-04
dynamodb: fix scaling factor limit

angethuy:
### Description

This PR fixes how capacity scaling limits are being checked in the DynamoDB service's `UpdateCapacity` function.

Context: The `UpdateCapacity` function has a configurable scaling limit (default 2.0x) that it checks target capacities against. If a target capacity request exceeds the scaling limit, an error is returned.

Bug: bad float math in formula. Lost precision in dividing int by double and casting the results to float e.g. `float32(100 / 2.0)` so the scaling factor limit wasn't applied to certain requested values unless they largely exceeded the scale factor. I also initially used bad unit test values so error wasn't caught ):

Fix: both operands typed as float32 before dividing so decimal precision is retained and accurately compared to scale factor limit

### Testing Performed

Unit testing with edge values


scarlettperry (approved):
lgtm! thanks for the detailed PR description and comments throughout

PR #2011, opened 2022-02-07
frontend: ProjectSelector - Load state error

jdslaugh:
### Description
- Reverts state change
- Adds all stored projects into api request, vs just the custom ones
- Will utilize partialFailures to track deprecated projects and remove them from the list
- Once we've loaded all projects, any missing upstreams or downstreams will then be removed as well

### Testing Performed
Manual, custom application storage


frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -63,28 +75,43 @@ const StyledProgressContainer = styled.div({
...
+
+  // we've received all data for our projects, check the upstreams / downstreams
+  const missing = upDownKeys.filter(p => !projectCheck(p));
+  if (missing.length) {
+    // we're missing upstreams / downstreams, remove them from our state
+    dispatch({
+      type: "HYDRATE_DEPRECATION",
+      payload: { result: { missing, type: "STREAMS" } },
+    });
+  }
```
scarlettperry, line 101: for my understanding, why is there a `"HYDRATE_DEPRECATION` call here and after the API call

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -93,17 +120,28 @@ const hydrateProjects = (state: State, dispatch: React.Dispatch<Action>) => {
...
       .post("/v1/project/getProjects", requestParams as IClutch.project.v1.GetProjectsRequest)
       .then(resp => {
-        const { results } = resp.data as IClutch.project.v1.GetProjectsResponse;
+        const { results, partialFailures } = resp.data as IClutch.project.v1.GetProjectsResponse;
+
+        if (partialFailures && partialFailures.length) {
+          const missing: string[] = [];
+          partialFailures.forEach(failure => {
+            (failure.details || []).forEach(detail => {
+              if (_.get(detail, ["data", "disabled"])) {
```
scarlettperry, line 134: since implementers of the projectAPI can stub the `data` field with any value, I think this would break for users who aren't handling these kinds of errors the way we are internally https://github.com/lyft/clutch/blob/main/api/core/project/v1/project.proto#L18

scarlettperry: since the `HYDRATE_DEPRECATION` error case may not be relevant to all project api users, should we instead handle the partial failure error in the `HYDRATE_ERROR` action (since the partial failure response will include a message like `"x project is disabled"`) and in that action remove those projects from the state?

jdslaugh: Mm, good call. Guess we won't be able to check that object then.

frontend/workflows/projectSelector/src/project-selector.tsx
jdslaugh: Once we've loaded our projects there is a chance that a given upstream or downstream that was previously stored is also deprecated and will need to be removed from the state so that it can proceed

jdslaugh: Modified this logic to utilize HYDRATE_ERROR to remove missing projects from the state

frontend/workflows/projectSelector/src/project-selector.tsx
jdslaugh: Removed HYDRATE_DEPRECATION and moved the logic into HYDRATE_ERROR, utilizing the message and the full details object. Might need to be even more generic

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -112,36 +153,34 @@ const hydrateProjects = (state: State, dispatch: React.Dispatch<Action>) => {
...
 
   // This hook updates the global dash state based on the currently selected projects for cards to consume (including upstreams and downstreams).
   React.useEffect(() => {
-    if (!allPresent(state)) {
+    if (!allPresent(state, dispatch)) {
       // Need to wait for the data.
       return;
     }
 
+    const interval = setInterval(() => hydrateProjects(state, dispatch), 30000);
```
jdslaugh, line 178: Moved the setInterval inside of our useEffect that reloads on state changes as I was seeing issues with the same state being sent every time the interval triggered. 

Think this should clearInterval on a re-render/state change and set a new interval to hydrate everything (testing worked at least).

dschaller: By moving this out it will be called every time there is any state change to the variable above. I think the way to handle this would be to continue using the useEffect hook but update the interval every time only the state changes.

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -9,18 +9,32 @@ import LayersOutlinedIcon from "@material-ui/icons/LayersOutlined";
...
+} from "./helpers";
 import ProjectGroup from "./project-group";
 import selectorReducer from "./selector-reducer";
-import { storeState } from "./storage";
+import { loadStoredState, storeState } from "./storage";
 import type { Action, DashState, State } from "./types";
 import { Group } from "./types";
 
+export interface ProjectSelectorError {
+  errors: IGoogle.rpc.IStatus[];
```
dschaller, line 25: Are these partial errors coming back from the server? If so it might be good to denote that in some way

frontend/workflows/projectSelector/src/project-selector.tsx
jdslaugh: Ah ok, thought it might've been cleaner having it inside the other useEffect but probably best to keep them separated. I will move it back into the original useEffect and set the dependency on the state.

frontend/workflows/projectSelector/src/project-selector.tsx
jdslaugh: Yeah, they're from the partialFailures response. Ill add a docstring around it, we can rename it though if we think that'll be better

dschaller (approved):

mikecutalo (approved):

PR #2020, opened 2022-02-11
FE: Input: Change Input Textfield to be fullWidth from small

smonero:
### Description

When using `small` the size doesn't fill up containers, so we can change it to `fullWidth` to get the behavior we want.
Note this is only for autocomplete. Without this, if you added autocomplete to a text field that previously did not have autocomplete, the size might change.

### Testing Performed

local

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Input/text-field.tsx
+++ b/frontend/packages/core/src/Input/text-field.tsx
@@ -163,6 +163,7 @@ export interface TextFieldProps
       | "defaultValue"
       | "disabled"
       | "error"
+      | "fullWidth"
       | "helperText"
       | "id"
       | "inputRef"
@@ -180,7 +181,6 @@ export interface TextFieldProps
     Pick<MuiInputProps, "readOnly" | "endAdornment"> {
   onReturn?: () => void;
   autocompleteCallback?: (v: string) => Promise<{ results: { id?: string; label: string }[] }>;
-  autocompleteFullWidth?: boolean;
 }
 
 const TextField = ({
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -252,7 +252,7 @@ const TextField = ({
     return (
       <Autocomplete
         freeSolo
-        size="small"
+        fullWidth
```
dschaller, line 255: does this impact any of the other uses of autocomplete?

smonero: no - tested downstreams and it seems the same. do you think we should pass it is a prop and have the default be small, or should it always be `fullWidth`? The reason why I was thinking it should always be fullWidth is that a lot of times people are using it inside some container so they can make it smaller there.

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -191,6 +192,7 @@ const TextField = ({
   endAdornment,
   autocompleteCallback,
   defaultValue,
+  autocompleteFullWidth = true,
```
jdslaugh, line 195: I would switch this to just be `fullWidth = true`

Since its a valid prop for both autocomplete and textfield, then you can just do `fullWidth` on both Autocomplete and TextField

jdslaugh (approved):

dschaller (approved):

PR #2030, opened 2022-02-15
Frontend: Toast component

jdslaugh:
### Description
- ~~Modified warning component to use built in alert component so we could utilize some features from it as well as work towards consistency on colors for the color palette~~
- Created new Toast component out of the old Warning component as the naming didn't fit the intended purpose any longer
- Added deprecation messaging around Warning component and extended it for backwards compatibility
- Updated storybook for both Warning and Toast

#### Before (w/o title)
Screen Shot 2022-02-15 at 2 42 54 PM

#### After (Warning w/o title)
Screen Shot 2022-02-15 at 2 41 42 PM

#### After (Warning w/title)
Screen Shot 2022-02-15 at 2 41 59 PM

#### After (Info)
Screen Shot 2022-02-15 at 4 49 57 PM

### Testing Performed
Manual


frontend/packages/core/src/Feedback/toast.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Warning component
+ * @param message the message to display in the warning
+ * @deprecated use Toast component
+ */
+export const Warning: React.FC<WarningProps> = ({ message, ...props }) => {
+  if (process.env.NODE_ENV === "development") {
+    // eslint-disable-next-line no-console
+    console.warn("Warning component is deprecated, please use Toast component instead");
```
jdslaugh, line 63: Not sure if there's a better way to do this, seemed to be one of the recommended ways (other than creating a helper class).

smonero (approved):

frontend/packages/core/src/Feedback/toast.tsx
```diff
@@ -0,0 +1,51 @@
...
+      open={open}
+      autoHideDuration={duration}
+      onExiting={onDismiss}
+      onClose={(_, reason) => {
+        // This way it will not auto close when clicking in the window, will instead wait on the timeout or onClose
+        if (reason !== "clickaway") {
+          setOpen(false);
+        }
+      }}
+      anchorOrigin={{ vertical: "bottom", horizontal: "right" }}
```
dschaller, line 38: Should we make this configurable? Wondering if we always want this in the bottom right

frontend/packages/core/src/Feedback/toast.tsx
```diff
@@ -0,0 +1,51 @@
...
+      onClose();
+    }
+    setOpen(false);
+  };
+
+  return (
+    <Snackbar
+      open={open}
+      autoHideDuration={duration}
+      onExiting={onDismiss}
```
dschaller, line 31: if someone closes the alert will `onDismiss` be triggered twice?

frontend/packages/core/src/Feedback/toast.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import { Snackbar } from "@material-ui/core";
+
+import type { AlertProps } from "./alert";
+import { Alert } from "./alert";
+
+export interface ToastProps extends AlertProps {
```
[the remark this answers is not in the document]
jdslaugh: Yeah, really only did that since it's a full wrapper around the Alert component but I'll limit it to a pick

frontend/packages/core/src/Feedback/toast.tsx
jdslaugh: Yeah it would probably be a good idea to make it customizable, and @material-ui already types it as well, so its easy

frontend/packages/core/src/Feedback/toast.tsx
jdslaugh: Ooh great catch, yeah it will. I'll add a check to see if it's open before triggering.

frontend/packages/core/src/Feedback/toast.tsx
jdslaugh: Extended the type from snackbarprops so its fully customizable but still defaulted

frontend/packages/core/src/Feedback/toast.tsx
jdslaugh: Added an `open` check in `onDismiss`

frontend/packages/core/src/Feedback/toast.tsx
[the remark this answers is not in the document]
jdslaugh: Modified this to limit extending props to only those necessary

dschaller (approved):

PR #2039, opened 2022-02-18
housekeeping: update react-hook-form dependencies

dschaller:
### Description

housekeeping: update react-hook-form dependencies

### Testing Performed

Manual


smonero (approved):

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -126,6 +128,7 @@ const SchemaResolver = ({ schema, expanded, onClick, submitHandler }: SchemaReso
 
   return (
     <Form noValidate onSubmit={schemaValidation.handleSubmit(() => submitHandler(data))}>
+      {/* {process.env.REACT_APP_DEBUG_FORMS && <DevTool control={schemaValidation.control} />} */}
```
jdslaugh, line 131: Is this supposed to be commented out when the other one isnt? Not really sure what it's doing

jdslaugh (approved):

frontend/packages/core/src/Resolver/input.tsx
dschaller: Yeah, I left it as a comment for now but we should figure out how to display a debug tool based on the resolver input type.

PR #2041, opened 2022-02-18
BE: aws: add safety check when accessing accounts map

smonero:
### Description

If the user supplies a value that doesn't exist in the accounts map, there would be a panic.
This PR adds a safety check to prevent that.

### Testing Performed

local
### GitHub Issue

Fixes #

### TODOs

- [x] add some unit tests for this function

PR #2043, opened 2022-02-20
BE: aws: handle cases like "aws:///" when applicable

smonero:
### Description

Output from kubectl when getting nodes has them in the format of: `aws:///us-east-1/i-0f8f8f8f`
To allow users to use this format in queries, we can first check if the account exists, and if it does, use it, but if it doesn't continue the original code path of fanout searching using the instance and not using the account and region.

### Testing Performed

local
### GitHub Issue

Fixes #


backend/resolver/aws/aws.go
```diff
@@ -164,6 +165,13 @@ func (r *res) Resolve(ctx context.Context, wantTypeURL string, input proto.Messa
 func (r *res) Search(ctx context.Context, typeURL, query string, limit uint32) (*resolver.Results, error) {
 	switch typeURL {
 	case typeURLInstance:
+		// Output from kubectl when getting nodes has them in the format of: aws:///us-east-1/i-0f8f8f8f
+		// To allow users to use this format in queries, we can remove the prefix.
+		// Note that if we don't do this, the aws:// will be considered the account due to the pattern matching
+		// done via the protobuf.
+		// TrimPrefix returns s without the provided leading prefix string. If s doesn't start with prefix, s is returned unchanged.
+		query = strings.TrimPrefix(query, "aws:///")
```
mikecutalo, line 173: I dont think hard coding here is the way to go just for kubectl output, we should think of a different way, maybe something thats configurable. Or just got back to the drawing board rethink this whole issue.

dschaller (commented):
Looks good to me. I think the only callout would be that if a user were to explicitly state an account / region / ID but the account were to fail this potentially could still match another instance with the same ID under a different account which might be confusing (e.g. the wizard would be showing a different account than the one specified). I wonder if we could denote that somehow on the response

smonero:
> Looks good to me. I think the only callout would be that if a user were to explicitly state an account / region / ID but the account were to fail this potentially could still match another instance with the same ID under a different account which might be confusing (e.g. the wizard would be showing a different account than the one specified). I wonder if we could denote that somehow on the response

hm yeah I can add a note or something

backend/resolver/aws/aws.go
```diff
@@ -168,10 +168,15 @@ func (r *res) Search(ctx context.Context, typeURL, query string, limit uint32) (
 		if err != nil {
 			return nil, err
 		}
+
 		if ok {
-			return r.instanceResults(ctx, patternValues["account"], patternValues["region"], []string{patternValues["instance_id"]}, limit)
+			// in the case of account not found, we continue to the below flow of only being concerned with the ID, rather than the account and region
+			if _, ok := r.client.AccountsAndRegions()[(patternValues["account"])]; ok {
```
smonero, line 174: Note that I chose to use the `AccountsAndRegions` function since `Accounts` gives me a list rather than a map and I wanted a map. I thought about this and realized that there are probably very few accounts, so it wouldn't really matter if it was a list or map. Or I could add a new function similar to `Accounts` but it returns a map instead. Many options here.

danielhochman, line 174: nit: redundant parenthesis
```suggestion
			if _, ok := r.client.AccountsAndRegions()[patternValues["account"]]; ok {
```

but you could also assign to a local var which would improve readability

backend/resolver/aws/aws.go
```diff
@@ -168,10 +168,15 @@ func (r *res) Search(ctx context.Context, typeURL, query string, limit uint32) (
 		if err != nil {
 			return nil, err
 		}
+
 		if ok {
-			return r.instanceResults(ctx, patternValues["account"], patternValues["region"], []string{patternValues["instance_id"]}, limit)
+			// in the case of account not found, we continue to the below flow of only being concerned with the ID, rather than the account and region
```
danielhochman, line 173: this is different behavior though, right? should we strip off `aws://` instead somewhere?

smonero: There was an attempt to do that earlier using `TrimPrefix` but it was decided that that is not a good solution as per Mike's comment:

> I dont think hard coding here is the way to go just for kubectl output, we should think of a different way, maybe something thats configurable. Or just got back to the drawing board rethink this whole issue.

Valid choices for accounts right now are things like `zimride` , `zimridestaging` etc. Without this PR, if the user enters an `aws:///` input, they would just get an error of `account not found`. This PR allows for them to not just get stuck at that error, but be able to continue using the default fan-out search. Note that a while ago before I added the error check PR (https://github.com/lyft/clutch/pull/2041) if you entered an `aws:///` input it would just panic.

danielhochman: I just wonder if we shouldn't internalize that to our implementation somehow, either by passing a strip function into the frontend (i think this is easiest maybe?) or extending the backend. i.e. the configurable option.

IMO, it would be strange for users to request `badprefix://account000/foo` only for the resolver to return `account999/foo`

smonero: lets sync on this tomorrow (thursday)

smonero: update: we synced and the plan is to have a configureable function that we can pass via the config to manipulate strings (takes in a string, returns a string) to be able to operate on resolvers

PR #2058, opened 2022-02-24
change feedback max length to 280 from 180

smonero:
### Description

We received feedback that 180 characters was not sufficient, so it has been bumped to 280. I thought about making it 500 or 360 but wanted to keep it as short as possible.
edit: Thanks to dschaller's suggestion of tweet length of 280

### Testing Performed

unit
### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/NPS/feedback.tsx
+++ b/frontend/packages/core/src/NPS/feedback.tsx
@@ -85,7 +85,7 @@ const FeedbackAlert = () => {
     </Alert>
   );
 };
-
+export const FEEDBACK_MAX_LENGTH = 360;
 /**
  * NPS feedback component which is the base for both Wizard and Anytime.
  * Will fetch given survey options from the server based on the given origin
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/NPS/feedback.tsx
```diff
@@ -85,7 +85,7 @@ const FeedbackAlert = () => {
     </Alert>
   );
 };
-
+export const FEEDBACK_MAX_LENGTH = 360;
```
dschaller, line 88: Just curious, what made you decide on 360? We could also make this tweet size at 280

smonero: I just thought doubling it would be decent. The individual who gave feedback's response was 264 characters long (it was split across 2 responses) so 280 could work as well.

mikecutalo (approved):

jdslaugh (approved):

dschaller (approved):

PR #2071, opened 2022-03-01
shortlink: refactor protos & add db table

mikecutalo:
### Description
Refactored the state protos slightly, switch from `any` -> `value` type for `state` as it will accept undefined json data.

Also included in this PR is the shortlink table migration scripts.


mikecutalo:
Yeah true i can do that, it will look a lot cleaner

danielhochman (approved):

jdslaugh (approved):

PR #2077, opened 2022-03-03
Replace nivo with recharts for pie chart in envoy triage

smonero:
### Description

The goal here is to replace the nivo pie chart with a recharts pie chart.

before
Screen Shot 2022-03-03 at 10 12 52 AM

after
Screen Shot 2022-03-03 at 10 12 35 AM

### Testing Performed

staging

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/envoy/src/remote-triage/dashboard.tsx
+++ b/frontend/workflows/envoy/src/remote-triage/dashboard.tsx
@@ -26,8 +26,6 @@ const FeaturedSummaryContainer = styled(Grid)({
 const PieContainer = styled("div")({
   display: "flex",
   justifyContent: "space-evenly",
-  width: "100%",
-  height: "100%",
 });
 
 const PieLegendContainer = styled("div")({
@@ -46,6 +44,10 @@ interface FeaturedSummaryProps {
   }[];
 }
 
+const PIECHART_OUTER_RADIUS = 100;
+const PIECHART_INNER_RADIUS = 70;
+const PIECHART_ANIMATION_DURATION_MS = 200;
+
 const FeaturedSummary = ({ summary }: { summary: FeaturedSummaryProps }) => {
   const values = summary?.data?.map(d => d.value);
   const total = values.reduce((t, value) => t + value);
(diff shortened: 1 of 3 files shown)


frontend/workflows/envoy/src/remote-triage/dashboard.tsx
```diff
@@ -47,20 +47,22 @@ interface FeaturedSummaryProps {
 const FeaturedSummary = ({ summary }: { summary: FeaturedSummaryProps }) => {
   const values = summary?.data?.map(d => d.value);
   const total = values.reduce((t, value) => t + value);
+  // We transform the data from the summary format to a format used for Recharts Library
+  const pieChartData = summary?.data?.map(d => {
```
jdslaugh, line 51: Is there a reason we're casting this to the same thing? It looks like the summary format is identical so couldn't we just do `summary?.data || []` as it was previously?

frontend/workflows/envoy/src/remote-triage/dashboard.tsx
```diff
@@ -47,20 +47,22 @@ interface FeaturedSummaryProps {
 const FeaturedSummary = ({ summary }: { summary: FeaturedSummaryProps }) => {
   const values = summary?.data?.map(d => d.value);
   const total = values.reduce((t, value) => t + value);
```
jdslaugh, line 49: I know this isn't part of the PR, but I'd adjust this a bit. We're iterating over values to pull them out to then iterate over them for a total.

```suggestion
  const total = (summary?.data || []).reduce((t, { value = 0 }) => t + value, 0);
```

frontend/workflows/envoy/src/remote-triage/dashboard.tsx
```diff
@@ -44,23 +44,31 @@ interface FeaturedSummaryProps {
...
+            <Pie
+              data={summary?.data || []}
+              dataKey="value"
+              innerRadius={PIECHART_INNER_RADIUS}
+              outerRadius={PIECHART_OUTER_RADIUS}
+              legendType="none"
+              animationDuration={PIECHART_ANIMATION_DURATION_MS}
+            >
+              {summary?.data?.map(d => (
+                <Cell key={`cell-${d.id}-${d.value}-${d.color}`} fill={d.color} />
```
danielhochman, line 68: i wouldn't include the value with the key, or probably the color. just the id.

danielhochman (approved):

jdslaugh (approved):
Thanks!

PR #2079, opened 2022-03-03
frontend: ProjectCatalog Workflow Scaffolding

jdslaugh:
### Description
- scaffold from running `make scaffold-workflow`, gated behind featureFlag

### Testing Performed
Manual verification that the workflow does not register without featureFlag present

### DIFF
--- a/frontend/workflows/projectCatalog/src/index.tsx
+++ b/frontend/workflows/projectCatalog/src/index.tsx
@@ -1,0 +1,25 @@
+import type { BaseWorkflowProps, WorkflowConfiguration } from "@clutch-sh/core";
+
+export interface WorkflowProps extends BaseWorkflowProps {}
+
+const register = (): WorkflowConfiguration => {
+  return {
+    developer: {
+      name: "Lyft",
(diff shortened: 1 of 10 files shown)


frontend/workflows/projectCatalog/src/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+      contactUrl: "mailto:hello@clutch.sh",
+    },
+    path: "catalog",
+    group: "Catalog",
+    displayName: "Project Catalog",
+    routes: {
+      catalog: {
+        path: "/",
+        displayName: "Project Catalog",
+        description: "Project Catalog",
```
danielhochman, line 18: Can you go ahead and update this description to something a little more verbose?

danielhochman (approved):

frontend/workflows/projectCatalog/src/index.tsx
jdslaugh: Updated to "A searchable catalog of services"

mikecutalo (approved):

PR #2086, opened 2022-03-04
shortlink: fix test json assert

mikecutalo:
### Description
Fixing weak assertion for json string compare, https://github.com/golang/protobuf/issues/1082 as this is discouraged. Switching to `assert.JSONEq()` which unmarshals into interfaces and then compares.


danielhochman (approved):
nice!

smonero (approved):

PR #2090, opened 2022-03-07
FE: Packages: Core: Time series wrapper for Recharts

smonero:
### Description

This is the first wrapper for timeseries charts. The next one will be a wrapper for scatter plots that will be used for the use-case of linear timeline views. This wrapper is specifically useful when the user wants to have a Y-Axis.

Note there are many options, most of which have defaults. The goal is to make the component as easy as possible to use. 
There are 3 examples showing different configurations in the storybook stories. Subsequent PRs will add support for zooming, automatic color choice support, and more styling options.
Screen Shot 2022-03-07 at 1 21 11 PM

### Testing Performed

local
storybook
### GitHub Issue

Fixes #


jdslaugh (commented):
Initial review, need to take another pass with the new changes

frontend/packages/core/src/Charts/timeseries.tsx
```diff
@@ -0,0 +1,172 @@
...
+export interface TimeseriesReferenceLineProps {
+  axis: ReferenceLineAxis;
+  coordinate: number;
+  color: string;
+  label?: string;
+}
+
+export interface LineProps {
+  dataKey: string;
+  color: string;
```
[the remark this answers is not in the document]
smonero: recharts defaults the colors to black. dataKey is required

frontend/packages/core/src/Charts/timeseries.tsx
```diff
@@ -0,0 +1,172 @@
...
+                <ReferenceLine
+                  key={index.toString() + refLine.axis + refLine.coordinate.toString()}
+                  y={refLine.coordinate}
+                  label={refLine.label}
+                  stroke={refLine.color}
+                  strokeDasharray="4 4"
+                />
+              );
+            })
+          : null}
```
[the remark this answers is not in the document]
smonero: I modified the suggestion slight because ` props.strokeDashArray ` gives an error about {} not having field `strokeDashArray`

jdslaugh (approved):
Thanks for putting this together!

frontend/packages/core/src/stories/timeseries.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+      />
+    </ChartContainer>
+  );
+};
+
+export const ReferenceLinesAndThingsDisabled = () => {
+  const mockData = [
+    {
+      lineName: "line1",
+      timestamp: 1546301000000,
```
danielhochman, line 139: i noticed that these millis since epoch just show up as raw numbers in the story. intentional?

frontend/packages/core/src/Charts/timeseries.tsx
```diff
@@ -0,0 +1,163 @@
...
+const TimeseriesChart = ({
+  data,
+  xAxisDataKey = "timestamp",
+  yAxisDataKey = "value",
+  lines,
+  refLines,
+  singleLineMode = true,
+  drawDots = true,
+  enableLegend = true,
+  enableGrid = true,
```
danielhochman, line 77: nit i would rename these 
```suggestion
  legend = true,
  grid = true,
```

frontend/packages/core/src/Charts/timeseries.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  return (
+    <ResponsiveContainer width="100%" height="100%">
+      <LineChart data={data}>
+        {enableGrid ? <CartesianGrid /> : null}
+        <XAxis
+          dataKey={xAxisDataKey}
+          type="number"
+          domain={[xAxisDomainMin, xAxisDomainMax]}
+          tickFormatter={tickFormatterFunc || null}
```
danielhochman, line 115: not sure how this could be null since it's a required prop. is this superfluous? 
```suggestion
          tickFormatter={tickFormatterFunc}
```

frontend/packages/core/src/Charts/timeseries.tsx
```diff
@@ -0,0 +1,163 @@
...
+            <YAxis type="number" domain={[yAxisDomainMin, yAxisDomainMax]} />
+          )
+        }
+        <Tooltip labelFormatter={tickFormatterFunc || null} />
+        {enableLegend ? <Legend /> : null}
+        {lines
+          ? lines.map((line, index) => {
+              return (
+                <Line
+                  key={index.toString() + line.dataKey + line.color}
```
danielhochman, line 133: i think we talked about this in another PR, can't remember if we came to a conclusion there. this key seems overly specific, especially to include the color.

danielhochman (approved):
nice! a few thought and comments. feel free to merge as you see fit from here.

frontend/packages/core/src/stories/timeseries.stories.tsx
smonero: Yes that is intentional, the purpose of that story is to show people that they don't have to be forced to use a specific time format, and can allow for the raw values as well.

smonero: added a comment to the story to help explain

PR #2096, opened 2022-03-09
feature: ShortLinks - StorageContext

jdslaugh:
### Description
- PR [1/x] for ShortLinks

This PR adds a WorkflowStorageContext wrapped around the workflow which provides helper functions for accessing the Hydrated Storage, Temporary Storage and Local Storage.

As part of ShortLinks we will have components write the state that needs to be saved into the Temporary Storage which the ShortLink can then utilize to write to the API. These functions will also directly write into LocalStorage so we can reduce the need to call two functions.

When a ShortLink has been hydrated into the application it will be kept in the 'Store'. This is the only time this portion of the state is used. During the retrieval phase the reducer will first check if we have any data here and if not will try to fetch data from local storage matching the same key. If it doesn't exist it has the option of returning defaultData if desired. This allows for the components to almost directly replace their 'fetch from local storage' functions with these and makes it easier to allow short links to be extendable for most use cases.

### NOTES

For this PR nothing is actually utilizing the context at this point, those changes will come in follow up PR's.

The only addition is the Alert that will be generated above a given workflow if it is on a ShortLink. Screenshot below:

Screen Shot 2022-03-31 at 7 43 19 AM

#### Follow Up

The original PR was a StorageContext wrapped around the entire application. This has been modified to be only wrapped around an individual workflow and should be unmounted as workflows change/update.

### Testing Performed
manual


frontend/packages/core/src/Contexts/storage-context/helpers.tsx
```diff
@@ -0,0 +1,58 @@
+import type { clutch as IClutch } from "@clutch-sh/api";
+
+import type { HydratedData, StorageState } from "./types";
+
+// API data comes in as an array, this will rotate it into an object usable by the StorageContext
+const rotateDataFromAPI = (data: IClutch.shortlink.v1.IShareableState[]): HydratedData => {
+  const hydrated: HydratedData = {};
+
+  data.forEach(({ key = "", state = {} }) => {
```
danielhochman, line 9: why the default value for key?

frontend/packages/core/src/Contexts/storage-context/reducer.tsx
```diff
@@ -0,0 +1,75 @@
+import { removeLocalData, rotateDataFromAPI, storeLocalData } from "./helpers";
+import type { Action, StorageState, UserPayload } from "./types";
+
+/**
+ * Reducer for the StorageContext
+ * This will act on the StorageState and add / remove items from the temporary storage
+ * as well as localStorage, this will keep all storage actions in one location for all
+ * components and allow for easier state hydration across the entire application
+ */
```
danielhochman, line 9: I'm not sure a global storage context is the right direction for the frontend. We use data-layouts per workflow (instead of something like Redux) to keep data isolated and simplify the understanding of data as it moves through the application. Sharing storage between components has the potential to become untenable if used improperly. Data is also destroyed when a workflow is unmounted (unless it's kept in local storage of course), which I think this might change in some cases.

I can see the value in a storage helper that comes from a hook that returns a scoped instance of storage so we don't have to repeat the same complex code everywhere to manage storage.

I can also see the value in a global storage context. But I think we should discuss a fair bit more before introducing it to  the frontend, and it's not necessary for the shortlink feature as it is scoped today (as far as I'm aware).

Definitely curious to get more of your thoughts on this!

frontend/packages/core/src/Contexts/storage-context/helpers.tsx
jdslaugh: Think I was just defaulting it all, I'll adjust

frontend/packages/core/src/Contexts/storage-context/reducer.tsx
jdslaugh: Lets sync in person, will set up some time to go over this.

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,67 @@
...
+      dispatch({ type: "REMOVE_DATA", payload: { componentName, key, localStorage } }),
+    retrieveData: (componentName: string, key: string, defaultData?: any) =>
+      retrieveData(state.store, componentName, key, defaultData),
+  };
+
+  return (
+    <WorkflowStorageContext.Provider value={storageProviderProps}>
+      {state.shortLinked && (
+        <StyledAlertContainer container direction="column" alignItems="center">
+          <Alert title="Short Link">Local Workflow Data will not be saved until reload</Alert>
```
jdslaugh, line 59: This comes out looking weird, think it might be worth just allowing the workflows to implement this so it can be rendered where it makes the most sense. Probably can go ahead and do the wizard though.

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }, [hydrateData]);
+
+  const storageProviderProps: WorkflowStorageContextProps = {
+    shortLinked: state.shortLinked,
+    storeData: (componentName: string, key: string, data: unknown, localStorage?: boolean) =>
+      dispatch({ type: "STORE_DATA", payload: { componentName, key, data, localStorage } }),
+    removeData: (componentName: string, key: string, localStorage?: boolean) =>
+      dispatch({ type: "REMOVE_DATA", payload: { componentName, key, localStorage } }),
+    retrieveData: (componentName: string, key: string, defaultData?: unknown) =>
+      retrieveData(state.store, componentName, key, defaultData),
```
dschaller, line 54: we will definitely want to have docstrings for these since they can be used by the workflows. Especially the retrieve data function since the fallback behavior is complex

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  return (
+    <WorkflowStorageContext.Provider value={storageProviderProps}>
+      {state.shortLinked && (
+        <Grid container direction="column" alignItems="flex-end">
+          <StyledAlert severity="warning">
+            Local Workflow Data will not be saved until reload
```
dschaller, line 62: nit
```suggestion
            Loaded shared state. Any local changes will not be preserved.
```

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,71 @@
...
+          </StyledAlert>
+        </Grid>
+      )}
+      {children}
+    </WorkflowStorageContext.Provider>
+  );
+};
+
+export default WorkflowHydrator;
```
dschaller, line 71: nit maybe call this something like `ShortlinkStateHydrator`

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { retrieveData } from "../Contexts/workflow-storage-context/helpers";
+import workflowStorageContextReducer from "../Contexts/workflow-storage-context/reducer";
+import type { WorkflowStorageContextProps } from "../Contexts/workflow-storage-context/types";
+import { defaultWorkflowStorageState } from "../Contexts/workflow-storage-context/types";
+import { Alert } from "../Feedback";
+import styled from "../styled";
+
+interface WorkflowHydratorProps {
+  children: React.ReactElement;
+  hydrateData: IClutch.shortlink.v1.IShareableState[];
```
dschaller, line 15: nit it might make more sense to call this something related to the type (e.g. `sharedState`)

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,71 @@
...
+  onClear: () => void;
```
dschaller, line 16: just from looking at the props it's unclear when or why this is called

frontend/packages/core/src/Contexts/workflow-storage-context/helpers.tsx
```diff
@@ -0,0 +1,56 @@
+import type { clutch as IClutch } from "@clutch-sh/api";
+
+import type { HydratedData } from "./types";
+
+// API data comes in as an array, this will rotate it into an object usable by the StorageContext
+const rotateDataFromAPI = (data: IClutch.shortlink.v1.IShareableState[]): HydratedData => {
```
dschaller, line 6: nit
```suggestion
const transformAPISharedState = (data: IClutch.shortlink.v1.IShareableState[]): HydratedData => {
```

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
```diff
@@ -0,0 +1,74 @@
...
+import type { Action, BackgroundPayload, UserPayload, WorkflowStorageState } from "./types";
+
+/**
+ * Reducer for the WorkflowStorageContext
+ * This will act on the WorkflowStorageState and add / remove items from the temporary storage
+ * as well as localStorage, this will (optionally) keep all storage actions in one location for all
+ * workflows and allow for easier state hydration
+ */
+const workflowStorageContextReducer = (
+  state: WorkflowStorageState,
```
dschaller, line 11: I would rename this since it's a bit general right now to all storage pertaining to workflows

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
```diff
@@ -0,0 +1,74 @@
...
+  action: Action
+): WorkflowStorageState => {
+  switch (action.type) {
+    // Will add data to our temporary storage as well as the local storage
```
dschaller, line 15: We might have already discussed this but just to refresh myself, is there a reason to add this to local storage and temporary?

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
```diff
@@ -0,0 +1,74 @@
...
+    case "STORE_DATA": {
+      const { componentName, key, data, localStorage = true } = action.payload as UserPayload;
```
dschaller, line 17: UserPayload is a bit confusing here, isn't this the component payload?

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  React.useEffect(() => {
+    if (hydrateData && hydrateData.length) {
+      dispatch({ type: "HYDRATE", payload: { data: hydrateData } });
+      onClear();
+    }
+  }, [hydrateData]);
+
+  const storageProviderProps: WorkflowStorageContextProps = {
+    shortLinked: state.shortLinked,
```
dschaller, line 48: nit: I think this is to denote the entry was from a short link, if so how about
```suggestion
    fromShortLink: state.fromShortLink,
```

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
```diff
@@ -0,0 +1,74 @@
...
+      const newState = { ...state };
+      const { shortLinked, tempStore } = newState;
+
+      if (!tempStore[componentName]) {
```
dschaller, line 21: We don't do a check on length of component name anywhere

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+      if (localStorage && !shortLinked) {
+        removeLocalData(key ?? componentName);
+      }
+
+      return newState;
+    }
+    // Will take a given input of data from an API and add it to the state as 'store', the only time this portion of the state should ever be modified
+    case "HYDRATE": {
+      const { data } = action.payload as BackgroundPayload;
```
dschaller, line 57: what's background payload?

jdslaugh: I stole the name from the ProjectSelector reducer. This data is what would come from the short link api when hydrating initially

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
jdslaugh: Yeah its meant to denote that the data has been loaded via a shortlink, in case anything needs to be modified for a component directly or any different actions need to be taken (ie showing the alert banner). I can rename it though

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
jdslaugh: Yeah that's a better name, let me update

frontend/packages/core/src/Contexts/workflow-storage-context/helpers.tsx
jdslaugh: Thats a better name, will update

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
jdslaugh: I like it

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
jdslaugh: Will do

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
jdslaugh: We synced on this but to make it transparent:

Example:

"ProjectSelector will save its state, which will write to temp and then it will forward it to local storage (since it needs local storage for its own use cases). So the temp storage could be used if you decide to create a short link at that moment in time and then you could make further changes and reload and it would pull from your local storage as normal."

"If you just want to be able to hydrate the state for a short link but not save anything in local storage you would just use the store the exact same but not pass the local storage option."

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
jdslaugh: Will do

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
jdslaugh: I'll modify the name though, maybe it should be HydratePayload since it's really the only time its used.

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
jdslaugh: Good call, I held off on adding too many docs if I was going to have to refactor a lot. Will add them in

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
jdslaugh: Modified the payload and storage names to be a bit clearer

frontend/packages/core/src/AppProvider/workflow-hydrator.tsx
jdslaugh: This is called so that we can clear the data temporarily stored in AppProvider after initialization. I've added docstrings around this as well as renamed it.

frontend/packages/core/src/Contexts/workflow-storage-context/reducer.tsx
jdslaugh: I thought this was pertaining to the original type name which I had modified to be `WorkflowStorageState`. Is the thought to rename 'state' to something else or the reducer?

dschaller (approved):

danielhochman (approved):

frontend/packages/core/src/Contexts/workflow-storage-context/helpers.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+};
+
+const retrieveData = (
+  store: HydratedData,
+  componentName: string,
+  key: string,
+  defaultData?: unknown
+): unknown => {
```
dschaller, line 44: Thought about this more after reviewing #2101 and realized we should probably be using generics here to avoid having to typecast.

jdslaugh: I've updated to use generics, tested it against the ProjectSelector and it seemed to work. If you dont mind giving it another pass and seeing if there's anything you'd like me to change about it.

dschaller (approved):

PR #2099, opened 2022-03-10
Scatter plot wrapper for linear timeline views

smonero:
### Description

We wrap the `ScatterPlot` Recharts component for use in linear timeline views. This is more useful than the wrapper for linecharts for this specific use case.

Screen Shot 2022-03-09 at 8 39 46 PM

### Testing Performed

story,
local

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Charts/linearTimeline.tsx
+++ b/frontend/packages/core/src/Charts/linearTimeline.tsx
@@ -14,11 +14,11 @@ import { calculateDomainEdges, calculateTicks, localTimeFormatter } from "./help
 
 export interface LinearTimelineDataPoint {
   timestamp: number;
-  // more to come
 }
 
 // Note that shape can be a set of shapes denoted by preset strings ("circle", "square", etc.)
 // or a custom SVG element
+// See https://recharts.org/en-US/api/Scatter#shape for more details
 export interface LinearTimelineDataPoints {
   points: LinearTimelineDataPoint[];
   shape?: any;
@@ -27,7 +27,11 @@ export interface LinearTimelineDataPoints {
 
 export interface LinearTimelineData {
   [lane: string]: LinearTimelineDataPoints;
-  // ...
+}
+
+interface CustomTooltipProps {
+  active: boolean;
+  payload: any; // A huge object that contains all the info for the data point and more
 }
 
 /**
@@ -92,7 +96,7 @@ const LinearTimeline = ({
   };
 
   // TODO: Allow for proper styling and make things less hacky than "payload[0]"
-  const formatTooltip = ({ active, payload, label }) => {
+  const formatTooltip = ({ active, payload }: CustomTooltipProps) => {
     if (active) {
       return <div>{localTimeFormatter(payload[0].value)}</div>;
     }
--- a/frontend/packages/core/src/Charts/linearTimeline.tsx
+++ b/frontend/packages/core/src/Charts/linearTimeline.tsx
@@ -2,64 +2,62 @@ import React from "react";
 import {
   CartesianGrid,
   Legend,
-  ScatterChart,
-  Scatter,
   ResponsiveContainer,
+  Scatter,
+  ScatterChart,
   Tooltip,
   XAxis,
   YAxis,
 } from "recharts";
-import { calculateTicks, calculateDomainEdges, localTimeFormatter } from "./helpers";
+
+import { calculateDomainEdges, calculateTicks, localTimeFormatter } from "./helpers";
 
 export interface LinearTimelineDataPoint {
-    timestamp: number;
-    // more to come
-  }
+  timestamp: number;
+  // more to come
+}
 
-  // Note that shape can be a set of shapes denoted by preset strings ("circle", "square", etc.)
-  // or a custom SVG element
-  export interface LinearTimelineDataPoints {
-    points: LinearTimelineDataPoint[];
-    shape?: any;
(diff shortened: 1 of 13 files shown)


frontend/packages/core/src/Charts/linearTimeline.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export interface LinearTimelineProps {
+  data: LinearTimelineData;
+  xAxisDataKey: string;
+  regularIntervalTicks?: boolean;
+  tickFormatterFunc?: (timestamp: number) => string;
+  legend?: boolean;
+}
+
+const LinearTimeline = ({
```
danielhochman, line 41: let's add some more documentation via comments.

frontend/packages/core/src/Charts/linearTimeline.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export interface LinearTimelineData {
+  [lane: string]: LinearTimelineDataPoints;
+  // ...
+}
+
+export interface LinearTimelineProps {
+  data: LinearTimelineData;
+  xAxisDataKey: string;
+  regularIntervalTicks?: boolean;
```
danielhochman, line 36: document this prop, not obvious at first glance what the behavior difference is for true v false.

frontend/packages/core/src/stories/linearTimeline.stories.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import type { Meta } from "@storybook/react";
+
+import type { LinearTimelineData } from "../Charts/linearTimeline";
+import LinearTimeline from "../Charts/linearTimeline";
+
+export default {
+  title: "Core/LinearTimeline",
```
danielhochman, line 9: can we nest all of the chart components under Core/Chart or similar?

frontend/packages/core/src/stories/linearTimeline.stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  height: 200,
+});
+
+export const Primary = () => {
+  const mockData: LinearTimelineData = {
+    deploys: { points: [{ timestamp: 1588884888 }], shape: "cross", color: "purple" },
+    "k8s events": { points: [{ timestamp: 1588985888 }], shape: "triangle", color: "blue" },
+    explosions: {
+      points: [
+        { timestamp: 1588788888 },
```
danielhochman, line 24: is it possible to annotate individual events with tooltips? i looked at the recharts docs briefly for `Scatter` but it wasn't clear to me what the possibilities were for the `data` prop. also was reading https://stackoverflow.com/a/66084216

danielhochman (commented):
looks really nice overall! good interface. a couple of questions and comments.

frontend/packages/core/src/stories/linearTimeline.stories.tsx
smonero: yes it is. The payload object contains all the data for the point so with custom tooltip renders we can put whatever we want. The thing that seemed difficult was that there was no default styling, so if you don't pass any styling info the text is just black with no background or box. We can fix that by passing styling info.

frontend/packages/core/src/Charts/linearTimeline.tsx
```diff
@@ -0,0 +1,139 @@
...
+      color: data[key].color,
+      laneID: index,
+    };
+  });
+  const formatIdsToNames = (value: string) => {
+    return <span>{mappingOfLaneIdsToNames[value]}</span>;
+  };
+
+  // TODO: Allow for proper styling and make things less hacky than "payload[0]"
+  const formatTooltip = ({ active, payload, label }) => {
```
jdslaugh, line 95: Do we have typings for this function?

frontend/packages/core/src/Charts/linearTimeline.tsx
```diff
@@ -0,0 +1,124 @@
...
+} from "recharts";
+
+import { calculateDomainEdges, calculateTicks, localTimeFormatter } from "./helpers";
+
+export interface LinearTimelineDataPoint {
+  timestamp: number;
+  // more to come
+}
+
+// Note that shape can be a set of shapes denoted by preset strings ("circle", "square", etc.)
```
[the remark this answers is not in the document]
smonero: they do but they don't export the type so I guess we can't import it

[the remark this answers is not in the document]
smonero: the type is specifically `SymbolType | ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>);`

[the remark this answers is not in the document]
smonero: I'll add a comment that talks about this

danielhochman (approved):
lgtm!

frontend/packages/core/src/Charts/linearTimeline.tsx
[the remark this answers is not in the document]
jdslaugh: Alright, well we can probably copy it at least. Especially the `ReactElement<SVGElement> | ((props: any) => ReactElement<SVGElement>);` portion, that way it's somewhat known what shape should be

jdslaugh (approved):
Thanks!

PR #2105, opened 2022-03-10
feature: ShortLinks - Collapsible Alert

jdslaugh:
### Description
- PR [5/x] for ShortLinks
- No reliance on other PRs
- Modifies the Alert to have a collapsible and hover option
- Also supports a click option that can be used in place of hover

#### Closed
Screen Shot 2022-03-10 at 1 29 25 PM

#### Open
Screen Shot 2022-03-10 at 1 29 37 PM

#### GIF
alert

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/Feedback/alert.tsx
+++ b/frontend/packages/core/src/Feedback/alert.tsx
@@ -10,37 +10,46 @@ import { Alert as MuiAlert, AlertTitle as MuiAlertTitle } from "@material-ui/lab
 import styled from "../styled";
 
 const backgroundColors = {
-  error: "linear-gradient(to right, #DB3615 8px, #FDE9E7 0%)",
-  info: "linear-gradient(to right, #3548D4 8px, #EBEDFB 0%)",
-  success: "linear-gradient(to right, #1E942E 8px, #E6F7EB 0%)",
-  warning: "linear-gradient(to right, #FFCC80 8px, #FFFDE6 0%)",
+  error: "#FDE9E7",
+  info: "#EBEDFB",
+  success: "#E6F7EB",
+  warning: "#FFFDE6",
 };
 
-const StyledAlert = styled(MuiAlert)<{ severity: MuiAlertProps["severity"] }>(
+const backgroundGradients = {
+  error: `linear-gradient(to right, #DB3615 8px, ${backgroundColors.error} 0%)`,
+  info: `linear-gradient(to right, #3548D4 8px, ${backgroundColors.info} 0%)`,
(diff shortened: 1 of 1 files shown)


smonero (approved):

danielhochman:
I was envisioning we put this within the project selector itself. That way it would be clear that it's talking specifically about project selector modifications. WDYT? Should we run this by Kanchan?

frontend/packages/core/src/Feedback/alert.tsx
```diff
@@ -76,14 +85,37 @@ const iconMappings = {
...
+  hover?: boolean;
 }
 
-export const Alert: React.FC<AlertProps> = ({ severity = "info", title, children, ...props }) => (
-  <StyledAlert severity={severity} iconMapping={iconMappings} {...props}>
-    {title && <AlertTitle>{title}</AlertTitle>}
-    {children}
-  </StyledAlert>
-);
+export const Alert: React.FC<AlertProps> = ({
```
danielhochman, line 93: i would make a separate collapsible alert component.

because collapsible could mean different things, we may want an alert later that collapses vertically (i.e. displays title with collapsible long text).

also because  the `hover` property does nothing if `collapsible` is false.

lyft-tla:
Messages in header may be interpreted as for all of Clutch, rather than a specific part of Clutch. If we have message that is specific to particular component/page, I would agree that we should either state that or have that message closer to the area that it applies to specifically.

jdslaugh:
> Messages in header may be interpreted as for all of Clutch, rather than a specific part of Clutch. If we have message that is specific to particular component/page, I would agree that we should either state that or have that message closer to the area that it applies to specifically.

That's fine. This PR doesn't affect the placement and was just used as an example.

frontend/packages/core/src/Feedback/alert.tsx
jdslaugh: Yeah that was my initial plan before I went this way, let me make it separate and clean it up

jdslaugh:
> I was envisioning we put this within the project selector itself. That way it would be clear that it's talking specifically about project selector modifications. WDYT? Should we run this by Kanchan?

I wouldn't put it in the project selector since it affects more than one component but we could move it into the page itself, the original goal was just to not have to replicate this as we use the feature across more workflows but it could live inside of the workflow directly instead of the header, that might work. Lets discuss it during our sync next week and then we can bring in Kanchan if we feel its necessary.

jdslaugh:
Converting to draft while changes are addressed

PR #2107, opened 2022-03-10
feature: ShortLinks

jdslaugh:
### Description
- PR [7/7] for ShortLinks
- Relies on all previous ShortLink PRs

This is the final PR for ShortLinks and adds the actual implementation and fetching details. For workflows that have provided the valid ShortLink attribute a link icon will be displayed in the header. Upon clicking on this link it will open with an option to generate a short link. Once clicked it will do:

- Take our temporary storage that any components have been writing into and send it down to the server
- The server will return a hash for the stored data
- The shortLinker component will then display a readable input field the user can copy the short link from
- Upon loading into a shortlink a ShortLinkHydrator component will be hit, this has no display as it is just an intermediate step
- The hydrator will call down with the given route hash and request the available short link state information
- This will be passed into our StorageContext for hydration information and then redirected to the given path
- Once redirected, an Alert will be displayed at the top of the page letting the user know that they are on a ShortLink and their local changes will not be saved without reloading, this is to not break their current state

### GIF
- available in team channel

### Testing Performed
manual


frontend/packages/core/src/AppProvider/workflow.tsx
```diff
@@ -24,10 +24,12 @@ interface BaseWorkflowConfiguration {
 }
 
 export interface Workflow extends BaseWorkflowConfiguration {
+  shortLink?: boolean;
```
dschaller, line 27: can you add docstrings to these?

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const [validWorkflow, setValidWorkflow] = React.useState<boolean>(false);
+  const [error, setError] = React.useState<ClutchError | null>(null);
+  const anchorRef = React.useRef(null);
+  const location = useLocation();
+
+  const checkValidWorkflow = () => {
+    const workflow = workflowByRoute(workflows, location.pathname);
+
+    setValidWorkflow(workflow?.shortLink ?? false);
+  };
```
dschaller, line 86: does valid here mean that is supports shortlinks? If so, can we denote that in someway?

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+    const requestData: IClutch.shortlink.v1.ICreateRequest = {
+      path: `${location.pathname}${location.search}`,
+      state: rotateStore(retrieveWorkflowSession()),
+    };
+
+    client
+      .post("/v1/shortlink/create", requestData)
+      .then(response => {
+        const { hash } = response.data as IClutch.shortlink.v1.ICreateResponse;
+        setShortLink(`${window.location.origin}/sl/${hash}`);
```
dschaller, line 130: Should we be defining the path of `/sl/{hash}` somewhere in a constant to share?

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+      })
+      .catch((err: ClutchError) => {
+        console.warn("failed to generate short link", err); // eslint-disable-line
```
dschaller, line 133: Do we need this console warn if we are showing an error to the users?

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  if (!validWorkflow) {
+    return null;
+  }
+
+  return (
+    <>
+      {error && (
+        <Toast severity="error" onClose={() => setError(null)}>
+          Unable to generate short link
```
dschaller, line 146: Should we show additional information from the error?

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  // Used for initial load to verify that once our workflows have loaded we are on a valid workflow
+  React.useEffect(() => {
+    if (workflows.length) {
+      checkValidWorkflow();
+    }
+  }, [workflows]);
+
+  const handleToggle = () => {
+    setOpen(!open);
```
dschaller, line 105: you'll want something like
```tsx
setOpen(o => !o);
```

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+    setShortLink(null);
```
dschaller, line 106: I'm confused by this since the docstring says that if there is an existing shortlink there shouldn't be an option to generate a new one. However, if the user opens the popper, generates a shortlink, and then closes it the shortlink will be set to null and a new one will need to be generated.

frontend/packages/core/src/AppProvider/short-link.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { useParams } from "react-router-dom";
+import type { clutch as IClutch } from "@clutch-sh/api";
+
+import { useNavigate } from "../navigation";
+import { client } from "../Network";
+import type { ClutchError } from "../Network/errors";
+
+interface ShortLinkProps {
+  hydrate: (data: IClutch.shortlink.v1.IShareableState[]) => void;
+  onError: (error: string) => void;
```
dschaller, line 11: Should we be using `ClutchError` here instead?

frontend/packages/core/src/AppProvider/short-link.tsx
```diff
@@ -0,0 +1,67 @@
...
+    .post("/v1/shortlink/get", requestData)
+    .then(response => {
+      const { path, state } = response.data as IClutch.shortlink.v1.IGetResponse;
+
+      hydrate(state);
+      return path;
+    })
+    .catch((error: ClutchError) => {
+      // eslint-disable-next-line no-console
+      console.error(error?.status?.text);
```
dschaller, line 28: if we pass back the error object we can probably remove this

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: This was because the toast doesn't contain a lot of information (below), so it was more for debugging purposes.

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: We can, wasn't sure if there was anything else we could really say in a toast

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Probably a good idea, I'll whip something up

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Yup, just that it has the defined property, I'll add some comments

frontend/packages/core/src/AppProvider/workflow.tsx
jdslaugh: Yup!

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Can do

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,193 @@
...
+      >
+        <LinkIcon />
+      </StyledLinkIcon>
+      <Popper open={open} anchorEl={anchorRef.current} transition placement="bottom-end">
+        {({ TransitionProps, placement }) => (
+          <Grow
+            {...TransitionProps}
+            placement={placement === "bottom" ? "center top" : "center bottom"}
+          >
+            <>
```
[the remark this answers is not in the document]
jdslaugh: They're present because it doesn't have a single parent element, so it'll fail. All of the header popper ones have fragments actually

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Which docstring are you referring to? The clearing was meant so that a user could build off of a shortlink. Generate 1, modify modify, generate 2, modify modify, etc etc. The docstring about `only time it is ever displayed` is more in line with what this does in that after you create it it will display it so you can copy and then it's gone (similar to how generating a private key works in most services).

Unless Im misunderstanding.

frontend/packages/core/src/AppProvider/short-link.tsx
jdslaugh: Can do

frontend/packages/core/src/AppProvider/short-link.tsx
jdslaugh: Originally it was because I was setting a string for the toast but I can return the error and let the parent handle it

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: So looking at the other poppers we use, it looks like they all just do `setOpen(!open)` as well (which is where I took it from I think)

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Added comments explaining what the function is/does. Let me know if thats enough :)

frontend/packages/core/src/AppProvider/short-link.tsx
```diff
@@ -0,0 +1,79 @@
+import React from "react";
```
dschaller, line 1: nit on the file name (should be camel case)

frontend/packages/core/src/AppLayout/shortLinker.tsx
dschaller: ah ok I see the confusion. The docstring implies that this link is going to be persisted.
This probably opens up another question which is, is there a reason to unset the shortlink if nothing in the state has changed? We don't really track that at the moment but if the state doesn't change the link should remain the same. Not something we have to worry about in this PR but it is something we should probably do as a follow up

frontend/packages/core/src/AppLayout/shortLinker.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  if (!validWorkflow) {
+    return null;
+  }
+
+  return (
+    <>
+      {error && (
+        <Toast severity="error" onClose={() => setError(null)}>
+          Unable to generate short link, please try again later
```
dschaller, line 152: Should we use the Error component here to get more robust error information

frontend/packages/core/src/AppProvider/short-link.tsx
```diff
@@ -0,0 +1,79 @@
...
+import type { clutch as IClutch } from "@clutch-sh/api";
+
+import { useNavigate } from "../navigation";
+import { client } from "../Network";
+import type { ClutchError } from "../Network/errors";
+
+/**
+ * The base for a short link route
+ */
+export const ShortLinkBaseRoute = "sl";
```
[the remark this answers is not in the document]
jdslaugh: Yeah goto would be fine. We could build it off an environment variable if we wanted, default to `goto` but if something else is available it could use that

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Will update, I'll leave the others alone for now. I eventually want to combine all these header components into a single parent component and reduce some of the duplication.

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -90,56 +100,80 @@ const ClutchApp: React.FC<ClutchAppProps> = ({
...
+                        })}
+                        <Route key={`${workflow.path}/notFound`} path="*" element={<NotFound />} />
+                      </Route>
+                    );
+                  })}
+                  <Route
+                    key="short-links"
+                    path={`/${ShortLinkBaseRoute}/:hash`}
+                    element={
+                      <ShortLink
```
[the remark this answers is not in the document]
jdslaugh: I like `ShortLinkProxy`, seems to specify it well

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Sure, I'll give it a shot

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Yeah that's probably a good idea, shouldn't be too hard to adjust either

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Modified to use the error component. Have to force it below the header but it works pretty well. Modified the component to also allow an `onClose` parameter so a user can get rid of it if they want to.

Screen Shot 2022-04-27 at 1 38 25 PM

frontend/packages/core/src/AppProvider/short-link.tsx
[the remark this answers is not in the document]
jdslaugh: Adjusted to use `goto`, we can expand this for an environment variable in the future though

frontend/packages/core/src/AppLayout/shortLinker.tsx
jdslaugh: Modified it to using a toast again and just deliberately putting the specific messaging in. Reverted the changes to the Error component.

Screen Shot 2022-04-27 at 5 59 51 PM

dschaller (approved):

PR #2108, opened 2022-03-11
FE: Text-field: prevent null onChange from breaking things

smonero:
### Description

I noticed that if onChange is null or undefined that can cause an error here because there is an assumption that it is not.

### Testing Performed

local

### GitHub Issue

Fixes #


jdslaugh (approved):
Thanks!

mikecutalo (approved):

PR #2110, opened 2022-03-13
[feedback] FE: add support for inputTransformers in term instance

smonero:
### Description

It is useful to be able to specify transformation functions in the config to be able to transform the input for aws instance termination.

### Testing Performed

local

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/ec2/src/terminate-instance.tsx
+++ b/frontend/workflows/ec2/src/terminate-instance.tsx
@@ -18,7 +18,7 @@ import { Wizard, WizardStep } from "@clutch-sh/wizard";
 
 import type { ConfirmChild, ResolverChild, WorkflowProps } from ".";
 
-const InstanceIdentifier: React.FC<ResolverChild> = ({ resolverType }) => {
+const InstanceIdentifier: React.FC<ResolverChild> = ({ resolverType, inputTransformer }) => {
   const { onSubmit } = useWizardContext();
   const resolvedResourceData = useDataLayout("resourceData");
 
@@ -28,7 +28,7 @@ const InstanceIdentifier: React.FC<ResolverChild> = ({ resolverType }) => {
     onSubmit();
   };
 
-  return <Resolver type={resolverType} searchLimit={1} onResolve={onResolve} />;
+  return <Resolver type={resolverType} searchLimit={1} onResolve={onResolve} inputTransformer={inputTransformer} />;
 };
 
 const InstanceDetails: React.FC<WizardChild> = () => {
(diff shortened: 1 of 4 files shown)


frontend/workflows/ec2/src/terminate-instance.tsx
```diff
@@ -109,7 +121,11 @@ const TerminateInstance: React.FC<WorkflowProps> = ({ heading, resolverType, not
 
   return (
     <Wizard dataLayout={dataLayout} heading={heading}>
-      <InstanceIdentifier name="Lookup" resolverType={resolverType} />
+      <InstanceIdentifier
+        name="Lookup"
+        resolverType={resolverType}
+        inputTransformer={inputTransformer}
```
danielhochman, line 127: it would be nice if the workflow automatically plugged this into a context somehow so we wouldn't have to prop drill every workflow. is that possible?

danielhochman (commented):
lgtm, question about possibly using a context though. i'm not sure if there's a place to automatically inject the value without reading the workflow registration and instantiation code more closely.

frontend/workflows/ec2/src/terminate-instance.tsx
smonero: hm not sure, I'll think about this and maybe ask derek as well

smonero: Synced with Derek, came to a couple of conclusions:

There are 2 separate issues here - 1. the publishing of the config, and 2. the consumption of the config. We will focus on the latter.
- For prop drilling, we know the types with our well defined interface, whereas if we were pulling from a context we would need to store the type information somewhere else. For example, we know what `notes` is at all times in each workflow. We could move this interface to a higher level (i.e the context)
- Potential Issues of re-rendering - how do we position the context properly to prevent re-renders down the component tree? This would seem to point to a per-workflow rather than global context.
- For a context approach, we could infer values based off a workflow name. Is that an assumption we want to go with?

In general, it is certainly doable to have a global or per workflow context from which we pull objects using hooks. It is a question that becomes one of prioritization as well as design.

smonero: I tried searching around for this topic and I found "component composition" in this: https://blog.logrocket.com/solving-prop-drilling-react-apps/ but I'm not sure that's a good solution either.

danielhochman: > Potential Issues of re-rendering - how do we position the context properly to prevent re-renders down the component tree? This would seem to point to a per-workflow rather than global context.

Wouldn't the context be static and read-only after initialization of the application?

smonero: we synced, but when I tried Daniel's solution from the sync I realized that we can't read it in from the config anymore.
Doc explaining the process https://docs.google.com/document/d/1kmLHmhCls6GbMvju-3CI9ajhMQqMTj53H_1EQB4R8QA/edit?usp=sharing

smonero:
@dschaller

PR #2123, opened 2022-03-16
[feedback] FE: envoy triage: allow for query params

smonero:
### Description

It is useful to allow for query hydration in the envoy triage flow.

### Testing Performed

local

### GitHub Issue

Fixes #


frontend/workflows/envoy/src/remote-triage/index.tsx
```diff
@@ -122,6 +130,9 @@ const TriageDetails: React.FC<WizardChild> = () => {
 };
 
 const RemoteTriage: React.FC<WorkflowProps> = ({ heading }) => {
+  const urlParams = new URLSearchParams(window.location.search);
+  const hostParam = urlParams.get("q");
```
dschaller, line 134: We use q for the resolver right now so we shouldn't overwrite it. Since this is scoped to only the remote triage workflow we should prefix it with `_`

dschaller (approved):

PR #2128, opened 2022-03-17
project api: option to compute project deps

mikecutalo:
### Description
As we introduce the project catalog search page, we dont always need to compute all project dependencies. This introduces an option to disable that computation which will result in only returning the projects that you have requested, or / and that are owned by the list of users.

### Testing Performed
n/a


api/project/v1/project.proto
```diff
@@ -22,6 +22,9 @@ service ProjectAPI {
 message GetProjectsRequest {
   repeated string projects = 1;
   repeated string users = 2;
+  // If this is set to true, we will compute all project dependencies for the requested projects.
```
danielhochman, line 25: by compute do you mean return? i'm not clear what this means.

mikecutalo: Yeah i guess return is a bit more clear, let me change this.

api/project/v1/project.proto
```diff
@@ -22,9 +22,9 @@ service ProjectAPI {
 message GetProjectsRequest {
   repeated string projects = 1;
   repeated string users = 2;
-  // If this is set to true, we will compute all project dependencies for the requested projects.
+  // If this is set to true, the api will return all project dependencies for the requested projects.
```
danielhochman, line 25: and to clarify dependencies, it's immediate upstreams and immediate downstreams, right?

api/project/v1/project.proto
```diff
@@ -22,9 +22,9 @@ service ProjectAPI {
...
   // By default this is set to true.
```
danielhochman, line 26: it's not possible with the current api to determine whether it was set explicitly to false or not

api/project/v1/project.proto
mikecutalo: Yeah I can link to the project proto here to make it more clear, but the literal dependencies field on the proto which does include all up/downstream.

https://github.com/lyft/clutch/blob/2304adf0e0189734d8a36803964214ee3bc73fbc/api/core/project/v1/project.proto#L19

api/project/v1/project.proto
mikecutalo: Oh yeah true since we dont implement in OSS, let me remove this.

api/project/v1/project.proto
```diff
@@ -22,6 +22,9 @@ service ProjectAPI {
 message GetProjectsRequest {
   repeated string projects = 1;
   repeated string users = 2;
+  // If this is set to true, the api will return all project dependencies for the requested projects.
+  // https://github.com/lyft/clutch/blob/2304adf0e0189734d8a36803964214ee3bc73fbc/api/core/project/v1/project.proto#L19
+  bool return_dependencies = 3;
```
danielhochman, line 27: last nit, make it include_dependencies (or exclude_dependencies if you want opposite default), return sounds like exclusively returning.

danielhochman (approved):

jdslaugh (approved):

PR #2129, opened 2022-03-18
FE: charting: change dashed grid lines to solid ones

smonero:
### Description

The goal of the charting library is that defaults should lead to an appearance similar to Grafana. This is the 1st of many that will change the appearance.

Also added more time formatting funcs for users into the stories.
before:
Screen Shot 2022-03-19 at 11 36 43 PM
after:
Screen Shot 2022-03-19 at 11 35 39 PM

### Testing Performed

local

### GitHub Issue

Fixes #


danielhochman (approved):
screenshot would be good if it's not too much trouble

jdslaugh (approved):

PR #2133, opened 2022-03-21
FE: add support for styling of charts

smonero:
### Description

Add more props to allow for coloring of charts (could also be used for dark mode)
I am aware the random colors I chose are extremely unpleasant to look at. Please disregard the colors and focus on the fact that you can change the colors.

Screen Shot 2022-03-20 at 9 19 44 PM
Screen Shot 2022-03-20 at 9 16 57 PM

### GitHub Issue

Fixes #


jdslaugh (approved):
Thanks for adding these in.

danielhochman (approved):
nice!

PR #2142, opened 2022-03-24
ProjectCatalog: Details Page

jdslaugh:
### Description
- Implements basic functionality for Project Catalog Details
- Adds a Header to the Details Page which is built from information pulled from Info Card
- Creates an extendable Detail Card type which is used to align cards based on Metadata (Left) or Dynamic (Right)
- BaseCard contains logic to optionally fetch and refresh data for the user or it can be handled manually from the creating component
- Details page wraps all children in a context to make the base project info available for all children and their api requests

#### Screenshot
Screen Shot 2022-03-29 at 8 34 38 AM

### Testing Performed
Manual

### TODOs
- [ ] Write unit tests
- [ ] Documentation
- [x] Clean up the BaseCard logic


frontend/packages/core/src/chip.tsx
```diff
@@ -72,9 +72,9 @@ const CHIP_COLOR_MAP = {
 const StyledChip = styled(MuiChip)<{
   $filled: ChipProps["filled"];
   $variant: ChipProps["variant"];
+  size: "small" | "medium";
```
dschaller, line 75: should we read this type from the ChipProps?

smonero (approved):

frontend/packages/core/src/chip.tsx
```diff
@@ -86,6 +86,7 @@ const StyledChip = styled(MuiChip)<{
     },
   },
   props => ({
+    height: props.size === "small" ? "24px" : "32px",
```
dschaller, line 89: Is this reflected in the design system (figma)?

frontend/workflows/projectCatalog/src/details/info/types.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export interface ProjectRepository {
+  name: string;
+  repo: string;
+  url?: string;
+  icon?: IconProp;
+  requests?: ProjectRequests;
+}
+
+export interface ProjectInfo {
```
dschaller, line 32: Should this project info come from the ProjectAPI definitions?

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+interface CardBodyProps {
+  children?: React.ReactNode;
+  loading?: boolean;
+  error?: ClutchError;
+}
+
+interface BaseCardProps extends CardTitleProps, CardBodyProps {
+  reloadInterval?: number;
+  autoReload?: boolean;
```
dschaller, line 25: can you add some docstrings to `reloadInterval` and `autoReload`

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+const CardBody = ({ loading, error, children }: CardBodyProps) => (
+  <>
+    <StyledRow>
+      {loading && (
+        <StyledProgressContainer>
+          <LinearProgress color="secondary" />
+        </StyledProgressContainer>
+      )}
+    </StyledRow>
+    {error ? <Error subject={error} /> : children}
```
dschaller, line 88: Should error be above the row?

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+            onError(err);
+          }
+          setCardError(err);
+        })
+        .finally(() => setIsLoading(false));
+    }
+  };
+
+  React.useEffect(() => {
+    const { autoReload, reloadInterval = 30000 } = props;
```
dschaller, line 121: should autoreload have a default value?

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
+import React from "react";
+import { Card, ClutchError, Error, Grid, styled, Typography } from "@clutch-sh/core";
+import { LinearProgress } from "@material-ui/core";
+
+export type CardTypes = "Dynamic" | "Metadata";
```
dschaller, line 5: nit
```suggestion
export type CardType = "Dynamic" | "Metadata";
```

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export interface DetailsCard {
```
dschaller, line 7: nit: this name is a bit confusing. Just to confirm, this is solely used to identify the card type right?

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -1,39 +1,122 @@
...
+import fetchProject from "./info/resolver";
+import type { ProjectInfo } from "./info/types";
+import type { DetailsCard } from "./card";
+import { MetaCard } from "./card";
+import { ProjectDetailsContext } from "./context";
+import ProjectHeader from "./header";
+import ProjectInfoCard from "./info";
+import QuickLinksCard from "./quick-links";
+
+type CardTyping = React.ReactNode | DetailsCard;
```
dschaller, line 20: Wouldn't this be either a meta or dynamic card?

frontend/workflows/projectCatalog/src/details/card.tsx
dschaller, line 7: If you can use meta and dynamic card types below you can remove the export here

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+      </CardBody>
+    </StyledCard>
+  );
+};
+
+const DynamicCard = (props: BaseCardProps) => <BaseCard type="Dynamic" {...props} />;
+
+const MetaCard = (props: BaseCardProps) => <BaseCard type="Metadata" {...props} />;
+
+export { DynamicCard, MetaCard, StyledCard };
```
dschaller, line 149: Why export StyledCard?

frontend/workflows/projectCatalog/src/details/header.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const TextLink = styled("a")({
+  textDecoration: "none",
+  color: "unset",
+});
+
+const ProjectHeader = ({
+  name,
+  route = "/catalog",
```
dschaller, line 19: It would be nice if we could read this in from the config. I think maybe in the immediate we might not want this to be clickable.

frontend/workflows/projectCatalog/src/details/header.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <Grid container direction="column" style={{ width: "100%", height: "100%" }}>
+      <Grid container item direction="row" alignItems="flex-end">
+        <Typography variant="body4">
+          <TextLink href={route}>{routeTitle}</TextLink>&nbsp;/&nbsp;
+        </Typography>
+        <Typography variant="caption2">{name}</Typography>
+      </Grid>
+      <div style={{ padding: "8px 0px 8px 0px" }}>
+        <Typography variant="h2">{capitalize(name)}</Typography>
+      </div>
```
dschaller, line 33: Can this div be extracted to use styled for consistency?

frontend/workflows/projectCatalog/src/details/helpers.tsx
```diff
@@ -0,0 +1,70 @@
...
+import TimeAgo from "react-timeago";
+import { Grid, Link, styled, Typography } from "@clutch-sh/core";
+import { faClock } from "@fortawesome/free-regular-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+const StyledLink = styled(Link)({
+  whiteSpace: "nowrap",
+});
+
+const unitformatter = (unit: string): string => {
```
dschaller, line 12:
```suggestion
const unitFormatter = (unit: string): string => {
```

dschaller, line 12: Should unit have stronger typings?

frontend/workflows/projectCatalog/src/details/helpers.tsx
```diff
@@ -0,0 +1,70 @@
...
+        </Grid>
+        <Grid item>
+          <Typography variant="body4">
+            <EventTime date={setMilliseconds(time)} /> ago
+          </Typography>
+        </Grid>
+      </>
+    )}
+  </>
+);
```
dschaller, line 68: I think you can do something like
```suggestion
const LastEvent = ({ time }: { time: number }) => {
  time && (
      <>
        <Grid item>
          <FontAwesomeIcon icon={faClock} />
        </Grid>
        <Grid item>
          <Typography variant="body4">
            <EventTime date={setMilliseconds(time)} /> ago
          </Typography>
        </Grid>
      </>
)};
```

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -1,39 +1,122 @@
...
+
+      React.Children.forEach(children, child => {
+        if (React.isValidElement(child)) {
+          const { type } = child?.props;
+
+          switch (type) {
+            case "Metadata":
+              tempMetaCards.push(child);
+              break;
+            case "Dynamic":
```
dschaller, line 53: Should we keep these as constants and export them to define them only in one place?

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -1,39 +1,122 @@
...
+
+  return (
+    <ProjectDetailsContext.Provider value={{ projectInfo }}>
+      <StyledContainer container direction="row" wrap="nowrap">
+        {/* Column for project details and header */}
+        <Grid container item direction="column">
+          <Grid item style={{ marginBottom: "22px" }}>
+            {/* Static Header */}
+            {projectInfo && (
+              <ProjectHeader name={projectInfo.name} description={projectInfo.description} />
```
dschaller, line 74: We should always have the project name from the URL right?

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -1,39 +1,122 @@
...
+              <Grid item xs={12}>
+                {/* Static Info Card */}
+                <MetaCard
+                  title={capitalize(projectInfo?.name)}
+                  titleIcon={<GroupIcon />}
+                  fetchDataFn={() => fetchProject(projectId)}
+                  onSuccess={setProjectInfo}
+                  autoReload
+                  endAdornment={
+                    projectInfo?.disabled ? <DisabledItem name={projectInfo?.name} /> : null
```
dschaller, line 88: More for my own edification but when is something disabled?

frontend/packages/core/src/chip.tsx
jdslaugh: Yeah, the figma has the chip as 24px and our default was 32 so I made the small == 24

frontend/workflows/projectCatalog/src/details/info/types.tsx
jdslaugh: Yeah that makes sense. This was a hold over from when I went super generic. I've removed the types files and utilize the ProjectAPI typings now.

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+interface CardBodyProps {
+  children?: React.ReactNode;
+  loading?: boolean;
+  error?: ClutchError;
+}
+
+interface BaseCardProps extends CardTitleProps, CardBodyProps {
+  reloadInterval?: number;
+  autoReload?: boolean;
+  fetchDataFn?: () => Promise<any>;
```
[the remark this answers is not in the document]
jdslaugh: Oh neat, never really looked into the differences between unknown and any, good to know. TIL. Adjusted.

frontend/workflows/projectCatalog/src/details/card.tsx
jdslaugh: Yup, added comments for the entire interface. Was planning on going back and adding documentation for everything anyways.

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+      )}
+    </StyledRow>
+    {error ? <Error subject={error} /> : children}
+  </>
+);
+
+const BaseCard = (props: CardProps) => {
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+  const [cardError, setCardError] = React.useState<ClutchError | undefined>(undefined);
+  const { children, title, titleIcon, endAdornment, loading, error } = props;
```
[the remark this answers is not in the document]
jdslaugh: I forgot I can do { prop1, prop2, ...props } in the definition. Was trying to keep it clean for the lower functions. Adjusted.

frontend/workflows/projectCatalog/src/details/card.tsx
jdslaugh: I thought booleans defaulted to false generally, I'll add one to be explicit though. Fixed.

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+interface CardTitleProps {
+  title: string;
+  titleIcon?: React.ReactNode;
+  endAdornment?: React.ReactNode;
+}
+
+interface CardBodyProps {
+  children?: React.ReactNode;
+  loading?: boolean;
```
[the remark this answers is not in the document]
jdslaugh: Did it this way so that a user could manually control the error/loading state separate from the auto-handled pieces. Open to suggestions though.

frontend/workflows/projectCatalog/src/details/card.tsx
jdslaugh: Yeah, it is a bit confusing. Wanted an interface that could be used as the child type interface which didn't interfere with the BaseCard. (even writing that was confusing lol).

Ultimately:

BaseCard inherits the type from DetailsCard (which is exported) via CardProps and BaseCardProps. 
The DynamicCard and MetaCard utilize the BaseCardProps to pass to the BaseCard and their individual types. 
Then any user can extend the DetailsCard to make a custom card with the required typing.

I think I ran out of names..

frontend/workflows/projectCatalog/src/details/index.tsx
jdslaugh: Yeah it should be, think this is a hold over from the original implementation. Adjusted

frontend/workflows/projectCatalog/src/details/card.tsx
jdslaugh: Yeah this was only so that it could be extended in custom cases, if we dont think we need that I can remove it though.

frontend/workflows/projectCatalog/src/details/card.tsx
jdslaugh: Forgot to delete it, was from the original implementation. Removed.

frontend/workflows/projectCatalog/src/details/header.tsx
jdslaugh: Yeah that was the original idea, I'll remove the linking for MVP

frontend/workflows/projectCatalog/src/details/header.tsx
jdslaugh: Yup, added

frontend/workflows/projectCatalog/src/details/helpers.tsx
jdslaugh: Probably, this was copied from internal and I need to export it into a reusable component. I have a ticket to track it and I'll make sure to type it better, need to make sure I understand all the use cases for it first.

frontend/workflows/projectCatalog/src/details/helpers.tsx
jdslaugh: Ah yeah, that is cleaner. Thanks!

frontend/workflows/projectCatalog/src/details/index.tsx
jdslaugh: Updated to use projectId. The thought was that projectId could be anything for users, hash etc, so I originally had it extending from projectInfo.

frontend/workflows/projectCatalog/src/details/index.tsx
jdslaugh: Yup, updated to use an enum

frontend/workflows/projectCatalog/src/details/card.tsx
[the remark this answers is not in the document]
dschaller: Yeah, I'm just trying to think of a use case where you'd want to control this from the parent. Maybe we can remove for now and add back if needed in the future.

frontend/workflows/projectCatalog/src/details/index.tsx
dschaller: Yeah that makes sense. I think for now we can use the ID and change later if needed since the old behavior would not show anything if projectInfo didn't exist which is kind of weird when you navigate to a identified page

frontend/workflows/projectCatalog/src/details/card.tsx
```diff
@@ -21,10 +21,15 @@ interface CardBodyProps {
...
 
 interface BaseCardProps extends CardTitleProps, CardBodyProps {
+  /** Number in ms to refresh the data from fetchDataFn */
   reloadInterval?: number;
+  /** Boolean representing whether the component should reload via the fetchDataFn */
   autoReload?: boolean;
+  /** Given promise which will be used to initially fetch data and optionally reload on intervals */
   fetchDataFn?: () => Promise<unknown>;
+  /** Function called when fetchDataFn returns successfully, returning the data */
   onSuccess?: (data: any) => void;
```
dschaller, line 34: Really minor nit but this should probably be `data: unknown`

frontend/workflows/projectCatalog/src/details/info/messengerRow.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <>
+      <Grid container item spacing={1}>
+        <Grid item>
+          <FontAwesomeIcon icon={icon} size="lg" />
+        </Grid>
+        <Grid item>{text && <LinkText text={text} link={link} />}</Grid>
+      </Grid>
+    </>
```
dschaller, line 40: nit:
```suggestion
    <Grid container item spacing={1}>
      <Grid item>
        <FontAwesomeIcon icon={icon} size="lg" />
      </Grid>
      <Grid item>{text && <LinkText text={text} link={link} />}</Grid>
    </Grid>
```

frontend/workflows/projectCatalog/src/index.tsx
```diff
@@ -1,9 +1,15 @@
 import type { BaseWorkflowProps, WorkflowConfiguration } from "@clutch-sh/core";
 
+import type { DetailsCard } from "./details/card";
+import type { ProjectInfoChip } from "./details/info/chipsRow";
 import Catalog from "./catalog";
 import Details from "./details";
 
 export interface WorkflowProps extends BaseWorkflowProps {}
+export interface DetailWorkflowProps {
```
dschaller, line 9: nit: this name should probably match with the export name `ProjectDetailsWorkflowProps`

frontend/workflows/projectCatalog/src/details/info/index.tsx
```diff
@@ -17,29 +18,37 @@ const StyledRow = styled(Grid)({
...
+      {data?.repository && (
+        <StyledRow container spacing={1} justify="flex-start" alignItems="center">
+          <RepositoryRow {...data.repository} />
+        </StyledRow>
+      )}
+      {data?.languages?.length ? (
+        <StyledRow container spacing={1} justify="flex-start" alignItems="flex-end">
+          <LanguageRow languages={data.languages} />
+        </StyledRow>
+      ) : null}
```
dschaller, line 59: super nit for consistency to mirror the length check below

frontend/workflows/projectCatalog/src/details/info/repositoryRow.tsx
```diff
@@ -0,0 +1,81 @@
...
+            break;
+          default:
+            setIcon(faCode);
+        }
+
+        // TODO (jslaughter): fetch open PR's count
+        // setRequests({
+        //   number: 0,
+        //   url: `https://${manager}/${project}/pulls`,
+        // });
```
dschaller, line 50: good callout. We will likely need to do this on the server since the remote could be private and require authentication

frontend/workflows/projectCatalog/src/details/info/repositoryRow.tsx
```diff
@@ -0,0 +1,81 @@
+import React from "react";
+import { Grid } from "@clutch-sh/core";
+import type { IconProp } from "@fortawesome/fontawesome-svg-core";
+import { faGithub } from "@fortawesome/free-brands-svg-icons";
+import { faCode, faCodeBranch } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+import { LinkText } from "../helpers";
+
+interface ProjectRequests {
```
dschaller, line 10: This is more for later but I think we could improve this name to be a bit more descriptive. I think it's ok to include something like pull requests in this name as it's more descriptive

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -1,39 +1,174 @@
...
+   * (ex: ["clutch-team@lyft.com"] -> "Clutch Team")
+   */
+  const getOwner = (owners: string[]): string => {
+    if (owners && owners.length) {
+      const firstOwner = owners[0];
+
+      return firstOwner
+        .replace(/@.*/, "")
+        .split("-")
+        .map(s => capitalize(s))
```
dschaller, line 103: I'm curious why do this instead of setting a text transform capitalize style on the text

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -1,39 +1,174 @@
...
+  /**
+   * Takes an array of owner emails and returns the first one capitalized
+   * (ex: ["clutch-team@lyft.com"] -> "Clutch Team")
+   */
+  const getOwner = (owners: string[]): string => {
+    if (owners && owners.length) {
+      const firstOwner = owners[0];
+
+      return firstOwner
+        .replace(/@.*/, "")
```
dschaller, line 101: nit: we might want this to be more specific (e.g. if someone were to want to include the @ sign for slack rooms or something). Maybe 
```suggestion
        .replace(/@.*\..*/, "")
```

frontend/workflows/projectCatalog/src/details/card.tsx
jdslaugh: Yup, missed that one. Will fix.

frontend/workflows/projectCatalog/src/details/info/messengerRow.tsx
jdslaugh: Woops, missed that when I adjusted. Thanks

frontend/workflows/projectCatalog/src/index.tsx
jdslaugh: Good call

frontend/workflows/projectCatalog/src/details/info/index.tsx
jdslaugh: Modified to make the checks match

frontend/workflows/projectCatalog/src/details/info/repositoryRow.tsx
jdslaugh: Yeah I was looking to see if we had anything setup but seems not currently. It stays hidden until we provide something though so it wont affect anything at least.

frontend/workflows/projectCatalog/src/details/info/repositoryRow.tsx
jdslaugh: Modified to be `ProjectPullRequests`, I had it as `ProjectRequests` originally since I was leaving it up to the user to specify the type, but I've removed that logic anyways.

frontend/workflows/projectCatalog/src/details/index.tsx
jdslaugh: Because I forgot that text transforms existed :) Updated!

frontend/workflows/projectCatalog/src/details/index.tsx
jdslaugh: Works for me!

dschaller (approved):

PR #2143, opened 2022-03-27
FE: add tests for chart helper funcs

smonero:
### Description

Pt 2 of many of adding tests to the FE

### Testing Performed

local

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Charts/helpers.tsx
+++ b/frontend/packages/core/src/Charts/helpers.tsx
@@ -10,7 +10,13 @@ export const dateTimeFormatter = (timestamp: number) => {
   return new Date(timestamp).toDateString();
 };
 
-const getMinAndMaxOfRangeUsingKey = (data: any, key: string) => {
+export const getMinAndMaxOfRangeUsingKey = (data: any, key: string) => {
+  if (!data) {
+    return {
+      min: null,
+      max: null,
+    };
+  }
   const filtered = data.filter(d => d[key]).map(d => d[key]);
   return { min: Math.min(...filtered), max: Math.max(...filtered) };
 };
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Charts/helpers.tsx
```diff
@@ -20,7 +26,17 @@ const getMinAndMaxOfRangeUsingKey = (data: any, key: string) => {
...
+  if (
+    edgeRatio <= 0 ||
+    min === null ||
+    max === null ||
+    min === Infinity ||
+    max === -Infinity ||
+    min === undefined ||
+    max === undefined ||
+    Number.isNaN(min) ||
+    Number.isNaN(max)
```
danielhochman, line 38: i believe this can be simplified using `Number.isFinite`

edit: same for other cases below

danielhochman (approved):

frontend/packages/core/src/Charts/tests/helpers.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+  calculateDomainEdges,
+  calculateTicks,
+  getLeftSideAndIntervalForTicks,
+  getMinAndMaxOfRangeUsingKey,
+  isoTimeFormatter,
+} from "../helpers";
+
+describe("helper functions for charts", () => {
+  describe("isoTimeFormatter", () => {
+    test("a timestamp", () => {
```
dschaller, line 11: super nit
```suggestion
    test("returns a timestamp", () => {
```

frontend/packages/core/src/Charts/tests/helpers.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+      meow: "100",
+      woof: "2",
+    },
+    {
+      meow: "50",
+      woof: "2",
+    },
+  ];
+  describe("getMinAndMaxOfRangeUsingKey", () => {
+    test("the key exists in the data", () => {
```
dschaller, line 31: nit: when writing these test blocks it helps to think about how this will read
```
> helper functions for charts
>> getMinAndMaxOfRangeUsingKey
>>> the key exists in the data ✅
```

Generally you'd want some information about expectations in the test names so it's easier to debug from a higher level. Reading through these tests it seems like what you're trying to assert is that different range lookup cases are handled without errors. Is that right?

jdslaugh (approved):

PR #2151, opened 2022-03-29
FE: Project Catalog: Quicklinks with images

smonero:
### Description

Adds support for images to the quick links.

### Testing Performed

Screen Shot 2022-03-29 at 12 57 18 PM

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/projectCatalog/src/details/quick-links.tsx
+++ b/frontend/workflows/projectCatalog/src/details/quick-links.tsx
@@ -1,7 +1,6 @@
 import React from "react";
-import { Link } from "react-router-dom";
 import type { clutch as IClutch } from "@clutch-sh/api";
-import { Card, Grid } from "@clutch-sh/core";
+import { Card, Grid, Tooltip, Link } from "@clutch-sh/core";
 
 const QuickLinksCard = (linkGroups: IClutch.core.project.v1.ILinkGroup[]) => {
   return (
@@ -17,10 +16,12 @@ const QuickLinksCard = (linkGroups: IClutch.core.project.v1.ILinkGroup[]) => {
         {linkGroups?.map(linkGroup => {
           return linkGroup.links?.map(link => {
             return (
-              <Grid item key={link.name}>
-                <Link to={link.url}>
-                  <img width="29px" height="29px" src={linkGroup.imagePath} alt={link.name} />
-                </Link>
+              <Grid item key={link.name} >
+                <Tooltip title={linkGroup.name}>
+                  <Link href={link.url}>
+                    <img width="29px" height="29px" src={linkGroup.imagePath} alt={link.name} />
+                  </Link>
+                </Tooltip>
               </Grid>
             );
           });
--- a/frontend/workflows/projectCatalog/src/details/quick-links.tsx
+++ b/frontend/workflows/projectCatalog/src/details/quick-links.tsx
@@ -11,13 +11,14 @@ import {
   Typography,
 } from "@clutch-sh/core";
 
+const ICON_SIZE = "32px";
 // If only a single link, then no popper is necessary
 const QuickLink = ({ link, linkGroupName, linkGroupImage }) => (
   <Grid item key={link.name}>
     <Tooltip title={linkGroupName}>
       <TooltipContainer>
         <Link href={link.url}>
-          <img width="32px" height="32px" src={linkGroupImage} alt={link.name} />
+          <img width={ICON_SIZE} height={ICON_SIZE} src={linkGroupImage} alt={link.name} />
         </Link>
       </TooltipContainer>
     </Tooltip>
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <Grid item key={link.name}>
+      <Tooltip title={linkGroupName}>
+        <Link href={link.url}>
+          <img width="32px" height="32px" src={linkGroupImage} alt={link.name} />
+        </Link>
+      </Tooltip>
+    </Grid>
+  );
+};
```
dschaller, line 16:
```suggestion
const QuickLink = ({ link, linkGroupName, linkGroupImage }) => (
  <Grid item key={link.name}>
    <Tooltip title={linkGroupName}>
      <Link href={link.url}>
        <img width="32px" height="32px" src={linkGroupImage} alt={link.name} />
      </Link>
    </Tooltip>
  </Grid>
);
```

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,99 @@
...
+// Have a popper in the case of multiple links per group
+const QuickLinkGroup = ({ linkGroupName, linkGroupImage, links }) => {
+  const anchorRef = React.useRef(null);
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <Grid item key={linkGroupName}>
+      <Tooltip title={linkGroupName}>
+        <TooltipContainer>
+          {/* eslint-disable */}
```
dschaller, line 36: just curious, why disable lint here?

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,99 @@
...
+  Typography,
+} from "@clutch-sh/core";
+
+// If only a single link, then no popper is necessary
+const QuickLink = ({ link, linkGroupName, linkGroupImage }) => (
+  <Grid item key={link.name}>
+    <Tooltip title={linkGroupName}>
+      <TooltipContainer>
+        <Link href={link.url}>
+          <img width="32px" height="32px" src={linkGroupImage} alt={link.name} />
```
dschaller, line 20: Should we extract out the size to a constant since it's used here and below?

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,99 @@
...
+              return (
+                <PopperItem key={link.name}>
+                  <Link href={link.url}>
+                    <Typography color="inherit" variant="body4">
+                      {link.name}
+                    </Typography>
+                  </Link>
+                </PopperItem>
+              );
+            })}
```
dschaller, line 56:
```suggestion
            {links.map(link => (
              <PopperItem key={link.name}>
                <Link href={link.url}>
                  <Typography color="inherit" variant="body4">
                    {link.name}
                  </Typography>
                </Link>
              </PopperItem>
            ))}
```

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </Popper>
+          </TooltipContainer>
+      </Tooltip>
+    </Grid>
+  );
+};
+export interface QuickLinksCardInput {
```
dschaller, line 63:
```suggestion
export interface QuickLinksProps {
```

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+const QuickLinksCard = ({ linkGroups }: QuickLinksCardInput) => (
+    <Card>
+      <Grid
+        container
+        item
+        direction="column"
+        alignItems="center"
+        spacing={1}
+        style={{ padding: "7px 0" }}
```
dschaller, line 75: You might have to adjust the padding a bit more from Kanchan's designs. I just went with this as the placeholder

frontend/workflows/projectCatalog/src/details/quick-links.tsx
smonero: We get the following errors:
            error  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
            error  Non-interactive elements should not be assigned mouse or keyboard event listeners               jsx-a11y/no-noninteractive-element-interactions

I tried the solutions I found on stackoverflow about assigning role to button, and a few other things, and none of those worked.
There's many directions we can go here.
Maybe a better solution is to use a `button` or another component (IconButton?)? I'm not sure. Just thought I'd disable until we come up with a better solution.

smonero: I can make it work if I wrap it in a `button` or `IconButton` but the problem is then I couldn't figure out how to fix the styling (it wouldn't override). When I do those things I get unwanted boxes and color.

smonero: So I got it to work with setting the background for an img. One issue is that the "hand" icon doesn't appear in that case. I wonder if that is ok or if that is a suboptimal UX

danielhochman: should be able to fix that with css `cursor: pointer`

smonero: yeah I tried that. Also I tried using Link but the Link from Clutch requires an href which doesn't fit this use case.

dschaller: I think you'll want something like 
```jsx
          <button
            type="button"
            style={{ padding: 0, background: "transparent", border: "0", cursor: "pointer" }}
            ref={anchorRef}
            onClick={() => setOpen(true)}
          >
            <img width={ICON_SIZE} height={ICON_SIZE} src={linkGroupImage} alt={linkGroupName} />
          </button>

```

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <Link href={link.url}>
+          <img width={ICON_SIZE} height={ICON_SIZE} src={linkGroupImage} alt={link.name} />
+        </Link>
+      </TooltipContainer>
+    </Tooltip>
+  </Grid>
+);
+
+// Have a popper in the case of multiple links per group
+const QuickLinkGroup = ({ linkGroupName, linkGroupImage, links }) => {
```
dschaller, line 29: We should probably have types on this.

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const anchorRef = React.useRef(null);
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <Grid item key={linkGroupName}>
+      <Tooltip title={linkGroupName}>
```
dschaller, line 35: If we don't have a name this will just show an empty dark field under the icon. We should probably not show a tooltip if there is no value here or enforce that this must be set.

smonero: yeah we can enforce that it is set in the proto

smonero:
made this for the blank tooltip thing https://github.com/lyft/clutch/pull/2153

jdslaugh (approved):
Looks great!

dschaller (approved):

PR #2158, opened 2022-03-31
ProjectCatalog: QuickLinks Patch

jdslaugh:
### Description
- Hides the quick links if there aren't any

#### Issue
Screen Shot 2022-03-31 at 4 35 27 AM

### Testing Performed
manual

### DIFF
--- a/frontend/workflows/projectCatalog/src/details/index.tsx
+++ b/frontend/workflows/projectCatalog/src/details/index.tsx
@@ -162,7 +162,7 @@ const Details: React.FC<ProjectDetailsWorkflowProps> = ({ children, chips }) => 
         <Hidden smDown>
           <Grid item direction="column">
             {projectInfo && (projectInfo.linkGroups ?? []).length > 0 && (
-              <QuickLinksCard linkGroups={projectInfo?.linkGroups ?? []} />
+              <QuickLinksCard linkGroups={projectInfo.linkGroups} />
             )}
           </Grid>
         </Hidden>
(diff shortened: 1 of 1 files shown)


smonero (approved):

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -161,7 +161,9 @@ const Details: React.FC<ProjectDetailsWorkflowProps> = ({ children, chips }) =>
         {/* Column for project quick links */}
         <Hidden smDown>
           <Grid item direction="column">
-            {projectInfo && <QuickLinksCard linkGroups={projectInfo?.linkGroups ?? []} />}
+            {projectInfo && (projectInfo.linkGroups ?? []).length > 0 && (
+              <QuickLinksCard linkGroups={projectInfo?.linkGroups ?? []} />
```
danielhochman, line 165: This can be simplified now
```suggestion
              <QuickLinksCard linkGroups={projectInfo.linkGroups} />
```

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -161,7 +161,9 @@ const Details: React.FC<ProjectDetailsWorkflowProps> = ({ children, chips }) =>
         {/* Column for project quick links */}
         <Hidden smDown>
           <Grid item direction="column">
-            {projectInfo && <QuickLinksCard linkGroups={projectInfo?.linkGroups ?? []} />}
+            {projectInfo && (projectInfo.linkGroups ?? []).length > 0 && (
```
danielhochman, line 164: wdyt about something like lodash isempty? might be cleaner
```js
_.isEmpty(projectInfo.?linkGroups) && (...)
```

danielhochman (approved):

frontend/workflows/projectCatalog/src/details/index.tsx
jdslaugh: Yeah that works, have to utilize a non-null assertion as well but with the additional check it should be good.

PR #2174, opened 2022-04-05
project catalog: fix autocomplete behavior

mikecutalo:
### Description
Previously when using autocomplete when you hit "return" to select an item from the list, this would submit the form before actually updating the value with the one you were trying to select.

This was the same issue we saw on the project selector, it was fixed in this exact way which i borrowed from https://github.com/lyft/clutch/pull/2061/files.

### Testing Performed
Locally


frontend/workflows/projectCatalog/src/catalog/index.tsx
```diff
@@ -60,19 +80,15 @@ const Catalog: React.FC<WorkflowProps> = ({ heading }) => {
...
-    });
+  const { handleSubmit } = useForm({
+    mode: "onSubmit",
+    reValidateMode: "onSubmit",
+    shouldFocusError: false,
+  });
 
-    return { results: response?.data?.results || [] };
+  const handleChanges = event => {
+    setCustomProject(event.target.value);
```
jdslaugh, line 90: I dont think you should need the extra customProject state, setting the value equal to `state.search` should be sufficient as long as the dispatch is correct.

mikecutalo: Sounds good let me remove it and see how it does.

mikecutalo: Seems to work just fine, i removed the extra state.

jdslaugh (approved):
Thanks!

frontend/workflows/projectCatalog/package.json
```diff
@@ -38,7 +38,8 @@
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-router-dom": "^6.0.0-beta",
-    "react-timeago": "^6.0.0"
+    "react-timeago": "^6.0.0",
+    "react-hook-form": "^6.11.0"
```
dschaller, line 42: super nit: this should be alphabetical

dschaller (approved):

frontend/workflows/projectCatalog/package.json
```diff
@@ -35,10 +35,11 @@
     "@material-ui/core": "^4.11.4",
     "@material-ui/icons": "^4.11.2",
     "lodash": "^4.17.21",
-    "react": "^17.0.2",
```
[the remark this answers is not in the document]
mikecutalo: I just sort with my editor i added by hand.

PR #2178, opened 2022-04-06
frontend: allow users to reset project catalog index state

dschaller:
### Description

Add ability to reset default state of project catalog index. This continues to allow users to remove all projects (even the ones they own). If all projects have been removed users will see a placeholder (see below). On initial load (e.g. no existing state) we will fetch all the projects a users owns.

Screenshot from 2022-04-05 20-14-34

Screenshot from 2022-04-05 20-14-51

Screenshot from 2022-04-08 09-08-29

### Testing Performed

manual

### DIFF
--- a/frontend/workflows/projectCatalog/src/catalog/index.tsx
+++ b/frontend/workflows/projectCatalog/src/catalog/index.tsx
@@ -6,6 +6,7 @@ import {
   Paper,
   TextField,
   Toast,
+  Tooltip,
   Typography,
   useNavigate,
 } from "@clutch-sh/core";
@@ -17,7 +18,7 @@ import type { WorkflowProps } from "..";
 
 import catalogReducer from "./catalog-reducer";
 import ProjectCard from "./project-card";
-import { addProject, clearProjects, getProjects, removeProject } from "./storage";
+import { addProject, clearProjects, getProjects, hasState, removeProject } from "./storage";
 import type { CatalogState } from "./types";
 
 const initialState: CatalogState = {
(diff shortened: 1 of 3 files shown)


smonero (approved):
Can we maybe add some text or tooltip or something making it obvious to the user that this is a reset projects button?

Compared to the alternative though I like this one better.

frontend/workflows/projectCatalog/src/catalog/index.tsx
```diff
@@ -103,13 +113,36 @@ const Catalog: React.FC<WorkflowProps> = ({ heading }) => {
...
+          <RestoreIcon />
+        </IconButton>
       </div>
       <Grid container direction="row" spacing={5}>
         {state.projects.map(p => (
           <Grid item onClick={() => navigateToProject(p)}>
-            <ProjectCard project={p} onRemove={() => triggerProjectRemove(p)} />
+            <ProjectCard
+              project={p}
+              onRemove={state.projects.length !== 1 ? () => triggerProjectRemove(p) : undefined}
```
jdslaugh, line 144: I feel like this behavior is going to be a little weird if we're not going to put up a message or something.

Might be better to just have a default div that shows up if there's no projects defined:

'No Projects Currently Added, Initialize Projects?' or something which could use the new reset features.

frontend/workflows/projectCatalog/src/catalog/index.tsx
```diff
@@ -20,6 +30,15 @@ const initialState: CatalogState = {
   error: undefined,
 };
 
+const Placeholder = () => (
+  <Paper>
+    <div style={{ margin: "32px", textAlign: "center" }}>
+      <Typography variant="h5">There is nothing to display here</Typography>
+      <Typography variant="body3">Please enter a namespace and clientset to proceed.</Typography>
```
jdslaugh, line 37: Should this say 'project name' to proceed? Currently the wording seems a little confusing since we dont search via namespace or anything correct?

dschaller: ah yes, good catch

jdslaugh (approved):
Looks great!

PR #2179, opened 2022-04-06
catalog: update card style on index search page

danielhochman:
### Description
- Remove language icon
- Adjust spacing
- Various minor updates to color and other properties

**Before**

**After**

### DIFF
--- a/frontend/workflows/projectCatalog/src/catalog/project-card.tsx
+++ b/frontend/workflows/projectCatalog/src/catalog/project-card.tsx
@@ -60,7 +60,7 @@ const ProjectCard = ({ project, onRemove }: ProjectCardProps) => {
         zeroMinWidth
       >
         <Typography variant="body2" color="rgba(13, 16, 48, 0.65)">
-          {project.data?.description}
+          {project?.data?.description}
         </Typography>
       </Grid>
       <Grid container item justify="flex-end" alignItems="center">
--- a/frontend/workflows/projectCatalog/src/catalog/project-card.tsx
+++ b/frontend/workflows/projectCatalog/src/catalog/project-card.tsx
@@ -60,7 +60,7 @@ const ProjectCard = ({ project, onRemove }: ProjectCardProps) => {
         zeroMinWidth
       >
         <Typography variant="body2" color="rgba(13, 16, 48, 0.65)">
-          Lorem ipsum baba chacha. Foo bar baz qux. Lorem ipsum bingo.
+          {project.data?.description}
         </Typography>
       </Grid>
       <Grid container justify="flex-end">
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectCatalog/src/catalog/project-card.tsx
```diff
@@ -42,34 +40,31 @@ const ProjectCard = ({ project, onRemove }: ProjectCardProps) => {
...
       </Grid>
       <Grid
         container
-        style={{ marginTop: "16px", paddingRight: "16px", flex: "1", overflow: "hidden" }}
+        style={{ marginTop: "8px", paddingRight: "16px", flex: "1", overflow: "hidden" }}
         zeroMinWidth
       >
-        <Typography variant="body2">{project.data?.description}</Typography>
+        <Typography variant="body2" color="rgba(13, 16, 48, 0.65)">
+          {project.data?.description}
```
smonero, line 63: should this be `project?.data?.description` (line 47 has a qmark after project)

smonero (approved):

dschaller:
Just want to be sure this is intentional since it doesn't match the designs

frontend/workflows/projectCatalog/src/catalog/project-card.tsx
```diff
@@ -42,34 +40,31 @@ const ProjectCard = ({ project, onRemove }: ProjectCardProps) => {
...
+        </Typography>
       </Grid>
-      <Grid container justify="flex-end">
-        <Grid item className="showOnHover">
-          <KeyboardArrowRightIcon />
-        </Grid>
+      <Grid container item justify="flex-end" alignItems="center">
+        <KeyboardArrowRightIcon color="secondary" />
```
dschaller, line 67: Why are we showing this on all cards always?

danielhochman: it conveys to the users that cards are clickable and go somewhere without having to hover over one.

danielhochman:
> This doesn't match the designs so I just want to be sure, is that intentional?

Yep, we decided to diverge from the designs when we removed the icon. The spacing in the design no longer really works without the icon (in my opinion). I also don't know that we ever really focused on the index page designs, though maybe we did when I was out.

I think we need to take another pass at the designs with tier badges and other info and really get it right. Then we can implement to match.

If you or others don't feel this is an improvement we can close this PR and do a simple one just to remove the icon.

frontend/workflows/projectCatalog/src/catalog/project-card.tsx
dschaller: maybe we can make it a bit lighter. It seems kind of overwhelming if there are several projects
Screenshot from 2022-04-06 11-36-56

dschaller:
sounds good. I like the smaller cards, I just wanted to be sure we meant to change things like the alignment of the remove icon.

danielhochman:
@dschaller good call, how's this?
Screenshot from 2022-04-06 13-55-04

dschaller (approved):

PR #2180, opened 2022-04-07
FE: packages: export charts and chart types

smonero:
### Description

Need to export these to use them internally.

### Testing Performed

no
### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/index.tsx
+++ b/frontend/packages/core/src/index.tsx
@@ -46,21 +46,10 @@ export * from "./Feedback";
 export * from "./Assets/emojis";
 export * from "./NPS";
 export * from "./chip";
+export * from "./Charts"
 
 export { default as ClutchApp } from "./AppProvider";
 
-export type {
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/index.tsx
```diff
@@ -1,6 +1,8 @@
 import { Grid } from "@material-ui/core";
 
 import { userId } from "./AppLayout/user";
+import LinearTimeline from "./Charts/linearTimeline";
```
dschaller, line 4: We should probably just have an index file in the charts folder that exports all these and can be re-exported here.

jdslaugh, line 4: You should be able to do:

```
export { default as LinearTimeline } from "./Charts/linearTimeline";
```

Although it might be better to create a Charts index file and just export everything from there instead.

jdslaugh (approved):

dschaller (approved):

PR #2182, opened 2022-04-08
frontend: allow for user override in dev

dschaller:
### Description

support overriding the userId from env variables

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/AppLayout/user.tsx
+++ b/frontend/packages/core/src/AppLayout/user.tsx
@@ -122,6 +122,13 @@ interface JwtToken {
 }
 
 const userId = (): string => {
+  if (process.env.NODE_ENV === "development") {
+    // eslint-disable-next-line
+    console.warn("No user token set in development");
+    if (process.env.REACT_APP_USER_ID) {
+      return process.env.REACT_APP_USER_ID;
+    }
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/AppLayout/user.tsx
```diff
@@ -122,6 +122,13 @@ interface JwtToken {
 }
 
 const userId = (): string => {
+  if (process.env.NODE_ENV === "development") {
+    // eslint-disable-next-line
+    console.warn("No user token set in development");
```
jdslaugh, line 127: Should this be after the if check? I'm confused what the warning is meant to entail.

dschaller: Oh it's basically to denote to users that dev will always be anonymous - I can also remove it

jdslaugh: Oh ok, yeah I'd say after the if check if we're not returning something it might be nice to say 'Returning `Anonymous` user for development'. I was confused at first when I had to figure out why it was getting that.

smonero (approved):

jdslaugh (approved):
Looks good!

PR #2191, opened 2022-04-12
FE: Project Selector: Add Quick Links

smonero:
### Description

Add the option for quicklinks to the project selector, such as links to logs, grafana, etc.
Note that there is a very slight issue with some of our source images - the k8s one has a slightly excessive border around it which leads to it looking slightly misaligned. However, I don't think anyone will really notice. There are also differences between the figma and reality because in figma there is differing padding due to the sizing of the icons.

I also thought about making the whole box clickable instead of the image and text but I wasn't sure if that would be confusing.

Screen Shot 2022-04-27 at 12 53 57 AM

showingspacing

Note I messed with the padding a fair amount and tried to end up with something that looks good. If you look closely at the inspection it looks like some values are 1 or 2 pixels off but I think they look fine. The top screenshot has no top padding on the sublinks (only bottom, left and right), whereas the bottom gif has 4px padding all around.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/projectSelector/src/project-group.tsx
+++ b/frontend/workflows/projectSelector/src/project-group.tsx
@@ -222,7 +222,10 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
                 </StyledClearIcon>
               </StyledHoverOptions>
               {!state?.loading && state?.projectData?.[key]?.linkGroups && (
-                <ProjectLinks linkGroups={state?.projectData?.[key]?.linkGroups} name={[key]} />
+                <ProjectLinks
+                  linkGroups={state?.projectData?.[key]?.linkGroups ?? []}
+                  name={[key]}
+                />
               )}
             </StyledMenuItem>
           ))}
--- a/frontend/workflows/projectSelector/src/project-group.tsx
+++ b/frontend/workflows/projectSelector/src/project-group.tsx
@@ -36,8 +36,9 @@ const StyledCount = styled.span({
   display: "inline-block",
 });
 
+// This div used to have `padding: "0 25px 0 8px"` but that made it look weird when we implemented quicklinks
+// because the "only" and "x" buttons are hidden when the popper is expanded and mouse is no longer hovering.
 const StyledMenuItem = styled.div({
-  padding: "0 25px 0 8px",
   height: "48px",
   display: "flex",
   alignItems: "center",
@@ -123,6 +124,10 @@ const StyledHoverOptions = styled.div({
   alignItems: "center",
 });
 
+const StyledFlexEnd = styled.div({
+  justifyContent: "right",
+  flexDirection: "row-reverse",
+});
 interface ProjectGroupProps {
   title: string;
   group: Group;
@@ -134,6 +139,7 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
   const state = useReducerState();
 
   const [collapsed, setCollapsed] = React.useState(false);
+  const [keyWithQLinksOpen, setKeyWithQLinksOpen] = React.useState("");
 
   const groupKeys = Object.keys(state?.[group] ?? {});
   const numProjects = groupKeys.length;
@@ -219,10 +225,16 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
                     </IconButton>
                   )}
                 </StyledClearIcon>
-                {state?.projectData?.key?.linkGroups && (
-                  <ProjectLinks linkGroups={state?.projectData?.key?.linkGroups} />
-                )}
               </StyledHoverOptions>
+              <StyledFlexEnd hidden={key !== keyWithQLinksOpen}>
+                {state?.projectData?.key?.linkGroups && (
+                  <ProjectLinks
+                    linkGroups={state?.projectData?.key?.linkGroups}
+                    name={key}
+                    setVisibleFn={setKeyWithQLinksOpen}
+                  />
+                )}
+              </StyledFlexEnd>
             </StyledMenuItem>
           ))}
         </div>
--- a/frontend/workflows/projectSelector/src/project-group.tsx
+++ b/frontend/workflows/projectSelector/src/project-group.tsx
@@ -7,7 +7,7 @@ import ClearIcon from "@material-ui/icons/Clear";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 
 import { deriveSwitchStatus, useDispatch, useReducerState } from "./helpers";
-import { ProjectLinks } from "./project-links";
+import ProjectLinks from "./project-links";
 import type { Group } from "./types";
 
 const StyledGroup = styled.div({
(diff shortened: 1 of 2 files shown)


miguelmolina95 (approved):

frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -133,6 +139,7 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
   const state = useReducerState();
 
   const [collapsed, setCollapsed] = React.useState(false);
+  const [keyWithQLinksOpen, setKeyWithQLinksOpen] = React.useState("");
```
danielhochman, line 142: nit: seems like we could push this state down a component level since it's only used in the one parent, just helps to minimize the amount of state to mentally work through when working on this component.

frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -219,6 +226,15 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
                   )}
                 </StyledClearIcon>
               </StyledHoverOptions>
+              <StyledFlexEnd hidden={key !== keyWithQLinksOpen}>
+                {state?.projectData?.key?.linkGroups && (
```
jdslaugh, line 230: Should this be wrapped around the `StyledFlexEnd`? That way we dont even render the entire element if there aren't any links

frontend/workflows/projectSelector/src/project-links.tsx
```diff
@@ -0,0 +1,168 @@
...
+        <StyledLinkTitle style={{ paddingLeft: "6px" }}>{linkGroupName}</StyledLinkTitle>
+      </Link>
+    </StyledMenuItem>
+  ) : null;
+
+interface QuickLinkGroupProps extends LinkGroupProps {
+  links: IClutch.core.project.v1.ILink[];
+}
+
+const QuickLinkGroup = ({ linkGroupName, linkGroupImage, links }: QuickLinkGroupProps) => {
```
jdslaugh, line 63: I wonder if there's a way to combine the `QuickLink` and `QuickLinkGroup` components. Right now it's somewhat duplicated but we could possibly have an optional children param which we could use for setting the styling/adding the extra links around the icon.

What do you think?

jdslaugh (commented):
This is great! Few comments.

Also, looking at the screenshot I see 2 possible nits:

Are all the icons centered? It looks like Slack is centered more than the others.

Also, is the spacing between the Title and the links the same from the mockups? Maybe it's just my eyes though, just looks like a lot of white space. :)

frontend/workflows/projectSelector/src/project-group.tsx
```diff
@@ -219,6 +226,15 @@ const ProjectGroup: React.FC<ProjectGroupProps> = ({ title, group, displayToggle
...
+                  <ProjectLinks
+                    linkGroups={state?.projectData?.key?.linkGroups}
```
danielhochman, line 232: this needs to reference the variable
```suggestion
                    linkGroups={state?.projectData?.[key]?.linkGroups}
```

edit: in two other places here as well

danielhochman:
Seeing two issues when testing
1. When the menu is open the menu icon is moving around based on hover
2. When there are no links the menu icon still appears

Peek 2022-04-21 11-00

smonero:
> This is great! Few comments.
> 
> Also, looking at the screenshot I see 2 possible nits:
> 
> Are all the icons centered? It looks like Slack is centered more than the others.
> 
> Also, is the spacing between the Title and the links the same from the mockups? Maybe it's just my eyes though, just looks like a lot of white space. :)

the slack one is the biggest. The original reasoning behind using external images was that it would save on bloat space. I think a decent solution would be to just import the icons that Kanchan made and not worry about how much space we add because of svg icons. Is that a reasonable solution or should we continue using external images (I would probably have to reupload and relink external images that are more equal in size?)

frontend/workflows/projectSelector/src/project-links.tsx
smonero: yeah I can combine them. One issue is that the Link wraps the single one, but not the group one, then that makes it so even the titles can't be shared.

smonero:
marking this as draft as I think it requires more thought

smonero:
figured it out - had to check the loading flag

smonero:
putting this back in draft until I add some more gifs explaining the choices behind the padding / spacing / alignment

smonero:
updated the padding and screenshots

frontend/workflows/projectSelector/src/project-links.tsx
```diff
@@ -0,0 +1,160 @@
...
+    })}
+  </div>
+);
+
+const StyledFlexEnd = styled.div({
+  justifyContent: "right",
+  flexDirection: "row-reverse",
+});
+
+const QuickLinksPopper = ({ linkGroups, anchorRef, open, setOpen, setKeyWithQLinksOpen }) => {
```
jdslaugh, line 116: Can we break out these props into an interface?

frontend/workflows/projectSelector/src/project-links.tsx
```diff
@@ -0,0 +1,160 @@
...
+        setOpen(false);
+        setKeyWithQLinksOpen("");
+      }}
+    >
+      <ExpandedLinks linkGroups={linkGroups} />
+    </Popper>
+  );
+};
+
+const ProjectLinks = ({ linkGroups, name }) => {
```
jdslaugh, line 131: Same about the props/interface

frontend/workflows/projectSelector/src/project-links.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const [validLinks, setValidLinks] = React.useState<IClutch.core.project.v1.ILink[]>([]);
+
+  React.useEffect(() => {
+    if (links) {
+      setValidLinks(links.filter(link => link?.url && link.url.length > 0));
+    }
+  }, [links]);
+  // In the case where there is only a singe link in the group, we make the title clickable.
+  // In the case where there are multiple links, the title is not clickable and has different styling.
+  return links.length === 1 ? (
```
jdslaugh, line 54: Should we be checking `validLinks` instead of links here?

jdslaugh: Also if so, will need to update the rest of the references below

frontend/workflows/projectSelector/src/project-links.tsx
```diff
@@ -0,0 +1,160 @@
...
+          <img width={ICON_SIZE} height={ICON_SIZE} src={linkGroupImage} alt={linkGroupName} />
+        </div>
+        <StyledLinkTitle>{linkGroupName}</StyledLinkTitle>
+      </div>
+      <div>
+        {validLinks.map(link => {
+          return (
+            link?.url && (
+              <React.Fragment key={link.url}>
+                <StyledMenuItem
```
jdslaugh, line 79: Can we extend the StyledMenuItem to take an optional prop for these styles? Would be nice if we could reduce some of the inline styles where possible.

jdslaugh (commented):
Looks really good! Some small nits but nothing major.

PR #2196, opened 2022-04-18
metrics api

smonero:
### GitHub Issue

Fixes #


api/metrics/v1/metrics.proto
```diff
@@ -0,0 +1,48 @@
...
+option go_package = "github.com/lyft/clutch/backend/api/metrics/v1;metricsv1";
+
+import "validate/validate.proto";
+import "api/v1/annotations.proto";
+import "google/api/annotations.proto";
+
+service MetricsAPI {
+  rpc GetMetrics(GetMetricsRequest) returns (GetMetricsResponse) {
+    option (google.api.http) = {
+      post : "/v1/metrics/getMetrics"
```
miguelmolina95, line 14: is there a reason we choose `getMetrics` over query range https://prometheus.io/docs/prometheus/latest/querying/api/#range-queries ?

miguelmolina95: oh I see the mockups now so `getMetrics` makes sense to show the CB values

miguelmolina95: actually I do not see this one in https://prometheus.io/docs/prometheus/latest/querying/api/

api/metrics/v1/metrics.proto
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+message Query {
+  // Prometheus expression query string
+  string query = 1 [ (validate.rules).string.min_len = 1 ];
+  // inclusive
+  int64 start_time_ms = 2;
+  // inclusive
+  int64 end_time_ms = 3;
```
mikecutalo, line 27: Are these defaulted? Or should the API require these fields?

api/metrics/v1/metrics.proto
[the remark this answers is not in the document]
miguelmolina95: second one I like better as I think having the word `get` in a `POST` method can be misleading a bit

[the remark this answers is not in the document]
mikecutalo: Every clutch api is a POST because it's over the GRPC gateway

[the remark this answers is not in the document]
mikecutalo: I do like the second one tho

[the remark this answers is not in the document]
miguelmolina95: > Every clutch api is a POST because it's over the GRPC gateway

TIL

but yeah I still second one is better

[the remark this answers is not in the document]
dschaller: I like the second one better but to bikeshed a bit more, is it confusing that the endpoint is called query and the body has a query? I think if we go with that we should change query to expression or something along those lines

api/metrics/v1/metrics.proto
```diff
@@ -0,0 +1,48 @@
...
+  // time step that starts from, a nice default is 1 minute (60000)
+  int64 step_ms = 4;
```
dschaller, line 29: Same question mike had above, should this just default to 1 minute?

api/metrics/v1/metrics.proto
```diff
@@ -0,0 +1,48 @@
...
+}
+
+message GetMetricsRequest {
+  repeated Query metric_queries = 1 [ (validate.rules).repeated = {min_items : 1} ];
+}
+
+message Metric {
+  int64 value = 1;
```
dschaller, line 37: Is this value always an int? I would imagine this could be a float

api/metrics/v1/metrics.proto
```diff
@@ -0,0 +1,48 @@
...
+  int64 timestamp_ms = 2;
+}
+
+message Metrics {
+  repeated Metric metrics = 1;
+}
+
+message GetMetricsResponse {
+  // Contains a mapping of individual queries to their respective results.
+  map<string, Metrics> querystring_to_metrics = 1;
```
dschaller, line 47: nit:
```suggestion
  map<string, Metrics> query_results = 1;
```

api/metrics/v1/metrics.proto
smonero: I'm not sure what you mean here. Proto3 does not allow default values, unless I'm missing something?

api/metrics/v1/metrics.proto
```diff
@@ -0,0 +1,49 @@
...
+    };
+    option (clutch.api.v1.action).type = READ;
+  }
+}
+
+message Query {
+  // Prometheus expression query string
+  string expression = 1 [ (validate.rules).string.min_len = 1 ];
+  // inclusive
+  int64 start_time_ms = 2 [ (validate.rules).int64.gte = 0 ];
```
smonero, line 25: When I looked at the protoc gen validate stuff, I didn't see a good way to make these required. For example, there is a case where people might want to get values before 1970, which would require the value to be negative. I could wrap the value in a message and make the message required but not sure if that was the way to go.

api/metrics/v1/metrics.proto
miguelmolina95: yeah I think the only workaround is making it an optional value but even then I think this reads easier

api/metrics/v1/metrics.proto
miguelmolina95: I think this is okay and not sure how data pre 1970 could be useful for clutch use cases but maybe I am not familiar with many

smonero: lol well yeah its probably not useful for clutch cases :). I guess I was just thinking about how there could potentially be a universe where people wanted something like that. But yeah I think you are right I don't think we need to worry about that.

dschaller (approved):

PR #2197, opened 2022-04-18
FE: textfield: allow onReturn to be used when onChange happens in Autocomplete

smonero:
### Description

This allows people that use autocomplete to pass in their submit function as onReturn, so that when a user clicks on an option or presses enter when selecting an option the submit function will be called. This is inline with how most autocompletes work in other projects / sites.

Note that we check if `v` is valid because we don't want to call it when `v` is null, which would mean the field is empty (as in, a user just deleted their input).

Also note that though this does not work for the project selector or the project catalog, but it does work for the dashboard case.

### Testing Performed

local
### GitHub Issue

Fixes #


frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -290,6 +290,11 @@ const TextField = ({
             {...props}
           />
         )}
+        onChange={(_e, v) => {
+          if (v && onReturn) {
+            onReturn();
```
jdslaugh, line 295: Wonder if we should modify onReturn to send back the value, might allow us to modify catalog/selector to handle the case at that point.

When you said it doesn't work for catalog/selector did that mean it breaks them or they still retain the same functionality and just not the extended functionality?

dschaller: Yeah I'm curious why this only works for some of the use cases

smonero: it means they still retain the same functionality, just not the extended functionality

smonero: @dschaller the reason why those are different is because they set the `value` prop. The resolver and k8s dashboard do not set this prop which is why they work. They also use `formRegistration` but I don't think that matters. All the other props are the same between the 4 cases (resolver and k8s dashboard work, proj catalog and proj selector do not work).

dschaller: Ok I think this is fine for a partial fix but can we add a TODO to come back and investigate the other use cases

dschaller (approved):

PR #2203, opened 2022-04-21
backend: align with Lyft golangci-lint baseline

aneeshusa:
### Description

`goimports` is not a strict superset of `gofmt`
as gofmt additionally performs simplification,
so Lyft's (internal) golangci-lint baseline has both linters enabled.

You can see the additional changes from adding `gofmt`
made automatically via `make backend-lint-fix` here.

None

### Testing Performed

`make backend-lint; echo $?`

### GitHub Issue

Fixes #

### DIFF
--- a/backend/.golangci.yml
+++ b/backend/.golangci.yml
@@ -7,6 +7,7 @@ linters:
   - exportloopref
   - gci
   - gocritic
+  - gofmt
   - goimports
   - gosec
   - gosimple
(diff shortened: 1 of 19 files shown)


dschaller (commented):
Thanks @aneeshusa! Can you elaborate more on the reasoning behind this change?

danielhochman (approved):

miguelmolina95 (approved):

aneeshusa:
@dschaller the impetus is that all of Lyft's other (internal) Go repos already have this baseline set up, so wanted to update Clutch as well, starting with OSS and then our internal repo. `gofmt` is the standard autoformatter for Go and we want to have consistency across our repos at Lyft.

PR #2204, opened 2022-04-21
proxy module: add special GET endpoint

danielhochman:
### Description
Read only roles are common for RBAC policies. However the existing proxy endpoint assumed that all requests were of action type `CREATE`. Here we add a separate GET endpoint that has action type READ for use with read-only roles. The client must elect to use the GET method for it to work.

I could see a future enhancement that blocks certain GET requests because the API developer had GET requests that modified a resource, but for now we will assume that APIs use non-GET requests for requests that modify resources, especially since it's susceptible to CSRF attacks.

The alternative implementation would be to treat the proxy request as special in the middleware and automatically infer the action from the embedded HTTP method in the request.

### Testing Performed
Unit

### TODOs
- [ ] Unit test


backend/module/proxy/proxy.go
```diff
@@ -171,6 +171,36 @@ func (m *mod) RequestProxy(ctx context.Context, req *proxyv1.RequestProxyRequest
 	return proxyResponse, nil
 }
 
+func (m *mod) RequestProxyGet(ctx context.Context, req *proxyv1.RequestProxyGetRequest) (*proxyv1.RequestProxyGetResponse, error) {
+	// Validate that it's a GET.
```
mikecutalo, line 175: You need to validate the request permitted based on the configuration like so

https://github.com/lyft/clutch/blob/65ec44815292479e0deeab59fe2febf9c282e6e6/backend/module/proxy/proxy.go#L72-L94

mikecutalo: I see this is a proxy for the proxy call :)

mikecutalo (approved):

miguelmolina95 (approved):
TIL

api/proxy/v1/proxy.proto
```diff
@@ -42,3 +52,25 @@ message RequestProxyResponse {
...
+
+message RequestProxyGetRequest {
+  // The name of a service that is configured
+  string service = 1 [ (validate.rules).string.min_len = 1 ];
+  // The HTTP method that will be used for the request
+  string http_method = 2 [ (validate.rules).string.min_len = 1 ];
+  // The URI path to call
+  string path = 3 [ (validate.rules).string.min_len = 1 ];
+  // The request body
+  google.protobuf.Value request = 4 [ (clutch.api.v1.log) = false ];
```
dschaller, line 64: Do we need this? A get request shouldn't support a body

PR #2206, opened 2022-04-23
FE: add support for location notes in delete pod

smonero:
### Description

It would be nice to support notes in different locations for the delete pod flow. This PR adds support for prop-drilling the notes from the config down to the delete pod workflow. Note that the delete pod workflow does not work in local dev because of the way k8s clusters are set up.

There are generally 2 ways that users get to the delete pod workflow - 1. is through the k8s dashboard, and 2. is through the normal sidebar. The link from the k8s dashboard will jump to the second step in the wizard. The notes can potentially show up on the first step and second steps of the wizard, depending on their `location` field. Thus, no matter which way the users are using the delete pod workflow, they will see the notes.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/k8s/src/delete-pod.tsx
+++ b/frontend/workflows/k8s/src/delete-pod.tsx
@@ -23,7 +23,6 @@ const PodIdentifier: React.FC<ResolverChild> = ({ resolverType, notes = [] }) =>
   const { onSubmit } = useWizardContext();
   const resolvedResourceData = useDataLayout("resourceData");
   const resolverInput = useDataLayout("resolverInput");
-  const resolverNotes = notes.filter(note => note.location === "resolver");
   const onResolve = ({ results, input }) => {
     // Decide how to process results.
     resolvedResourceData.assign(results[0]);
@@ -32,7 +31,7 @@ const PodIdentifier: React.FC<ResolverChild> = ({ resolverType, notes = [] }) =>
   };
 
   return (
-    <Resolver type={resolverType} searchLimit={1} onResolve={onResolve} notes={resolverNotes} />
+    <Resolver type={resolverType} searchLimit={1} onResolve={onResolve} notes={notes} />
   );
 };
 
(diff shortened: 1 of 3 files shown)


frontend/workflows/k8s/src/delete-pod.tsx
```diff
@@ -8,40 +8,44 @@ import {
...
 
-import type { ConfirmChild, ResolverChild, WorkflowProps } from ".";
+import type { ConfirmChild, ResolverChild, VerifyChild, WorkflowProps } from ".";
 
 const PodIdentifier: React.FC<ResolverChild> = ({ resolverType, notes = [] }) => {
   const { onSubmit } = useWizardContext();
   const resolvedResourceData = useDataLayout("resourceData");
   const resolverInput = useDataLayout("resolverInput");
-
+  const resolverNotes = notes.filter(note => note.location === "resolver");
```
dschaller, line 26: This pattern is probably ok for now but I think this comes back to the earlier thread around using context for complex components like the resolver.

However, this will break existing functionality for anyone using this workflow that doesn't specify a location.

smonero: Do you think it is ok if we show the new notes in both places? Like on the resolver we show both notes, then only the new ones at the next step. I think that could be a good fix and still keep backwards functionality.

miguelmolina95 (approved):

frontend/workflows/k8s/src/delete-pod.tsx
```diff
@@ -34,14 +33,16 @@ const PodIdentifier: React.FC<ResolverChild> = ({ resolverType, notes = [] }) =>
...
+const PodDetails: React.FC<VerifyChild> = ({ notes = [] }) => {
   const { onSubmit, onBack } = useWizardContext();
   const resourceData = useDataLayout("resourceData");
   const instance = resourceData.displayValue() as IClutch.k8s.v1.Pod;
+  const locationNotes = notes.filter(note => note.location === "verify");
 
   return (
     <WizardStep error={resourceData.error} isLoading={resourceData.isLoading}>
       <strong>Pod Details</strong>
+      <NotePanel notes={locationNotes} />
```
jdslaugh, line 45: Should this be a conditional render if there are no notes?

dschaller: I think it'll just render a grid with no children so to the end user it's not a huge change.

dschaller (approved):

PR #2215, opened 2022-04-27
API: change metric tags from list of strings to map of string string

smonero:
### GitHub Issue

Fixes #


api/metrics/v1/metrics.proto
```diff
@@ -36,7 +36,8 @@ message GetMetricsRequest {
 message Metric {
   double value = 1;
   int64 timestamp_ms = 2;
-  repeated string metric_tags = 3;
+  /** metric labels (tags) have a key and a value, https://prometheus.io/docs/concepts/data_model/#notation */
```
mikecutalo, line 39: Idk if this is the best doc to link, as the metric API here could be aggregating from maybe backends beyond just prom

mikecutalo (approved):

dschaller (approved):

miguelmolina95 (approved):

PR #2218, opened 2022-04-28
FE: proj catalog: add button for configuration and overflow for qlinks

smonero:
### Description

In the case where there are more than 5 links, there is an overflow popout now. There is also a button that will allow for configuration.

### Testing Performed

local
### GitHub Issue

Fixes #
th

### TODOs

- [x] possibly remove some of the redundant styling

### DIFF
--- a/frontend/workflows/projectCatalog/src/details/quick-links.tsx
+++ b/frontend/workflows/projectCatalog/src/details/quick-links.tsx
@@ -38,7 +38,7 @@ const QuickLinkContainer = ({ key, name, children }: QuickLinkContainerProps) =>
   );
 
   return (
-    <Grid item key={key ?? ""}>
+    <Grid item key={key ?? ""} style={{padding: "8px"}}>
       {name ? container : children}
     </Grid>
   );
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -130,8 +131,48 @@ const QuickLinksCard = ({ linkGroups }: QuickLinksProps) => (
...
+        item
+        direction="row"
+        alignItems="center"
+        spacing={1}
+        style={{ padding: "10px" }}
+      >
+        <SlicedLinkGroup slicedLinkGroups={firstFive} />
+        {overflow.length > 0 && (
+          <>
+            <IconButton size="small" ref={anchorRef} onClick={() => setOpen(true)}>
```
jdslaugh, line 159: We have a neutral variant for the IconButton, can we try that here?

jdslaugh (commented):
I like it. 

Few small comments, mostly can we get rid of the blue background on the overflow icon. Also can we add a bit of spacing in between the icons? They seem a bit cluttered. Shouldn't need to be much but it looks like the design might have around 8px between.

And then we spoke of this earlier, but getting rid of the spacing on the right where the quick links used to reside.

frontend/workflows/projectCatalog/src/details/quick-links.tsx
smonero: oh yeah oops I had forgotten you added that a while back. thanks yeah that works

jdslaugh (approved):
Thanks!

miguelmolina95 (approved):

PR #2219, opened 2022-04-29
frontend: ProjectCatalog - Config Page Skeleton

jdslaugh:
### Description
- First pass at a skeleton for Project Configuration page
- Will build the page based off of children in an extended component
- Modified header to build usable breadcrumbs

- Still some bits to clean up and want to rework the routing/breadcrumbs as well as the page error handler

Modified the routing to be explicit as it seems that react-router in v6 removes regex routing which would allow for optional parameters. 

https://reactrouter.com/docs/en/v6/faq#what-happened-to-regexp-routes-paths
https://github.com/remix-run/react-router/issues/8381

#### Screenshots
Screen Shot 2022-04-29 at 3 08 18 PM

### Testing Performed
manual

PR #2220, opened 2022-04-30
FE: fix alignment of icons and add feature flag for settings button

smonero:
### Description

Since settings button isn't working yet we can gate it using a feature flag. We also use flex to put the icons at the same height.
before:
Screen Shot 2022-04-29 at 10 55 19 PM

after:

Screen Shot 2022-04-29 at 10 53 16 PM

edit: might not need the feature flag because https://github.com/lyft/clutch/pull/2219 will probably be merged in soon

Also fixed the text to be `noWrap` to prevent wrapping of long links, and set spacing to 1 for that grid in the corner.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/projectCatalog/src/details/index.tsx
+++ b/frontend/workflows/projectCatalog/src/details/index.tsx
@@ -48,7 +48,7 @@ const QuickLinksAndSettingsBtn = ({ linkGroups }) => {
         <Grid item>
           <QuickLinksCard linkGroups={linkGroups} />
         </Grid>
-        <SimpleFeatureFlag feature="projectSettings">
+        <SimpleFeatureFlag feature="projectCatalogSettings">
           <Grid item>
             <IconButton onClick={() => {}}>
               <SettingsIcon />
--- a/frontend/workflows/projectCatalog/src/details/index.tsx
+++ b/frontend/workflows/projectCatalog/src/details/index.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { useParams } from "react-router-dom";
 import type { clutch as IClutch } from "@clutch-sh/api";
-import { client, Grid, IconButton, styled, Tooltip } from "@clutch-sh/core";
+import { client, Grid, IconButton, SimpleFeatureFlag, styled, Tooltip } from "@clutch-sh/core";
 import { faLock } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import GroupIcon from "@material-ui/icons/Group";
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -39,11 +39,13 @@ const QuickLinksAndSettingsBtn = ({ linkGroups }) => {
         <Grid item>
           <QuickLinksCard linkGroups={linkGroups} />
         </Grid>
-        <Grid item style={{ padding: "10px" }}>
-          <IconButton onClick={() => {}}>
-            <SettingsIcon />
-          </IconButton>
-        </Grid>
+        <SimpleFeatureFlag feature="projectSettings">
```
jdslaugh, line 42: Can we rename this to projectCatalogSettings? That way its inline with the projectCatalog feature flag

jdslaugh (approved):

frontend/workflows/projectCatalog/src/details/index.tsx
```diff
@@ -35,15 +35,26 @@ const DisabledItem = ({ name }: { name: string }) => (
...
     <>
-      <Grid container direction="row" style={{ padding: "8px", justifyContent: "flex-end" }}>
+      <Grid
+        container
+        direction="row"
+        style={{
+          padding: "8px",
+          justifyContent: "flex-end",
+          display: "flex",
+          alignItems: "center",
```
dschaller, line 45: Grid should provide props for justify and align and flex should be set automatically.

smonero: when I remove them I get this issue (even with adding justify="flex-end" for the grid component

smonero: oh nvm I get it. I'll fix it now

frontend/workflows/projectCatalog/src/details/quick-links.tsx
```diff
@@ -94,15 +94,17 @@ const QuickLinkGroup = ({ linkGroupName, linkGroupImage, links }: QuickLinkGroup
...
       >
         {validLinks.map(link => (
           <PopperItem key={link.name}>
             {link?.url && (
               <Link href={link.url}>
-                <Typography color="inherit" variant="body4">
-                  {link.name}
-                </Typography>
+                <span style={{ whiteSpace: "nowrap" }}>
+                  <Typography color="inherit" variant="body4">
```
dschaller, line 104: Do we want to remove the color from Link here?

dschaller (approved):

PR #2222, opened 2022-05-01
FE: allow for callback to set value in Tabs component

smonero:
### Description

It is convenient to allow for a callback function to communicate back the tab index to a parent.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/tab.tsx
+++ b/frontend/packages/core/src/tab.tsx
@@ -81,12 +81,16 @@ const TabPanel = styled(MuiTabPanel)({
 
 export interface TabsProps extends Pick<MuiTabsProps, "value" | "variant"> {
   children: React.ReactElement<TabProps> | React.ReactElement<TabProps>[];
+  setIndex?: (index: string) => void;
 }
 
-export const Tabs = ({ children, value, variant }: TabsProps) => {
+export const Tabs = ({ children, value, variant, setIndex }: TabsProps) => {
   const [selectedIndex, setSelectedIndex] = React.useState((value || 0).toString());
   const onChangeMiddleware = (_, v: string) => {
     setSelectedIndex(v);
+    if (setIndex) {
+      setIndex(v);
+    }
   };
 
   return (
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/tab.tsx
```diff
@@ -81,12 +81,16 @@ const TabPanel = styled(MuiTabPanel)({
 
 export interface TabsProps extends Pick<MuiTabsProps, "value" | "variant"> {
   children: React.ReactElement<TabProps> | React.ReactElement<TabProps>[];
+  setIndex?: (index: string) => void;
```
dschaller, line 84: nit: reading this implies that it's triggering behavior instead of on an action (e.g. `onChange`) and index implies this should be an integer. I know we use the word index internally to this component but it should probably be value.

frontend/packages/core/src/tab.tsx
```diff
@@ -81,12 +81,16 @@ const TabPanel = styled(MuiTabPanel)({
 
 export interface TabsProps extends Pick<MuiTabsProps, "value" | "variant"> {
   children: React.ReactElement<TabProps> | React.ReactElement<TabProps>[];
+  setValueCallback?: (index: string) => void;
```
jdslaugh, line 84: Should `index` be a string here? I would assume it should be a `number`

jdslaugh (approved):

frontend/packages/core/src/tab.tsx
```diff
@@ -81,12 +81,18 @@ const TabPanel = styled(MuiTabPanel)({
 
 export interface TabsProps extends Pick<MuiTabsProps, "value" | "variant"> {
   children: React.ReactElement<TabProps> | React.ReactElement<TabProps>[];
+  // To allow for callback functionality, use `onChange`
+  // Note the value is referring to the index, like "0", "1", "2", etc.
+  onChange?: (value: string) => void;
```
dschaller, line 86: Actually I'm a bit confused as to why this is a string to begin with. The [docs](https://mui.com/material-ui/api/tabs/) are a bit unclear since it's an any. Should this be an int?
```
value: We default to the index of the child (number)
```

smonero: I think the docs are wrong. If you set it as an any and log the type when it executes it comes out as a string

smonero: Screen Shot 2022-05-03 at 2 53 26 PM
In other places though it says its a string

dschaller (approved):

PR #2226, opened 2022-05-03
frontend: ProjectCatalog Config params patch

jdslaugh:
### Description
- Making `onError` optional since it is currently set during `cloneElement`
- Making `defaultRoute` optional
- Adjusted onError handler and will check if the user has already defined one and call that one before setting the error

### Testing Performed
manual

### DIFF
--- a/frontend/workflows/projectCatalog/src/index.tsx
+++ b/frontend/workflows/projectCatalog/src/index.tsx
@@ -28,7 +28,7 @@ export interface ProjectDetailsWorkflowProps extends WorkflowProps {
 
 export interface ProjectDetailsConfigWorkflowProps extends WorkflowProps {
   children?: ProjectConfigPage | ProjectConfigPage[];
-  defaultRoute: string;
+  defaultRoute?: string;
 }
 
 const register = (): WorkflowConfiguration => {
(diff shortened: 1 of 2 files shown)


frontend/workflows/projectCatalog/src/index.tsx
```diff
@@ -12,7 +12,7 @@ type DetailCard = CatalogDetailsCard | typeof DynamicCard | typeof MetaCard;
 export interface ProjectConfigProps {
   title: string;
   path: string;
-  onError: (error: ClutchError) => void;
+  onError?: (error: ClutchError) => void;
```
dschaller, line 15: Ah yeah, we might want to have an internal version of our props for this purpose since even if users set an onError prop we would overwrite it with out own callback

jdslaugh: Thinking...should we even worry about a top level error or keep it confined into the individualized components? Then it can be handled as they see fit. If not I can just remove this logic entirely.

dschaller: I think having consistent error handling is nice but I see the added complexity here. If each component / page were to define it's own error handler we might end up with inconsistencies in the experience between tabs

jdslaugh: Alright, I modified it to keep the same props but will check if the child has already set an `onError` handler and call that before setting the error

dschaller (approved):

smonero (approved):

PR #2228, opened 2022-05-04
frontend: add package scope support to workflow registrar

dschaller:
### Description

Add support for package scopes to the workflow registrar. This allows for packages with conflicting names (@a/bar and @b/bar) to both be registered via the Clutch config utilizing the package scope for uniqueness.

### Testing Performed

manual


smonero (approved):
should we update some docs to call this out somewhere?

miguelmolina95 (approved):
good idea on docs

dschaller:
>should we update some docs to call this out somewhere?
This is purely an under the hood change which aligns with the [existing documentation](https://clutch.sh/docs/configuration/#frontend)

PR #2230, opened 2022-05-06
FE: add a test to Tabs component

smonero:
### Description

Another test, this one for the Tabs component. This part of an wider effort to add more FE tests to Clutch.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/tests/tab.test.tsx
+++ b/frontend/packages/core/src/tests/tab.test.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import { Tab, Tabs } from "../tab";
+
+describe("Tabs component aka TabGroup", () => {
+  describe("basic rendering", () => {
+    let component;
+    beforeEach(() => {
+      component = shallow(
+        <Tabs value={1}>
+          <Tab label="meow" />
+          <Tab label="mix" />
+        </Tabs>
+      );
+    });
+
+    it("renders", () => {
+      expect(component.find(Tabs)).toBeDefined();
+    });
+
+    it("renders 2 Tabs", () => {
+      expect(component.find(Tab)).toHaveLength(2);
+    });
+
+    it("has the second tab selected (index 1)", () => {
+      expect(component.find("TabContext").prop("value")).toBe("1");
+    });
+  });
+});
--- a/frontend/packages/core/src/tests/tab.test.tsx
+++ b/frontend/packages/core/src/tests/tab.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { shallow } from "enzyme";
+import { mount, shallow } from "enzyme";
 
 import { Tab, Tabs } from "../tab";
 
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/tests/tab.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { shallow } from "enzyme";
+
+import { Tab, Tabs } from "../tab";
+
+describe("Tabs component aka TabGroup", () => {
```
dschaller, line 6: I think it's ok to just call this Tabs component

frontend/packages/core/src/tests/tab.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+  describe("basic rendering", () => {
```
dschaller, line 7: nit:
```suggestion
  describe("with a value set", () => {
```

frontend/packages/core/src/tests/tab.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+          <Tab label="mix" />
+        </Tabs>
+      );
+    });
+
+    it("renders", () => {
+      expect(component.find(Tabs)).toBeDefined();
+    });
+
+    it("renders 2 Tabs", () => {
```
dschaller, line 22:
```suggestion
    it("renders children tabs", () => {
```

frontend/packages/core/src/tests/tab.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+      expect(component.find(Tab)).toHaveLength(2);
+    });
+
+    it("has the second tab selected (index 1)", () => {
```
dschaller, line 26:
```suggestion
    it("displays the tab from the specified value", () => {
```

frontend/packages/core/src/tests/tab.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+      expect(component.find("TabContext").prop("value")).toBe("1");
```
dschaller, line 27: Should we be expecting the mix tab to be selected and displayed?

smonero: yes

smonero: Are you saying I should add another test for that? I remember Josh suggested that we only have 1 `expect` per test to keep things organized

dschaller: Yeah I would add a test that it's displayed as well

smonero: here is the debug() output for the component:
```
<div style={{...}}>
@clutch-sh/core:       <TabContext value="1">
@clutch-sh/core:         <Styled(Component) variant={[undefined]} onChange={[Function: onChangeMiddleware]}>
@clutch-sh/core:           <Tab label="meow" value="0" />
@clutch-sh/core:           <Tab label="mix" value="1" />
@clutch-sh/core:         </Styled(Component)>
@clutch-sh/core:         <Styled(WithStyles(ForwardRef(TabPanel))) value="0" />
@clutch-sh/core:         <Styled(WithStyles(ForwardRef(TabPanel))) value="1" />
@clutch-sh/core:       </TabContext>
@clutch-sh/core:     </div>
```
as you can see it doesn't give an indication that "mix" tab is selected other than the tabContext . An alternative thing I could do is maybe use the react-test renderer like Josh does in the NPS thing instead. I'll ask Josh today about it.

jdslaugh (approved):

PR #2231, opened 2022-05-07
FE: assets: upstream all icons

smonero:
### Description

In an attempt to make Clutch more friendly to contribute to, we upstream the icons so that other people can have access to them. This also allows us to centralize in one location so when we reuse them we don't have to duplicate, as they are currently.

Screen Shot 2022-05-13 at 11 43 16 AM

### Testing Performed

local, storybook
### GitHub Issue

Fixes #


frontend/packages/core/src/Assets/icons/ExperimentIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react";
+
+import StyledSvg from "./helpers";
+
+const ExperimentIcon = (props: React.SVGProps<SVGSVGElement>) => (
```
dschaller, line 5: We should probably use the `SVGProps` from within Clutch since it will allow you to remove the height and width and rely on the `size` prop instead with predetermined height and widths

dschaller:
Should the diamond icon be bottom aligned to match the other icons?

frontend/packages/core/src/Assets/icons/ExperimentIcon.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react";
```
dschaller, line 1: Can we get Kanchan to provide this without the holes in the beaker?

smonero: yeah I was thinking we should remove all the holes and replace them with white or off-white or whatever - the flame has a hole in the middle, the rocket ship has a hole for the window, the gem has holes as edges, and the beaker has holes as bubbles.

smonero:
> Should the diamond icon be bottom aligned to match the other icons?

we can ask Kanchan to make it taller or whatever. Generally for all the icons originally I just copied and pasted the svgs from figma.

frontend/packages/core/src/Assets/icons/ExperimentIcon.tsx
dschaller: I think it's ok to keep the cutouts for the various icons but I think the beaker already has lines on neck which is fine, the bubbles are just hard to see at smaller sizes

frontend/packages/core/src/Assets/icons/ExperimentIcon.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import type { SVGProps } from "../global";
+
+import StyledSvg from "./helpers";
+
+const ExperimentIcon = (props: SVGProps) => (
+  <StyledSvg
+    width="24"
+    height="24"
```
dschaller, line 10: You can remove the height and width since it's set by the SVGProps `size` prop

smonero: when I do that they get massive. I tried a bunch of things like passing various sizes in ("small", "medium", "large" etc) removing / adding the viewbox, and setting width and height from props and none of it changes. I also tried using `StyledSVG` instead of `StyledSvg` so that I could pass the size in but that doesn't work either.

I see in the `SVGProps` in global.ts it sets the default to be small (aka 24) but its not getting set on the icons. Either I'm doing something wrong or there is some issue with the current SVGProps that makes it not work for this case.
Screen Shot 2022-05-11 at 5 42 19 PM
Screen Shot 2022-05-11 at 5 42 07 PM

dschaller: Ok let me take a deeper look today

jdslaugh: So looking into this, the issue is that `StyledSvg` in the icons is coming from `helpers.tsx` which is not the same as the `StyledSVG` from the `global.ts` file. Modifying that and changing the icon to extend the size it works as expected.

```diff
--- a/frontend/packages/core/src/Assets/icons/ExperimentIcon.tsx
+++ b/frontend/packages/core/src/Assets/icons/ExperimentIcon.tsx
@@ -1,17 +1,14 @@
 import * as React from "react";

 import type { SVGProps } from "../global";
+import { StyledSVG } from "../global";

-import StyledSvg from "./helpers";
-
-const ExperimentIcon = (props: SVGProps) => (
-  <StyledSvg
-    width="24"
-    height="24"
+const ExperimentIcon = ({ size, ...props }: SVGProps) => (
+  <StyledSVG
+    size={size}
     viewBox="0 0 24 24"
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
-    hoverFill="#00809D"
     {...props}
   >
     <path
@@ -20,7 +17,7 @@ const ExperimentIcon = (props: SVGProps) => (
       fillRule="evenodd"
       clipRule="evenodd"
     />
-  </StyledSvg>
+  </StyledSVG>
 );
```

frontend/packages/core/src/Assets/stories/icons.stories.tsx
```diff
@@ -11,13 +11,32 @@ import SlackIcon from "../icons/SlackIcon";
 
 export const AllIcons = () => (
```
jdslaugh, line 12: You can clean this up a little easier so you're not trying to support it all.

```suggestion
export const AllIcons: React.FC<SVGProps> = ({ size }) => (
  <div>
    <FireIcon size={size} />
    <PlusIcon size={size} />
    <ExperimentIcon size={size} />
    <GemIcon size={size} />
    <RocketIcon size={size} />
    <SirenIcon size={size} />
    <SlackIcon size={size} />
  </div>
);

export default {
  title: "Core/Assets/Icons",
  component: AllIcons,
  argTypes: {
    size: {
      options: VARIANTS,
      control: { type: "select" },
    },
  },
} as Meta;
```

dschaller (approved):

jdslaugh (approved):
Looks great, thanks for adding this in!

PR #2232, opened 2022-05-09
k8s: add direct clientset access

mikecutalo:
### Description
Add direct client access for K8s clientsets, which may be needed if you wanted to implement custom logic with the K8s client in a users private gateway.

### Testing Performed
Unit

### DIFF
--- a/backend/mock/service/k8smock/k8smock.go
+++ b/backend/mock/service/k8smock/k8smock.go
@@ -355,6 +355,10 @@ func (*svc) Clientsets(ctx context.Context) ([]string, error) {
 	return []string{"fake-user@fake-cluster"}, nil
 }
 
+func (*svc) GetK8sClientset(clientset string) (k8sservice.ContextClientset, error) {
+	return nil, nil
+}
+
 func New() k8sservice.Service {
 	return &svc{}
 }
--- a/backend/service/k8s/clientset.go
+++ b/backend/service/k8s/clientset.go
@@ -22,6 +22,7 @@ const (
 type ClientsetManager interface {
 	Clientsets(ctx context.Context) (map[string]ContextClientset, error)
 	GetK8sClientset(ctx context.Context, clientset, cluster, namespace string) (ContextClientset, error)
+	GetK8sClientsetSimple(clientset string) (ContextClientset, error)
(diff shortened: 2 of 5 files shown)


smonero (approved):
✅  Correct
✅  Secure
✅  Readable
✅  Elegant

backend/mock/service/k8smock/k8smock.go
```diff
@@ -355,6 +355,10 @@ func (*svc) Clientsets(ctx context.Context) ([]string, error) {
 	return []string{"fake-user@fake-cluster"}, nil
 }
 
+func (*svc) GetK8sClientset(clientset string) (k8sservice.ContextClientset, error) {
```
dschaller, line 358: Was this just missing from a previous commit?

backend/service/k8s/clientset.go
```diff
@@ -142,6 +143,14 @@ func (m *managerImpl) Clientsets(ctx context.Context) (map[string]ContextClients
 	return ret, nil
 }
 
+func (m *managerImpl) GetK8sClientsetSimple(clientset string) (ContextClientset, error) {
```
dschaller, line 146: It seems like this should be `GetK8sClientset` and the function below should be `GetK8sClientsetWithFallback` but I know it would be a lot of updates to make. Maybe we can add some docstrings to denote the differences to consumers?

backend/mock/service/k8smock/k8smock.go
mikecutalo: Let me double check this actually

backend/service/k8s/clientset.go
mikecutalo: Yeah I'm not in favor of breaking that API at the moment, but ill add some comments to clarify its usage.

backend/mock/service/k8smock/k8smock.go
mikecutalo: Oh yeah sorry its been a few days since i looked at this code, this is the new public accessor method that i added to the k8s service interface, previously it was not possible to get at access a clientset before this pr.

dschaller (approved):

PR #2235, opened 2022-05-12
frontend: ShortLinks Patch for ProjectSelector

jdslaugh:
### Description
Ran into an error when an error is returned via onError it wasn't being cleared. So if you successfully make a change afterwards you would get the same initial error. This PR modifies the logic so after calling `onError` it will immediately dispatch a change to clear the error from the state.

### Testing Performed
manual

To test, search for a deprecated project (which will generate a toast), let it clear. Then look up anything else. You'll get the same initial error as the first one.

### DIFF
--- a/frontend/workflows/projectSelector/src/selector-reducer.tsx
+++ b/frontend/workflows/projectSelector/src/selector-reducer.tsx
@@ -279,7 +279,7 @@ const selectorReducer = (state: State, action: Action): State => {
         error: action?.payload?.result,
       };
     }
-    case "CLEAR_ERRORS": {
+    case "CLEAR_PROJECT_ERRORS": {
       return { ...state, projectErrors: [] };
     }
     default:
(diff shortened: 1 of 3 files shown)


smonero (approved):

frontend/workflows/projectSelector/src/selector-reducer.tsx
```diff
@@ -280,6 +279,9 @@ const selectorReducer = (state: State, action: Action): State => {
         error: action?.payload?.result,
       };
     }
+    case "CLEAR_ERRORS": {
```
dschaller, line 282: It's a bit unclear by the name what errors are being cleared (e.g. it looks like there are project errors vs a global error)

jdslaugh: Good call, renamed this event to `CLEAR_PROJECT_ERRORS` to make it clear what is happening.

dschaller (approved):

PR #2236, opened 2022-05-15
FE: input: toggle button group

smonero:
### Description

A toggle button group component that will be useful.

I know in the figma that the words are not all caps, but I decided to do that because that's the default with material so I thought the default was fine.

demo1

### Testing Performed

storybook

### GitHub Issue

Fixes #


frontend/packages/core/src/Input/stories/toggle-button-group.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+import type { Meta } from "@storybook/react";
+
+import ToggleButtonGroup from "../toggle-button-group";
+
+export default {
+  title: "Core/Input/ToggleButtonGroup",
+  component: ToggleButtonGroup,
+} as Meta;
+
+export const ExclusiveDemo = () => {
```
katpioro, line 11: whats the difference between a stories file and a regular file?

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,39 @@
...
+      color: "#3548D4",
+      backgroundColor: "#3548D41F",
+    },
+  },
+  padding: "6px 16px 16px 6px",
+});
+
+// TODO(smonero): add some tests
+const ToggleButtonGroup = ({
+  exclusive = true,
```
katpioro, line 27: what does exclusive represent?

smonero: the exclusive represents whether or not there can be only one option selected at a time. If exclusive is false, you can select multiple options simultaneously (ex. think about editing text, where you can select both bold and italic at the same time). If exclusive is true, user is forced to only select one

frontend/packages/core/src/Input/stories/toggle-button-group.stories.tsx
smonero: when the files have `.stories` in the name storybook picks it up and displays the exports in the page

frontend/packages/core/src/Input/toggle-button-group.tsx
jdslaugh: I think it might be a little more intuitive to use something like `multiple?: boolean` which should default to false. That way it works as expected for the base case and if you need multi-select you could optionally set it.

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const StyledMuiToggleButtonGroup = styled(MuiToggleButtonGroup)({
+  ".MuiToggleButton-root": {
+    "&.Mui-selected": {
+      color: "#3548D4",
+      backgroundColor: "#3548D41F",
+    },
+  },
+  padding: "6px 16px 16px 6px",
+});
```
jdslaugh, line 23: Could we adjust the CSS to add some inset properties? To make it look a bit more like the button is actually pressed vs just selected. Unless this is based off of Figma already.

smonero: The figma doesn't seem to have an inset, but here's what it looks like with 1px all around. do you think that works? or maybe it should just be vertical inset not horizontal?
inset1
?

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+import {
+  ToggleButton as MuiToggleButton,
+  ToggleButtonGroup as MuiToggleButtonGroup,
+} from "@material-ui/lab";
+
+export interface ToggleButtonGroupProps {
+  /** A boolean to decide if the user can select multiple options
+   * simultaneously. If true, users will be able to select multiple options,
+   * whereas if false, an exclusive choice will be enforced.
+   */
```
dschaller, line 12: nit:
```suggestion
  /** If true, multiple children options can be selected simultaneously.*/
```

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+  multiple?: boolean;
+  /** The current value that is selected in the group. Usually this will be
+   * a useState value.
+   */
```
dschaller, line 16:
```suggestion
  multiple?: boolean;
  /** The current value that is selected in the group.*/
```

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+  currentValue: string;
+  /** The onChange function that gets called when a user changes selection */
+  onChange: (event: React.ChangeEvent<{}>, value: string) => void;
```
dschaller, line 19: can you pick this from the material ui toggle button group props?

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+  /** An array of strings where each string will be one toggle button */
+  toggleButtonValues: string[];
```
dschaller, line 21: Should this just be a children prop of a specific react element type? That would probably give users the most flexibility if they wanted to say render an icon instead of text.

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+export interface ToggleButtonGroupProps {
+  /** A boolean to decide if the user can select multiple options
+   * simultaneously. If true, users will be able to select multiple options,
+   * whereas if false, an exclusive choice will be enforced.
+   */
+  multiple?: boolean;
+  /** The current value that is selected in the group. Usually this will be
+   * a useState value.
+   */
+  currentValue: string;
```
dschaller, line 17: can you pick this from the material ui toggle button group `value` prop instead of using a custom `currentValue` prop?

frontend/packages/core/src/Input/stories/toggle-button-group.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [value, setValue] = React.useState("MEOW");
+  const onChange = (_event: React.ChangeEvent<{}>, newValue: string) => {
```
dschaller, line 13: nit: 
```suggestion
  const onChange = (_: React.ChangeEvent<{}>, newValue: string) => {
```

frontend/packages/core/src/Input/stories/toggle-button-group.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+    // Note that we check for null because we want to enforce that
+    // at least value is active
+    if (newValue !== null) {
+      setValue(newValue);
+    }
+  };
```
dschaller, line 19: This makes me think that in the component, if the value is specified as an invalid option, we should handle null values such that either we leave the existing value(s) or if there is no value and exclusive is true we should default to the first option

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from "react";
+import styled from "@emotion/styled";
+import {
+  ToggleButton as MuiToggleButton,
+  ToggleButtonGroup as MuiToggleButtonGroup,
+} from "@material-ui/lab";
+
+export interface ToggleButtonGroupProps {
```
dschaller, line 8: Should we expose the ToggleButtonGroup disabled and orientation props as well?

frontend/packages/core/src/Input/toggle-button-group.tsx
jdslaugh: Yeah maybe try horizontal, them changing height is a little odd lol. But it definitely looks pressed so I think it's a step in the right direction.

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+  /** The onChange function that gets called when a user changes selection */
+  onChange: (event: React.ChangeEvent<{}>, value: string) => void;
+  /** An array of strings where each string will be one toggle button */
+  toggleButtonValues: string[];
+}
+
+const StyledMuiToggleButtonGroup = styled(MuiToggleButtonGroup)({
+  ".MuiToggleButton-root": {
+    "&.Mui-selected": {
+      color: "#3548D4",
```
dschaller, line 27: I think there is opacity in the designs of 12%

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+      backgroundColor: "#3548D41F",
```
dschaller, line 28: It looks like the designs have this background color as `rgba(53, 72, 212, 0.12)`

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+      inset: "1px 1px 1px 1px",
```
dschaller, line 29: I don't think the material toggle button groups have insets. I saw Josh's comment but maybe we should mirror material or error on the side of following the designs for now

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+    },
+    textTransform: "none",
+  },
+  padding: "6px 16px 16px 6px",
```
dschaller, line 33: trick: you can do something like
```suggestion
  padding: "6px 16px",
```

This will infer the padding for the bottom and left edges

frontend/packages/core/src/Input/stories/toggle-button-group.stories.tsx
smonero: Are we saying that we should iterate over the children checking if there is a possible `value` prop, put those into an object, then check if the key exists? I guess I'm confused because the component doesn't seem to know about its chlidren

smonero: Oh I think I get it. You mean like the `onChangeMiddleware` func

dschaller: Here I'm trying to say that we shouldn't need to do a null value check in the parent components that control the value being sent to `ToggleButtonGroup`. In the `ToggleButtonGroup`, if the `value` prop is null the component should either:
 * if the `multiple` prop is `true`, allow a value of null for the `value` prop
 * if the `multiple` prop is `false`, default to the first child since at least one value must be selected.

frontend/packages/core/src/Input/toggle-button-group.tsx
dschaller: Oh I was looking at V5 docs, we can not include that for now since it doesn't exist yet.

smonero:
hm for some reason children[0] is always undefined even though children seems to have 3 entries in it. Need to investigate more. Can't print the object though because apparently its a circular reference.

frontend/packages/core/src/Input/stories/toggle-button-group.stories.tsx
smonero: hm for some reason children[0] is always undefined even though children seems to have 3 entries in it. Need to investigate more. Can't print the object though because apparently its a circular reference.

smonero: thought more about this and it seems the proposed behavior is unintuitive. If a user clicks a choice to deselect it but they can't (because its not multiple), then the first value will become selected. It seems the more intuitive behavior would be that they just wouldn't be able to deselect any (swallow nulls if exclusive is true (multiple is false))

smonero: it works for the exclusive case if I do `value={value || (multiple ? value : React.Children.toArray(children)[0].props.value)}` . However, this breaks when `multiple` is true because internally the way the component works is that its splicing together the selection choices.

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,50 @@
...
+} from "@material-ui/lab";
+
+export interface ToggleButtonGroupProps {
+  /** A boolean to decide if the user can select multiple options
+   * simultaneously. If true, users will be able to select multiple options,
+   * whereas if false, an exclusive choice will be enforced.
+   */
+  multiple?: boolean;
+  /** The current value that is selected in the group. Usually this will be
+   * a useState value.
```
dschaller: I think you can remove the `Usually` bit here since it's really not relevant to consumers what react hook they use to store their value.

frontend/packages/core/src/Input/toggle-button-group.tsx
```diff
@@ -0,0 +1,52 @@
...
+    "value" | "children" | "size" | "orientation" | "onChange"
+  > {
+  /** If true, multiple children options can be selected simultaneously. */
+  multiple?: boolean;
+}
+
+const StyledMuiToggleButtonGroup = styled(MuiToggleButtonGroup)({
+  ".MuiToggleButton-root": {
+    "&.Mui-selected": {
+      color: "rgba(53, 72, 212, 1)",
```
dschaller, line 20: Oh I think I misread the designs when I made the original opacity comment. Not a huge deal either way but for posterity in case we want to update this later the hex value is `#3548D4`

smonero: i thought it was the same color. ill change it

dschaller: It is, one is just a hex value vs rgba

katpioro (approved):

PR #2237, opened 2022-05-16
add configureable oidc claims

povilasv:
oidc: allow users to map oidc subject & group claims

### Description

Users can now configure Subject and Groups mappings via the clutch config file:
```
      oidc:
        issuer: ...
        client_id: ${CREDENTIALS_OIDC_CLIENT_ID}
        client_secret: ${CREDENTIALS_OIDC_CLIENT_SECRET}
        redirect_url: "${BASE_URL}/v1/authn/callback"
        subject_claim_name: email
        groups_claim_name: group
      session_secret: ${CREDENTIALS_SESSION_SECRET}
```

If these arguments are not set, the old mapper [DefaultClaimsFromOIDCToken](https://pkg.go.dev/github.com/lyft/clutch/backend/service/authn#DefaultClaimsFromOIDCToken) is used.

### Testing Performed

tested locally in a private setup
also added unit tests

### GitHub Issue

Discussed in slack, these flags are similar to Kubernetes API Server's:
```
   - --oidc-groups-claim=groups
   - --oidc-username-claim=email
```


dschaller:
@povilasv mind filling out the description above? We have a template [here](https://github.com/lyft/clutch/blob/main/.github/pull_request_template.md)

povilasv:
Uuups somehow forgot to fill the template. Thanks @dschaller

backend/service/authn/oidc.go
```diff
@@ -296,6 +296,64 @@ func DefaultClaimsFromOIDCToken(ctx context.Context, t *oidc.IDToken) (*Claims,
...
+		if !ok {
+			return nil, fmt.Errorf("claims did not deserialize with %s field", cc.groupsClaim)
+		}
+		groupsIntSlice, ok := groupsInt.([]interface{})
+		if !ok {
+			return nil, fmt.Errorf("group claim did not deserialize with %s field, groups claim must be a list", cc.groupsClaim)
+		}
+
+		for _, v := range groupsIntSlice {
+			group, ok := v.(string)
```
dschaller, line 340: I'm curious how you're currently planning on using this? I'm a bit worried about casting this to a string instead of using a specific provider group type (e.g. an Okta group).

api/config/service/authn/v1/authn.proto
```diff
@@ -14,6 +14,9 @@ message OIDC {
   string redirect_url = 4 [ (validate.rules).string = {min_bytes : 1} ];
 
   repeated string scopes = 5;
+
+  string subject_claim_name = 6;
+  string groups_claim_name = 7;
```
mikecutalo, line 19: Should we do something like this to indicate that this is overriding the default? Also can you add a comment with more detail description of the behavior so it's more clear to readers, i don't think will be able to be expressive enough with just the name of the variable.

as an example maybe something like this?
```
  string override_subject_claim_name= 6;
  string override_groups_claim_name = 7;
```

dschaller:
This is great! 🎉

backend/service/authn/oidc.go
povilasv: I feel like if we follow the Kubernetes approach we can assume that the groups claim is a slice of strings:

image

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuring-the-api-server 

I personally,

povilasv:
Made a new PR with only subject claim -> https://github.com/lyft/clutch/pull/2255

RichardWLaub:
@dschaller can we revisit this group claim implementation? I'm trying to create some authz policies with okta group principals. I might be missing something but that currently doesn't seem possible since the `grp` field of cluth's JWT token is always being set to `[""]`. I tried with @povilasv 's original changes and authz worked like a charm.

PR #2243, opened 2022-05-17
k8s: refactor direct clientset access

mikecutalo:
### Description
In https://github.com/lyft/clutch/pull/2232 i made changes to allow for direct cientset access. However upon looking back the implementation could of been simpler without needing to change the `ClientsetManager` interface at all. This refactor uses the existing `GetK8sClientset` func with defaults. I also updated the the k8s interface to include a context for this new function as its now required.

### Testing Performed
Unit.

### DIFF
--- a/backend/service/k8s/k8s.go
+++ b/backend/service/k8s/k8s.go
@@ -140,7 +140,8 @@ func (s *svc) Clientsets(ctx context.Context) ([]string, error) {
 }
 
 func (s *svc) GetK8sClientset(clientset string) (ContextClientset, error) {
-	return s.manager.GetK8sClientsetDirect(clientset)
+	// Dont specify cluster or namespace, were simply looking for the clientset.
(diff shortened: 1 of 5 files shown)


smonero (approved):

dschaller (approved):

backend/service/k8s/k8s.go
```diff
@@ -139,8 +139,9 @@ func (s *svc) Clientsets(ctx context.Context) ([]string, error) {
 	return ret, nil
 }
 
-func (s *svc) GetK8sClientset(clientset string) (ContextClientset, error) {
-	return s.manager.GetK8sClientsetDirect(clientset)
+func (s *svc) GetK8sClientset(ctx context.Context, clientset string) (ContextClientset, error) {
```
sbansal7, line 142: Can we optionally take cluster and namespace also as arguments to this function?

PR #2247, opened 2022-05-20
FE: add test for toggle button group component

smonero:
### Description

relies on https://github.com/lyft/clutch/pull/2236

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
+++ b/frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
@@ -32,7 +32,7 @@ describe("ToggleButtonGroup", () => {
   });
   const searchStringForComponent = "Styled(WithStyles(ForwardRef(ToggleButton)))";
 
-  describe("basic rendering", () => {
+  describe("rendering", () => {
     it("renders", () => {
       expect(multipleFalseWrapper.find(ToggleButtonGroup)).toBeDefined();
     });
--- a/frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
+++ b/frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
@@ -1,0 +1,80 @@
+import * as React from "react";
+import { ToggleButton } from "@material-ui/lab";
+import { shallow } from "enzyme";
+
+import ToggleButtonGroup from "../toggle-button-group";
+
+// Note interaction tests will happen when there is a parent component that
+// can maintain state and select a default
+describe("ToggleButtonGroup", () => {
+  let multipleFalseWrapper;
+  let multipleTrueWrapper;
+  beforeEach(() => {
+    multipleFalseWrapper = shallow(
+      <ToggleButtonGroup value="foo" onChange={() => {}} size="large">
+        <ToggleButton value="foo">foo</ToggleButton>
+        <ToggleButton value="Ingress">Ingress</ToggleButton>
+        <ToggleButton value="Egress">Egress</ToggleButton>
+      </ToggleButtonGroup>
+    );
+
+    multipleTrueWrapper = shallow(
+      <ToggleButtonGroup
+        multiple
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+import ToggleButtonGroup from "../toggle-button-group";
+
+// Note interaction tests will happen when there is a parent component that
+// can maintain state and select a default
+describe("ToggleButtonGroup", () => {
+  let multipleFalseWrapper;
+  let multipleTrueWrapper;
+  beforeEach(() => {
+    multipleFalseWrapper = shallow(
```
jdslaugh, line 13: nit: It might be nicer to turn these wrappers into a function that accepts props on the ButtonGroup and returns the shallow wrapped component. This way we can expand the usage of this easily and we don't need two separate pieces to represent multi/non-multi.

frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+      expect(multipleFalseWrapper.find(ToggleButtonGroup)).toBeDefined();
+    });
+
+    it("renders with 3 toggles", () => {
+      expect(multipleFalseWrapper.find(ToggleButton)).toHaveLength(3);
+    });
+
+    it("renders with 2 toggles", () => {
+      expect(multipleTrueWrapper.find(ToggleButton)).toHaveLength(2);
+    });
```
jdslaugh, line 46: These tests are identical so I'm not sure we're going to glean much from it. We should likely be testing that if there is no length that it doesn't render or renders 'x' etc.

frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+      expect(multipleFalseWrapper.find(searchStringForComponent).prop("exclusive")).toEqual(true);
+    });
+
+    it("renders with size being large", () => {
+      expect(multipleFalseWrapper.find(searchStringForComponent).prop("size")).toEqual("large");
+    });
+
+    it("renders with the default size being medium", () => {
+      expect(multipleTrueWrapper.find(searchStringForComponent).prop("size")).toEqual("medium");
+    });
```
jdslaugh, line 66: Another case for separating the wrappers into a function, right now it's difficult to understand why this test is different than the one above it because the wrapper defaults is the only thing that changes.

frontend/packages/core/src/Input/tests/toggle-button-group.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+      expect(multipleFalseWrapper.find(searchStringForComponent).prop("value")).toEqual("foo");
+    });
+
+    it("renders with exclusive false when multiple is true", () => {
+      expect(multipleTrueWrapper.find(searchStringForComponent).prop("exclusive")).toEqual(false);
+    });
+
+    it("renders with exclusive true when multiple is false", () => {
+      expect(multipleFalseWrapper.find(searchStringForComponent).prop("exclusive")).toEqual(true);
+    });
```
jdslaugh, line 58: So one thing is that while it's good that we are testing for the property to exist, we're not actually testing if this property has any effect. We should also have a test, or extend this test, where we try and select multiple and are unable to because exclusive is true, similar for the test above.

jdslaugh (commented):
Some comments. I think it's a good start but I'd say this these tests will need to be more involved for us to gain some useful information from them in case of a breakage.

Since we've been discussing MUIv5, let's think about what tests we would need to create where we could confidently say that most likely nothing broke this component during the upgrade because the tests were all successful.

PR #2250, opened 2022-05-23
API: add case insensitive to autocomplete request

smonero:
### Description

It would be useful to have autocomplete be able to allow for case insensitivity. Note this is just the proto change - the actual backend implementation change will come later and is more involved, but I wanted to split things like usual for fast PR reviews.

### GitHub Issue

Fixes #


dschaller (approved):

jdslaugh (approved):
From my limited knowledge of go this looks pretty straightforward :)

smonero:
realized we can actually just do this on the FE

PR #2251, opened 2022-05-25
FE: project catalog: adjust spacing

smonero:
### Description

The final design is like the second pic but with 8px padding added to the top

Another potential option is to use `justify-content` to center the grid items rather than having things left aligned like they are now. Note that the project catalog search page has things left aligned, not centered, so I guess it is good to be consistent.

Before:
Screen Shot 2022-05-24 at 8 37 00 PM

After [FINAL DESIGN]:

Screen Shot 2022-05-25 at 2 15 40 PM

[old design] After (before shifting things over for Josh's comment):
Screen Shot 2022-05-24 at 8 36 27 PM

justify content strategy (could also center the top grid row as well, you can see the lower one is centered but it doesn't line up with the header anymore):

Screen Shot 2022-05-24 at 8 18 11 PM

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/projectCatalog/src/details/index.tsx
+++ b/frontend/workflows/projectCatalog/src/details/index.tsx
@@ -45,7 +45,7 @@ const QuickLinksAndSettingsBtn = ({ linkGroups }) => {
         justify="flex-end"
         spacing={1}
         style={{
-          paddingRight: "32px",
+          padding: "8px 32px 0px 0px",
         }}
       >
         <Grid item>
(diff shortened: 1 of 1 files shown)


jdslaugh:
@smonero I like the second one better than the 3rd. Think we can shift it a bit so that the settings icon button lines up with the lower grid?

For future, I wonder if we should investigate moving the settings button onto the left vs the right during design. I think it would look a little nicer personally.

jdslaugh (approved):

PR #2253, opened 2022-05-25
tools: update pre-flight check script for local yarn binary

dschaller:
### Description

This updates the pre-flight script to look for the local yarn binary that's installed into the local bin directory instead of globally. As part of this modification the script needs to read in a local directory override since it can be run from any gateway and needs to check that gateway for the yarn binary.

### Testing Performed

manual


smonero:
Just for my learning, why is it better to use the local yarn binary rather than the global one? Why did we originally choose to use the global one?

smonero (approved):

jdslaugh (approved):

PR #2256, opened 2022-05-27
FE: autocompletes: lowercase search strings for case insensitivity

smonero:
### Description

Before sending the autocomplete request, we lowercase the strings. This allows people to do searches that are case-insensitive. Everything in the db is lowercase.

There was a previous PR that somewhat fixed this but that was only for the k8s dash. There was also investigation into doing this on the backend instead (sending a bool in the request to specify if it was case-insensitive) but it was decided this was a quicker easier way.

### Testing Performed

local

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Resolver/input.tsx
+++ b/frontend/packages/core/src/Resolver/input.tsx
@@ -41,9 +41,10 @@ const autoComplete = async (type: string, search: string): Promise<any> => {
     return { results: [] };
   }
 
+  // We lowercase the search to make the lookup case insensitive
   const response = await client.post("/v1/resolver/autocomplete", {
     want: `type.googleapis.com/${type}`,
-    search,
+    search: search.toLowerCase(),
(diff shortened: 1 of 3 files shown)


jdslaugh (approved):

frontend/packages/core/src/Resolver/input.tsx
```diff
@@ -41,9 +41,10 @@ const autoComplete = async (type: string, search: string): Promise<any> => {
     return { results: [] };
   }
 
+  // We lowercase the search to make the lookup case insensitive
```
dschaller, line 44: I thought we could set a flag on the request to have this be case insensitive. It seems odd that you have to pass along a lowercase string for it to be case insensitive

smonero: synced in standup. We will make it on the backend.

PR #2270, opened 2022-06-08
BE: Case-insensitive autocomplete

smonero:
### Description

Using the API changes, we can now use `ILIKE` for things to be case-insensitive.

An assumption is that this only applies to filters, not the sort search requests. It could go in either place in the protobuf - the filter message, or the search request message.

### Testing Performed

local

### GitHub Issue

Fixes #


miguelmolina95 (approved):

backend/service/topology/topology.go
```diff
@@ -187,7 +188,11 @@ func (c *client) Search(ctx context.Context, req *topologyv1.SearchRequest) ([]*
 	return results, nextPageTokenStr, nil
 }
 
-func (c *client) Autocomplete(ctx context.Context, typeURL, search string, limit uint64) ([]*topologyv1.Resource, error) {
+func (c *client) Autocomplete(ctx context.Context, typeURL, search string, limit uint64, caseInsensitive bool) ([]*topologyv1.Resource, error) {
+	if caseInsensitive {
+		search = strings.ToLower(search)
```
danielhochman, line 193: i'm just curious how this works without being passed down to the actual search function. haven't been in the loop on this bug fix, so maybe it's already been documented somewhere else.

danielhochman: but the database just happens to be lowercase at Lyft, is that right? couldn't it be mixed case when dealing with data from other sources?

smonero: here was an alternative PR https://github.com/lyft/clutch/pull/2256 that just lowercased the FE. But that was decided to be the wrong approach after a discussion.

smonero: I thought about this more - lets have 2 modes, fast and slow. Fast just lowercases the string. Slow adds the `LIKE` keyword with collation. So instead of `caseInsensitive` being a bool it could be an enum instead. Does this plan sound acceptable?

danielhochman: i wouldn't overcomplicate it. is the performance of ILIKE really unacceptable for our data set?

alternatively the postgres docs suggest just using `lower()` in the query https://www.postgresql.org/docs/current/indexes-expressional.html, which you can also create an index on if performance is still not acceptable. i would go that direction i think.

api/topology/v1/topology_api.proto
```diff
@@ -70,6 +70,13 @@ message SearchRequest {
     string type_url = 2;
     // Any valid metadata on the cache Resource object
     map<string, string> metadata = 3;
+
+    // If case-insensitive, will use `ILIKE`. Else, will use `LIKE`
+    // Notes about ILIKE from postgres docs:
+    // The key word ILIKE can be used instead of LIKE to make the match case-insensitive
+    // according to the active locale. This is not in the SQL standard but is a
+    // PostgreSQL extension.
+    bool case_insensitive = 5;
```
danielhochman, line 79: i would switch this around and make case insensitive the default 
```suggestion
    bool case_sensitive = 5;
```

danielhochman (approved):

PR #2294, opened 2022-07-05
frontend: Short Link Patch

jdslaugh:
### Description
- Due to the redirect when loading a short link it was found to be difficult to remember which short link a user had loaded. This patch adds support for displaying the currently utilized short link in the alert message.

#### Alert
Screen Shot 2022-07-06 at 2 13 39 PM

### Testing Performed
manual


miguelmolina95 (approved):

frontend/packages/core/src/AppProvider/short-link-state-hydrator.tsx
```diff
@@ -103,7 +105,7 @@ const ShortLinkStateHydrator = ({
       {state.fromShortLink && (
         <Grid container direction="column" alignItems="flex-end">
           <StyledAlert severity="warning">
-            Loaded shared state. Any local changes will not be preserved.
+            Loaded shared state &quot;{state.hash}&quot;. Any local changes will not be preserved.
```
danielhochman, line 108: can you make it a link so people can easily right click, copy link address?

jdslaugh: Sure, let me update it

frontend/packages/core/src/Contexts/workflow-storage-context/types.tsx
```diff
@@ -24,12 +24,17 @@ export interface HydratedData {
   };
 }
 
+export interface HydrateState {
+  hash: string;
+  state: IClutch.shortlink.v1.IShareableState[];
```
dschaller, line 29: Should this be optional? Just asking because of the optional chaining above

frontend/packages/core/src/AppProvider/short-link-state-hydrator.tsx
```diff
@@ -103,7 +108,13 @@ const ShortLinkStateHydrator = ({
       {state.fromShortLink && (
         <Grid container direction="column" alignItems="flex-end">
           <StyledAlert severity="warning">
-            Loaded shared state. Any local changes will not be preserved.
+            <div style={{ display: "flex" }}>
+              Loaded shared state &quot;
+              <Link href={generateShortLinkRoute(window.location.origin, state.hash)}>
+                {state.hash}
+              </Link>
```
dschaller, line 115: Should we have an alternative state if `state.hash` is an empty string to prevent showing something like `Loaded shared state "". Any local changes will not be preserved.`?

jdslaugh: Good call out, modified it to handle if hash is empty and it'll just look like it was previously.

dschaller (approved):

PR #2296, opened 2022-07-08
open source markdown drawer component

smonero:
### GitHub Issue

Fixes #


smonero:
there is an issue because react-markdown is an ESM package, it seems support for those was only added in 3.1 which we don't use? Or maybe I'm not reading right.

https://yarnpkg.com/getting-started/qa#does-yarn-support-esm
`Loaders aren't perfect yet, and until they are ESM-only packages cannot be recommended, but there's a way forward and as we work together we'll get there. We just have to be careful not to push people aside as we run towards our goal.`

When I add the `"react-markdown": "<rootDir>/node_modules/react-markdown/react-markdown.min.js",` for some reason it doesn't work.

PR #2302, opened 2022-07-18
Project Selector: add refresh toggle

smonero:
### Description

It is nice to be able to control the refresh. Also once we can control it then we can pass it to the cards through the context and they can also be controlled :)

This is also important because users want to be able to pause refreshing when zooming or out on the timeline, or looking at a specific set of data.

Kapture 2022-07-25 at 13 54 27

For the task to be finished, it requires 2 more PRs - 1 for saving the auto-refresh choice to the context, and another in clutch-private that will read from the context the choice and pass it to each of the cards

### Testing Performed

use ENABLE_DASH_REFRESH_TOGGLE to test

### TODOs

- [x] discuss with team the styling and placement (maybe its too big, it could be shrunk, but we would need to make it a custom select rather than the clutch core select. We could also move it to the top next to the info icon, or have it below downstreams, or have it inside the timeline)


jdslaugh (approved):
Looks good, some minor comments

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -295,12 +299,28 @@ const ProjectSelector = ({ onError }: ProjectSelectorProps) => {
...
+  const updateRefreshToggle = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setAutoRefresh(event.target.checked);
+  };
+
+  // useEffect that handles auto-refreshing / reloading
+  React.useEffect(() => {
+    if (autoRefresh) {
+      // hardcode to 30 seconds
+      const interval = setInterval(() => hydrateProjects(state, dispatch, fromShortLink), 30000);
+      return () => clearInterval(interval);
```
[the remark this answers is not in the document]
smonero: We clear the interval every time we set it. Its not possible for it to be set and not be cleared. I'm not 100% sure what you mean by "in the middle of a 30 second refresh", but the useEffect happens when there is change in the toggle, and we only care about when its turned on (as in, when autoRefresh is true).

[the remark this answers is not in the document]
jdslaugh: Ah good call, for whatever reason I was thinking it was holding the lock, so nevermind there. Moving along :)

smonero:
waiting for consensus from https://lyft.slack.com/archives/C02TW2BB57X/p1658429124888349

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -295,48 +301,77 @@ const ProjectSelector = ({ onError }: ProjectSelectorProps) => {
...
+                maxWidth="400px"
+                placement="right-start"
+              >
+                <InfoOutlinedIcon fontSize="small" />
+              </Tooltip>
+            </FlexCenterAlignContainer>
+            <FlexCenterAlignContainer>
+              <SimpleFeatureFlag feature="dashRefreshSelect">
+                <FeatureOn>
+                  <Tooltip title="Toggle Refresh" interactive maxWidth="400px" placement="bottom">
```
scarlettperry, line 368: jst a suggestion
```suggestion
                  <Tooltip title="Toggle data refresh" interactive maxWidth="400px" placement="bottom">
```

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -295,48 +300,83 @@ const ProjectSelector = ({ onError }: ProjectSelectorProps) => {
...
+  // useEffect that handles auto-refreshing / reloading.
+  React.useEffect(() => {
+    // This way of autorefreshing is also used in other places.
+    if (autoRefresh) {
+      // hardcode to 30 seconds
+      const interval = setInterval(() => hydrateProjects(state, dispatch, fromShortLink), 30000);
+      return () => clearInterval(interval);
+    }
+    return () => {};
+  }, [hasError, state, autoRefresh]);
```
scarlettperry, line 315: for my understanding - why is `hasError` included?

scarlettperry (approved):
lgtm! just 1 non-blocking Q

frontend/workflows/projectSelector/src/project-selector.tsx
smonero: oh yeah actually I realized that its just coming from the `state`, so we don't actually need it. good catch, thanks.

PR #2288, opened 2022-06-22
housekeeping: Material UI v5

jdslaugh:
### Description
This PR updates Clutch to Material UI v5. 
- Resolves #2097 

Most of the changes are import path adjustments as well as spacing / sizing changes. The goal of this PR is to keep consistency between the current version of Clutch with @material-ui and @mui. Further improvements utilizing newer features brought by @mui will be done in later updates.

Most of the changes were made via codemods provided by @mui here: https://mui.com/material-ui/migration/migration-v4/

##### Text Field (Before)
Screen Shot 2022-06-24 at 10 51 58 AM

##### Text Field (After)
Screen Shot 2022-06-24 at 10 51 43 AM

##### Select (Before)
Screen Shot 2022-06-24 at 10 52 01 AM

##### Select (After)
Screen Shot 2022-06-24 at 10 51 48 AM

### Testing Performed
manual

PR #2306, opened 2022-07-19
muiv5 - Breaking Apart - Core/Input Changes

jdslaugh:
### Description
- Breaking apart https://github.com/lyft/clutch/pull/2288
- Contains core/input changes

### Testing Performed
manual


frontend/packages/core/src/Input/select.tsx
```diff
@@ -62,7 +56,7 @@ const SelectIcon = (props: any) => (
 
 const BaseSelect = ({ className, ...props }: MuiSelectProps) => (
   <MuiSelect
-    disableUnderline
+    // disableUnderline
```
smonero, line 59: just curious - what are the tradeoffs of commenting this out vs just deleting it?

smonero (approved):

frontend/packages/core/src/Input/select.tsx
jdslaugh: Probably just forgot to delete it honestly :) Let me remove it

frontend/packages/core/src/Input/form.tsx
```diff
@@ -9,6 +9,7 @@ const Form = styled.form({
 
 const FormRow = styled.div({
```
scarlettperry, line 10: the form rows get misaligned when you type into them
Screen Shot 2022-07-26 at 11 59 12 AM

scarlettperry: oh actually i think it's due to the new styling changes to the text field

scarlettperry:
don't know whether some of these are due to the select or text field updates so listing what I found here (compared w/ https://storybook.clutch.sh):

The label above the select and text field are creeping into the text field area vs before the label was fully above the text field area. The label size also looks smaller than before.
Screen Shot 2022-07-26 at 12 41 02 PM

For the select dropdown, when you hold down on an option selection, there's a ripple effect.

For the select dropdown, the popper/paper is misaligned with the textfield area. I only noticed this in storybook and not in the workflow themselves.
Screen Shot 2022-07-26 at 12 29 54 PM

The autocomplete text field width is smaller. I only noticed this in storybook and not in the workflow themselves.
Screen Shot 2022-07-26 at 12 47 19 PM

For the select dropdown "With Start Adornment", the icon is oversized.
Screen Shot 2022-07-26 at 12 30 47 PM

jdslaugh:
> don't know whether some of these are due to the select or text field updates so listing what I found here (compared w/ https://storybook.clutch.sh):
> 
> The label above the select and text field are creeping into the text field area vs before the label was fully above the text field area. The label size also looks smaller than before. 
> 
> For the select dropdown, when you hold down on an option selection, there's a ripple effect.
> 
> For the select dropdown, the popper/paper is misaligned with the textfield area. I only noticed this in storybook and not in the workflow themselves. 
> 
> The autocomplete text field width is smaller. I only noticed this in storybook and not in the workflow themselves. 
> 
> For the select dropdown "With Start Adornment", the icon is oversized. 
> 
> For the select dropdown, the initial selected color of an option is blue instead of the clutch purple
Thanks for the feedback!

 - Labels in the text field is intentional and is a MUIv5 change
 - I'll take a look at the rest though!

frontend/packages/core/src/Input/form.tsx
jdslaugh: Hmm, didn't see this one, I'll take a look. Which workflow was this?

scarlettperry: the k8s dashboard

jdslaugh:
@scarlettperry 

> For the select dropdown, when you hold down on an option selection, there's a ripple effect.
Never saw this one, but it may have been fixed with the others.
 
> For the select dropdown, the popper/paper is misaligned with the textfield area. I only noticed this in storybook and not in the workflow themselves. 
Should be working now.

> The autocomplete text field width is smaller. I only noticed this in storybook and not in the workflow themselves. Screen Shot 2022-07-26 at 12 47 19 PM
Should be fixed.

> For the select dropdown "With Start Adornment", the icon is oversized. Screen Shot 2022-07-26 at 12 30 47 PM
Haha yeah, missed that. There's a `height: auto` that was being applied that I needed to remove/replace, so it should be fixed now.

> For the select dropdown, the initial selected color of an option is blue instead of the clutch purple Screen Shot 2022-07-26 at 12 54 33 PM
Should be fixed.

scarlettperry:
here's a gif of the ripple/animation effect on the select dropdown. i checked out https://mui.com/material-ui/react-select/ and see it also shows up in their demos but from a quick scan didn't see mention of a prop to disable it
Kapture 2022-08-02 at 15 44 33

frontend/packages/core/src/Input/form.tsx
jdslaugh: Alright the secondary issue should be fixed now, I need to see if theres a better way to do this however because I'm currently not a fan.

jdslaugh:
> jst 2 things:
> 
> here's a gif of the ripple/animation effect on the select dropdown. i checked out https://mui.com/material-ui/react-select/ and see it also shows up in their demos but from a quick scan didn't see mention of a prop to disable it (for example something like [this](https://github.com/lyft/clutch/blob/8fea20d80299841f820c03c8034b84e616abaa51/frontend/packages/core/src/Table/table.tsx#L227)) so seems like we'll need to target the css selector for it to change it Kapture 2022-08-02 at 15 44 33 [ Kapture 2022-08-02 at 15 44 33 ](https://user-images.githubusercontent.com/39421794/182460490-dc62fca3-de85-4203-aacf-7df6a7d9efcc.gif) [ ](https://user-images.githubusercontent.com/39421794/182460490-dc62fca3-de85-4203-aacf-7df6a7d9efcc.gif)
> 
> & looks like the icon on the select "With Start Adornment" story in storybook is still oversized

So the ripple looks to be fixed in the latest merge, so I think we're good there. I modified the icon to have a max height equaling that of the select (20px) which fixes the other issue.

frontend/packages/core/src/Input/form.tsx
jdslaugh: Alright, was able to modify it and removed some extra code, looks like we should be good now.

scarlettperry (approved):
nice! 🎉

PR #2308, opened 2022-07-19
muiv5 - Breaking Apart - Core Changes

jdslaugh:
### Description
- Breaking apart https://github.com/lyft/clutch/pull/2288
- Contains core changes

### Testing Performed
manual


frontend/packages/core/src/Feedback/tooltip.tsx
```diff
@@ -26,7 +26,7 @@ const StyledTooltip = styled(BaseTooltip)((props: { maxwidth?: string }) => ({
   },
 }));
 
-export interface TooltipProps extends Pick<MuiTooltipProps, "interactive" | "placement"> {
```
scarlettperry, line 29: is the interactive prop not available in muiv5?

frontend/packages/core/src/tab.tsx
```diff
@@ -48,14 +48,15 @@ const StyledTabs = styled(TabList)({
   },
 });
 
-export interface TabProps extends Pick<MuiTabProps, "label" | "selected" | "value" | "onClick"> {
+export interface TabProps extends Pick<MuiTabProps, "label" | "value" | "onClick"> {
   children?: React.ReactNode;
   startAdornment?: React.ReactNode;
+  selected?: boolean;
```
scarlettperry, line 54: should we pass in `selected` like we do `startAdornment`?

frontend/packages/core/src/button.tsx
```diff
@@ -174,12 +175,11 @@ const StyledIconButton = styled(MuiIconButton)<{
...
 }));
 
-// TODO: (jslaughter) Update when large sizing is available with material-ui@5
-export interface IconButtonProps extends Pick<MuiIconButtonProps, "disabled" | "type" | "onClick"> {
+export interface IconButtonProps
+  extends Pick<MuiIconButtonProps, "disabled" | "type" | "onClick" | "size"> {
   /** The button variantion. Defaults to primary. */
   variant?: ButtonVariant;
   children: React.ReactElement;
-  size?: IconButtonSize;
```
scarlettperry, line 182: should we delete the type reference as well?

frontend/packages/core/src/stories/icon-button.stories.tsx
```diff
@@ -18,7 +18,7 @@ export default {
 } as Meta;
 
 const Template = (props: IconButtonProps) => (
-  <IconButton {...props}>
+  <IconButton {...props} size="large">
```
scarlettperry, line 21: noticed the large size is being set in a few places - is that the new default we're going with? if so, should we update that in the iconbutton component instead?

personal preference would be to default to the medium size tho as the large looks too big to use as the default 🤔

Screen Shot 2022-08-01 at 11 54 17 AM



Screen Shot 2022-08-01 at 11 54 25 AM

frontend/packages/core/src/navigation.tsx
```diff
@@ -8,7 +8,7 @@ import {
 } from "react-router-dom";
 import type { PartialPath, State } from "history";
 
-const UTM_PARAMS = ["utm_source", "utm_medium", "utm_campaign", "utm_term", "utm_content"];
```
scarlettperry, line 11: just curious, why are these 3 types being removed?

frontend/packages/core/src/Feedback/tooltip.tsx
jdslaugh: Interactive was removed as they're all interactive by default. Now the only interactive option is `disableInteractive`. Which probably makes sense to add on, let me fix.

frontend/packages/core/src/tab.tsx
jdslaugh: As in pull it onto the tab component and apply it? We could, but it's already applied by spreading the props. Unless you're referring to something else.

This was only added so that it would setup the `mui-selected` class.

frontend/packages/core/src/button.tsx
jdslaugh: I left it for the moment since it's being used in other spots. Wanted to keep it simple initially.

frontend/packages/core/src/stories/icon-button.stories.tsx
jdslaugh: https://mui.com/material-ui/migration/v5-component-changes/#%E2%9C%85-update-size-prop

That's the new default by mui, which keeps the same 48px as was the default before. The only IconButtons that should've been updated were those that were imported directly from MUI vs our internal one. (So this one was a mistake :) )

frontend/packages/core/src/navigation.tsx
jdslaugh: Copy paste error, should be fixed :)

frontend/packages/core/src/tab.tsx
scarlettperry: ahh i see, nvm!

scarlettperry (approved):
nice!

PR #2309, opened 2022-07-19
frontend: Catalog Config Patches

jdslaugh:
### Description
- ~Adds fullWidth functionality to Error/Alert components~
- Re-Exports the ToggleButton from @material-ui/lab
- Adds textTransform functionality to typography
- Moves linkGroup check inside of QuickLinks component
- Keeps navigate params for redirects

### Testing Performed
manual


smonero (approved):

frontend/packages/core/src/Feedback/alert.tsx
```diff
@@ -76,10 +77,17 @@ const iconMappings = {
 export interface AlertProps
   extends Pick<MuiAlertProps, "severity" | "action" | "onClose" | "elevation" | "variant"> {
   title?: React.ReactNode;
+  fullWidth?: boolean;
```
dschaller, line 80: hmmm I'm curious about this. When looking at the [storybook component](https://storybook.clutch.sh/?path=/story/core-feedback-alert--with-title&args=variant:standard) its rendering as full width of the available space so we might not need this.

frontend/workflows/projectCatalog/src/config/index.tsx
```diff
@@ -80,20 +91,24 @@ const Config: React.FC<ProjectDetailsConfigWorkflowProps> = ({ children, default
...
+            description="Edit your projects' settings. Some changes may require additional approval from another owner of the project."
           />
         </Grid>
         {configPages && configPages.length > 1 ? (
           <Grid item xs={12}>
             <Tabs value={selectedPage} centered>
               {configPages.map((page, i) => (
-                <Tab label={page.props.title} onClick={() => setSelectedPage(i)} />
+                <Tab
+                  key={page.props.title}
```
dschaller, line 108: 🎉

frontend/packages/core/src/Feedback/alert.tsx
jdslaugh: For whatever reason I couldn't get it to go full width. The alert was fine, but the error rendering the alert was breaking it and I couldn't get it to override. I'll take another pass though.

jdslaugh: Bah figured it out, was an issue with the grid. I've removed the properties now so should be good :)

dschaller: 👍

dschaller (approved):

PR #2314, opened 2022-07-21
Hooks for auto refresh

smonero:
### Description

This adds hooks, which is part 2 of fixing the issue of users being able to disable refresh on DASH. Part 3 will be done in private, using the hook to pass the value to the card, and enabling the feature flag.

Part 1: https://github.com/lyft/clutch/pull/2302

### Testing Performed

local, used `useRefreshRateState` to put the value in the cards, and it worked.
### GitHub Issue

Fixes #


frontend/workflows/projectSelector/src/dash.tsx
```diff
@@ -44,6 +48,10 @@ const initialState: DashState = {
   projectData: {},
 };
 
+const initialRefreshState: RefreshRateState = {
+  refreshRate: 30000,
```
scarlettperry, line 52: can a comment be added that this is 30 seconds since the unit might not be obvious

scarlettperry (approved):
lgtm!

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -300,17 +300,24 @@ const ProjectSelector = ({ onError }: ProjectSelectorProps) => {
     setCustomProject(event.target.value);
   };
 
+  const { updateRefreshRate } = useRefreshUpdater();
+
   // Autorefresh will eventually be passed via hooks to the cards to enforce their update intervals
   const [autoRefresh, setAutoRefresh] = React.useState<boolean>(true);
-
+  const REFRESH_RATE_MS = 30000;
```
dschaller, line 307: should we read this value in from the context since it will have a default value? This would also mean that the project selector wouldn't overwrite the refresh rate when auto refresh / state change.

frontend/workflows/projectSelector/src/dash.tsx
```diff
@@ -74,6 +83,16 @@ const dashReducer = (state: DashState, action: DashAction): DashState => {
   }
 };
 
+const refreshReducer = (state: RefreshRateState, action: RefreshRateAction): RefreshRateState => {
+  switch (action.type) {
+    case "UPDATE": {
+      return action.payload;
```
dschaller, line 89: Should we do some validation here against the refresh interval? e.g. it probably shouldn't be negative. Do we want to allow 0 as a value?

smonero: hm yeah I think we want only vals greater than 0, something else I thought of as well is that it is kind of awkward to use `null` as a value denoting that we want to turn off refreshing. Maybe we could use an extra bool or something for actually turning refreshing off? idk

miguelmolina95 (approved):

frontend/workflows/projectSelector/src/project-selector.tsx
smonero: I guess my line of thinking was that because this is setting the context value, it wouldn't need to read it as well. I'll update it so that it initializes based off the read context value.

smonero: hm actually not sure why when I try it like this it doesn't work, or how this would work.

frontend/workflows/projectSelector/src/dash.tsx
dschaller: yeah I think that 0 would denote no refreshing because the refresh interval would be 0 :)

frontend/workflows/projectSelector/src/project-selector.tsx
smonero: actually I think I figured it out. awesome!

frontend/workflows/projectSelector/src/dash.tsx
smonero: something interesting I read about was that in browsers if you nest beyond 5 setIntervals, it has a built in mechanism to make the smallest value something like 4 ms, because the idea is that 0-3 ms is too fast.

https://developer.mozilla.org/en-US/docs/Web/API/setInterval#delay_restrictions

frontend/workflows/projectSelector/src/project-selector.tsx
```diff
@@ -300,19 +300,37 @@ const ProjectSelector = ({ onError }: ProjectSelectorProps) => {
     setCustomProject(event.target.value);
   };
 
-  // Autorefresh will eventually be passed via hooks to the cards to enforce their update intervals
-  const [autoRefresh, setAutoRefresh] = React.useState<boolean>(true);
+  const { updateRefreshRate } = useRefreshUpdater();
+
+  const initialRefreshRateState = useRefreshRateState();
+  const { refreshRate } = initialRefreshRateState;
```
dschaller, line 306: super nit: you can actually combine these into
```tsx
const { refreshRate } = useRefreshRateState();
```

dschaller (approved):

PR #2329, opened 2022-08-03
[DS-9408] Enhancing Describe Table & Adding BatchGetItem functionality

Streadz:
### Description
This PR does the following:
* Updates DescribeTable so that Attributes and KeySchemas can be returned
* Adds functionality for performing a batch get

The DescribeTable API is updated because this functionality will be leveraged within our internal DDBQuerier workflow, specifically in the UI where the table properties are hydrated. The hydrated values will be used to make the UI more friendly (i.e. select which attributes you need vs having to explicitly list them out as a user).

The BatchGet functionality is added as a general AWS functionality (and not needed in an RPC call) because it'll be invoked in the backend service component of DDBQuerier for gathering records to be displayed to the user. 

### Testing Performed
Added automated tests that verified client behavior. Additionally I'll be [linking](https://github.com/lyft/clutch-private/#developing) Clutch private to this (while it's in review) and doing additional verification


Streadz:
👀  #core-datastores-prs @angethuy @dschaller

Streadz:
👀  #core-datastores-prs @angethuy @dschaller

Streadz:
👀  @angethuy

api/aws/dynamodb/v1/dynamodb.proto
```diff
@@ -104,6 +104,12 @@ message Table {
 
   // The aws acccount
   string account = 7;
+
+  // Key schema details for the table
+  repeated KeySchema key_schema = 8;
```
danielhochman, line 109:
```suggestion
  repeated KeySchema key_schemas = 8;
```

angethuy (approved):

backend/service/aws/dynamodb.go
```diff
@@ -307,3 +340,16 @@ func generateIndexUpdates(cl *regionalClient, t *dynamodb.DescribeTableOutput, i
...
+	client, err := c.getAccountRegionClient(account, region)
+	if err != nil {
+		return nil, err
+	}
+	result, err := client.dynamodb.BatchGetItem(ctx, input)
+
+	if err != nil {
+		return nil, err
+	}
+	return result, nil
```
danielhochman, line 354:
```suggestion
	return client.dynamodb.BatchGetItem(ctx, input)
```

api/aws/dynamodb/v1/dynamodb.proto
```diff
@@ -151,6 +157,28 @@ message Throughput {
   int64 read_capacity_units = 2;
 }
 
+// Represents the Key structure for the table
+message KeySchema {
+  enum Type {
+    // Undetermined value
+    UNSPECIFIED = 0;
```
danielhochman, line 164: you should add UNKNOWN for the case when the type cannot be determined coming from the AWS SDK.

danielhochman (approved):
lgtm, small comments

miguelmolina95 (approved):

api/aws/dynamodb/v1/dynamodb.proto
Streadz: Ok, will do!

Streadz:
💨

PR #2337, opened 2022-08-12
housekeeping: Adding npmrc file

jdslaugh:
### Description
Adding an npmrc file


.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org/
```
scarlettperry, line 1: nit, add new line?

scarlettperry (approved):

.npmrc
jdslaugh: Added it, but it doesn't show up in the file changes, so thats interesting

scarlettperry: the 🛑 symbol shows up if there isn't a new line so it's good now

smonero (approved):

miguelmolina95 (approved):

PR #2345, opened 2022-08-16
[DS-9347] Adding schema info in mock describe table call

Streadz:
Adding schema info in mock describe table call. By default this would have been nil values, and it's more helpful with some test data. Also updated some of the variable setting into inline instantiation for easier readability and since the variables weren't being modified in a significant way.


Streadz:
👀  #core-datastores-prs @angethuy

angethuy (approved):

scarlettperry (approved):

Streadz:
💨

PR #2352, opened 2022-08-24
aws: expose access to aws.Config from DirectClient

sbansal7:
### Description

AWS has too many services and it is impossible for clutch to support all
of them. If someone wants to write a private extension for a given
service, today there is no way to get the config/credentials to
construct a client. This patch exposes a new method to get regional
config so that clients to an unsupported service can be created.

### Testing Performed

All unit tests performed. No impact to any additional interfaces

### DIFF
--- a/backend/service/aws/aws.go
+++ b/backend/service/aws/aws.go
@@ -201,8 +201,8 @@ type Client interface {
 // AWS SDK's struct types and not an interface that can be substituted for. It is recommended following initial
 // development of a feature that you add the calls to a service interface so they can be tested more easily.
 type DirectClient interface {
-	Config() *aws.Config
 	Autoscaling() *autoscaling.Client
+	Config() *aws.Config
 	DynamoDB() *dynamodb.Client
 	EC2() *ec2.Client
 	IAM() *iam.Client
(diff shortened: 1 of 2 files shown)


sbansal7:
I just signed the CLA (after creating the PR), but the PR is not reflecting that.

mikecutalo:
Going to close and open for the CLA check to run

backend/service/aws/aws.go
```diff
@@ -232,6 +234,10 @@ type regionalClient struct {
 	sts         stsClient
 }
 
+func (r *regionalClient) Config() *aws.Config {
```
mikecutalo, line 237: I know its basic but lets add a simple test to assert this functionality

mikecutalo (commented):
Thanks for contributing! Ive gone ahead a left a couple nits. 

As I think more about leaky abstractions and exposing too much of the internal this seems like an ok path forward, as you state its not possible to keep up with all of the AWS services and this just gets in the way of the implementer.

backend/service/aws/aws.go
sbansal7: Added a new test.

backend/service/aws/aws.go
```diff
@@ -200,6 +201,7 @@ type Client interface {
 // AWS SDK's struct types and not an interface that can be substituted for. It is recommended following initial
 // development of a feature that you add the calls to a service interface so they can be tested more easily.
 type DirectClient interface {
+	Config() *aws.Config
```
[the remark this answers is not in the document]
sbansal7: Sorted

mikecutalo (approved):

PR #2355, opened 2022-08-26
FE: core: export keyframes and css so we can animate things

smonero:
### Description

It would be useful to have these for internal components.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/index.tsx
+++ b/frontend/packages/core/src/index.tsx
@@ -52,6 +52,8 @@ export { default as Code } from "./text";
 export { Typography } from "./typography";
 export { default as ClutchApp } from "./AppProvider";
 
+export { css, keyframes } from "@emotion/react";
+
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/index.tsx
```diff
@@ -52,6 +52,8 @@ export { default as Code } from "./text";
 export { Typography } from "./typography";
 export { default as ClutchApp } from "./AppProvider";
 
+export { css, keyframes } from "@emotion/react";
```
jdslaugh, line 55: Wonder if it would be useful to rename these to something like `EMOTION_CSS` or `EMOTION_KEYFRAMES`. That way its obvious where they come from in case we need those names when we export our theming.

jdslaugh (approved):

scarlettperry (approved):

PR #2360, opened 2022-08-31
k8s: add DescribeJob API

sbansal7:
### Description
Add a new API for describing k8s jobs. Added proto definitions, and module and service implementations.

### Testing Performed
All unit tests are passing. Added a new test for new API.

### GitHub Issue

Fixes #


backend/service/k8s/job_test.go
```diff
@@ -54,6 +54,24 @@ func testJobClientset() k8s.Interface {
...
+		manager: &managerImpl{
+			clientsets: map[string]*ctxClientsetImpl{"testing-clientset": {
+				Interface: cs,
+				namespace: "testing-namespace",
+				cluster:   "testing-cluster",
+			}},
+		},
+	}
+
+	cron, err := s.DescribeJob(context.Background(), "testing-clientset", "testing-cluster", "testing-namespace", "testing-job-name")
```
scarlettperry, line 70:
```suggestion
	job, err := s.DescribeJob(context.Background(), "testing-clientset", "testing-cluster", "testing-namespace", "testing-job-name")
```

backend/service/k8s/job_test.go
```diff
@@ -54,6 +54,24 @@ func testJobClientset() k8s.Interface {
...
+	assert.NoError(t, err)
+	assert.NotNil(t, cron)
```
scarlettperry, line 72:
```suggestion
	assert.NotNil(t, job)
```

backend/service/k8s/job_test.go
```diff
@@ -54,6 +54,24 @@ func testJobClientset() k8s.Interface {
...
+
+	s := &svc{
+		manager: &managerImpl{
+			clientsets: map[string]*ctxClientsetImpl{"testing-clientset": {
+				Interface: cs,
+				namespace: "testing-namespace",
+				cluster:   "testing-cluster",
+			}},
+		},
+	}
```
scarlettperry, line 68: looks like we had previously been duplicating the same thing [here](https://github.com/lyft/clutch/blob/main/backend/service/k8s/job_test.go#L109-L119) and [here](https://github.com/lyft/clutch/blob/main/backend/service/k8s/job_test.go#L130-L140). can  a helper be created like this [one](https://github.com/lyft/clutch/blob/main/backend/service/k8s/cronjob_test.go#L17-L37) and reused across the delete/create job tests?

backend/service/k8s/job_test.go
scarlettperry, line 70: can we also add assertions for the not found and found more than 1 job case?

scarlettperry (commented):
looks great! some small comments

backend/service/k8s/job_test.go
sbansal7: Made the suggested change.

backend/service/k8s/job_test.go
sbansal7: Added assertion for not found case and changed implementation to avoid more than 1 job case. Since name is key for the object, duplicate jobs cannot be created.

sbansal7:
@scarlettperry thanks for your review. I have made the suggested changes. Please review again when you get a chance,

Thanks,
Sachin

backend/service/k8s/job_test.go
scarlettperry: ah yes that's correct! however, the pattern we've historically set is to have our `Describe` APIs internally make a `List` K8s API call. This was due to an issue with the omni search (full details [here](https://github.com/lyft/clutch/issues/257)) failing to look up a resource and the [fix](https://github.com/lyft/clutch/pull/271) was to use fieldSelectors to search cluster wide for a resource's metadata.name.

@mikecutalo - thoughts on this API making a `List` vs `Get` call?

sbansal7: Thanks for the background on why `List` is used in place of `Get` in all these APIs. I would revert to the `List` based implementation if that is desired, however, IMHO, using `DescribePod` when the intention is to search across namespaces might not be appropriate. Perhaps, we could call a version of `List` API that allows specifying `FieldSelector` as option.

sbansal7: Updated the API implementation to use `List` and changed tests accordingly.

scarlettperry: thank you for the feedback! will share it with the team.

scarlettperry (approved):
thank you for the contribution!

backend/service/k8s/job_test.go
mikecutalo: Thank you @scarlettperry! And thank you @sbansal7 for changing over to the list API!

mikecutalo (approved):

PR #2369, opened 2022-09-09
Enable partial failures for listing pods

Yiwen-Gao:
This pull request updates the list pods response to contain a field for partial failures.

### Description

This allows the K8s dashboard to show best-effort results: some pods will be displayed even if others weren't able to be retrieved. In the case of partial failures, users can see an incomplete list instead of no list at all. 

### GitHub Issue

Fixes #


api/k8s/v1/k8s.proto
```diff
@@ -423,6 +424,7 @@ message ListPodsResponse {
   option (clutch.api.v1.reference).fields = "pods";
 
   repeated Pod pods = 1 [ (clutch.api.v1.log) = false ];
+  repeated google.rpc.Status partial_failures = 2;
```
Yiwen-Gao, line 427: This is based on partial failures from the projects proto:
https://github.com/lyft/clutch/blob/16d1aff019f4b370a304db4afecefc9f63d0cfd9/api/project/v1/project.proto#L42

Yiwen-Gao: But should this be a single value instead of a repeated one? It seems like the implementation in our [private repository](https://github.com/lyft/clutch-private/blob/bb130c122183b22d438a9f8a7dedbe06bad21703/backend/service/lyftk8s/pod.go#L54-L67) returns one error (see L129).

Yiwen-Gao: Also, I didn't make any changes to the OSS version of `ListPods()` because it seems like partial failures aren't relevant and it's all-or-nothing:

https://github.com/lyft/clutch/blob/16d1aff019f4b370a304db4afecefc9f63d0cfd9/backend/service/k8s/pods.go#L46-L69

smonero: I think to do this we might need to update the error handling code https://github.com/lyft/clutch-private/blob/bb130c122183b22d438a9f8a7dedbe06bad21703/backend/service/lyftk8s/k8s.go#L69 , as in, right now it concatenates all the errors into one, and we need a system where we can have multiple errors. I'm not sure if that breaks the original intention though (would have to ask Hochman).

Yiwen-Gao: True, or should we make the failures field a single value instead of a list? Maybe if there are partial failures, we can just set it on the response:
```golang
return &corev1.PodList{Pods: pods, Failures: st.Err()}, nil
```
and if it's a complete failure, we return it directly:
```golang
return nil, st.Err()
```
I'm not sure if this follows Clutch's patterns, but let me know what you think.

frontend/workflows/k8s/src/k8s-dashboard.tsx
```diff
@@ -63,7 +63,7 @@ const defaultRequestData = inputData => {
 };
 
 const KubeDashboard: React.FC<WorkflowProps> = () => {
-  const [error, setError] = React.useState<ClutchError | undefined>(undefined);
+  const [errors, setErrors] = React.useState<ClutchError[] | undefined>(undefined);
```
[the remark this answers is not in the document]
scarlettperry: looks like there aren't changes to the OSS version of ListPods() so we don't need to make changes to the OSS K8s dashboard?

the internal implementation has its own K8s dashboard

api/k8s/v1/k8s.proto
scarlettperry: yeah it'd be good to sync with Hochman on this. our pattern of using partial errors is more recent than the K8s `NewMulticlusterStatus` error flow so perhaps it'd be a good idea to refactor that.

for this change though, i like the idea suggested in the code snippet of 1) setting the partial error on the response if both pods and errors are returned or 2) returning just the error if no pods are found. 

For single value instead of list, I think that even though it's just 1 error (technically all errors concatenated in a string), we could still keep it as a list type with 1 entry in case we support multiple errors in the future

frontend/workflows/k8s/src/k8s-dashboard.tsx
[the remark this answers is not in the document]
Yiwen-Gao: Makes sense, I'll undo these changes.

api/k8s/v1/k8s.proto
Yiwen-Gao: Hmm since Hochman isn't back until October, should this PR just wait until then?

dschaller: `corev1.PodList` is a [k8s struct](https://pkg.go.dev/k8s.io/api/core/v1#PodList) and doesn't have a `Failures` field.
If I remember correctly `MulticlusterStatus` was before the frontend supported partial failures and was meant to provide a human readable error message to users. I believe we use it pretty heavily in the resolver since that shows a single error message.

scarlettperry: yea the k8s struct doesn't; the PR adds the failure field to our list pods API response proto

scarlettperry (approved):

jdslaugh (approved):
LGTM from my limited GO experience :)

PR #2372, opened 2022-09-09
scaffolding: Non-Wizard Workflow

jdslaugh:
### Description
- We realized there is no path to create a workflow through the scaffolding which is a non-wizard flow, this alleviates that by templating between a wizard vs non wizard workflow and adding any necessary dependencies

### Testing Performed
manual


tools/scaffolding/scaffolder.go
```diff
@@ -295,6 +306,8 @@ func main() {
 	defer os.RemoveAll(tmpout)
 	log.Println("Using tmpdir", tmpout)
 
+	helloWorldModule := string(reflect.Indirect(reflect.ValueOf(data)).FieldByName("HelloWorldModule").String())
```
jdslaugh, line 309: Found this way to access the data in the struct, not sure if there's a better way currently

tools/scaffolding/scaffolder.go
```diff
@@ -34,6 +34,7 @@ type workflowTemplateValues struct {
 	DeveloperEmail string
 	URLRoot        string
 	URLPath        string
+	WizardTemplate bool
```
smonero, line 37: I think in some cases when we have bools we prepend them with `is`, like IsWizardTemplate

smonero (approved):

scarlettperry:
ran through it locally. the wizard flow works fine but when i go through the non-wizard flow, I get this error:
```
2022/09/12 10:44:47 Compiling workflow, this may take a few minutes...
2022/09/12 10:44:47 cd /var/folders/zv/36htt8mj4ls_gfkjdc4s2rd80000gp/T/clutch-scaffolding-3807014797 && yarn --frozen-lockfile && yarn tsc && yarn compile
yarn run v1.22.19
$ /private/var/folders/zv/36htt8mj4ls_gfkjdc4s2rd80000gp/T/clutch-scaffolding-3807014797/node_modules/.bin/tsc
error TS2688: Cannot find type definition file for 'long'.
  The file is in the program because:
    Entry point of type library 'long' specified in compilerOptions
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

2022/09/12 10:45:09 `yarn tsc` returned the above error
exit status 1
make: *** [scaffold-workflow] Error 1
```

tools/scaffolding/scaffolder.go
jdslaugh: Ah good, call, I'll update that

jdslaugh:
> ran through it locally. the wizard flow works fine but when i go through the non-wizard flow, I get this error:
> 
> ```
> 2022/09/12 10:44:47 Compiling workflow, this may take a few minutes...
> 2022/09/12 10:44:47 cd /var/folders/zv/36htt8mj4ls_gfkjdc4s2rd80000gp/T/clutch-scaffolding-3807014797 && yarn --frozen-lockfile && yarn tsc && yarn compile
> yarn run v1.22.19
> $ /private/var/folders/zv/36htt8mj4ls_gfkjdc4s2rd80000gp/T/clutch-scaffolding-3807014797/node_modules/.bin/tsc
> error TS2688: Cannot find type definition file for 'long'.
>   The file is in the program because:
>     Entry point of type library 'long' specified in compilerOptions
> error Command failed with exit code 2.
> info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> 
> 2022/09/12 10:45:09 `yarn tsc` returned the above error
> exit status 1
> make: *** [scaffold-workflow] Error 1
> ```
> 
> Perhaps the `long` type ([here](https://github.com/lyft/clutch/blob/main/tools/scaffolding/templates/frontend/tsconfig.json#L16)) isn't needed for the non-wizard workflow? Looks like it was added in this PR for changes mades to the wizard workflow [f618d2a](https://github.com/lyft/clutch/commit/f618d2a35704974d8cb8e75ba790b155d0777667)

Thanks for finding this, found the issue was that since I was no longer including the wizard library it wasnt importing @clutch-sh/api which is why long was needed

```
└─┬ @clutch-sh/api@2.0.0-beta -> ./api
  └─┬ protobufjs@6.11.3
    └── @types/long@4.0.1
```

Switched it to use the core import if we're not a wizard component, which I think makes sense anyways.

scarlettperry (approved):

PR #2379, opened 2022-09-14
frontend: Override for landing page

jdslaugh:
### Description
- For custom gateways, need a way to override the landing page if so desired. This PR adds a check in a workflow loop that already exists to see if there is a workflow with no path (ie a landing), if so it will set a boolean variable to not render the current landing page and allow the workflow route to take over.

#### Screenshot (workflow path of "examples")
Screen Shot 2022-09-14 at 2 21 50 PM

#### Screenshot (workflow path of "")
Screen Shot 2022-09-14 at 2 22 00 PM

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/AppProvider/workflow.tsx
+++ b/frontend/packages/core/src/AppProvider/workflow.tsx
@@ -30,6 +30,7 @@ interface BaseWorkflowConfiguration {
   group: string;
   /**
    * The path (subordinate to the group) where the workflow will exist.
+   * (optionally) use "" to override the landing page
    */
   path: string;
   /**
--- a/frontend/packages/core/src/AppProvider/workflow.tsx
+++ b/frontend/packages/core/src/AppProvider/workflow.tsx
@@ -30,7 +30,7 @@ interface BaseWorkflowConfiguration {
   group: string;
   /**
    * The path (subordinate to the group) where the workflow will exist.
-   * (optionally) use "" to override the landing page
+   * (optionally) use "" to override the landing page, will need to also update the route path
    */
   path: string;
   /**
@@ -70,6 +70,7 @@ export interface Route {
   component: React.FC<any>;
   description: string;
   displayName?: string;
+  /** (optionally) use "" to override the landing page */
   path: string;
   /** Properties required by the Component that are set only via the config. */
   requiredConfigProps?: string[];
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/AppProvider/workflow.tsx
```diff
@@ -30,6 +30,7 @@ interface BaseWorkflowConfiguration {
   group: string;
   /**
    * The path (subordinate to the group) where the workflow will exist.
+   * (optionally) use "" to override the landing page
```
scarlettperry, line 33: do we also want to add this comment [here](https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/AppProvider/workflow.tsx#L72)?

i noticed that with the echo workflow, which is structured like
```
    path: "examples",
    group: "Examples",
    displayName: "Echo",
    routes: {
      echo: {
        path: "echo",
        component: EchoWorkflow,
        displayName: "Demo Workflow",
        description: "Workflow for demonstration purposes.",
      },
    },
```
both paths had to be `""` in order for the workflow to show up as the landing page

scarlettperry:
unrelated to this PR but something it reminded me of - we should also hide the search bar on the header if no workflows are visible. Because if all are set to `hideNav` then the search isn't able to find the workflows

scarlettperry:
clarifying Q - the landing page path being set to `""` wouldn't affect query param/short link support right? in the scenario where the landing page is wizard workflow and we want to deep link the user to the second step

jdslaugh:
> clarifying Q - the landing page path being set to `""` wouldn't affect query param/short link support right? in the scenario where the landing page is a wizard workflow and we want to deep link the user to the second step

It shouldn't because it pulls the route directly from the window

jdslaugh:
> unrelated to this PR but something it reminded me of - we should also hide the search bar on the header if no workflows are visible. Because if all are set to `hideNav` then the search isn't able to find the workflows

Oh that's a good call, let me add a new PR for that

frontend/packages/core/src/AppProvider/workflow.tsx
jdslaugh: Ah yeah, I had mine set as "/" so I forgot about it, let me add another comment

scarlettperry (approved):

dschaller (approved):

PR #2391, opened 2022-09-29
frontend: simply experimentation fields and fix css

dschaller:
### Description

simplify experimentation workflows by the removing radio button option. This was only used in one spot at the moment and it was easily swapped out with a select which both aesthetically looks nicer and also makes sense given users can only select one option.

This allowed for the removal of the wrapping form element which was there to apply centering to the page for the smaller radio buttons which then let's the form styles properly cascade down providing proper margins.

Screen Shot 2022-09-29 at 12 42 10 PM

Screen Shot 2022-09-29 at 12 46 43 PM

### Testing Performed

manual


jdslaugh (approved):
Thanks for taking care of this, I agree that the select looks better!

scarlettperry (approved):

PR #2397, opened 2022-10-05
backend: return partial errors from ListPods

Yiwen-Gao:
### Description

If there were partial errors, return them in the response in addition to the pods that were found. If there was a complete failure, return `nil` as the response and the error as a separate value.

### GitHub Issue

Fixes #

### DIFF
--- a/backend/module/k8s/pods.go
+++ b/backend/module/k8s/pods.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	k8sapiv1 "github.com/lyft/clutch/backend/api/k8s/v1"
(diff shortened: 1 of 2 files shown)


Yiwen-Gao:
Are there any suggestions to mock an error for the test clientset on L195?
https://github.com/lyft/clutch/blob/cf39f3c8f0430b1a8b309ded6dc565c2cfcdb18c/backend/service/k8s/pods_test.go#L192-L204

mikecutalo (changes requested):
blocking on if partial errors are even possible in client-go

PR #2399, opened 2022-10-05
gateway: allow for stats prefix to be configurable

scarlettperry:
### Description
Currently the gateway sets the stats prefix to `clutch`; however, there could be cases where teams are running multiple Clutch gateways, in which case the stats prefix should be configurable to allow differentiating the metrics among the gateways.

PR adds a `prefix` field in the gateway config and if no prefix gets set, `clutch` remains the default.

### Testing Performed
Manual & unit test

### TODOs
- [x] test with internal gateways


dschaller (approved):

miguelmolina95 (approved):

danielhochman:
Can we write an effective test for this?

dschaller (approved):

PR #2418, opened 2022-10-12
frontend: Toast Bug Fix

jdslaugh:
### Description
- With the upgrade to MUIv5 the Toast component was no longer working and instead throwing an error `Cannot read property scrollTop of null`
- Fix was to wrap the children in a div to correctly forward the ref, taken from https://mui.com/material-ui/migration/troubleshooting/#cannot-read-property-scrolltop-of-null

NOTE: Tried to use React.forwardRef but was unable to get it to work correctly, wrapping the alert component mitigates the issue without any other changes

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/Feedback/toast.tsx
+++ b/frontend/packages/core/src/Feedback/toast.tsx
@@ -40,15 +40,17 @@ const Toast: React.FC<ToastProps> = ({
         }
       }}
     >
-      <Alert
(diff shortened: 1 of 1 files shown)


miguelmolina95 (approved):
example linked helped make review easier

kyle-xiao (approved):

PR #2421, opened 2022-10-14
backend: only emit github ratelimit gauge on valid response

dschaller:
### Description

Fixes possible panic if `resp` is nil from the RoundTrip call when trying to evaluate the response headers from a GitHub GraphQL query to gauge the remaining GitHub rate limit.

### Testing Performed

unit

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -51,9 +51,11 @@ type StatsRoundTripper struct {
 func (st *StatsRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp, err := st.Wrapped.RoundTrip(req)
 
-	if hdr := resp.Header.Get("X-RateLimit-Remaining"); hdr != "" {
-		if v, err := strconv.Atoi(hdr); err == nil {
-			st.scope.Gauge("rate_limit_remaining").Update(float64(v))
+	if resp != nil {
+		if hdr := resp.Header.Get("X-RateLimit-Remaining"); hdr != "" {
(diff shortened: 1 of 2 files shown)


backend/service/github/github.go
```diff
@@ -51,9 +51,11 @@ type StatsRoundTripper struct {
 func (st *StatsRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	resp, err := st.Wrapped.RoundTrip(req)
 
-	if hdr := resp.Header.Get("X-RateLimit-Remaining"); hdr != "" {
-		if v, err := strconv.Atoi(hdr); err == nil {
-			st.scope.Gauge("rate_limit_remaining").Update(float64(v))
+	if resp != nil {
```
scarlettperry, line 54: hmm i wonder if we should be checking for `if err != nil` instead?

from the package code comments it says

```
	// RoundTrip should not attempt to interpret the response. In
	// particular, RoundTrip must return err == nil if it obtained
	// a response, regardless of the response's HTTP status code.
	// A non-nil err should be reserved for failure to obtain a
	// response. Similarly, RoundTrip should not attempt to
	// handle higher-level protocol details such as redirects,
	// authentication, or cookies.
```

so a non nil error would let us know if there was an error getting a response

dschaller: yeah I originally had the check for error being non-nil but I realized it makes an assumption that resp is always nil when err is set when under the hood we rely on resp

danielhochman (approved):
i'm fine with resp or err check

PR #2424, opened 2022-10-17
frontend: NPS - Add ability to specify custom sub routes

jdslaugh:
### Description
Modifies the NPS header component, which already reads in the workflows, to check for component props on each visible route and add any custom nps additions found under its route. 

NOTE: This is only visible to routes where hideNav: false as it is hidden from the workflow context otherwise

Example:
```
  "@clutch-sh/project-catalog": {
    catalog: {
      trending: true,
      componentProps: {
        additionalNPS: [
          { label: "New Config", value: "/config/new config" },
        ],
      },
    },
  },
```

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/NPS/header.tsx
+++ b/frontend/packages/core/src/NPS/header.tsx
@@ -7,7 +7,7 @@ import {
   Paper as MuiPaper,
   Popper as MuiPopper,
 } from "@mui/material";
-import { sortBy } from "lodash";
+import { get, sortBy } from "lodash";
 
 import type { Workflow } from "../AppProvider/workflow";
 import { IconButton } from "../button";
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/NPS/header.tsx
```diff
@@ -62,12 +62,16 @@ export const generateFeedbackTypes = (workflows: Workflow[]): SelectOption[] =>
...
+      typeMap[group].push({
         label: route.displayName || displayName,
         value: `/${path}/${route.path}`.replace(/\/\/+/g, "/"),
-      }))
-    );
+      });
+
+      if (get("route.componentProps.customNPS", [])) {
+        typeMap[group].push(...get("route.componentProps.customNPS", []));
+      }
```
dschaller, line 73: nit:
```suggestion
      const customNPS = get("route.componentProps.customNPS", []);
      if (customNPS) {
        typeMap[group].push(...customNPS);
      }
```

frontend/packages/core/src/NPS/header.tsx
```diff
@@ -62,12 +62,16 @@ export const generateFeedbackTypes = (workflows: Workflow[]): SelectOption[] =>
...
-      ...routes.map(route => ({
+    routes.forEach(route => {
+      typeMap[group].push({
         label: route.displayName || displayName,
         value: `/${path}/${route.path}`.replace(/\/\/+/g, "/"),
-      }))
-    );
+      });
+
+      if (get("route.componentProps.customNPS", [])) {
```
dschaller, line 71: Really minor but what do you think about something like `additionalNPS`? My only concern with custom is it's unclear they are being appended

dschaller (approved):

Yiwen-Gao (approved):

PR #2426, opened 2022-10-20
github/workflows: migrate CI to Go 1.19 and fix linting issues

fsouza:
### Description

We want to make sure we can safely migrate to 1.19, so let's run it in CI. This PR ended-up being a bit big because it does multiple things (see individual commits for more details):

1. upgrade golangci-lint to 1.50.0 (latest version) and fix violations
2. fix GHA actions (`set-output` is now deprecated)
3. Update GHA workflows to run Go 1.19, and add go version to all cache keys to make future upgrades easier (note: `setup-go` has built-in [cache support](https://github.com/actions/setup-go/blob/main/README.md#caching-dependency-files-and-build-outputs), we probably don't need to use `actions/cache` directly, but I didn't want to change the config too much)
4. run gofmt with Go 1.19 to reduce future diffs as folks upgrade to Go 1.19
5. regen protos and configure the api job to use Go 1.19
6. fix golangci-lint issues
    1. `golangci-lint run --fix` for violations that can be fixed automatically
    2. some search and replace for the `ioutil` deprecation
    3. two manual fixes (one `nolint` for a false positive and one code change where we pass `ReadHeaderTimeout` to `http.Server`)
    4. remove deprecated linters from `.golangci.yml`

I'd be happy to split it apart. Each of the items above could be a separate PR, but to be honest since many of the changes are either automatic or simple search and replace, I wonder if that's really necessary? Let me know either way and I'll adjust.

One note is that clutch is already distributed with Go 1.19 via the Docker image, but I'm unsure about who are the consumers of that.

### Testing Performed

GitHub will test it for us :)


tools/golangci-lint.sh
```diff
@@ -6,9 +6,9 @@ BUILD_ROOT="${REPO_ROOT}/build"
...
 
 NAME=golangci-lint
-RELEASE=v1.42.1
-OSX_RELEASE_256=9c0042e91218dc1dd4eb7b54e29c7331eff081b3ac3f88b0d5df89b976fcd45c
-LINUX_RELEASE_256=214b093c15863430c4b66dd39df677dab6e38fc873ded147e331740d50eea51f
+RELEASE=v1.47.3
+OSX_RELEASE_256=75ff9491f1ac88ab8594d0498d2a86168ad20e3f4b958dd21be8a6c2988d50cf
+LINUX_RELEASE_256=67b5f7a2275d94034c5c87ff7655ba4b20bdb5f83c43c5d22476f2bf5c6766ef
 
 ARCH=amd64
```
fsouza, line 13: curious if anyone ever runs this locally or only in CI? I'd be happy to implement the required logic to download the arm64 tarball when running on Apple Silicon.

tools/golangci-lint.sh
```diff
@@ -6,13 +6,8 @@ BUILD_ROOT="${REPO_ROOT}/build"
...
 
 NAME=golangci-lint
-RELEASE=v1.42.1
-OSX_RELEASE_256=9c0042e91218dc1dd4eb7b54e29c7331eff081b3ac3f88b0d5df89b976fcd45c
-LINUX_RELEASE_256=214b093c15863430c4b66dd39df677dab6e38fc873ded147e331740d50eea51f
-
-ARCH=amd64
-
-RELEASE_BINARY="${BUILD_BIN}/${NAME}-${RELEASE}"
+RELEASE=v1.47.3
```
danielhochman, line 9: i saw you tried latest but then went back to this version, what was the problem with 1.50.0?

fsouza: @danielhochman starting with 1.48.0, Go 1.18 is required.

tools/golangci-lint.sh
```diff
@@ -26,31 +21,7 @@ ensure_binary() {
...
-
-    if [[ ! -f "${RELEASE_BINARY}" ]]; then
-      find "${BUILD_BIN}" -maxdepth 0 -regex '.*/'${NAME}'-[A-Za-z0-9\.]+$' -exec rm {} \;  # cleanup older versions
-      mv "${release_tmp_dir}/${NAME}" "${RELEASE_BINARY}"
-    fi
-
-    # Cleanup stale resources.
-    rm "${release_archive}"
-    rm -rf "${release_tmp_dir}"
+    env GOBIN="${BUILD_BIN}" go install github.com/golangci/golangci-lint/cmd/golangci-lint@"${RELEASE}"
```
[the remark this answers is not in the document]
fsouza: There's no ABI compatibility in Go, so there's no guarantee that a the golangci-lint binary will work with a different version of the go tooling. IIRC, that's what I saw when trying it, I can try to bring the binary back to see if things work.

With caching, the actual time difference is negligible, we can do this in a dedicated step to confirm.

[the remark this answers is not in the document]
danielhochman: Ah ok I wasn't aware that golangci-lint relied on the go tooling, I thought it was a standalone binary. The cache itself in actions is just a tarball, I don't see how it could be any faster. I was mainly concerned for local dev.

tools/golangci-lint.sh
danielhochman: I'd rather drop 1.17 then and go with 1.19. I have some other core changes planned that rely on generics anyways.

backend/gateway/gateway.go
```diff
@@ -303,10 +302,11 @@ func RunWithConfig(f *Flags, cfg *gatewayv1.Config, cf *ComponentFactory, assets
...
 
 	srv := &http.Server{
-		Handler:      mux.InsecureHandler(rpcMux),
-		Addr:         addr,
-		ReadTimeout:  timeout,
-		WriteTimeout: timeout,
+		Handler:           mux.InsecureHandler(rpcMux),
+		Addr:              addr,
+		ReadTimeout:       timeout,
+		ReadHeaderTimeout: timeout,
```
dschaller, line 308: I don't think you need to set `ReadHeaderTimeout`:
```
If ReadHeaderTimeout is zero, the value of ReadTimeout is used. If both are zero, there is no timeout.
```

fsouza: oh yeah let me revert this, this was from a warning in Go 1.17 only I think

backend/service/k8s/k8s_test.go
```diff
@@ -134,9 +132,9 @@ func TestInterceptError(t *testing.T) {
 }
 
 func TestGetK8sClientset(t *testing.T) {
-	tempfile, _ := ioutil.TempFile("", "")
+	tempfile, _ := os.CreateTemp("", "")
 	defer os.Remove(tempfile.Name())
-	_ = ioutil.WriteFile(tempfile.Name(), []byte(testConfig), 0500)
+	_ = os.WriteFile(tempfile.Name(), []byte(testConfig), 0o500)
```
dschaller, line 137: Just for my understanding why `s/0500/0o500/`

backend/retry/retry.go
```diff
@@ -34,12 +34,13 @@ Do retries a given function with delays between each attempts as defined by the
 or their corresponding defaults.
 
 err := retry.Do(
-			ctx,
-			logger,
-			scope,
-			func() error { return nil },
-			Backoff(ExponentialBackoff),
-		)
+
```
dschaller, line 37: is this newline generated?

backend/gateway/meta/meta.go
```diff
@@ -320,7 +321,9 @@ func APIBody(body interface{}) (*anypb.Any, error) {
 	return anypb.New(ClearLogDisabledFields(m))
 }
 
-/* ToValue converts custom types to a structpb.Value. This helper was added
+/*
+	ToValue converts custom types to a structpb.Value. This helper was added
+
```
dschaller, line 326:
```suggestion
```

backend/gateway/meta/meta.go
```diff
@@ -188,19 +187,21 @@ func HydratedPatternForProto(pb proto.Message) (string, error) {
...
 // And the value of "mycluster/mynamespace/nameofresource"
 // we transform the pattern into a regex and map the values to the pattern names
 //
 // The output for this example is:
-// map[string]string{
-//  cluster: mycluster
-//  namespace: mynamespace
-//  name: nameofresource
-// }
+//
```
dschaller, line 199:
```suggestion
```

backend/gateway/meta/meta.go
dschaller, line 326: Some weird indentation going on here

fsouza: Yeah that's a change in `gofmt` in Go 1.18. I didn't make any changes to source code (aside from `ReadHeaderTimeout`). I just upgraded `golangci-lint` and ran `golangci-lint run --fix`.

backend/service/k8s/k8s_test.go
fsouza: I was gonna say `gofmt` does it, but looks like it doesn't yet (https://github.com/golang/go/issues/33363). So it was probably `gofumpt` running in my editor. Let me revert it.

backend/cmd/migrate/migrate.go
```diff
@@ -13,13 +13,12 @@ import (
...
 	_ "github.com/lib/pq"
-	"github.com/uber-go/tally/v4"
-	"go.uber.org/zap"
-
 	gatewayv1 "github.com/lyft/clutch/backend/api/config/gateway/v1"
 	postgresv1 "github.com/lyft/clutch/backend/api/config/service/db/postgres/v1"
 	"github.com/lyft/clutch/backend/gateway"
 	clutchpg "github.com/lyft/clutch/backend/service/db/postgres"
+	"github.com/uber-go/tally/v4"
+	"go.uber.org/zap"
```
danielhochman, line 21: why are imports no longer grouped properly?

fsouza: Hmm, I usually see folks using only two groups (stdlib and non-stdlib), but I can adjust it to make it Python-style again. Let me do that.

fsouza: Actually, `gci` is making this change. So if we want to revert it we'd need to remove gci from `.golangci.yml`.

danielhochman (approved):

colek319 (approved):

PR #2436, opened 2022-10-21
tools: use yarn script in scaffolder

dschaller:
### Description

give preference to yarn script in scaffolder and fallback to the global yarn installation.

### Testing Performed

manual

### DIFF
--- a/tools/scaffolding/scaffolder.go
+++ b/tools/scaffolding/scaffolder.go
@@ -79,6 +79,23 @@ func determineGoPath() string {
 	return strings.TrimSpace(string(goPathOut))
 }
 
+func determineYarnPath() string {
+	path, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+	yarnScript := filepath.Join(path, "..", "..", "build", "bin", "yarn.sh")
+	scriptPath, err := exec.LookPath(yarnScript)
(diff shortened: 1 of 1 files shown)


tools/scaffolding/scaffolder.go
```diff
@@ -79,6 +79,23 @@ func determineGoPath() string {
...
+		log.Fatal(err)
+	}
+	yarnScript := filepath.Join(path, "..", "..", "build", "bin", "yarn.sh")
+	scriptPath, err := exec.LookPath(yarnScript)
+	if err != nil {
+		path, err := exec.LookPath("yarn")
+		if err != nil {
+			log.Fatal("could not find yarn executable to use")
+		}
+		return strings.TrimSpace(path)
```
danielhochman, line 94: what's with the TrimSpace?

dschaller: I was following the pattern from `determineGoPath` which I assume did it to be safe

danielhochman: that just comes from the fact that it's executing a command that returns with a newline. i don't think any of your commands do that.

danielhochman (approved):
i think you can remove trimspace otherwise lgtm

miguelmolina95 (approved):

PR #2438, opened 2022-10-24
audit: add stats to sink writer

danielhochman:
### Description
We're seeing significant lag in our pipeline and want to measure the behavior of the writer.

I also made some minor readability changes that amount to whitespace changes, which makes the diff look more complicated than it is. I can undo these changes if preferred by the reviewer.

### Testing Performed
Existing unit


Yiwen-Gao (approved):
Looks good to me!

backend/service/audit/audit.go
```diff
@@ -137,29 +139,44 @@ func (c *client) ReadEvents(ctx context.Context, start time.Time, end *time.Time
...
-	readAndFanout := func(ctx context.Context) {
-		// TODO(maybe): Backpressure on continued failure.
-		events, err := c.storage.UnsentEvents(ctx)
-		if err != nil {
-			return
-		}
+func (c *client) readAndFanout(ctx context.Context) {
+	// TODO(maybe): Backpressure on continued failure.
+
+	c.sinkWriterScope.Counter("events_fetch").Inc(1)
```
mikecutalo, line 145: i was going to suggest an `events_fetch` success & error, but with gauge and counter i think it will make it pretty clear if we are receiving anything at all.

backend/service/audit/audit.go
```diff
@@ -137,29 +139,44 @@ func (c *client) ReadEvents(ctx context.Context, start time.Time, end *time.Time
...
-					)
-				}
+	numEvents := len(events)
+	c.sinkWriterScope.Gauge("events_received").Update(float64(numEvents))
+	c.sinkWriterScope.Counter("events_total").Inc(int64(numEvents))
+
+	flushTimer := c.sinkWriterScope.Timer("unsent_events_write_all").Start()
+	for _, event := range events {
+		lag := time.Since(event.OccurredAt.AsTime())
+		c.sinkWriterScope.Timer("event_lag").Record(lag)
```
mikecutalo, line 158: nice

backend/service/audit/audit.go
```diff
@@ -137,29 +139,44 @@ func (c *client) ReadEvents(ctx context.Context, start time.Time, end *time.Time
...
+
+		for _, s := range c.sinks {
+			c.sinkWriterScope.Counter("event_write.attempt")
```
mikecutalo, line 161: do you think that this is superfluous? We should be able to compute this with success and fail metrics, however this is easier and less computation is required.

backend/service/audit/audit.go
```diff
@@ -179,11 +199,15 @@ func (c *client) poll() {
 		}
 
 		if isLocked {
+			c.scope.Counter("lock_acquired").Inc(1)
```
mikecutalo, line 202: nice

mikecutalo (approved):

backend/service/audit/audit.go
danielhochman: i'll drop it for now. i have seen alignment issues with counters where they don't add up correctly due to flushing on the edge of a window but that's such a rare case.

backend/service/audit/audit.go
danielhochman: i added error just in case

PR #2440, opened 2022-10-25
K8s Resize HPA: add label for text fields on mix/max size row

scarlettperry:
### Description
For the K8s Resize HPA flow, a label is added above the text fields on the Mix/Max Size to clarify what they are. To allow this customization, PR adds a new prop to the metadata-table component for providing these labels.

### Testing Performed
Local
Screen Shot 2022-10-25 at 1 55 08 PM

**Example of a workflow where the new prop isn't set**
Screen Shot 2022-10-25 at 1 00 36 PM

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -19,6 +19,7 @@ import { object } from "yup";
 import { useWizardContext } from "../Contexts";
 import TextField from "../Input/text-field";
 import styled from "../styled";
+import { Typography } from "../typography";
 
 interface RowData {
   input?: {
@@ -28,6 +29,7 @@ interface RowData {
   };
   name: string;
   value: unknown;
+  disabledFieldlabel?: string;
 }
 
 interface IdentifiableRowData extends RowData {
@@ -137,6 +139,9 @@ interface MutableRowProps extends ImmutableRowProps {
   validation: UseFormReturn<FieldValues, object>;
 }
 
+// TODO: fix Grid styling to account for labels over the text field or revisit implementation
+// (ex. instead of a disabled text field and editable text field, remove disabled field and have a reset icon next to text field
+// to reset field to the default value
 const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, validation }) => {
   const error = validation?.formState?.errors?.[data.name];
 
@@ -149,6 +154,9 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
     <TableRow key={data.id}>
       <KeyCell data={data} />
       <TableCell>
+        {data.disabledFieldlabel && (
+          <Typography variant="body4">{data.disabledFieldlabel}</Typography>
+        )}
         <Grid>
           <div className="textfield-disabled">
             <TextField disabled id={data.id} name={data.name} defaultValue={data.value} />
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -149,6 +154,9 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
     <TableRow key={data.id}>
       <KeyCell data={data} />
       <TableCell>
+        {data.disabledFieldlabel && (
+          <Typography variant="body4">{data.disabledFieldlabel}</Typography>
+        )}
```
scarlettperry, line 159: this is a hack as it should really go inside here https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/Table/metadata-table.tsx#L153-L155 but ran into styling issues. I was able to fix some but those new changes then caused other styling issues (examples below), so went with the hack for now. left a TODO to come back to fix the styling or change the design to be a text field and reset icon (like we discussed in the meeting)

**ex 1**
Screen Shot 2022-10-25 at 9 55 29 AM

**ex 2**
Screen Shot 2022-10-25 at 9 56 19 AM

**ex 3**
Screen Shot 2022-10-25 at 9 56 49 AM

frontend/workflows/k8s/src/resize-hpa.tsx
```diff
@@ -74,6 +74,7 @@ const HPADetails: React.FC<WizardChild> = () => {
           {
             name: "Min Size",
             value: hpa.sizing.minReplicas,
+            disabledFieldlabel: "Current size",
```
scarlettperry, line 77: we have a few workflows that have mutable rows on the metadata table so thought it'd be better to make the label for the disabled text field be prop based so that workflows can customize it/enable it (rather than setting `Current` by default) . lmk if you have other preferences for the prop name

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -75,7 +76,7 @@ const TableCell = styled(MuiTableCell)({
   fontSize: "14px",
   fontWeight: "normal",
   height: "48px",
-  padding: "0 16px",
+  padding: "8px 16px",
```
scarlettperry, line 79: added a bit of top and bottom padding, otherwise the the label the row looks like this:

Screen Shot 2022-10-25 at 10 52 40 AM

danielhochman:
If we're going to label the greyed out box I think we should also label the new value box.

scarlettperry:
> If we're going to label the greyed out box I think we should also label the new value box.

good idea, i'll update

frontend/workflows/k8s/src/resize-hpa.tsx
```diff
@@ -74,6 +74,10 @@ const HPADetails: React.FC<WizardChild> = () => {
           {
             name: "Min Size",
             value: hpa.sizing.minReplicas,
+            textFieldLabels: {
+              disabledField: "Current size",
```
danielhochman, line 78: I think Current Min, New Min is easier to follow so each row has different labels

scarlettperry: updated

Screen Shot 2022-10-25 at 1 55 08 PM

danielhochman (approved):

jdslaugh (approved):

PR #2441, opened 2022-10-25
audit sink sql: use dedicated connection for advisory locks

mikecutalo:
### Description
Use a dedicated connection for the audit sink advisory lock. Previously we were using the db session pool for all advisory lock requests, the issue with this is that in order for a session level advisory lock to be unlock, the unlock request must come from the same session. This [post](https://engineering.qubecinema.com/2019/08/26/unlocking-advisory-locks.html) summarizes the issue we are running into nicely.

In this pr I also moved the `readAndFanout()` before calling `ReleaseLock()`, this was the original intent but this was not setup correctly.

### Testing Performed
unit

### DIFF
--- a/backend/service/audit/storage/sql/sql.go
+++ b/backend/service/audit/storage/sql/sql.go
@@ -230,6 +230,12 @@ func (c *client) ReleaseLock(ctx context.Context, lockID uint32) (bool, error) {
 		c.logger.Error("Unable to perform an advisory unlock", zap.Error(err))
 		return false, err
 	}
+
+	// conn should never be nil but this guards against a panic in the event that it is
+	if conn != nil {
+		conn.Close()
+	}
+
 	return unlock, nil
 }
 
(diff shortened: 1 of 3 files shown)


backend/service/audit/storage/sql/sql.go
```diff
@@ -187,18 +187,46 @@ func (c *client) query(ctx context.Context, query string, args ...interface{}) (
...
 
 func (c *client) ReleaseLock(ctx context.Context, lockID uint32) (bool, error) {
+	conn, err := c.getAdvisoryConn()
+	if err != nil {
+		return false, err
+	}
+
 	var unlock bool
-	if err := c.db.QueryRowContext(ctx, "SELECT pg_advisory_unlock($1)", lockID).Scan(&unlock); err != nil {
+	if err := conn.QueryRowContext(ctx, "SELECT pg_advisory_unlock($1)", lockID).Scan(&unlock); err != nil {
```
danielhochman, line 229: one optimization would be to invoke Close() after unlocking so the connection is returned to the pool so each worker is not always holding one connection constantly for this.

danielhochman (approved):
lgtm!

miguelmolina95 (approved):

PR #2442, opened 2022-10-25
audit: save filter configuration on service

danielhochman:
### Description
We were not filtering before because the filter configuration was not saved on the struct.


lyft-tla (approved):
+1

PR #2454, opened 2022-11-04
dynamodb: use chips for status indicators [frontend]

angethuy:
### Description

This PR adds chips to the status indicators for table/indexes, for more clarity and easier parsing in the table results page. 

image

### Testing Performed

Tested locally with custom gateway

### GitHub Issue

Fixes #


frontend/workflows/dynamodb/src/update-capacity.tsx
```diff
@@ -55,6 +56,18 @@ const TableDetails: React.FC<TableDetailsChild> = ({ enableOverride, notes = []
...
+      case status.toString() === "ACTIVE":
+        return "active";
+      case status.toString() === "UPDATING" || status.toString() === "CREATING":
+        return "pending";
+      case status.toString() === "DELETING":
+        return "warn";
+      default:
+        return "neutral";
+    }
+  };
```
jdslaugh, line 70: Would probably adjust this to something like below, makes it a bit easier to read and extend

```suggestion
    switch (status.toString()) {
      case "ACTIVE":
        return "active";
      case "UPDATING":
      case "CREATING":
        return "pending";
      case "DELETING":
        return "warn";
      default:
        return "neutral";
    }
  };
```

danielhochman (approved):

Issue #37, opened 2020-07-18
core: display audit trail in UI


danielhochman:
**Description**

Add an audit log viewer to the frontend, on top of the audit API defined in https://github.com/lyft/clutch/blob/master/api/audit/v1/audit.proto.

Complexity [S/M/L]: M

**Wireframes, Mockups, or Screenshots (if applicable)**
TODO

autisticvegan:
Hi, I will be joining Lyft on Feb 8th full-time and was pointed towards this issue since I will be joining this team and had some extra time :).  Here is my doc for my plan:  
https://docs.google.com/document/d/1xmX_gNwxevH4QORS5ujUqzkpQo5lf_O7Smy5MYqC6Uc/edit?usp=sharing

I will schedule office hours for next Tuesday to ask about this.

PR #2468, opened 2022-11-17
backend: add audit get event API and pagination support

dschaller:
### Description

Add implementation for Audit Module get event API as well as pagination support for get events API

### Testing Performed

Unit / Manual

### GitHub Issue

Partial Fix for #37


backend/module/audit/audit.go
```diff
@@ -75,12 +76,49 @@ func (m *mod) GetEvents(ctx context.Context, req *auditv1.GetEventsRequest) (*au
...
+		end := offset + limit
+		if end < len(resp.Events) {
+			resp.NextPageToken = strconv.FormatInt(int64(page+1), 10)
+		}
+
+		if end > len(resp.Events) {
+			end = len(resp.Events)
+		}
+
+		resp.Events = resp.Events[offset:end]
```
danielhochman, line 108: should we try to push this to the query layer instead?

dschaller: yeah, we could but I was thinking that would be a breaking change since the query would need to take the start and stop props. Maybe I can do it in a separate PR, what do you think?

danielhochman: i'd go ahead and do it now, an audit query could take down the DB or service otherwise with the potential for the amount of data it could return

dschaller: couldn't they currently take it down now if the range is large enough?

miguelmolina95 (approved):
reminds of a certain interview question

backend/service/audit/storage/sql/sql.go
```diff
@@ -132,17 +132,80 @@ func (c *client) UnsentEvents(ctx context.Context) ([]*auditv1.Event, error) {
...
+			readEventsRangeStatement = fmt.Sprintf(`
+			%s
+			LIMIT $3
+			`, readEventsRangeStatement)
+			args = append(args, options.Limit)
+		case options.Offset != 0:
+			readEventsRangeStatement = fmt.Sprintf(`
+			%s
+			OFFSET $4
+			`, readEventsRangeStatement)
```
danielhochman, line 187: nit: Can we change these to: 
```suggestion
			readEventsRangeStatement = fmt.Sprintf(`%s OFFSET $4`, readEventsRangeStatement)
```

I'm not sure the spacing really helps with readability.

backend/module/audit/audit.go
```diff
@@ -52,35 +51,71 @@ func (m *mod) Register(r module.Registrar) error {
...
+		p, err := strconv.Atoi(req.PageToken)
+		if err != nil || p < 0 {
+			return nil, fmt.Errorf("invalid page token: %s", req.PageToken)
+		}
+		page = p
+	}
+
+	limit := int(req.Limit)
+	if req.Limit == 0 {
+		limit = 1
```
danielhochman, line 87: is this the proper default?

backend/module/audit/audit.go
```diff
@@ -52,35 +51,71 @@ func (m *mod) Register(r module.Registrar) error {
...
+	options := &audit.ReadOptions{Offset: int64(startIdx)}
+	if req.Limit != 0 {
+		options.Limit = additionalEventsNumber
+	}
+	events, err := m.client.ReadEvents(ctx, start, end, options)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(events) != 0 && req.Limit != 0 && len(events) == int(additionalEventsNumber) {
```
danielhochman, line 101: please add comments throughout these functions but especially places like this

backend/module/audit/audit.go
```diff
@@ -52,35 +51,71 @@ func (m *mod) Register(r module.Registrar) error {
...
+
+	limit := int(req.Limit)
+	if req.Limit == 0 {
+		limit = 1
+	}
+
+	startIdx := page * limit
+	additionalEventsNumber := int64(limit + 1)
+	options := &audit.ReadOptions{Offset: int64(startIdx)}
+	if req.Limit != 0 {
```
danielhochman, line 93: is this check necessary? don't we have a default limit?

danielhochman (commented):
nice, getting close, a couple of questions

backend/module/audit/audit.go
dschaller: at the moment we don't have a default. you get all the events back. In a prior commit I was defaulting to 10 but that would modify the existing behavior

backend/module/audit/audit.go
dschaller: This is only used when determining the page offset if the request limit is 0. I left a response to your comment below about the limit but this behavior maintains backwards compatibility with todays behavior where if you request audit events all are returned.

dschaller:
ptal @danielhochman

danielhochman (approved):

PR #2469, opened 2022-11-17
frontend: Timeago Component

jdslaugh:
### Description
- Moves timeago component out of ProjectCatalog and into core

#### With short=true
Screen Shot 2022-11-17 at 2 25 37 PM

#### With short=false
Screen Shot 2022-11-17 at 2 25 45 PM

#### Story
Screen Shot 2022-11-18 at 8 29 16 AM

### Testing Performed
manual


frontend/packages/core/src/time.tsx
```diff
@@ -0,0 +1,41 @@
...
+const parseTimestamp = (timestamp?: number | Long | null): number => {
+  return parseInt(timestamp?.toString() || "0", 10);
+};
+
+const setMilliseconds = (timestamp?: number | Long | null): number => {
+  const ts = new Date(0);
+  return ts.setUTCMilliseconds(parseTimestamp(timestamp));
+};
+
+const EventTime = ({ short = true, onClick, ...props }: EventTimeProps) => (
```
dschaller, line 29: Should we call this something like `TimeAgo` or `ElapsedTime`? `EventTime` seems a bit generic and doesn't denote that it has a live count.

dschaller (commented):
Can you add a story and tests?

frontend/packages/core/src/time.tsx
jdslaugh: Yeah TimeAgo makes sense, will update

frontend/packages/core/src/timeago.tsx
```diff
@@ -0,0 +1,47 @@
+import React from "react";
+import type { ReactTimeagoProps as TimeAgoProps } from "react-timeago";
+import ReactTimeago from "react-timeago";
+
+interface EventTimeProps extends Pick<TimeAgoProps, "date"> {
+  onClick?: () => void;
+  short?: boolean;
+  live?: boolean;
```
scarlettperry, line 8: what does `live` do? i toggled it and didn't see a change

jdslaugh: It really only works if you're in the seconds (put in a new timestamp of now from epoch converter and toggle it then). It just auto increments the time for you instead of being static.

Yeah I'll add a comment

scarlettperry (approved):

frontend/packages/core/src/timeago.tsx
dschaller: maybe we should call it static and invert the defaults / logic

frontend/packages/core/src/timeago.tsx
```diff
@@ -1,10 +1,11 @@
 import React from "react";
 import type { ReactTimeagoProps as TimeAgoProps } from "react-timeago";
-import TimeAgo from "react-timeago";
+import ReactTimeago from "react-timeago";
 
 interface EventTimeProps extends Pick<TimeAgoProps, "date"> {
   onClick?: () => void;
```
dschaller, line 6: Should we pick `onClick` from `TimeAgoProps` as well?

frontend/packages/core/src/timeago.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+/**
+ * Will take a millsecond timestamp in and calculate the timeago for it
+ * @param date the millisecond timestamp
+ * @param short (default) will shorten the unit string (day -> d)
+ * @param live (default) will auto increment based on a given time
+ * @returns react component representing the timeago
+ */
+const TimeAgo = ({ short = true, onClick, ...props }: EventTimeProps) => (
```
dschaller, line 37: I'm wondering if it makes sense to have date be a `Date` type and then we don't have to rely on users passing in a number that may or may not be correct

frontend/packages/core/src/timeago.tsx
dschaller: Shouldn't this also work in the minutes and hours if you have the page open long enough?

jdslaugh: It would, its just easiest to see from a testing perspective in seconds

jdslaugh: Static is a reserved keyword, so that seems to not work

frontend/packages/core/src/timeago.tsx
jdslaugh: Picked it, but it made it a required field so I reverted it back for now

frontend/packages/core/src/timeago.tsx
jdslaugh: Actually it does seem to be a date type already from Timeago, so I've modified the comment.

miguelmolina95 (approved):

PR #2470, opened 2022-11-18
frontend: add support for table overflow configuration

dschaller:
### Description

add a new prop to `Table` for controlling overflow layout.

### Testing Performed

manual

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -31,6 +31,7 @@ const StyledTable = styled(MuiTable)<{
   $columnCount: number;
   $compress: boolean;
   $responsive?: TableProps["responsive"];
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Table/table.tsx
```diff
@@ -122,6 +125,8 @@ interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   actionsColumn?: boolean;
   /** Make table responsive */
   responsive?: boolean;
```
scarlettperry, line 127: is this not needed here anymore as well?

dschaller: It's still being picked from the TableProps

scarlettperry (approved):

jdslaugh (approved):

PR #2480, opened 2022-12-05
FE: Kinesis: make shard count unique

smonero:
### Description

Screenshot 2022-12-05 at 10 24 54 AM
Right now, users could potentially be confused as to why the same shard count values are showing up multiple times in the select. This PR makes it so that only unique values are shown.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/kinesis/src/update-shard-count.tsx
+++ b/frontend/workflows/kinesis/src/update-shard-count.tsx
@@ -55,17 +55,20 @@ const StreamDetails: React.FC<WizardChild> = () => {
     resourceData.updateData("targetShardCount", value);
   };
 
-  const values = [
-    Math.ceil(stream.currentShardCount * 0.5),
-    Math.ceil(stream.currentShardCount * 0.75),
-    Math.ceil(stream.currentShardCount * 1),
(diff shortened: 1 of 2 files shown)


frontend/workflows/kinesis/src/update-shard-count.tsx
```diff
@@ -55,17 +55,12 @@ const StreamDetails: React.FC<WizardChild> = () => {
...
-    Math.ceil(stream.currentShardCount * 2),
-  ];
-  const options = values.map(value => {
-    return { label: value.toString() };
+  // These will be used to multiply the currentShardCount, which will populate the choices
+  const multipliers = [0.5, 0.75, 1, 1.25, 1.5, 1.75, 2];
+
+  const values = multipliers.map(m => Math.ceil(stream.currentShardCount) * m);
+  const options = values.map((value, index) => {
+    return { label: `${value.toString()}(ceil of current * ${multipliers[index]})` };
```
jdslaugh, line 63: I wonder if we should do this or just uniq the list so we dont have duplicates, probably doesn't make a lot of sense to have the dupes anyways

jdslaugh (approved):

PR #2481, opened 2022-12-06
FE: Metadata Table: add tool tip prop for disabled text field

smonero:
### Description

Screenshot 2022-12-06 at 12 42 04 PM

Its nice to be able to have a tooltip for a disabled text field.

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -11,13 +11,13 @@ import {
   TableCell as MuiTableCell,
   TableContainer as MuiTableContainer,
   TableRow,
-  Tooltip,
 } from "@mui/material";
 import _ from "lodash";
 import type { BaseSchema } from "yup";
 import { object } from "yup";
 
 import { useWizardContext } from "../Contexts";
+import { Tooltip } from "../Feedback/tooltip";
 import TextField from "../Input/text-field";
 import styled from "../styled";
 
@@ -33,7 +33,6 @@ interface RowData {
   };
   name: string;
   value: unknown;
-  disabledFieldlabel?: string;
   disabledFieldTooltip?: string;
 }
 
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -152,19 +154,27 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
   const updateCallback = (e: React.ChangeEvent<HTMLTextAreaElement | HTMLInputElement>) =>
     error ? () => {} : onUpdate(e);
 
+  const disabledTextFieldComponent = (
+    <TextField
+      disabled
+      id={data.id}
+      name={data.name}
+      defaultValue={data.value}
+      label={data.textFieldLabels?.disabledField ?? data.textFieldLabels?.disabledField}
```
Yiwen-Gao, line 163: Should the second value be something else?
```suggestion
      label={data.textFieldLabels?.disabledField}
```

Yiwen-Gao:
Would the main use case just be that the tooltip explains why the textfield is disabled?

frontend/packages/core/src/Table/metadata-table.tsx
scarlettperry: shawna & synced in standup as i originally wrote that. yeah we can change to the suggestion or `data.textFieldLabels?.disabledField ?? ""`

smonero:
> 

I think so

jdslaugh:
So my understanding from the issue was that because the field was disabled there was no way to see the entire value of the field (because it was also cut off), so the goal of adding a tooltip to the disabled field would just be to show the current value on hover.

Yiwen-Gao:
Hmm in that case, wouldn't it be better to just add a test that the value inside the textfield is fully visible, even if it's many characters long?

jdslaugh:
I think that could be useful but the tooltip would probably still be nice, just have it map to the value directly instead of being custom defined.

jdslaugh (approved):
I like it

PR #2482, opened 2022-12-06
FE: Resize HPA: add a warning on confirmation if new min or max replicas are different by more than 50%

smonero:
### Description

In the case where the new min or new max replicas are different than the old min or max by more than 50%, display a warning note indicating as such.

### Testing Performed

Tested on staging (see private PR for screenshot)

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/workflows/k8s/src/resize-hpa.tsx
+++ b/frontend/workflows/k8s/src/resize-hpa.tsx
@@ -137,12 +137,12 @@ const HPADetails: React.FC<WizardChild> = () => {
   );
 };
 
-// TODO (sperry): possibly show the previous size values
 const Confirm: React.FC<ConfirmChild> = ({ notes }) => {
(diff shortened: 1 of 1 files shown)


smonero:
> it looks like this note is shown at the end of the flow. I think it'd be better to have an alert popup when they're setting the new sizes to raise this warning sooner, wdyt?

I made a different ticket, I decided this one will just be the note on the confirmation page, then there will be another PR that will have a warning on the resize page.

frontend/workflows/k8s/src/resize-hpa.tsx
```diff
@@ -44,6 +44,8 @@ const HPADetails: React.FC<WizardChild> = () => {
     hpaData.updateData(key, value);
   };
 
+  const oldHpaData = useDataLayout("oldHpaData");
```
dschaller, line 47:
```suggestion
  const currentHpaData = useDataLayout("currentHpaData");
```

miguelmolina95 (approved):

PR #2487, opened 2022-12-09
API: add sizing field for resize HPA request

smonero:
### Description

It is useful to be able to add some non-required fields to the Resize HPA Request.

1. The old sizing (just like the new sizing, but contains the old values)

### GitHub Issue

Fixes #


api/k8s/v1/k8s.proto
```diff
@@ -553,6 +553,11 @@ message ResizeHPARequest {
     uint32 max = 2;
   }
   Sizing sizing = 5 [ (validate.rules).message = {required : true} ];
+
+  Sizing old_sizing = 6;
+
+  // A flag to add sirens to the audit message or not
+  bool add_warning = 7;
```
scarlettperry, line 560: and can an example be added (ie if the new sizing is much greater/less than old sizing)
```suggestion
  bool resize_warning = 7;
```

scarlettperry (approved):

api/k8s/v1/k8s.proto
```diff
@@ -553,6 +553,11 @@ message ResizeHPARequest {
     uint32 max = 2;
   }
   Sizing sizing = 5 [ (validate.rules).message = {required : true} ];
+
+  Sizing old_sizing = 6;
+
+  // A flag to add sirens to the audit message or not
+  bool resize_warning = 7;
```
dschaller, line 560: Just curious, why have the client determine if a warning should be shown?

api/k8s/v1/k8s.proto
```diff
@@ -553,6 +553,11 @@ message ResizeHPARequest {
     uint32 max = 2;
   }
   Sizing sizing = 5 [ (validate.rules).message = {required : true} ];
+
+  Sizing old_sizing = 6;
```
dschaller, line 557: nit 
```suggestion
  Sizing current_sizing = 6;
```

api/k8s/v1/k8s.proto
smonero: Not really a good reason, but we can't do math inside the YAML file, so we would need to add some interception to the audit / aimbot functionality, which I figured was more involved than we wanted. Maybe I could just split things and have this PR just be the `current_sizing` and leave the warning for another?

miguelmolina95 (approved):

PR #2490, opened 2022-12-12
Frontend: Add set height for Card component

jecr:
### Description
Card components in the landing page adjusted their height according to the card's description length. That caused disparity in the cards sizes. This update sets the Card height to a fixed 214px.

Before:
image

After:
image

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/card.tsx
+++ b/frontend/packages/core/src/card.tsx
@@ -176,6 +176,7 @@ interface CardContentProps extends SpacingProps {
   collapseAction?: CardContentCollapsibleProps;
   /** The max height of the card content. The default is none. */
   maxHeight?: number | "none";
+  className?: string;
 }
 
 const CardContent = ({
@@ -244,7 +245,7 @@ const StyledLandingCard = styled(Card)({
   height: "214px",
   maxHeight: "100%",
 
-  "& .cardActionArea, & .cardActionArea > div, & .cardActionArea > div > div": {
+  "& .cardActionArea": {
     height: "inherit",
     maxHeight: "inherit",
   },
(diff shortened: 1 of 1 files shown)


smonero:
what happens if the text is too long to fit inside the card? As in, if there is a super long description or title could that cause problems?

jecr:
> what happens if the text is too long to fit inside the card? As in, if there is a super long description or title could that cause problems?

image
The text will get cut at the card's max-height and some of if will be left out of view, I tried getting the nested elements to max out at the card's height using flexbox, in order to add `text-overflow: ellipsis`, but I couldn't prevent the inner elements from overflowing.

jdslaugh:
> 

So looking into this a bit, it looks like adding ellipsis is rather difficult through pure CSS for multiline text. So we have two options, set a maximum length and truncate it manually using `substring` or we can try doing it via css `line-clamp`.

To do that you would add a styled component to the typography and do something like this:

```
const TruncatedText = styled(Typography)({
  display: "-webkit-box",
  "-webkit-box-orient": "vertical",
  "-webkit-line-clamp": "3",
});
...
          <TruncatedText color="rgba(13, 16, 48, 0.6)" variant="body2">
            {description}
          </TruncatedText>
```

We'll need to verify this works in multiple browsers before implementing, however I think it's the cleanest and most extensible variation.

jecr:
Added @jdslaugh's suggested wrapper for Typography the description. According to MDN it has [full support](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) in browsers.

image

smonero (approved):

jdslaugh (approved):

PR #2491, opened 2022-12-13
Housekeeping: Updated docs with paths for local development of features

jecr:
### Description
Adding a few paths that were unclear while following the Feature development guide. Both the path_to_gateway and the location of clutch.config.js differ when locally developing a feature without using a custom gateway.

image

image

### Testing Performed
manual

### DIFF
--- a/docs/development/feature.md
+++ b/docs/development/feature.md
@@ -48,6 +48,10 @@ For the reasons stated above, the API defined below will largely look identical 
 
 Clutch uses protobuf for interface definitions. An empty definition is provided below.
 
+:::info
+When adding a feature directly in the clutch project, `path_to_gateway` should be `lyft/clutch`
+:::
+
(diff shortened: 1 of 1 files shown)


smonero (approved):

docs/development/feature.md
```diff
@@ -787,6 +791,10 @@ export default register;
 
 Next, open your `clutch.config.js` file and add the amiibo configuration highlighted below:
 
+:::info
+When adding a feature directly in the clutch project, the configuration file will be located in `frontend/packages/app/src/clutch.config.js`
```
dschaller, line 795: Same comment above applies here

jecr: This case is more generic I think, since the org scope is not used. It was not immediately clear when following the guide of what file should be changed if a custom gateway was not used. But like the comment above, maybe there is another way of identifying the config file without adding here the specific route?

PR #2492, opened 2022-12-13
frontend: Button Sizing Extension

jdslaugh:
### Description
MUI supports multiple button sizes (https://mui.com/material-ui/react-button/#sizes) which we did not currently support. Adding support for them here as well as rotating the tests to use React Testing Library.

This will also be utilized as part of the NPS Banner component.

#### Small
Screenshot 2022-12-13 at 9 47 17 AM

#### Medium (current default)
Screenshot 2022-12-13 at 9 47 20 AM

#### Large
Screenshot 2022-12-13 at 9 47 23 AM

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/button.tsx
+++ b/frontend/packages/core/src/button.tsx
@@ -95,18 +95,34 @@ const colorCss = (palette: ButtonPalette) => {
   };
 };
 
-const StyledButton = styled(MuiButton)<{ palette: ButtonPalette }>(
+const BUTTON_SIZE_MAP = {
+  small: {
+    height: "32px",
+    padding: "7px 32px",
+  },
+  medium: {
+    height: "48px",
+    padding: "14px 32px",
+  },
+  large: {
+    height: "64px",
+    padding: "21px 32px",
+  },
+};
+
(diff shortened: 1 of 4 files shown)


smonero (approved):

frontend/packages/core/src/button.tsx
```diff
@@ -144,20 +160,23 @@ const variantPalette = (variant: ButtonVariant): ButtonPalette => {
 };
 
 export interface ButtonProps
-  extends Pick<MuiButtonProps, "disabled" | "endIcon" | "onClick" | "startIcon" | "type"> {
+  extends Pick<
+    MuiButtonProps,
+    "id" | "disabled" | "endIcon" | "onClick" | "startIcon" | "type" | "size"
```
scarlettperry, line 165: curious why `id` is picked

jdslaugh: To target the button in the NPS Banner component so that I could detect whether it was being pressed or somewhere in the body and cancel the normal close action that the Header uses

scarlettperry (approved):

PR #2494, opened 2022-12-14
frontend: NPS Banner Component

jdslaugh:
### Description
An NPS banner component which will trigger the NPSHeader component to open via the ApplicationContext.

Utilized the ApplicationContext since it is not exported from core so it is safer to touch for NPS. Attempted to make the functionality extensible so it could be used by any component in the Header, would love ideas if anyone has any improvements.

Also added stories for both the NPS Wizard and NPS Header, which required some modifications to get working due to feature flags.

Relies on: https://github.com/lyft/clutch/pull/2492 & https://github.com/lyft/clutch/pull/2493

#### StoryBook
Screenshot 2022-12-13 at 5 21 56 PM

#### GIF
Kapture 2022-12-13 at 17 21 04

### Testing Performed
manual


frontend/packages/core/src/Contexts/app-context.tsx
```diff
@@ -2,9 +2,21 @@ import React from "react";
...
 import type { Workflow } from "../AppProvider/workflow";
 
+export type HeaderItems = "NPS";
+
+export interface TriggeredHeaderData {
+  [key: string]: {
+    open: boolean;
+    [key: string]: unknown;
+  };
+}
```
jdslaugh, line 12: Currently made this very generic so we could support other options easily. However I can see a case for it being strongly typed as we're unlikely to have a large amount of items in the Header so I'm open to opinions

frontend/packages/core/src/paper.tsx
```diff
@@ -14,6 +14,8 @@ const StyledPaper = styled(MuiPaper)({
   minHeight: "inherit",
 });
 
-const Paper = ({ children }: PaperProps) => <StyledPaper>{children}</StyledPaper>;
+const Paper = ({ children, ...props }: PaperProps) => (
+  <StyledPaper {...props}>{children}</StyledPaper>
+);
```
jdslaugh, line 19: Not honestly sure why this is necessary but it would allow me to override the padding using styled components without it. Originally I had added this to use the `sx` prop but it wound up working for this use case so I left it.

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -120,9 +122,22 @@ const ClutchApp: React.FC<ClutchAppProps> = ({
...
-  const appContextValue = React.useMemo(() => ({ workflows: discoverableWorkflows }), [
-    discoverableWorkflows,
-  ]);
+  const appContextValue = React.useMemo(
+    () => ({
+      workflows: discoverableWorkflows,
+      triggerHeaderItem: (item: HeaderItems, open: boolean, data: unknown) =>
+        // Will set the open status and spread any additional data onto the property named after the item
+        setTriggeredHeaderData({
+          ...triggeredHeaderData,
```
scarlettperry, line 131: should we rename these to include `feedback` or `nps` or see this as being reusable for non-feedback flows?

frontend/packages/core/src/Contexts/app-context.tsx
```diff
@@ -2,9 +2,21 @@ import React from "react";
 
 import type { Workflow } from "../AppProvider/workflow";
 
+export type HeaderItems = "NPS";
```
scarlettperry, line 5: will this be a list of items in the future? just checking b/c the naming is a lil unclear

scarlettperry (commented):
some small comments

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -15,6 +15,14 @@ import { UserInformation } from "./user";
 
 export const APP_BAR_HEIGHT = "64px";
 
+/**
+ * Properties used to allow stories to override featureflag settings
+ */
+interface HeaderProps extends AppConfiguration {
+  showShortLinks?: boolean;
+  showNPS?: boolean;
```
dschaller, line 23: can we add some docstrings to these so users know what they mean (specifically NPS since I think it might be unclear what that could stand for)

frontend/packages/core/src/AppLayout/header.tsx
```diff
@@ -40,7 +48,12 @@ const StyledLogo = styled("img")({
   verticalAlign: "middle",
 });
 
-const Header: React.FC<AppConfiguration> = ({ title = "clutch", logo = <Logo /> }) => {
+const Header: React.FC<HeaderProps> = ({
+  title = "clutch",
+  logo = <Logo />,
+  showNPS = false,
+  showShortLinks = false,
```
dschaller, line 55: Aren't shortlinks only visible if the page supports it? Would it be weird to have this conditional be true but have links not show on some pages? Maybe we should call this something like `enableShortLinks` instead of show just to make it clear we are enabling functionality but it doesn't mean that there will always be short links shown.

frontend/packages/core/src/AppProvider/index.tsx
jdslaugh: I made this initially so it could be reused for non-feedback flows, if we think it wont ever be reused though I can rename and strongly type it though

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -120,9 +122,22 @@ const ClutchApp: React.FC<ClutchAppProps> = ({
     [workflowSessionStore]
   );
 
-  const appContextValue = React.useMemo(() => ({ workflows: discoverableWorkflows }), [
-    discoverableWorkflows,
-  ]);
+  const appContextValue = React.useMemo(
+    () => ({
+      workflows: discoverableWorkflows,
+      triggerHeaderItem: (item: HeaderItems, open: boolean, data: unknown) =>
```
dschaller, line 128: Would it make sense to have this solely be for opening header items? My thought here is that if a header item is already open then there is probably no point in having this be used to close it since a click away would probably do that anyways. It would also simplify this logic since we would only need to support triggering an open.

frontend/packages/core/src/Contexts/app-context.tsx
dschaller, line 5:
```suggestion
export type HeaderItem = "NPS";
```

dschaller: I think this will end up being a union e.g. `"NPS" | "Foobar" but Josh can confirm

frontend/packages/core/src/Contexts/app-context.tsx
```diff
@@ -2,9 +2,21 @@ import React from "react";
...
+
+export interface TriggeredHeaderData {
+  [key: string]: {
```
dschaller, line 8: won't the key here be a `HeaderItem`?

frontend/packages/core/src/NPS/stories/banner.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+                componentProps: { resolverType: "clutch.aws.ec2.v1.AutoscalingGroup" },
+                description: "Resize an autoscaling group.",
+                displayName: "Resize Autoscaling Group",
+                path: "asg/resize",
+                requiredConfigProps: ["resolverType"],
+              },
+            ],
+          },
+        ],
+        triggerHeaderItem: (item: HeaderItems, open: boolean, data: unknown) =>
```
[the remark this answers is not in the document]
jdslaugh: All of the properties are handled via the docs tab in storybook, trying to get away from adding an individual story if it can be handled already

Kapture 2023-01-09 at 11 13 59

frontend/packages/core/src/NPS/banner.tsx
```diff
@@ -0,0 +1,98 @@
...
+export const Banner = ({
+  integrated = false,
+  icon = <HappyEmoji />,
+  feedbackText = "Enjoying this feature? We would like your feedback!",
+  feedbackButtonText = "Give Feedback",
+  defaultOption,
+}: BannerFeedbackProps) => {
+  const { triggerHeaderItem } = useAppContext();
+  const banner = (
+    <Grid container direction="row" spacing={1} alignItems="center">
```
dschaller, line 57: really minor but wdyt about center justifying the whole thing? On larger screens I feel like there could be a ton of white space on the right with the current left justification

frontend/packages/core/src/Contexts/app-context.tsx
jdslaugh: Yep, @dschaller is correct, I can add a comment to clarify

frontend/packages/core/src/NPS/banner.tsx
```diff
@@ -0,0 +1,98 @@
...
+import Paper from "../paper";
+import styled from "../styled";
+import { Typography } from "../typography";
+
+export interface BannerFeedbackProps {
+  /**
+   * Whether this component should appear integrated into the page versus in its own container
+   * @defaultValue false
+   */
+  integrated?: boolean;
```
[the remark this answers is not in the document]
dschaller: 👍 should we call this something like `elevated` to mirror the prop naming for material's paper?

frontend/packages/core/src/NPS/banner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  integrated = false,
+  icon = <HappyEmoji />,
+  feedbackText = "Enjoying this feature? We would like your feedback!",
+  feedbackButtonText = "Give Feedback",
+  defaultOption,
+}: BannerFeedbackProps) => {
+  const { triggerHeaderItem } = useAppContext();
+  const banner = (
+    <Grid container direction="row" spacing={1} alignItems="center">
+      <Grid item sx={{ marginTop: "4px" }} data-testid="nps-banner-icon">
```
dschaller, line 58: I think you might have added this margin to make the icon center aligned vertically with the text. You should be able to add `display: flex` to the icon and have this render the same without needing to hard code the margin.

frontend/packages/core/src/NPS/banner.tsx
```diff
@@ -0,0 +1,98 @@
...
+          {feedbackText}
+        </Typography>
+      </Grid>
+      <Grid item sx={{ marginLeft: "16px" }}>
+        <Button
+          id="nps-banner-button"
+          data-testid="nps-banner-button"
+          variant="neutral"
+          text={feedbackButtonText}
+          size="small"
```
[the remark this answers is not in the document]
jdslaugh: I can see it both ways, the button is meant to draw attention but it does look slightly large compared to the text. I'll see if I can play around with it. Material-UI doesn't have an XSmall so it'll have to be custom sizing

frontend/packages/core/src/AppLayout/header.tsx
jdslaugh: Actually good call out, it wont work anyways since the ShortLinks component actually does the check and returns null if it's not a valid workflow. I'll remove it.

frontend/packages/core/src/AppProvider/index.tsx
jdslaugh: Yeah I was going back and forth on this one, trying to think of an instance where we might want to close it outside of the header. I'll remove the open check and simplify the logic and if we run into an instance where it's necessary we can always extend it

frontend/packages/core/src/Contexts/app-context.tsx
jdslaugh: Hah yeah, it should be. Let me update

frontend/packages/core/src/NPS/banner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * Whether this component should appear integrated into the page versus in its own container
+   * @defaultValue false
+   */
+  integrated?: boolean;
+  /**
+   * The icon to display or null if none
+   * @defaultValue <HappyEmoji />
+   */
+  icon?: EmojiType | React.ReactNode | null;
```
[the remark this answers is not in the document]
jdslaugh: It's optional but since it has a default we need to be able to override it

frontend/packages/core/src/NPS/banner.tsx
```diff
@@ -0,0 +1,98 @@
...
+  ) : (
+    <StyledPaper data-testid="nps-banner-container">{banner}</StyledPaper>
+  );
+};
+
+/**
+ * An NPS Feedback Banner which will ask the user for feedback then open up the NPSHeader
+ * NOTE: requires the NPSHeader to be enabled
+ */
+const BannerFeedback = ({ ...props }: BannerFeedbackProps) => (
```
dschaller, line 90:
```suggestion
const FeedbackBanner = ({ ...props }: FeedbackBannerProps) => (
```

frontend/packages/core/src/NPS/banner.tsx
dschaller, line 57: Smaller screens have no padding on top.
Screen Shot 2023-01-09 at 12 02 30 PM

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
dschaller: ah I see, when you say override do you mean explicitly setting it to null in the prop value?

frontend/packages/core/src/NPS/banner.tsx
jdslaugh: I think it looks pretty good, I'll modify it

Screenshot 2023-01-09 at 2 26 08 PM

jdslaugh: Modified it to use a breakpoint and add padding at that point

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
jdslaugh: Yeah, `<NPSBanner icon={null} />`, we could remove the default though and leave it up to the implementing component as well.

frontend/packages/core/src/AppProvider/index.tsx
jdslaugh: Removed the open boolean and am using whether the data is present or not to signify it's been triggered (which it was already doing anyways).

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
jdslaugh: What about this? Made an xsmall size with a height of `24px` and a padding of `0px 16px`.

Small is normally a height of `32px` with padding of `7px 32px`.

Screenshot 2023-01-09 at 2 43 27 PM

frontend/packages/core/src/Contexts/app-context.tsx
jdslaugh: Added a comment and an example to indicate usability

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
scarlettperry: yeah it looks too small 🤔 the button size in the screenshot above looks better

scarlettperry (approved):
looks good to me! not sure if @dschaller also wants to take another pass

i left 1 comment about the button size being too small but that can be adjusted in another PR. maybe we should sync with Kanchan on the xsmall sizing

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
jdslaugh: My thinking exactly, lets discuss during the design sync

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
dschaller: @jdslaugh what do you think about the naming here?

dschaller:
Looks good to me as well! Reopened one comment about prop naming for `elevated` vs `integrated`

dschaller (approved):

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
jdslaugh: Totally missed this comment, whoops. Didn't actually realize MUI had a prop for elevated, yeah I think that makes sense since elevated=0 is basically what we're achieving here. Let me adjust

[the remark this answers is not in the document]
jdslaugh: Adjusted to have elevated be the default now, and false would be what used to be integrated.

frontend/packages/core/src/NPS/banner.tsx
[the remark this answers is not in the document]
jdslaugh: Adjusted the button sizing back to 24px, left padding at `0px 8px` currently

PR #2501, opened 2022-12-16
frontend: migrate all uses of @emotion/styled to @clutch-sh/styled

jdslaugh:
### Description
Since we have an internal wrapper for @emotion/styled, converting all instances of @emotion/styled to use the wrapper component and removing any references to imports for @emotion/styled. This should allow for easier upgrades as time goes on.

### Testing Performed
manual


smonero (approved):

dschaller (approved):
Nice! Thanks for the housekeeping

PR #2502, opened 2022-12-19
frontend: add ellipsis prop to text fields

jecr:
### Description
The numeric input fields, both disabled and enabled truncated their contents with no indication that additional text could be hidden after the cut. This could lead to a wrong reading of the stored value.
A prop `truncateWithEllipsis` was added to the TextField component to add an ellipsis (`...`) whenever the content of the input overflows. The prop is leveraged by MetadataTable by passing a flag `truncateWithEllipsis` in the body of `RowData`.

Max Size without the flag enabled, Desired Size with the flag enabled
image

### Testing Performed
Manual

### DIFF
--- a/frontend/packages/core/src/Table/metadata-table.tsx
+++ b/frontend/packages/core/src/Table/metadata-table.tsx
@@ -34,6 +34,7 @@ interface RowData {
   name: string;
   value: unknown;
   disabledFieldTooltip?: string;
+  truncateWithEllipsis?: boolean;
 }
 
 interface IdentifiableRowData extends RowData {
@@ -160,6 +161,7 @@ const MutableRow: React.FC<MutableRowProps> = ({ data, onUpdate, onReturn, valid
       name={data.name}
       defaultValue={data.value}
       label={data.textFieldLabels?.disabledField}
+      truncateWithEllipsis={data.truncateWithEllipsis}
     />
   );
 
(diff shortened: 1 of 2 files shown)


smonero (approved):

frontend/packages/core/src/Table/metadata-table.tsx
```diff
@@ -34,6 +34,7 @@ interface RowData {
   name: string;
   value: unknown;
   disabledFieldTooltip?: string;
+  truncateWithEllipsis?: boolean;
```
dschaller, line 37: should we have this be extended from the text field props

danielhochman:
Do we need a prop for this? Seems like we could blanket apply it. I can't think of a scenario where you would want it disabled.

jecr:
> How does a user know the exact value? Do you need to hover on the input or something? If so add a screenshot to demonstrate please.

There is a tooltip for disabled input fields, it is at the MetadataTable component level.
image

jecr:
> Do we need a prop for this? Seems like we could blanket apply it. I can't think of a scenario where you would want it disabled.

Maybe we can reverse the flag and use it to disable the feature to maintain that flexibility, wdyt?

dschaller:
I was just rereading this and was just think the same thing as @danielhochman. I think let's just always show it since it's removes the reliance on users to set the prop

jecr:
Adjusted to show ellipsis on text overflow by default for all text inputs.

image

jdslaugh:
> Adjusted to show ellipsis on text overflow by default for all text inputs.
> 
> image

I think we might want to limit this to only disabled fields, it might be confusing on a field that can accept input if all of a sudden the input has an ellipsis

jdslaugh (approved):

PR #2505, opened 2022-12-28
frontend: hoist alert stories

jecr:
### Description
The Alert component is subdivided by severity and title in Storybook, adding items to the stories navigation panel. By using controls the stories were unified to a single entry with the same title/severities options.

Before:
image

After:
image

### Testing Performed
Manual

### DIFF
--- a/frontend/packages/core/src/Assets/global.ts
+++ b/frontend/packages/core/src/Assets/global.ts
@@ -32,3 +32,5 @@ export const StyledSVG = styled.svg<{ size?: IconSizeVariant }>(props => ({
 export interface SVGProps extends React.SVGProps<SVGSVGElement> {
   size?: IconSizeVariant;
 }
+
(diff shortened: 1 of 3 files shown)


smonero (approved):

frontend/packages/core/src/Assets/global.ts
```diff
@@ -32,3 +32,5 @@ export const StyledSVG = styled.svg<{ size?: IconSizeVariant }>(props => ({
 export interface SVGProps extends React.SVGProps<SVGSVGElement> {
   size?: IconSizeVariant;
 }
+
+export const SEVERITIES = ["error", "info", "success", "warning"];
```
jdslaugh, line 36: Lets keep this in the alert component file instead of the global assets. We can also do:

```
export const SEVERITIES = Object.keys(iconMappings);
```

jdslaugh (approved):

PR #2506, opened 2022-12-29
frontend: hoist note stories

jecr:
### Description
The Note component is subdivided by severity in Storybook, adding items to the stories navigation panel. By using controls the stories were unified to a single entry with the same severities options.

Before:
image

After:
image

### Testing Performed
Manual

### DIFF
--- a/frontend/packages/core/src/Assets/global.ts
+++ b/frontend/packages/core/src/Assets/global.ts
@@ -32,3 +32,5 @@ export const StyledSVG = styled.svg<{ size?: IconSizeVariant }>(props => ({
 export interface SVGProps extends React.SVGProps<SVGSVGElement> {
   size?: IconSizeVariant;
 }
+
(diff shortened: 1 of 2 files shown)


smonero (approved):

frontend/packages/core/src/Assets/global.ts
```diff
@@ -32,3 +32,5 @@ export const StyledSVG = styled.svg<{ size?: IconSizeVariant }>(props => ({
 export interface SVGProps extends React.SVGProps<SVGSVGElement> {
   size?: IconSizeVariant;
 }
+
+export const SEVERITIES = ["error", "info", "success", "warning"];
```
jdslaugh, line 36: Lets either drop this into the note component or hard code it into the note.stories, the Assets/global is more so for SVG Assets, so it doesn't really apply here

jdslaugh (approved):

PR #2528, opened 2023-01-06
frontend: Select - support multiple

colek319:
### Description

Modifies Select component to support multiple values. MUI Select uses an array type as the value (https://mui.com/material-ui/react-select/#multiple-select) when multiple is true. This PR changes the internal representation of selected values to an array, even when multiple is False, to support type consistency for the component.

Screenshot 2023-01-06 at 16 33 35

### Testing Performed

Manual.

### DIFF
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -189,6 +189,7 @@ export interface SelectProps extends Pick<MuiSelectProps, "disabled" | "error"> 
   name: string;
   options: SelectOption[];
   onChange?: (value: string) => void;
+  multiple?: boolean;
 }
 
 export const Select = ({
@@ -200,6 +201,7 @@ export const Select = ({
   name,
   options,
   onChange,
+  multiple = false,
 }: SelectProps) => {
   // Flattens all options and sub grouped options for easier retrieval
   const flatOptions: BaseSelectOptions[] = flatten(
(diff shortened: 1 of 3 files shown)


frontend/packages/core/src/Input/select.tsx
```diff
@@ -241,8 +249,18 @@ export const Select = ({
...
-    setSelectedIdx(flatOptions.findIndex(opt => opt?.value === value || opt?.label === value));
-    onChange && onChange(value);
+
+    if (multiple) {
+      const newSelectedIdxs = value.map(val =>
+        flatOptions.findIndex(opt => opt.value === val || opt.label === val)
+      );
+      setSelectedIdxs(newSelectedIdxs);
+    } else {
+      const index = flatOptions.findIndex(opt => opt.value === value || opt.label === value);
```
colek319, line 259: nit: maybe add a comment here for clarity

frontend/packages/core/src/Input/select.tsx
```diff
@@ -241,8 +249,18 @@ export const Select = ({
...
+      setSelectedIdxs([index]);
+    }
```
jdslaugh, line 261: We can probably simplify this a bit:

```suggestion
const findIndex = (val) => flatOptions.findIndex(opt => opt.value === val || opt.label === val);
setSelectedIdxs(multiple ? value.map(val => findIndex(val)) : [findIndex(value)]);
```

frontend/packages/core/src/Input/select.tsx
```diff
@@ -218,20 +220,26 @@ export const Select = ({
...
+  // Returns the list of selected values
+  const selectedValues = () => {
+    return selectedIdxs.map(idx => flatOptions[idx].value || flatOptions[idx].label);
+  };
 
   React.useEffect(() => {
     if (flatOptions.length !== 0) {
-      onChange && onChange(flatOptions[selectedIdx]?.value || flatOptions[selectedIdx].label);
+      const onChangeInput: string[] = selectedValues();
+      onChange && onChange(onChangeInput.join(","));
```
jdslaugh, line 242: Probably dont need this variable

```suggestion
      onChange && onChange((selectedValues() ?? []).join(","));
```

frontend/packages/core/src/Input/select.tsx
```diff
@@ -218,20 +220,26 @@ export const Select = ({
...
+    return index >= 0 ? [index] : [0];
   };
 
-  const [selectedIdx, setSelectedIdx] = React.useState(calculateDefaultOption());
+  const [selectedIdxs, setSelectedIdxs] = React.useState(calculateDefaultOption());
+
+  // Returns the list of selected values
+  const selectedValues = () => {
+    return selectedIdxs.map(idx => flatOptions[idx].value || flatOptions[idx].label);
+  };
```
jdslaugh, line 237: nit:
```suggestion
  const selectedValues = () => selectedIdxs.map(idx => flatOptions[idx].value || flatOptions[idx].label);
```

jdslaugh (approved):

PR #2532, opened 2023-01-10
frontend: convert core tests to rtl

jecr:
### Description
As part of a migration of unit tests to React Test Library, the tests for the Landing, Tabs and NotFound components were moved away from the use of enzyme.

image

### Testing Performed
manual


frontend/packages/core/src/tests/tab-context.test.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import "@testing-library/jest-dom";
+
+import { Tab, Tabs } from "../tab";
+
+jest.mock("@mui/lab/TabContext", () => {
+  return jest.fn(() => null);
+});
+
+test("TabContext is called with specified value", () => {
```
jdslaugh, line 13: From what we talked about, lets just remove this test as it tries to test internal logic and isn't necessary for verification of functionality or the presentation layer. We'll extend the tabs test functionality after we've migrated.

frontend/packages/core/src/tests/landing.test.tsx
```diff
@@ -1,23 +1,22 @@
...
+import { render } from "@testing-library/react";
+
+import "@testing-library/jest-dom";
 
 import * as appContext from "../Contexts/app-context";
 import Landing from "../landing";
 
-describe("Landing component", () => {
-  let component;
+beforeAll(() => {
```
jdslaugh, line 10: Is the beforeAll required here? Since we're only mocking it once it should be fine to leave it as the base level right?

frontend/packages/core/src/tests/tab.test.tsx
```diff
@@ -1,43 +1,28 @@
...
+      <Tab label="mix" />
+    </Tabs>
+  );
+});
 
-    it("renders children tabs", () => {
-      expect(component.find(Tab)).toHaveLength(2);
-    });
+test("renders correctly", () => {
+  expect(screen.getByTestId("styled-tabs")).toBeInTheDocument();
```
jdslaugh, line 18: We should probably make this `.toBeVisible()` so that we can make sure it's not only in the DOM but also visible to the user.

jdslaugh (commented):
Some small suggestions, looks great!

jdslaugh (approved):
Looks great!

PR #2535, opened 2023-01-12
frontend: adding warning message if new value is more than 50% different

roier:
### Description
Adding warning message if the change in min or max values is more than 50% difference between current and new values

https://user-images.githubusercontent.com/1994723/212132285-e7fe2aea-2a39-4538-b593-bdb91771a2f1.mov

### GitHub Issue

Fixes #

### DIFF
--- a/frontend/packages/core/src/Input/stories/text-field.stories.tsx
+++ b/frontend/packages/core/src/Input/stories/text-field.stories.tsx
@@ -3,7 +3,7 @@ import SearchIcon from "@mui/icons-material/Search";
 import type { Meta } from "@storybook/react";
 
 import type { TextFieldProps } from "../text-field";
-import TextField from "../text-field";
+import { TextField } from "../text-field";
 
 export default {
   title: "Core/Input/TextField",
--- a/frontend/packages/core/src/Input/stories/text-field.stories.tsx
+++ b/frontend/packages/core/src/Input/stories/text-field.stories.tsx
@@ -8,6 +8,13 @@ import TextField from "../text-field";
 export default {
   title: "Core/Input/TextField",
   component: TextField,
+  argTypes: {
+    color: {
+      options: ["primary", "secondary", "error", "info", "success", "warning"],
+      control: { type: "select" },
+      defaultValue: "primary",
+    },
+  },
 } as Meta;
 
 const Template = (props: TextFieldProps) => <TextField {...props} />;
@@ -15,6 +22,8 @@ const Template = (props: TextFieldProps) => <TextField {...props} />;
 export const Primary = Template.bind({});
 Primary.args = {
   label: "My Label",
+  color: "primary",
+  error: false,
   placeholder: "This is a placeholder, start typing",
 };
 
--- a/frontend/packages/core/src/Input/text-field.tsx
+++ b/frontend/packages/core/src/Input/text-field.tsx
@@ -2,6 +2,7 @@ import * as React from "react";
 import type { FieldValues, UseFormRegister } from "react-hook-form";
 import styled from "@emotion/styled";
 import ErrorIcon from "@mui/icons-material/Error";
+import WarningIcon from "@mui/icons-material/Warning";
 import type {
   InputProps as MuiInputProps,
   StandardTextFieldProps as MuiStandardTextFieldProps,
@@ -46,6 +47,7 @@ const StyledAutocomplete = styled(Autocomplete)({
 
 const StyledTextField = styled(BaseTextField)({
   "--error-color": "#DB3615",
+  "--warning-color": "#FFCC80",
   height: "unset",
 
   ".MuiInputLabel-root": {
@@ -59,6 +61,10 @@ const StyledTextField = styled(BaseTextField)({
     "&.Mui-error": {
       color: "var(--error-color)",
     },
+  },
+
+  "&.Mui-warning .MuiInputLabel-root:not(.Mui-error), .MuiFormHelperText-root": {
+    color: "var(--warning-color)",
   },
 
   ".MuiInputBase-root": {
@@ -99,6 +105,12 @@ const StyledTextField = styled(BaseTextField)({
         opacity: 1,
       },
     },
+  },
+
+  "&.Mui-warning .MuiInputBase-root:not(.Mui-error) fieldset": {
+    "--input-border-width": "1px",
+    borderColor: "var(--warning-color)",
+    borderWidth: "var(--input-border-width)",
   },
 
   ".MuiInputBase-adornedEnd": {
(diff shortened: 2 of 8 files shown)


frontend/workflows/k8s/src/resize-hpa.tsx
```diff
@@ -49,6 +50,18 @@ const HPADetails: React.FC<WizardChild> = () => {
   const metadataAnnotations = [];
   const metadataLabels = [];
 
+  const [warnings, setWarnings] = React.useState<{
+    minSize: NoteConfig;
+    maxSize: NoteConfig;
+  }>({ minSize: undefined, maxSize: undefined });
+
+  const warningThreshold = 0.5;
```
smonero, line 58: let's put constants like this in caps i.e. WARNING_THRESHOLD

smonero (commented):
looks great! I'll let josh take a look if he's interested but if he's busy then I'll approve at the end of the day. Thanks!

frontend/workflows/k8s/src/resize-hpa.tsx
roier: should I update the `alertLevel` variable as well?

smonero: hm yeah sorry about that I probably should've done that. I guess I did that because it was inside the useEffect rather than outside. But yeah lets have both be all caps. Thanks!

jdslaugh (commented):
Great work!

Some thoughts, I wonder if we should handle this in a different way. MUI TextFields already handle a color property (https://mui.com/material-ui/api/text-field/) which supports a warning level. We should probably extend that onto the TextField and have a custom color for it. 

You can see how our Button component does this here: https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/button.tsx#L123

This would allow us to support warnings via `color="warning"` as well as any future colors we see fit.

jdslaugh (commented):
Thanks for the updates!

Still thinking we can do this different however.

Here's my thinking:

We would have a function which takes a color and styles all the relevant css props correctly, such as below:
```
const COLOR_MAP = {
  warning: "1234",
};

const colorTextField = (color: MuiStandardTextFieldProps["color"]) => {
  ".MuiInputLabel-root": {
    color: COLOR_MAP[color],
  },
  ...
};
```

This would be an extension to the current styled field like this:
```
const StyledTextField = styled(BaseTextField)<{ color?: MuiStandardTextFieldProps["color"] }>(
  {
    ...
  },
  props => ({
    ...(props.color ? colorTextField(props.color) : {}),
  })
);
```

Then we would remove the `warning` prop from the TextField and just utilize `color` by extending the pick from `MuiStandardTextFieldProps`. This would mean that the resize-hpa flow would no longer set `warning` but instead `color="warning"`. 

We could even have a mapping of icons to display based on the color (warning would be the only one necessary for now).

The purpose of this would be to support things like 'success' or 'info' or possibly any arbitrary color with a smaller amount of overhead work.  

Let me know if you want to discuss this in a 1:1 and we can dive a little deeper. Thanks!

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -44,39 +45,42 @@ const StyledAutocomplete = styled(Autocomplete)({
   },
 });
 
-const StyledTextField = styled(BaseTextField)({
-  "--error-color": "#DB3615",
-  height: "unset",
+const TEXT_FIELD_COLOR_MAP = {
```
jdslaugh, line 48: Lets go ahead and add options for `primary`, `secondary`, `info` and `success` that way they're supported as possible options. 

Right now they're available via typing but do nothing which could be a little odd.

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -113,9 +117,9 @@ const StyledTextField = styled(BaseTextField)({
     marginTop: "7px",
     lineHeight: "16px",
     marginLeft: "0px",
-
+    color: `${TEXT_FIELD_COLOR_MAP[props.color] || TEXT_FIELD_COLOR_MAP.default}`,
```
jdslaugh, line 120: Not relevant to this line but we also need to update the placeholder color to pull from the map now

frontend/packages/core/src/Input/text-field.tsx
```diff
@@ -44,39 +45,42 @@ const StyledAutocomplete = styled(Autocomplete)({
...
 
-    "& fieldset": {
-      borderColor: "var(--input-default-color)",
-    },
     "&.Mui-focused fieldset": {
-      borderColor: "#3548d4",
+      borderColor: `${TEXT_FIELD_COLOR_MAP[props.color] || TEXT_FIELD_COLOR_MAP.focused}`,
       borderWidth: "var(--input-border-width)",
     },
+    "&:not(.Mui-focused) fieldset": {
```
jdslaugh, line 79: Looks like this is breaking the hover border color change when there isn't a color selected, you can see the difference here: https://storybook.clutch.sh/?path=/story/core-input-textfield--primary

jdslaugh (commented):
Some small nits just to button everything up

jdslaugh:
Thanks for updating the secondary color, only issue Im seeing now is that error is not overwriting the color property. It looks like the hover effect is working correctly for the error property but as you can see in the screenshots the border is not updating.

#### Error=true with color=primary
Screenshot 2023-01-24 at 9 44 16 AM

#### Error=true
Screenshot 2023-01-24 at 9 44 41 AM

frontend/packages/core/src/Input/stories/text-field.stories.tsx
```diff
@@ -8,13 +8,22 @@ import TextField from "../text-field";
 export default {
   title: "Core/Input/TextField",
   component: TextField,
+  argTypes: {
```
jdslaugh, line 11: So to make this easier, we can auto import the available options for storybook. It's a bit hacky for some of them but the gist is this:

https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/Input/text-field.tsx#L385
```diff
- const TextField = React.forwardRef<HTMLDivElement, TextFieldProps>(TextFieldRef);
+ export const TextField = React.forwardRef<HTMLDivElement, TextFieldProps>(TextFieldRef);
```

https://github.com/lyft/clutch/blob/main/frontend/packages/core/src/Input/stories/text-field.stories.tsx#L6
```diff
- import TextField from "../text-field";
+ import { TextField } from "../text-field";
```

There's a weird bug that doesn't allow default exports to work correctly and they must be a named export

jdslaugh (approved):
Left 1 minor comment

PR #2537, opened 2023-01-12
frontend: migrate input tests to rtl

jecr:
### Description
As part of the migration of unit tests to React Testing Library, the input/ component tests for Select and DateTimePicker were updated.

image

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/Input/tests/date-time.test.tsx
+++ b/frontend/packages/core/src/Input/tests/date-time.test.tsx
@@ -29,7 +29,7 @@ test("onChange is called when valid value", () => {
   expect(onChange).toHaveBeenCalled();
 });
 
-test("onChange is not called when valid value", () => {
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/tests/date-time.test.tsx
```diff
@@ -1,81 +1,70 @@
...
-    let onChange: () => void;
-    let component;
-    beforeAll(() => {
-      date = new Date();
-      onChange = () => {};
-      component = mount(
-        <DateTimePicker value={date} onChange={onChange} label="testing" disabled={false} />
-      );
-    });
+test("onChange is not called when valid value", () => {
```
jdslaugh, line 32:
```suggestion
test("onChange is not called when invalid value", () => {
```

jdslaugh (approved):
Small nit

PR #2539, opened 2023-01-12
frontend: Select - support multiple

colek319:
### Description

Creates a new select component that supports multiple values.

https://user-images.githubusercontent.com/22361209/223369219-43464568-2f8e-4f00-accf-7e6547d582c8.mov

### Testing Performed

Added a happy path test.

### DIFF
--- a/frontend/packages/core/src/Input/select.tsx
+++ b/frontend/packages/core/src/Input/select.tsx
@@ -215,6 +215,31 @@ const renderSelectItems = (option: SelectOption) => {
   return menuItemFromOption(option);
 };
 
+// Will take an array of strings or integers and attempt to find the indexes where they exist based on the flattened items
+const calculateDefaultOptions = (
+  defaultOptions: Array<number> | Array<string>,
+  flattenedOptions: BaseSelectOptions[]
+): Array<number> => {
+  const options = [];
+
+  if (defaultOptions === undefined || defaultOptions.length === 0) {
+    return options;
+  }
+
+  defaultOptions.forEach(option => {
+    if (Number.isInteger(option)) {
+      options.push(option < flattenedOptions.length && option > 0 ? option : 0);
+    }
+
+    // we're a string, lets look it up based on the value/label and default to 0 if none
+    const index = flattenedOptions?.findIndex(
+      opt => opt?.value === option || opt?.label === option
+    );
+    options.push(index >= 0 ? index : 0);
+  });
+  return options;
+};
+
 export interface SelectProps extends Pick<MuiSelectProps, "disabled" | "error"> {
   defaultOption?: number | string;
   helperText?: string;
@@ -222,7 +247,6 @@ export interface SelectProps extends Pick<MuiSelectProps, "disabled" | "error"> 
   name: string;
   options: SelectOption[];
   onChange?: (value: string) => void;
-  flex?: boolean;
 }
 
 const Select = ({
@@ -234,31 +258,17 @@ const Select = ({
   name,
   options,
   onChange,
-  flex = false,
 }: SelectProps) => {
   // Flattens all options and sub grouped options for easier retrieval
   const flatOptions: BaseSelectOptions[] = flattenBaseSelectOptions(options);
+  const defaultOptions = calculateDefaultOptions(
+    [defaultOption] as Array<number> | Array<string>,
+    flatOptions
+  );
 
-  // Will take a string or an integer and attempt to find the index where it exists based on the flattened items
-  const calculateDefaultOption = () => {
-    let option = defaultOption;
-
-    // handle empty case
-    if (option === undefined || option === "") {
-      option = 0;
-    }
-
-    if (Number.isInteger(option)) {
-      return option < flatOptions.length && option > 0 ? option : 0;
-    }
-
-    // we're a string, lets look it up based on the value/label and default to 0 if none
-    const index = flatOptions?.findIndex(opt => opt?.value === option || opt?.label === option);
-
-    return index >= 0 ? index : 0;
-  };
-
-  const [selectedIdx, setSelectedIdx] = React.useState(calculateDefaultOption());
+  const [selectedIdx, setSelectedIdx] = React.useState(
+    defaultOptions.length > 0 ? defaultOptions[0] : 0
+  );
 
   React.useEffect(() => {
     if (flatOptions.length !== 0) {
@@ -310,7 +320,6 @@ export interface MultiSelectProps extends Pick<MuiSelectProps, "disabled" | "err
   name: string;
   selectOptions: SelectOption[];
   onChange?: (values: Array<string>) => void;
-  flex?: boolean;
 }
 
 const MultiSelect = ({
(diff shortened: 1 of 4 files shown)


frontend/packages/core/src/Input/select.tsx
```diff
@@ -241,8 +245,10 @@ export const Select = ({
     if (!value) {
       return;
     }
-    setSelectedIdx(flatOptions.findIndex(opt => opt?.value === value || opt?.label === value));
-    onChange && onChange(value);
+    const findIndex = val => flatOptions.findIndex(opt => opt.value === val || opt.label === val);
+    setSelectedIdxs(multiple ? value.map(val => findIndex(val)) : [findIndex(value)]);
+
+    onChange && onChange(selectedIdxs.join(","));
```
jdslaugh, line 251: I think this is the problem, we're returning the selectedIdxs joined and not their actual values like we do above

```suggestion
    onChange && onChange((selectedValues() ?? []).join(","));
```

colek319: yeah thats def it lol. Let me add a breaking test and we can re-review after the fix

jdslaugh: Sounds good!

frontend/packages/core/src/Input/select.tsx
```diff
@@ -218,31 +219,34 @@ export const Select = ({
...
 
     // we're a string, lets look it up based on the value/label and default to 0 if none
     const index = flatOptions?.findIndex(opt => opt?.value === option || opt?.label === option);
 
-    return index >= 0 ? index : 0;
+    return index >= 0 ? [index] : [0];
   };
 
-  const [selectedIdx, setSelectedIdx] = React.useState(calculateDefaultOption());
+  const [selectedIdxs, setSelectedIdxs] = React.useState(calculateDefaultOption());
```
dschaller, line 231: super nit: should we rename this function to `calculateDefaultOptions` now?

frontend/packages/core/src/Input/select.tsx
```diff
@@ -279,8 +283,9 @@ export const Select = ({
       {label && <StyledInputLabel>{label}</StyledInputLabel>}
       {flatOptions.length && (
         <StyledSelect
+          multiple={multiple}
           id={`${name}-select`}
-          value={flatOptions[selectedIdx]?.value || flatOptions[selectedIdx].label}
+          value={selectedValues()}
           onChange={updateSelectedOption}
```
dschaller, line 289: If we are adding multiselect support should we change the onChange callback to pass in a list of strings instead of a single one? This way users don't need to manage state themselves.

colek319: We are going to implement MultiSelect over Select to avoid version conflcits.

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+  selectOptions,
+  onChange,
+  flex = false,
+}: MultiSelectProps) => {
+  // Flattens all options and sub grouped options for easier retrieval
+  const flatOptions: BaseSelectOptions[] = flatten(
+    selectOptions.map((option: SelectOption) =>
+      option.group ? option.group.map(groupOption => groupOption) : option
+    )
+  );
```
colek319, line 329: could probably move this to a separate function

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+  error,
+  helperText,
+  label,
+  name,
+  selectOptions,
+  onChange,
+  flex = false,
+}: MultiSelectProps) => {
+  // Flattens all options and sub grouped options for easier retrieval
+  const flatOptions: BaseSelectOptions[] = flatten(
```
jdslaugh, line 325: We should probably pull this out into a helper function since its now duplicated

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+    selectOptions.map((option: SelectOption) =>
+      option.group ? option.group.map(groupOption => groupOption) : option
+    )
+  );
+
+  // Will take a string or an integer and attempt to find the index where it exists based on the flattened items
```
jdslaugh, line 331:
```suggestion
  // Will take an array of strings or integers and attempt to find the indexes where they exist based on the flattened items
```

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+    return menuItem(option);
+  };
+
+  return (
+    <MuiFormControl id={name} key={name} disabled={disabled} error={error} fullWidth>
+      {label && <StyledInputLabel>{label}</StyledInputLabel>}
+      {flatOptions.length && (
+        <StyledSelect
+          multiple
+          id={`${name}-select`}
```
jdslaugh, line 409:
```suggestion
          id={`${name}-multi-select`}
```

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+
+  const renderSelectItems = option => {
+    if (option.group) {
+      return [
+        <ListSubheader>{option.label}</ListSubheader>,
+        option.group.map(opt => menuItem(opt)),
+      ];
+    }
+    return menuItem(option);
+  };
```
jdslaugh, line 401: I think we can bring these both out into helper functions as well since they're duplicated.

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+}: MultiSelectProps) => {
+  // Flattens all options and sub grouped options for easier retrieval
+  const flatOptions: BaseSelectOptions[] = flatten(
+    selectOptions.map((option: SelectOption) =>
+      option.group ? option.group.map(groupOption => groupOption) : option
+    )
+  );
+
+  // Will take a string or an integer and attempt to find the index where it exists based on the flattened items
+  const calculateDefaultOption = (): Array<number> => {
```
jdslaugh, line 332: Im tempted to say we pull this out into a helper function as well. The biggest difference is that it takes an array input which would be easy enough to support by casting it beforehand.

frontend/packages/core/src/Input/select.tsx
```diff
@@ -300,4 +300,128 @@ export const Select = ({
...
+const MultiSelect = ({
+  defaultOptions = [],
+  disabled,
+  error,
+  helperText,
+  label,
+  name,
+  selectOptions,
+  onChange,
+  flex = false,
```
jdslaugh, line 322: I dont actually see this used anywhere, and have no idea what it does. We can probably just remove it.

frontend/packages/core/src/Input/stories/multi-select.stories.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from "react";
+import type { Meta } from "@storybook/react";
+
+import type { MultiSelectProps } from "../select";
+import { MultiSelect } from "../select";
+
+export default {
+  title: "Core/Input/MultiSelect",
+  component: MultiSelect,
+  argTypes: {
```
jdslaugh, line 10: I dont think you need this block, it should pull from the params correctly

frontend/packages/core/src/Input/select.tsx
colek319: going to try it out

jdslaugh (approved):
Thanks for taking the time to do this!

PR #2541, opened 2023-01-13
ensure that yarn is installed

alileza:
### Description
the documentation [here](https://clutch.sh/docs/getting-started/local-build) doesn't seem to include the part where we need to run `tools/install-yarn.sh` to generate the build directory. This update should make new people that just the repo an try to clone easier

### Testing Performed
tested locally make

### GitHub Issue
n/a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ PROJECT_ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 	@grep -E '^\.PHONY: [a-zA-Z_-]+ .*?# .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = "(: |#)"}; {printf "%-30s %s\n", $$2, $$3}'
 
 .PHONY: all # Generate API, Frontend, and backend assets.
-all: preflight-checks api frontend backend-with-assets
+all: install-yarn preflight-checks api frontend backend-with-assets
 
 .PHONY: api # Generate API assets.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,7 +16,7 @@ PROJECT_ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 	@grep -E '^\.PHONY: [a-zA-Z_-]+ .*?# .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = "(: |#)"}; {printf "%-30s %s\n", $$2, $$3}'
 
 .PHONY: all # Generate API, Frontend, and backend assets.
-all: preflight-checks api frontend backend-with-assets
+all: install-yarn preflight-checks api frontend backend-with-assets
```
jdslaugh, line 19: Looks like we already have a command for installing yarn `yarn-ensure` so we should be able to just swap out `install-yarn` with `yarn-ensure` and remove the extra make command.

Makefile
```diff
@@ -197,3 +197,5 @@ preflight-checks:
 util-bump-aws:
 	cd backend && go list -m -f '{{if not .Indirect}}{{.Path}}{{end}}' all | grep aws-sdk-go | tr '\n' ' ' | xargs go get -u
 
+install-yarn:
```
jdslaugh, line 200: See above about removing this

jdslaugh (commented):
Thanks for doing this, small comments

PR #2549, opened 2023-01-23
frontend: audit workflow fix state bug

dschaller:
### Description

Fixes two issues with the audit workflow:
- allows users to resubmit queries even after selecting and submitting a time range (the only limitation is users cannot query within a second)
- If the server sends back a response with the next page token empty it's stored in state and when a subsequent query is submitted we attempt to reset the page token but this doesn't happen before `fetch` is invoked which triggers the quick return since it believes there is no next page token to fetch. This is fixed by explicitly calling fetch with a zero page token.

### Testing Performed

manual

### DIFF
--- a/frontend/workflows/audit/src/logs/event-row.tsx
+++ b/frontend/workflows/audit/src/logs/event-row.tsx
@@ -171,16 +171,17 @@ const EventRows = ({
 
   const containerRef = React.useRef(null);
 
-  const fetch = () => {
-    if (pageToken === "" || isLoading) {
+  const fetch = (page?: number) => {
+    if ((page === undefined && pageToken === "") || isLoading) {
       return;
     }
+    const tkn = page || pageToken;
     onFetch();
     setIsLoading(true);
     const data = {
       range: { startTime, endTime },
       limit: 10,
-      pageToken,
+      pageToken: tkn,
     } as IClutch.audit.v1.IGetEventsRequest;
     client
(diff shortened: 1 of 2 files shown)


frontend/workflows/audit/src/logs/index.tsx
```diff
@@ -97,13 +97,18 @@ const AuditLog: React.FC<AuditLogProps> = ({ heading, detailsPathPrefix, downloa
             onQuickSelect={(start, end) => {
               setStartTime(start);
               setEndTime(end);
-              setTimeRangeKey(`${startTime}-${endTime}`);
+              setTimeRangeKey(`${startTime}-${endTime}-${new Date().toString()}`);
```
jdslaugh, line 100: Should we break this out into it's own function that way you're not having to update 3 places?

frontend/workflows/audit/src/logs/event-row.tsx
```diff
@@ -171,16 +171,17 @@ const EventRows = ({
 
   const containerRef = React.useRef(null);
 
-  const fetch = () => {
-    if (pageToken === "" || isLoading) {
+  const fetch = (page?: string) => {
+    if ((page === undefined && page === "" && pageToken === "") || isLoading) {
```
jdslaugh, line 175: Can page be both undefined and empty at the same time? Should this be an OR?

jdslaugh (approved):

PR #2563, opened 2023-01-27
frontend: Modifying filtered workflows to be handled at the component level

jdslaugh:
### Description
Noticed we were having some issues if `hideNav: true` was set in our configuration then it would not show up in the NPS Header, however this was a necessary step to remove the item from the drawers and eventually hide them.

So this PR does a few things:

- Adds a new filterHiddenRoutes to our utils file in AppLayout, this function has the same functionality as the old one in AppProvider
- Removed the filtering in AppProvider and left the landing page check
- Modified the Search Component to use the new filterHiddenRoutes function
- Modified the Drawers Component to use the new filterHiddenRoutes function
- Modified the NPS Header Component to add to the dropdown if either `hideNav: false` or `additionalNPS` is set

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/AppLayout/utils.tsx
+++ b/frontend/packages/core/src/AppLayout/utils.tsx
@@ -127,8 +127,8 @@ const searchIndexes = (workflows: Workflow[]): SearchIndex[] => {
 };
 
 /** Filter out all of the workflows that are configured to be `hideNav: true`.
- * This prevents the workflows from being discoverable by the user from the UI,
- * both search and drawer navigation.
+ * This prevents the workflows from being discoverable by the user from the UI where used.
+ * Some example usages are in the search and drawer navigation components.
  *
  * The routes for all configured workflows will still be reachable
  * by manually providing the full path in the URI.
--- a/frontend/packages/core/src/AppLayout/utils.tsx
+++ b/frontend/packages/core/src/AppLayout/utils.tsx
@@ -1,4 +1,5 @@
 import type { Route, Workflow } from "../AppProvider/workflow";
+import * as _ from "lodash";
 
 interface GroupedRoutes {
   [category: string]: {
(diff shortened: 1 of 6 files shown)


frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -92,25 +91,12 @@ const ClutchApp: React.FC<ClutchAppProps> = ({
...
   React.useEffect(() => {
-    /** Filter out all of the workflows that are configured to be `hideNav: true`.
-     * This prevents the workflows from being discoverable by the user from the UI,
-     * both search and drawer navigation.
-     *
-     * The routes for all configured workflows will still be reachable
-     * by manually providing the full path in the URI.
-     */
-    const pw = _.cloneDeep(workflows).filter(workflow => {
+    const filteredWorkflows = workflows.filter(workflow => workflow.path === "");
```
dschaller, line 94: super nit: 
```suggestion
    const landingWorkflows = workflows.filter(workflow => workflow.path === "");
```

frontend/packages/core/src/NPS/header.tsx
```diff
@@ -58,19 +58,24 @@ export const generateFeedbackTypes = (workflows: Workflow[]): SelectOption[] =>
...
       const additionalNPS = get(route, "componentProps.additionalNPS", []);
-      if (additionalNPS) {
-        typeMap[group].push(...additionalNPS);
+
+      if (route.hideNav === undefined || route.hideNav === false || additionalNPS.length > 0) {
+        if (!typeMap[group]) {
+          typeMap[group] = [];
+        }
+
+        if (route.hideNav === undefined || route.hideNav === false) {
```
dschaller, line 69: super nit
```suggestion
      const showRoute = route.hideNav === undefined || route.hideNav === false;
      if (showRoute || additionalNPS.length > 0) {
        if (!typeMap[group]) {
          typeMap[group] = [];
        }

        if (showRoute) {
```

dschaller (approved):

frontend/packages/core/src/AppLayout/utils.tsx
```diff
@@ -124,4 +126,27 @@ const searchIndexes = (workflows: Workflow[]): SearchIndex[] => {
   return indexOptions;
 };
 
-export { routesByGrouping, searchIndexes, sortedGroupings, workflowByRoute, workflowsByTrending };
+/** Filter out all of the workflows that are configured to be `hideNav: true`.
+ * This prevents the workflows from being discoverable by the user from the UI,
+ * both search and drawer navigation.
```
dschaller, line 131: may want to change this message now that it's on a function.
```suggestion
 * This prevents the workflows from being discoverable by the user from the UI where used.
 * Some example usages are in the search and drawer navigation components.
```

jdslaugh: Ah good call out, I completely forgot to update that comment

frontend/packages/core/src/NPS/header.tsx
jdslaugh: Haha that's so much nicer, I was going through if logic hell when I wrote it :)

PR #2570, opened 2023-02-03
frontend: add children prop to error component

jecr:
### Description
In order to allow additional content to complement an error message, this prop allows to pass children to the Error component.

image

### Testing Performed
manual


danielhochman:
Can we add a story for this?

frontend/packages/core/src/Feedback/error/index.tsx
```diff
@@ -61,6 +62,7 @@ const Error = ({ subject: error, onRetry }: ErrorProps) => {
     return (
       <Alert severity="error" title={error.status?.text} action={action}>
         {error.message}
+        {children}
```
jdslaugh, line 65:
```suggestion
        {children && children}
```

jdslaugh (approved):
Small nits but looks good, thanks for adding the story

PR #2588, opened 2023-02-22
frontend: NPS Header fixing close behavior

jdslaugh:
### Description
Was found that the NPS header was not closing when clicking outside of the container after submission. This was due to the new NPS Feedback Banner open behavior which was being retriggered and keeping it open. Also decided to put in a timeout to auto-close the feedback submission.

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/NPS/header.tsx
+++ b/frontend/packages/core/src/NPS/header.tsx
@@ -7,7 +7,7 @@ import {
   Paper as MuiPaper,
   Popper as MuiPopper,
 } from "@mui/material";
-import { get, sortBy } from "lodash";
+import { get, isEmpty, sortBy } from "lodash";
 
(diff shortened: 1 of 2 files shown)


scarlettperry (approved):

frontend/packages/core/src/NPS/header.tsx
```diff
@@ -99,11 +99,18 @@ const HeaderFeedback = () => {
...
+  const timer = React.useRef(null);
 
   const handleToggle = () => {
     setOpen(!open);
   };
 
+  const timedClose = () => {
+    timer.current = setTimeout(() => {
+      setOpen(false);
+    }, 1500);
```
scarlettperry, line 111: is this milliseconds so 1.5 seconds? is that too fas if the user has opened it?

PR #2603, opened 2023-03-01
frontend: allow table row to receive extra props

jecr:
### Description
The TableRow component currently can't receive additional props, which can be useful when trying to apply styles at this component's level in a table, such as a grayed out look for a disabled state. This PR adds such behavior.

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -206,8 +206,9 @@ const TableRow = ({
   cellDefault,
   responsive = false,
   colSpan,
+  ...props
 }: TableRowProps) => (
-  <StyledTableRow onClick={onClick} $responsive={responsive}>
+  <StyledTableRow onClick={onClick} $responsive={responsive} {...props}>
     {React.Children.map(children, (value, index) => (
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Table/table.tsx
```diff
@@ -206,8 +206,9 @@ const TableRow = ({
   cellDefault,
   responsive = false,
   colSpan,
+  ...props
 }: TableRowProps) => (
-  <StyledTableRow onClick={onClick} $responsive={responsive}>
+  <StyledTableRow onClick={onClick} $responsive={responsive} {...props}>
```
jdslaugh, line 211: Is this just so we can use `sx` or something else?

jecr: It is to allow TableRow to be customised with styled and add a disabled style to the entire row, i.e. lower the opacity of the contents, my alternative was to wrap it with a styled component and target the MUI class but I figured this could be reusable.

jdslaugh: Makes sense

jdslaugh (approved):

PR #2623, opened 2023-03-22
housekeeping: Upgrading to Node v18

jdslaugh:
### Description
Now that Node v18 is in LTS, upgrading our repo to support it.

### Testing Performed
manual


danielhochman (approved):
nice!

PR #2626, opened 2023-03-23
authn: add configureable oidc groups claim

RichardWLaub:
### Description

Reviving the group portion of https://github.com/lyft/clutch/pull/2237

### Testing Performed

```yaml
      oidc:
        issuer: ...
        client_id: ${CREDENTIALS_OIDC_CLIENT_ID}
        client_secret: ${CREDENTIALS_OIDC_CLIENT_SECRET}
        groups_claim_name_override: "groups"
        redirect_url: "http://localhost:8080/v1/authn/callback"
        subject_claim_name_override: "name"
        scopes:
        - openid
        - email
        - groups
        - profile
      session_secret: this_is_my_secret
```

Configured clutch as above and authenticated via an Okta OIDC app. I decoded the jwt token cookie and saw the okta groups my user belongs to in the payload:

```json
{
  "exp": 1679606766,
  "iat": 1679603166,
  "iss": "https://myokta.okta.com",
  "sub": "Richard Laub",
  "grp": [
    "my-squad",
    "Everyone"
  ]
}
```

Also, added the same test from https://github.com/lyft/clutch/pull/2237


backend/service/authn/oidc.go
```diff
@@ -451,11 +453,33 @@ func (cc *ClaimsConfig) ClaimsFromOIDCToken(ctx context.Context, t *oidc.IDToken
 	if subject == "" {
 		return nil, fmt.Errorf("claims field %s is empty", cc.subjectClaimName)
 	}
+	var groups []string
+	if cc.groupsClaimName != "" {
+		groupsInt, ok := claims["groups"]
```
danielhochman, line 458:
```suggestion
		groupsInt, ok := claims[cc.groupsClaimName]
```

and let's add a test to catch this

api/config/service/authn/v1/authn.proto
```diff
@@ -18,6 +18,7 @@ message OIDC {
   // The subject is mapped from the JWT token's email claim by default.
   // Set this field to the JWT token's claim name to override the subject.
   string subject_claim_name_override = 6;
+  string groups_claim_name_override = 7;
```
danielhochman, line 21: Please add a comment documenting the default behavior and what the override does (like the above comment).

backend/service/authn/oidc.go
RichardWLaub: Removed the hardcoded `"groups"` and added a test that will catch that.

danielhochman (approved):

PR #2671, opened 2023-05-31
ci: use node-version for cache key

shigwata:
### Description
node-version not used for cache key.

### Testing Performed

CI

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Set up Node
         uses: actions/setup-node@v3
+        id: setup-node
         with:
           node-version: '16.x'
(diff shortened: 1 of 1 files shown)


danielhochman:
I don't see any changes related to the cache key containing node-version in these changes. Can you elaborate on the change here?

shigwata:
Since there is no 'id' in the step, the node-version is not being used in the cache key.

```
key: ${{ runner.os }}-${{ steps.setup-node.outputs.node-version }}-node-modules-${{ hashFiles('**/yarn.lock') }}
```

[before](https://github.com/lyft/clutch/actions/runs/5134697859/jobs/9239067735#step:5:10
)
[after](https://github.com/lyft/clutch/actions/runs/5128893940/jobs/9239058799?pr=2671#step:5:10)

PR #2677, opened 2023-06-06
frontend: Adding Icon Support for Drawers

jdslaugh:
### Description
Adds Icon support for Drawers. This will default to using the given path which can be locally hosted or an external URL. If not present will default to the normal avatar with the first letter.

Configured by modifying the `clutch-config.js` and adding a property `icon: { path: string }`

#### Example
```
  "@clutch-sh/ec2": {
    icon: {
      path: "/example.jpg", OR "http://fake-url.com/image.png"
    },
    terminateInstance: {
      ...
    },
  },
```

#### Screenshot
Screenshot 2023-06-06 at 8 56 05 AM

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/AppProvider/index.tsx
+++ b/frontend/packages/core/src/AppProvider/index.tsx
@@ -19,9 +19,14 @@ import { Theme } from "./themes";
 import type { ConfiguredRoute, Workflow, WorkflowConfiguration } from "./workflow";
 import ErrorBoundary from "./workflow";
 
+export interface WorkflowIcon {
+  path: string;
+}
+
 export interface UserConfiguration {
   [packageName: string]: {
-    [key: string]: ConfiguredRoute;
+    icon: WorkflowIcon;
+    [key: string]: WorkflowIcon | ConfiguredRoute;
   };
 }
 
--- a/frontend/packages/core/src/AppProvider/registrar.tsx
+++ b/frontend/packages/core/src/AppProvider/registrar.tsx
@@ -10,8 +10,11 @@ const workflowRoutes = (
 ): ConfiguredRoute[] => {
   const workflowConfig = configuration?.[workflowId] || {};
   const allRoutes = Object.keys(workflowConfig).map(key => {
-    // if workflow does not contain route with user-specified key return an empty object
-    if (workflow.routes[key] === undefined) {
+    // if workflow contains an icon, return an empty object
+    if (key === "icon") {
+      return {} as ConfiguredRoute;
+      // if workflow does not contain route with user-specified key return an empty object
+    } else if (workflow.routes[key] === undefined) {
       /* eslint-disable-next-line no-console */
       console.warn(
         `[${workflowId}][${key}] Not registered: Invalid config - route does not exist. Valid routes: ${Object.keys(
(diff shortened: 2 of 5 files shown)


scarlettperry:
would specifying an `icon` perhaps go better in the `componentprops` object since that's where we also specify if there are NPS overrides, workflow notes, etc?

jdslaugh:
> would specifying an `icon` perhaps go better in the `componentprops` object since that's where we also specify if there are NPS overrides, workflow notes, etc?

I considered that, but componentprops are workflow specific and I wanted this to be at the group level that way it didn't have to be re-specified for each workflow (and then there would be issues figuring out which one to show if they're not identical).

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -19,9 +19,14 @@ import { Theme } from "./themes";
...
 
+export interface WorkflowIcon {
+  path: string;
+}
+
 export interface UserConfiguration {
   [packageName: string]: {
-    [key: string]: ConfiguredRoute;
+    icon: WorkflowIcon;
+    [key: string]: WorkflowIcon | ConfiguredRoute;
```
scarlettperry, line 29: curious, why does WorkflowIcon need to be set in 2 places rather than just `icon` or vice versa?

jdslaugh: Yeah this one was weird. It's because we're defining `[key: string]` as an unknown property, which technically will match on `icon` as well. So it has to be OR'ed into the typing to alleviate any potential errors. We could technically remove the `icon` property and just look into the workflow configuration for it, but that seemed counter-intuitive when we know what it is.

frontend/packages/core/src/AppProvider/registrar.tsx
```diff
@@ -10,7 +10,11 @@ const workflowRoutes = (
 ): ConfiguredRoute[] => {
   const workflowConfig = configuration?.[workflowId] || {};
   const allRoutes = Object.keys(workflowConfig).map(key => {
-    // if workflow does not contain route with user-specified key return an empty object
+    // if workflow contains an icon, return an empty object
+    if (key === "icon") {
+      return {} as ConfiguredRoute;
+      // if workflow does not contain route with user-specified key return an empty object
+    }
```
scarlettperry, line 17: similar q to https://github.com/lyft/clutch/pull/2677/files#r1219985997 curious why icon is handled as part of configuredroute?

scarlettperry: answered above

frontend/packages/core/src/AppProvider/workflow.tsx
```diff
@@ -54,6 +56,7 @@ export interface Workflow extends BaseWorkflowConfiguration, WorkflowShortlinkCo
    * on homepage) and `componentProps` which allow the passing of workflow/route
    * specific props.
    */
+  icon: WorkflowIcon;
```
scarlettperry, line 59: can a code comment be added?

jdslaugh: Will do!

PR #2693, opened 2023-06-14
Add new rpc for namespace events

smonero:
### Description

We are going to have a function that lists events in a namespace rather than for a pod or given object. We could also modify the current ListEvents but I feel it is more organized to make another specific function.

### GitHub Issue

Fixes #


scarlettperry (commented):
some small comments / 1 q

scarlettperry (approved):

PR #2694, opened 2023-06-16
api: add DescribeContinuousBackups endpoint to AWS - Dynamodb

angethuy:
### Description

This PR adds a new `DescribeContinuousBackups` endpoint to the AWS dynamodb API and implements it in the dynamodb service and module. This endpoint can be used to retrieve backup information about a given table. 

### Testing Performed

Unit tests for success/error cases

### GitHub Issue

Fixes #


api/aws/dynamodb/v1/dynamodb.proto
```diff
@@ -182,6 +191,35 @@ message AttributeDefinition {
   string attribute_type = 2;
 }
 
+// Represents the description for a table's continuous backups
```
scarlettperry, line 194:
```suggestion
// Represents the description for a table's continuous backups and point in time recovery
```

backend/service/aws/dynamodb.go
```diff
@@ -140,6 +141,15 @@ func newProtoForTable(t *types.TableDescription, account, region string) *dynamo
 	return ret
 }
 
+func newProtoForContinuousBackups(t *types.ContinuousBackupsDescription, account string, region string) *dynamodbv1.ContinuousBackups {
```
scarlettperry, line 144: from the docs, looks like `ContinuousBackupsStatus` is required to be in the description output. is that true for `PointInTimeRecoveryDescription`, do we need check for nil?

scarlettperry (commented):
looks good! just 1 q

backend/service/aws/dynamodb.go
angethuy: Good catch tyty, I'll add nil checks for `PointInTimeRecoveryDescription` and its fields

scarlettperry (approved):

PR #2699, opened 2023-06-19
impl ListNamespaceEvents()

smonero:
### Description

Tested in private / staging
Have a method that will get all the events in a namespace and can be filtered by normal, warning, or error.

### GitHub Issue

Fixes #


backend/service/k8s/event.go
```diff
@@ -57,9 +56,50 @@ func ProtoForEvent(cluster string, k8sEvent *corev1.Event) *k8sapiv1.Event {
 	}
 }
 
-func (s *svc) ListNamespaceEvents(ctx context.Context, clientset, cluster, namespace string) ([]*k8sapiv1.Event, error) {
-	// TODO: actually implement the function correctly
-	return nil, errors.New("not implemented")
+func convertTypesToFieldSelector(types []k8sapiv1.EventType) string {
```
scarlettperry, line 59: no field selector would query for all event types right? perhaps we can add a code comment

scarlettperry (approved):
lgtm!

backend/service/k8s/event.go
smonero: 👍

PR #2701, opened 2023-06-22
fix field selector generation

smonero:
### Description

due to https://github.com/kubernetes/kubernetes/issues/32946 we need to do the field selection multiple times when having multiple types of items to get

### Testing Performed

unit

### GitHub Issue

Fixes #


backend/service/k8s/event.go
```diff
@@ -91,15 +88,21 @@ func (s *svc) ListNamespaceEvents(ctx context.Context, clientset, cluster, names
 	}
 
 	// Note if field selector is blank it will return everything
-	eventList, err := cs.CoreV1().Events(cs.Namespace()).List(ctx, metav1.ListOptions{FieldSelector: convertTypesToFieldSelector(types)})
-	if err != nil {
-		return nil, err
+	totalEventList := []corev1.Event{}
+	fs := convertTypesToFieldSelectors(types)
+	for _, f := range fs {
```
scarlettperry, line 93: if the request didn't pass filter types then this wouldn't make make the api call as `fs` would be empty, unless im misunderstanding

backend/service/k8s/event.go
```diff
@@ -91,15 +88,21 @@ func (s *svc) ListNamespaceEvents(ctx context.Context, clientset, cluster, names
...
+		eventList, err := cs.CoreV1().Events(cs.Namespace()).List(ctx, metav1.ListOptions{FieldSelector: f})
```
scarlettperry, line 94: since we're now collecting the event results of separate API calls, do we need to sort the events list by timestamp?

backend/service/k8s/event.go
```diff
@@ -34,51 +35,65 @@ func (s *svc) ListEvents(ctx context.Context, clientset, cluster, namespace, obj
 	return events, nil
 }
 
-func ProtoForEvent(cluster string, k8sEvent *corev1.Event) *k8sapiv1.Event {
-	clusterName := GetKubeClusterName(k8sEvent)
-	if clusterName == "" {
-		clusterName = cluster
```
scarlettperry, line 40: curious, why was this removed? (https://github.com/lyft/clutch/pull/2704/files)

scarlettperry (approved):

backend/service/k8s/event.go
smonero: oops yeah I accidently removed that when i merged in main branch

PR #2702, opened 2023-06-22
api: k8s probe validate rules

fran2344:
Adding new validate rules to attributes

### Description
Adding validation rules to attributes on proto probe k8s components

No running curl yet that test with the backend 

```
curl -X POST http://localhost:8080/v1/k8s/updateDeployment
-H 'Content-Type: application/json' 
--data 
'{"clientset": "clientset", 
"cluster": "mycluster",
"namespace": "default", 
"name": "two-container-app", 
"fields": 
{ 
"containerProbes": 
[
{"containerName": "container2", 
"readinessProbe": 
  {"initialDelaySeconds": "6", "periodSeconds": "4" }
}]}}'
```


api/k8s/v1/k8s.proto
```diff
@@ -634,12 +634,12 @@ message Probe {
...
-  optional int32 period_seconds = 7;
-  optional int32 success_threshold = 8;
-  optional int32 failure_threshold = 9;
-  optional int64 termination_grace_period_seconds = 10;
+  int32 initial_delay_seconds = 5 [ (validate.rules).int32.gte = 0 ];
+  int32 timeout_seconds = 6 [ (validate.rules).int32.gte = 0 ];
+  int32 period_seconds = 7 [ (validate.rules).int32.gte = 0 ];
+  int32 success_threshold = 8 [ (validate.rules).int32.gte = 0 ];
+  int32 failure_threshold = 9 [ (validate.rules).int32.gte = 0 ];
+  int64 termination_grace_period_seconds = 10 [ (validate.rules).int64.gte = 0 ];
```
mikecutalo, line 642: Just wanted to double check, these are all required by kube right?

fran2344: Those are optional but all of them have a default value if we don't defined them

mikecutalo (approved):

PR #2703, opened 2023-06-23
Backend update probes

fran2344:
Backend update probes with validation attributes

### Description
Adding update method to update liveness and readiness probes.
 
https://github.com/lyft/clutch/pull/2702

No frontend changes

### Testing Performed

```
curl -X POST http://localhost:8080/v1/k8s/updateDeployment
-H 'Content-Type: application/json' 
--data 
'{"clientset": "clientset", 
"cluster": "mycluster",
"namespace": "default", 
"name": "two-container-app", 
"fields": 
{ 
"containerProbes": 
[
{"containerName": "container2", 
"readinessProbe": 
  {"initialDelaySeconds": "6", "periodSeconds": "4" }
}]}}'
```

### DIFF
--- a/backend/service/k8s/deployment.go
+++ b/backend/service/k8s/deployment.go
@@ -311,6 +311,10 @@ func (s *svc) UpdateDeployment(ctx context.Context, clientset, cluster, namespac
 		return err
 	}
 
+	if err := updateContainerProbes(newDeployment, fields); err != nil {
+		return err
+	}
+
 	patchBytes, err := GenerateStrategicPatch(oldDeployment, newDeployment, appsv1.Deployment{})
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


backend/service/k8s/deployment.go
```diff
@@ -101,12 +101,144 @@ func ProtoForDeploymentSpec(deploymentSpec appsv1.DeploymentSpec) *k8sapiv1.Depl
 			resourceRequests[string(res)] = quantity.String()
 		}
 
+		LivenessProbeObject := &k8sapiv1.Probe{}
+		ReadinessProbeObject := &k8sapiv1.Probe{}
```
mikecutalo, line 105: lets move this probe generation code into its own function to keep this `ProtoForDeploymentSpec` smaller and lets add some unit test coverage for that new function.

backend/service/k8s/deployment.go
```diff
@@ -101,12 +101,144 @@ func ProtoForDeploymentSpec(deploymentSpec appsv1.DeploymentSpec) *k8sapiv1.Depl
...
+
+			if container.LivenessProbe.ProbeHandler.GRPC != nil {
+				LivenessProbeGRPC := &k8sapiv1.GRPCAction{
+					Port:    container.LivenessProbe.ProbeHandler.GRPC.Port,
+					Service: *container.LivenessProbe.ProbeHandler.GRPC.Service,
+				}
+				HandlerObj.Handler = &k8sapiv1.Probe_Grpc{
+					Grpc: LivenessProbeGRPC,
+				}
+			}
```
mikecutalo, line 158: nit: maybe we should use a switch here? only one of these can be configured at a time.

backend/service/k8s/deployment.go
```diff
@@ -101,12 +101,144 @@ func ProtoForDeploymentSpec(deploymentSpec appsv1.DeploymentSpec) *k8sapiv1.Depl
...
+
+			if container.ReadinessProbe.ProbeHandler.GRPC != nil {
+				ReadinessProbeGRPC := &k8sapiv1.GRPCAction{
+					Port:    container.ReadinessProbe.ProbeHandler.GRPC.Port,
+					Service: *container.ReadinessProbe.ProbeHandler.GRPC.Service,
+				}
+				HandlerObj.Handler = &k8sapiv1.Probe_Grpc{
+					Grpc: ReadinessProbeGRPC,
+				}
+			}
```
mikecutalo, line 221: nit: maybe we should use a switch here? only one of these can be configured at a time.

backend/service/k8s/deployment.go
```diff
@@ -236,3 +372,87 @@ func updateContainerResources(deployment *appsv1.Deployment, fields *k8sapiv1.Up
...
+	}
+}
+
+func setOptionalValue(source *int32, target *int32) {
+	if source != nil {
+		*target = *source
+	}
+}
+
+func updateContainerProbes(deployment *appsv1.Deployment, fields *k8sapiv1.UpdateDeploymentRequest_Fields) error {
```
mikecutalo, line 388: lets add unit tests

backend/service/k8s/deployment.go
fran2344: Move code to a new method and add test function

backend/service/k8s/deployment.go
fran2344: Added unit test

backend/service/k8s/deployment.go
fran2344: I try but I think it is not possible to switch with this type of structure I moved to a new method and reduce the amount of code

backend/service/k8s/deployment.go
fran2344: That is not possible I cannot filter by type this structure

backend/service/k8s/deployment.go
```diff
@@ -87,6 +87,62 @@ func ProtoForDeployment(cluster string, deployment *appsv1.Deployment) *k8sapiv1
 	return k8sDeployment
 }
 
+func ProcessObjProbe(objProbe *v1.Probe) *k8sapiv1.Probe {
```
mikecutalo, line 90: Lets add some test for this function, also do we want this function to be public? i think will probably want to keep it private.

backend/service/k8s/deployment_test.go
```diff
@@ -364,3 +365,78 @@ func TestProtoForDeploymentStatus(t *testing.T) {
...
+									},
+									LivenessProbe: &v1.Probe{
+										ProbeHandler: v1.ProbeHandler{
+											HTTPGet: &v1.HTTPGetAction{
+												Path: "/",
+												Port: intstr.IntOrString{
+													IntVal: 8080,
+												},
+											},
+										},
```
mikecutalo, line 402: This is a good start but it only handles one case, live probe http action. Lets add more tests here for for each new case we added for live probe all handlers and the same for ready probes.

backend/service/k8s/deployment_test.go
```diff
@@ -364,3 +365,78 @@ func TestProtoForDeploymentStatus(t *testing.T) {
...
+									Resources: v1.ResourceRequirements{
+										Limits: v1.ResourceList{
+											"cpu":    resource.MustParse("500m"),
+											"memory": resource.MustParse("128Mi"),
+										},
+										Requests: v1.ResourceList{
+											"cpu":    resource.MustParse("250m"),
+											"memory": resource.MustParse("64Mi"),
+										},
+									},
```
mikecutalo, line 393: Do you need this for the test?

backend/service/k8s/deployment_test.go
```diff
@@ -364,3 +365,78 @@ func TestProtoForDeploymentStatus(t *testing.T) {
...
+
+func TestProtoForDeploymentSpecWithProbes(t *testing.T) {
+	t.Parallel()
+	var terminationVar int64 = 30
+	var deploymentTestCases = []struct {
+		id         string
+		deployment *appsv1.Deployment
+	}{
+		{
+			id: "foo",
```
mikecutalo, line 377: when were creating more tests lets use names to reflect what were testing, in this case `LivenessProbe / HTTPGetAction` for example.

backend/service/k8s/deployment_test.go
fran2344: Removed

backend/service/k8s/deployment_test.go
fran2344: Gotcha update the name of each of test cases

backend/service/k8s/deployment_test.go
fran2344: Adding all 4 cases with all handlers

backend/service/k8s/deployment.go
fran2344: Remove public method and add this method to testing

backend/service/k8s/deployment_test.go
```diff
@@ -364,3 +365,569 @@ func TestProtoForDeploymentStatus(t *testing.T) {
...
 	}
 }
+
+func TestProtoForDeploymentSpecWithProbesLivenessHTTPGet(t *testing.T) {
+	t.Parallel()
+	var terminationVar int64 = 30
+	var deploymentTestCases = []struct {
+		id         string
+		deployment *appsv1.Deployment
+	}{
```
mikecutalo, line 375: If your going to only have one test case per test func, you can remove all of this extra setup.

fran2344: Remove structure and just keep the deployment

backend/service/k8s/deployment_test.go
```diff
@@ -364,3 +365,481 @@ func TestProtoForDeploymentStatus(t *testing.T) {
...
+										Port:    portGRPC,
+										Service: &serviceGRPC,
+									},
+								},
+								InitialDelaySeconds:           4,
+								PeriodSeconds:                 24,
+								TimeoutSeconds:                2,
+								SuccessThreshold:              4,
+								FailureThreshold:              5,
+								TerminationGracePeriodSeconds: &terminationVar,
```
mikecutalo, line 808: Generally should we be testing that if these values are not set, they are at least defaulted correctly?

fran2344: Yes, in most tests I include some attributes and remove others to validate that some are using the default value

K8s include most of the default attributes

mikecutalo (approved):

PR #2704, opened 2023-06-23
k8s: replace usage of clusterName to for k8s upgrade

mikecutalo:
### Description

Clutch relies on cluster information throughout the application, previously we relied on the [clusterName](https://github.com/kubernetes/apimachinery/blob/2456ebdaba229616fab2161a615148884b46644b/pkg/apis/meta/v1/types.go#L266-L270) field on the objectmeta which has since been [deprecated](https://github.com/kubernetes/kubernetes/commit/331525670b772eb8956b7f5204078c51c00aaef3).

To replace this we utilize our own label `cluster.clutch.sh/name` to denote which cluster the object belongs to, these helper functions are used to standardize the getting and setting of this field.

This unblocks our ability to upgrade to the latest k8s client.

### Testing Performed
Unit


backend/service/k8s/lightweightinformer.go
```diff
@@ -90,7 +91,8 @@ func NewLightweightInformer(
 						if err := cacheStore.Add(lightweightObj); err != nil {
 							return err
 						}
-						h.OnAdd(d.Object)
+						// (mikecutalo) : double check what isInInitialList is doing
+						h.OnAdd(d.Object, false)
```
mikecutalo, line 95: i need to double check this new `isInInitialList` option but otherwise the rest is ready for review.

backend/service/k8s/meta_test.go
```diff
@@ -94,49 +90,3 @@ func TestApplyListOptions(t *testing.T) {
 		})
 	}
 }
-
-func TestApplyClusterMetadata(t *testing.T) {
```
mikecutalo, line 98: I simply moved this test to `kube_cluster_test.go`

samrabelachew (approved):

scarlettperry (approved):

smonero (approved):

PR #2715, opened 2023-07-05
change to v1 from v1beta1

smonero:
### Description

Currently we see warnings about using the v1beta1 version so we can use the v1 to fix it
`13 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob`

### GitHub Issue

Fixes #


Yiwen-Gao (approved):
Nice! 

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#cronjob-v125

PR #2724, opened 2023-07-12
Add last time to event and fix partial errors issue for namespace events

smonero:
### Description

We could use more fields in the clutch k8s event struct so this PR adds them - lastTimestampMillis - and also changes the type to a string since there could be more in the future than just NORMAL WARNING and ERROR. Also, there is a fix for partial errors because in the case no events are found, a 404 is returned, but we don't want to fail the whole call in case one category of types is 404 and the others are found.

### GitHub Issue

Fixes #


api/k8s/v1/k8s.proto
```diff
@@ -1294,13 +1294,14 @@ message Event {
...
+  // marks when the object is created.
   // https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta
   int64 creation_time_millis = 10;
 
-  // type is normal, warning, or error
-  EventType type = 11;
+  // type is string so that in the future it is open-ended
+  string type = 11;
+
+  int64 last_timestamp_millis = 12;
```
scarlettperry, line 1304: nit: code comment

scarlettperry (approved):

PR #2731, opened 2023-07-17
proxy config: support path regex in allowed requests

scarlettperry:
### Description
currently we perform [isallowedRequest](https://github.com/lyft/clutch/blob/main/backend/module/proxy/proxy.go#L204-L213) on each proxy API request to confirm the request path exactly matches an allowed request path in the proxy config. this set up works for paths with query parameters (i.e. `/car?brand=foo`) but not path parameters (ie `/car/{brand}`).

PR adds support in the proxy config to specify either an exact path or a path regex. If it's the latter, in `isallowedRequest` we check if the request path string matches the path regex.

Note: the updates to the config are backwards compatible and won't need to modify existing use cases of `path`.

### Testing Performed
manually
unit tests


backend/module/proxy/proxy.go
```diff
@@ -35,12 +36,17 @@ func New(cfg *any.Any, log *zap.Logger, scope tally.Scope) (module.Module, error
...
-	// Validate that each services constructs a parsable URL
 	for _, service := range config.Services {
 		for _, ar := range service.AllowedRequests {
-			_, err := url.Parse(fmt.Sprintf("%s%s", service.Host, ar.Path))
-			if err != nil {
-				return nil, fmt.Errorf("unable to parse the configured URL for service [%s]", service.Name)
+			switch t := ar.PathType.(type) {
+			case *proxyv1cfg.AllowRequest_Path:
+				// For exact path type, validate that each services constructs a parsable URL
+				_, err := url.Parse(fmt.Sprintf("%s%s", service.Host, t.Path))
```
scarlettperry, line 44: decided to just do this validation when it's an exact path type as a path regex might contain characters like `^` which would fail `url.Parse`

backend/module/proxy/proxy.go
mikecutalo, line 1: When we read in the config in the constructor, maybe we should also validate the regex rules are valid?

backend/module/proxy/proxy.go
```diff
@@ -205,13 +211,22 @@ func isAllowedRequest(services []*proxyv1cfg.Service, service, path, method stri
...
+					if parsedUrl.Path == t.Path && strings.EqualFold(method, ar.Method) {
+						return true, nil
+					}
+				case *proxyv1cfg.AllowRequest_PathRegex:
+					r := regexp.MustCompile(t.PathRegex)
+					if r.MatchString(path) {
+						return true, nil
+					}
+				default:
+					return false, fmt.Errorf("path type not supported: %s", t)
```
mikecutalo, line 229: think we want to print the type here with `%T` right?

backend/module/proxy/proxy.go
scarlettperry: updated & added test

mikecutalo (approved):
Nice!

smonero (approved):

PR #2790, opened 2023-09-13
frontend: Adding User Preferences Context and Customizable Header

jdslaugh:
### Description
- Modifies the header to take customizable settings for what to display/hide.
- Can now pass in a header with a `type="HEADER"` as a child to `<ClutchApp />` and it will take over as the main header, this allows for customizing the header entirely
- Can pass children to the newly passed in Header (if you just reuse the exported one from Clutch) that will place them inside the header component
- The userInfo component now also takes children so small options or buttons can be expanded for use on the dropdown

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/AppProvider/index.tsx
+++ b/frontend/packages/core/src/AppProvider/index.tsx
@@ -4,7 +4,7 @@ import Bugsnag from "@bugsnag/js";
 import BugsnagPluginReact from "@bugsnag/plugin-react";
 
 import AppLayout from "../AppLayout";
-import { ApplicationContext, ShortLinkContext } from "../Contexts";
+import { ApplicationContext, ShortLinkContext, UserPreferencesProvider } from "../Contexts";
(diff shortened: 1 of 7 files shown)


frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -129,77 +132,103 @@ const ClutchApp: React.FC<ClutchAppProps> = ({
     [discoverableWorkflows, triggeredHeaderData]
   );
 
+  React.useEffect(() => {
+    if (children) {
+      React.Children.forEach(children, child => {
+        if (React.isValidElement(child)) {
+          const { type } = child?.props || {};
+
+          if (type === "HEADER") {
```
jecr, line 141: Do we have a way of hinting the possibility to use this type prop?, maybe adding a type

jecr (approved):

jecr (approved):

PR #2792, opened 2023-09-14
feat: allow passing date and time picker props

taisph:
This change allows control of date and time picker properties, such as
`ampm` which is used to switch time picker between 12-hour and 24-hour
clock input.

Only the listed properties are still required and should not break
existing usages.


jdslaugh (approved):

taisph:
I don't think I understand what's going on in the build but I can reproduce the build error locally in a clean checkout. However, I can patch the `@clutch-sh/core` components directly in my scaffold-gateway created frontend project and still build a typescript workflow with `make frontend` and `tsc -b` as compile script without any errors. 🤔

jdslaugh:
> I don't think I understand what's going on in the build but I can reproduce the build error locally in a clean checkout. However, I can patch the `@clutch-sh/core` components directly in my scaffold-gateway created frontend project and still build a typescript workflow with `make frontend` and `tsc -b` as compile script without any errors. 🤔

Yeah Im not sure, I haven't looked much into combining Omit and Pick. I can understand the sentiment but it might need something different.

PR #2904, opened 2024-01-18
frontend: type instances of theme in styled components

jecr:
When using the `theme` prop in `styled` components, it is not typed, causing it to not have type checking. 

This PR replaces all instances of `({ theme })` an adds its type as `({ theme }: { theme: Theme })`

### Testing Performed
manual, unit tests


jdslaugh (approved):
Hmm, unfortunate that we have to do this but it makes some sense. Thanks for digging into it!

lucechal14 (approved):

PR #2910, opened 2024-01-22
frontend: Change Tabs to be a controlled component

septum:
### Description

Make the value prop to control the index of the tabs 

### GitHub Issue

Manual

### TODOs

N/A


septum:
This is no longer needed.

PR #2951, opened 2024-03-12
frontend: add support for icons in quick links

jecr:
This PR adds updates to the QuickLinks component:

- Reduce padding for dropdown elements
before:
image

after:
image

- Prepend an icon if passed in the `link` object
image

### Testing Performed
manual

### DIFF
--- a/frontend/packages/core/src/quick-links.tsx
+++ b/frontend/packages/core/src/quick-links.tsx
@@ -30,7 +30,9 @@ const StyledButton = styled("button")({
 });
 
 const StyledPopperItem = styled(PopperItem)({
-  height: "auto !important",
+  "&&&": {
+    height: "auto",
+  },
   "& span.MuiTypography-root": {
     padding: "0",
   },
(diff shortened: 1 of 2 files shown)


lucechal14 (approved):

frontend/packages/core/src/quick-links.tsx
```diff
@@ -29,8 +29,14 @@ const StyledButton = styled("button")({
   display: "flex",
 });
 
-const StyledSpan = styled("span")({
-  whiteSpace: "nowrap",
+const StyledPopperItem = styled(PopperItem)({
+  height: "auto !important",
```
jdslaugh, line 33: Is this `!important` necessary?

jecr: `PopperItem` has this high specificity selector, hence my `!important` and also since the parent is a `Popper` that doesn't really take in styles for its children.
So TIL in styled components we are able to nest `&` selectors and increase specificity that way.

jdslaugh: Oh interesting, I had no idea as well lol. But my CSS knowledge is limited anyways haha. 

Yeah I was thinking maybe it was similar to before where we had to make sure we allowed the className prop type like I had in a previous PR. Ultimately just want to avoid `!important` when possible :)

jdslaugh (approved):
Looks good thanks!

PR #2954, opened 2024-03-14
housekeeping: frontend: Updating Yarn to v4 to better support versioning system

jdslaugh:
### Description
- Updates the system to utilize Yarn v4 which will allow for the unlock of workspace versioning of `workspace:^` which will allow the versioning system to be a little more resilient for any submodules that use it.
- Implements constraints checking so that package versions can be consistent
- Adds a pre-commit hook configuration which will be utilized for frontend commits to run verification checks

### Testing Performed
manual


frontend/packages/tools/package-command.js
jecr, line 1: just wondering, what does this file do 😅?

jecr (approved):

frontend/packages/tools/package-command.js
jdslaugh: Accidental commit from something I had considered doing lol

PR #2958, opened 2024-03-21
frontend: Add noDefault flag to Select component

lucechal14:
Scope: Adding a noDefault flag to the Select component to avoid triggering the onChange method on component did update for when we don't want the first option in the list to be set as default.


jdslaugh (approved):
Thanks!

PR #2974, opened 2024-04-09
[Backend] Attempt to move watcher data from heap to the stack so we can reduce memory usage

miguelmolina95:
### Description

I have noticed memory usage increase in this flow when running the Lyft private version and attempting to letting some of these variables be freed from the heap sooner since as far as I can tell we do not do any manipulation of the pointer objects.

### GitHub Issue

Fixes #


miguelmolina95:
closing this I think we can optimize in better ways

PR #2981, opened 2024-04-16
frontend: Project Catalog - Allowing disabled projects to be found and returned

jdslaugh:
### Description
- Adjusts the projectCatalog to accept an additional componentProp in its configuration of `allowDisabled`. This will be propagated to the API requests and allow the system to accept and return disabled projects so that they can be navigated to.
- Modifies the settings button to accept additional links and renders a popper to show them
- Adjusts the breadcrumbs to reference the project now

#### Breadcrumbs
Screenshot 2024-04-17 at 12 05 51 PM
Screenshot 2024-04-17 at 12 05 47 PM

#### Catalog Settings Button
Screenshot 2024-04-17 at 12 05 41 PM

### Testing Performed
manual

### DIFF
--- a/frontend/workflows/projectCatalog/src/details/helpers.tsx
+++ b/frontend/workflows/projectCatalog/src/details/helpers.tsx
@@ -22,12 +22,16 @@ const fetchProjectInfo = (
 
       if (_.isEmpty(projectResults) && allowDisabled && partialFailures) {
         // Will add disabled projects to the list of projects to display if requested
-        let failuresMap = partialFailures
+        const failuresMap = partialFailures
           .map(p => {
             if ((_.get(p, "message", "") ?? "").includes("disabled")) {
-              const project = _.get(p, "details.[0]");
-              _.set(project, ["data", "description"], `${project.name} is disabled.`);
-              return project;
+              const disabledProject = _.get(p, "details.[0]");
+              _.set(
+                disabledProject,
+                ["data", "description"],
+                `${disabledProject.name} is disabled.`
+              );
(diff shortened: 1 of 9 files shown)


frontend/workflows/projectCatalog/src/details/helpers.tsx
```diff
@@ -3,18 +3,46 @@ import type { clutch as IClutch } from "@clutch-sh/api";
 import { client, Grid, Link, styled, TimeAgo as EventTime, Typography } from "@clutch-sh/core";
 import { faClock } from "@fortawesome/free-regular-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import * as _ from "lodash";
```
khtruong, line 6: Are bundle sizes still ok with importing all of this, aka things are tree shaken? I know that lodash is quite a large library.

khtruong (commented):
How about a screenshot and/or video?

khtruong (approved):

jecr (approved):

frontend/workflows/projectCatalog/src/details/helpers.tsx
jdslaugh: They should be, but we currently do this in multiple places in Clutch so it won't affect anything more than it does already. It's on my list to do some more investigation and I'll adjust the imports as necessary to handle it after.

jdslaugh:
> How about a screenshot and/or video?

Updated to include screenshots. Not much I'm able to show in a video with the mock data.

PR #3002, opened 2024-05-06
DateTimePicker - trigger onChange method when value is empty

lucechal14:
Scope: Update onChange trigger validation to allow trigger when the value of the DateTimePicker is empty

### DIFF
--- a/frontend/packages/core/src/Input/date-time.tsx
+++ b/frontend/packages/core/src/Input/date-time.tsx
@@ -20,7 +20,7 @@ const PaddedTextField = styled(TextField)({
 
 export interface DateTimePickerProps
   extends Pick<MuiDateTimePickerProps<Date, Date>, "disabled" | "value" | "onChange" | "label"> {
-  allowEmpty: boolean;
+  allowEmpty?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


frontend/packages/core/src/Input/date-time.tsx
```diff
@@ -26,8 +26,9 @@ const DateTimePicker = ({ onChange, ...props }: DateTimePickerProps) => (
     <MuiDateTimePicker
       renderInput={inputProps => <PaddedTextField {...inputProps} />}
       onChange={(value: Dayjs | null) => {
-        if (value && value.isValid()) {
-          onChange(value.toDate());
+        if (!value || value.isValid()) {
```
jdslaugh, line 29: Should we do this in the time picker as well?

jdslaugh (commented):
Lets do a test deploy of this and make sure it doesn't have any issues with our current implementations of this.

frontend/packages/core/src/Input/date-time.tsx
lucechal14: i don't have a scenario for the time picker where we need it but I can certainly do it too

jecr (approved):

jdslaugh (approved):
Looks good. We can address the TimePicker in a future PR if we wind up needing it.

PR #3007, opened 2024-05-14
DS-10153: Add Functionality to aws client for calling S3GetAccessPointPolicy

tolujimoh:
### Description

This PR adds functionality for calling S3GetAccessPointPolicy to support [clutch-private#2500](https://github.com/lyft/clutch-private/pull/2500)

### DIFF
--- a/backend/service/aws/aws.go
+++ b/backend/service/aws/aws.go
@@ -212,7 +212,6 @@ type DirectClient interface {
 	IAM() *iam.Client
 	Kinesis() *kinesis.Client
 	S3() *s3.Client
-	S3Control() *s3control.Client
 	STS() *sts.Client
 }
 
@@ -267,8 +266,6 @@ func (r *regionalClient) Kinesis() *kinesis.Client {
 func (r *regionalClient) S3() *s3.Client {
 	return r.s3.(*s3.Client)
 }
-
-func (r *regionalClient) S3Control() *s3control.Client { return r.s3control.(*s3control.Client) }
 
 func (r *regionalClient) STS() *sts.Client {
 	return r.sts.(*sts.Client)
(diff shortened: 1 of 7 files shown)


backend/service/aws/aws.go
```diff
@@ -178,6 +180,8 @@ type Client interface {
 	S3GetBucketPolicy(ctx context.Context, account, region, bucket, accountID string) (*s3.GetBucketPolicyOutput, error)
 	S3StreamingGet(ctx context.Context, account, region, bucket, key string) (io.ReadCloser, error)
 
+	S3GetAccessPointPolicy(ctx context.Context, account, region, accessPointName string, accountID string) (*s3control.GetAccessPointPolicyOutput, error)
+
```
mikecutalo, line 184:
```suggestion
	S3GetAccessPointPolicy(ctx context.Context, account, region, accessPointName, accountID string) (*s3control.GetAccessPointPolicyOutput, error)
```

consecutive params of the same type only need a single type definition

backend/service/aws/aws.go
```diff
@@ -263,6 +269,8 @@ func (r *regionalClient) S3() *s3.Client {
 	return r.s3.(*s3.Client)
 }
 
+func (r *regionalClient) S3Control() *s3control.Client { return r.s3control.(*s3control.Client) }
+
```
mikecutalo, line 273: We already expose all clients via the `GetDirectClient()` on the aws client struct, we dont need to expose them individually.

mikecutalo (commented):
Nice! I just did a first pass and left a few comments.

backend/service/aws/aws.go
tolujimoh: Good catch! Updated.

backend/service/aws/aws.go
tolujimoh: Ah! Thanks for pointing, will update.

mikecutalo (approved):
Approved!

PR #3012, opened 2024-05-15
backend: Extending GitHub service to have the ability to get files based on a path

jdslaugh:
### Description
Currently the GitHub client has the ability to get a file at a specific path but if we need to fetch the contents of a directory we were out of luck, this addresses that and allows us to navigate around based on paths.

#### Test Query against `clutch/frontend`
```
HTTP/1.1 200 OK
Content-Type: application/json
Grpc-Metadata-Content-Type: application/grpc
Grpc-Metadata-Trailer: Grpc-Status
Grpc-Metadata-Trailer: Grpc-Message
Grpc-Metadata-Trailer: Grpc-Status-Details-Bin
Date: Wed, 15 May 2024 20:11:29 GMT
Content-Length: 653

{
  "path": "frontend",
  "lastModifiedTime": "2024-05-15 18:34:45 +0000 UTC",
  "lastModifiedSha": "3bc4eb7da42ee2adc089bc9db2441233503b9c9e",
  "entries": [
    { "name": ".gitignore", "type": "blob" },
    { "name": ".nvmrc", "type": "blob" },
    { "name": ".storybook", "type": "tree" },
    { "name": ".yarnrc.yml", "type": "blob" },
    { "name": "api", "type": "tree" },
    { "name": "license-linter.js", "type": "blob" },
    { "name": "netlify", "type": "tree" },
    { "name": "package.json", "type": "blob" },
    { "name": "packages", "type": "tree" },
    { "name": "tsconfig.base.json", "type": "blob" },
    { "name": "tsconfig.json", "type": "blob" },
    { "name": "workflows", "type": "tree" },
    { "name": "yarn.config.cjs", "type": "blob" },
    { "name": "yarn.lock", "type": "blob" }
  ]
}

```

### Testing Performed
local

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -96,7 +96,7 @@ type File struct {
 	LastModifiedSHA  string
 }
 
-type DirectoryEntry struct {
+type Entry struct {
 	Name string
 	Type string
 }
@@ -105,7 +105,7 @@ type Directory struct {
 	Path             string
 	LastModifiedTime time.Time
 	LastModifiedSHA  string
-	Files            []*DirectoryEntry
+	Entries          []*Entry
 }
 
 // Client allows various interactions with remote repositories on GitHub.
@@ -505,9 +505,9 @@ func (s *svc) GetDirectory(ctx context.Context, ref *RemoteRef, path string) (*D
 		return nil, errors.New("directory not found")
 	}
 
-	var entries []*DirectoryEntry
+	var entries []*Entry
 	for _, obj := range q.Repository.Object.Tree.Entries {
-		entries = append(entries, &DirectoryEntry{
+		entries = append(entries, &Entry{
 			Name: string(obj.Name),
 			Type: string(obj.Type),
 		})
@@ -515,7 +515,7 @@ func (s *svc) GetDirectory(ctx context.Context, ref *RemoteRef, path string) (*D
 
 	d := &Directory{
 		Path:             path,
-		Files:            entries,
+		Entries:          entries,
 		LastModifiedTime: q.Repository.Ref.Commit.History.Nodes[0].CommittedDate.Time,
 		LastModifiedSHA:  string(q.Repository.Ref.Commit.History.Nodes[0].OID),
 	}
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -96,22 +96,22 @@ type File struct {
 	LastModifiedSHA  string
 }
 
-type FileEntry struct {
+type DirectoryEntry struct {
 	Name string
 	Type string
 }
 
-type FilePath struct {
+type Directory struct {
 	Path             string
 	LastModifiedTime time.Time
 	LastModifiedSHA  string
-	Files            []*FileEntry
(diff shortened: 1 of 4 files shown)


backend/service/github/github.go
```diff
@@ -96,9 +96,22 @@ type File struct {
...
+	Path             string
+	LastModifiedTime time.Time
+	LastModifiedSHA  string
+	Files            []*FileEntry
+}
+
 // Client allows various interactions with remote repositories on GitHub.
 type Client interface {
 	GetFile(ctx context.Context, ref *RemoteRef, path string) (*File, error)
+	GetFilePath(ctx context.Context, ref *RemoteRef, path string) (*FilePath, error)
```
khtruong, line 114: What about calling this `GetDirectory`?

scarlettperry: +1 can we rename the structs as well?

backend/service/github/github.go
```diff
@@ -96,9 +96,22 @@ type File struct {
 	LastModifiedSHA  string
 }
 
+type FileEntry struct {
+	Name string
+	Type string
+}
```
scarlettperry, line 102: just curious, is a FileEntry diff from a File from github's POV?

scarlettperry: in terms of the endpoint flow, with this info returned, is the plan then to make another API call for the contents?

backend/service/github/github.go
jdslaugh: Renamed to GetDirectory and adjusted the structs as well.

backend/service/github/github.go
jdslaugh: I dont 'think' so? Other than a File normally having the contents fetched already it's really just a summary of what's in a directory.

Yeah, my thinking is that we may only want to know what files are available and then another call/calls can be made for the contents, versus fetching everything already which could greatly increase the response size with data that may not be needed,

backend/service/github/github.go
```diff
@@ -96,9 +96,22 @@ type File struct {
 	LastModifiedSHA  string
 }
 
+type DirectoryEntry struct {
```
scarlettperry, line 99: nit -- since the entry would either be a file entry or directory entry
```suggestion
type Entry struct {
```

backend/service/github/github.go
```diff
@@ -96,9 +96,22 @@ type File struct {
...
+	Name string
+	Type string
+}
+
+type Directory struct {
+	Path             string
+	LastModifiedTime time.Time
+	LastModifiedSHA  string
+	Files            []*DirectoryEntry
```
scarlettperry, line 108:
```suggestion
	Entries            []*DirectoryEntry
```

scarlettperry (approved):
some more small comments re naming but approving!

backend/service/github/github.go
jdslaugh: Updated!

backend/service/github/github.go
jdslaugh: Updated!

PR #3022, opened 2024-05-21
aws: proto definition for s3 and iam resources for topology cache

tolujimoh:
### Description

This CR contains the proto definition for `Role`, `Bucket` and `AccessPoint`.


api/aws/iam/v1/iam.proto
```diff
@@ -0,0 +1,21 @@
...
+message Role {
+  option (clutch.api.v1.id).patterns = {
+    type_url : "clutch.aws.iam.v1.Role",
+    pattern : "{account}/{region}/{name}"
+  };
+
+  string name = 1;
+  string id = 2;
+  string arn = 3;
+  string created_date = 4;
```
mikecutalo, line 18: should we use `google.protobuf.Timestamp` here?

api/aws/s3/v1/s3.proto
```diff
@@ -0,0 +1,34 @@
...
+message AccessPoint {
+  option (clutch.api.v1.id).patterns = {
+    type_url : "clutch.aws.s3.v1.AccessPoint",
+    pattern : "{account}/{region}/{name}"
+  };
+
+  string name = 1;
+  string bucket = 2;
+  string alias = 3;
+  string creation_date = 4;
```
mikecutalo, line 29: same as above

api/aws/iam/v1/iam.proto
tolujimoh: Make sense! Will update

mikecutalo (approved):

PR #3041, opened 2024-06-12
frontend: Add sort logic to table component

lucechal14:
Props and components were added to the table component to enable sort behavior

https://github.com/lyft/clutch/assets/25833665/7090f92b-333c-4b7e-8877-f9f23c26b926

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -4,6 +4,7 @@ import type {
   TableCellProps as MuiTableCellProps,
   TableProps as MuiTableProps,
   TableRowProps as MuiTableRowProps,
+  TableSortLabelProps as MuiTableSortLabelProps,
   Theme,
 } from "@mui/material";
 import {
@@ -116,11 +117,9 @@ const TableContainer = ({ children }: TableContainerProps) => (
   </MuiTableContainer>
 );
 
-export type Order = "asc" | "desc";
-
 interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   /** The names of the columns. This must be set (even to empty string) to render the table. */
-  columns: string[];
+  columns: string[] | { id: string; title: string; sortable?: boolean }[];
   /** The breakpoint at which to compress the table rows. By default the small breakpoint is used. */
   compressBreakpoint?: Breakpoint;
   /** Hide the header. By default this is false. */
@@ -135,8 +134,7 @@ interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   children?:
     | (React.ReactElement<TableRowProps> | null | undefined | {})[]
     | React.ReactElement<TableRowProps>;
-  order?: Order;
-  orderBy?: string;
+  defaultSort?: [string, MuiTableSortLabelProps["direction"]];
   onRequestSort?: (event: React.MouseEvent<unknown>, property: string) => void;
 }
 
@@ -150,8 +148,7 @@ const Table: React.FC<TableProps> = React.forwardRef(
       responsive = false,
       overflow = "scroll",
       children,
-      order,
-      orderBy,
+      defaultSort,
       onRequestSort,
       ...props
     },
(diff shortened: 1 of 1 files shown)


frontend/packages/core/src/Table/table.tsx
```diff
@@ -132,64 +136,84 @@ interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   children?:
     | (React.ReactElement<TableRowProps> | null | undefined | {})[]
     | React.ReactElement<TableRowProps>;
+  order?: Order;
+  orderBy?: string;
+  onRequestSort?: (event: React.MouseEvent<unknown>, property: string) => void;
 }
 
-const Table: React.FC<TableProps> = React.forwardRef(
```
jdslaugh, line 137: Is it intentional not to forward the ref anymore?

frontend/packages/core/src/Table/table.tsx
```diff
@@ -115,6 +117,8 @@ const TableContainer = ({ children }: TableContainerProps) => (
   </MuiTableContainer>
 );
 
+export type Order = "asc" | "desc";
```
jdslaugh, line 120: We can probably pull this from the TableSortLabelProps and just extend it onto the props.

frontend/packages/core/src/Table/table.tsx
```diff
@@ -132,64 +136,84 @@ interface TableProps extends Pick<MuiTableProps, "stickyHeader"> {
   children?:
     | (React.ReactElement<TableRowProps> | null | undefined | {})[]
     | React.ReactElement<TableRowProps>;
+  order?: Order;
+  orderBy?: string;
```
jdslaugh, line 140: Im confused why we need these but maybe I'm missing something.

I would think we would have a column prop of 'sortable' or something. If you click on it then it would call the `onRequestSort` if available and then internally keep track of the order/orderBy, if `onRequestSort` isn't available then we could just sort it internally as if it has all the data.

We could have a `defaultSort` param if necessary but I don't think the caller should need this data externally.

frontend/packages/core/src/Table/table.tsx
lucechal14: nop, my bad, I'll put it back

frontend/packages/core/src/Table/table.tsx
lucechal14: I based the implementation on the example from the MUI docs, basically only one column is active at a time, that's the orderBy prop, I'll add a video in the description of the PR so it's more clear

frontend/packages/core/src/Table/table.tsx
lucechal14: in the example from the docs it was define like this, not imported from MUI library

frontend/packages/core/src/Table/table.tsx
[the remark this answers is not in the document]
lucechal14: I'll add it in the description of the PR

frontend/packages/core/src/Table/table.tsx
jdslaugh: Which docs, material's?

Ultimately it's just one piece we don't have to manually handle since the prop exists and thats what we're passing it onto anyways. It's not a huge deal since it's only two possibilities though :)

frontend/packages/core/src/Table/table.tsx
```diff
@@ -166,20 +194,41 @@ const Table: React.FC<TableProps> = React.forwardRef(
...
+            managedColumns.filter(h => h.title.length !== 0).length !== 0 && (
+              <MuiTableHead>
+                <StyledTableHeadRow>
+                  {managedColumns.map(h => (
+                    <StyledTableCell
+                      key={typeof h === "string" ? h : h?.id}
+                      $responsive={responsive}
+                      align="left"
+                      sortDirection={
+                        h?.sortable && defaultSort?.[0] === h?.id ? defaultSort[1] : false
```
jdslaugh, line 208: Just double checking but 'false' just means 'not sortable' right?

Just want to make sure that if it isn't set we don't have it turned on by default :)

lucechal14: yes, false means that is not sortable

PR #3044, opened 2024-06-17
Add JSX.Element as a column type for Table component

lucechal14:
JSX.Element has been added as a posible type for the column props, if the column is in fact a React Element then it will be render directly without any sort logic.


jdslaugh (approved):
Thanks for the quick update!

PR #3046, opened 2024-06-17
network: Create `proxyClient` to handle errors on API calls done via the proxy API

ajimenezlyft:
### Description
API calls done via the proxy API always return 200 unless there's a 5XX error. There's a `httpStatus` property that's returned inside the `data` property on Axios calls which contains the actual HTTP status.

The new `proxyClient` now will be able to handle those errors correctly.

### Testing Performed
Unit testing and tested on internal UI.

### DIFF
--- a/frontend/packages/core/src/Network/index.ts
+++ b/frontend/packages/core/src/Network/index.ts
@@ -21,6 +21,13 @@ export interface HttpStatus {
   text: string;
 }
 
+interface CreateClientProps {
+  response?: {
+    success: (response: AxiosResponse) => AxiosResponse | Promise<never>;
+    error: (error: AxiosError) => Promise<ClutchError>;
+  };
+}
+
 const successInterceptor = (response: AxiosResponse) => {
   return response;
 };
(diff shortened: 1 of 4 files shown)


frontend/packages/core/src/Network/index.ts
```diff
@@ -25,6 +25,23 @@ const successInterceptor = (response: AxiosResponse) => {
   return response;
 };
 
+const successProxyInterceptor = (response: AxiosResponse) => {
+  // Handle proxy errors
+  if (response.data.httpStatus >= 400) {
+    const error = {
+      status: {
+        code: response.data.httpStatus,
+        text: HTTP_CODE_MAPPING[response.data.httpStatus] || "Unknown",
```
scsjonatan, line 34: Maybe we could create a function that return the right HTTP code status or "Unknown", just to avoid use `"|| Unknown"` everytime.

scsjonatan (approved):

frontend/packages/core/src/Network/index.ts
```diff
@@ -81,6 +98,20 @@ const createClient = () => {
   return axiosClient;
 };
 
+const createProxyClient = () => {
```
jdslaugh, line 101: So we're recreating logic here to create this client so I think we can just make it a bit more customizable to reduce the duplication.

Something like:

```
const successInterceptor = (response: AxiosResponse): AxiosResponse => {
...

const successProxyInterceptor = (response: AxiosResponse): AxiosResponse | Promise<never> => {
...
interface CreateClientProps {
  response?: {
    success: (response: AxiosResponse) => AxiosResponse | Promise<never>;
    error: (error: AxiosError) => Promise<ClutchError>;
  };
}

const createClient = ({ response }: CreateClientProps): AxiosInstance => {
  const axiosClient = axios.create({
    // n.b. the client will treat any response code >= 400 as an error and apply the error interceptor.
    validateStatus: status => {
      return status < 400;
    },
  });

  if (response) {
    axiosClient.interceptors.response.use(response.success, response.error);
  }

  return axiosClient;
};

const client = createClient({ response: { success: successInterceptor, error: errorInterceptor } });
const proxyClient = createClient({
  response: { success: successProxyInterceptor, error: errorInterceptor },
});
```

Haven't tested this specifically but it should work and then we're just telling it which interceptors to use versus recreating the instance.

jdslaugh (commented):
Minor adjustment but looks good otherwise!

frontend/packages/core/src/Network/index.ts
ajimenezlyft: Changes done!

jdslaugh (approved):
Looks good!

PR #3050, opened 2024-06-24
Frontent fix stepper height

scsjonatan:
### Description
* Fix height and scroll bar for vertical stepper

### Testing Performed

**Before**
bad

**After**
good

### DIFF
--- a/frontend/packages/wizard/src/wizard.tsx
+++ b/frontend/packages/wizard/src/wizard.tsx
@@ -281,4 +281,3 @@ const Wizard = ({
 };
 
 export default Wizard;
-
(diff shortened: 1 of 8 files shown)


jdslaugh (changes requested):
Doesn't look like this solution is working correctly. Testing in some workflows with a stepper it looks off.

It 'looks' like it might be because of no orientation being set automatically (at least that's what I'm gleaning from storybook). So we'll need to do some more testing and verification on other workflows.

Not sure I'm a huge fan of using an absolute here anyways, so something else to consider.

scsjonatan:
@jdslaugh We could take a better second look later.
Closing this PR to avoid noise.

PR #3058, opened 2024-07-04
frontend: Add options to the table header component

lucechal14:
Similar to sorting, the user can send attributes (_options?: JSX.Element_) in the columns list to the **Table** component to specify if they want to display options (e.g. filters) while also providing the _component_ that needs to be rendered.

An example could be something like the following:
Screenshot 2024-07-04 at 5 43 42 p m

Screenshot 2024-07-08 at 10 28 45 a m

**Filtering logic and UI is 100% responsibility of the father component, table header only renders the component provided**

### DIFF
--- a/frontend/packages/core/src/Table/components/TableHeader.tsx
+++ b/frontend/packages/core/src/Table/components/TableHeader.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import type { Theme } from "@mui/material";
+import type { TableSortLabelProps as MuiTableSortLabelProps, Theme } from "@mui/material";
 import {
   TableHead as MuiTableHead,
   TableRow as MuiTableRow,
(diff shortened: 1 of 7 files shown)


frontend/packages/core/src/Table/components/TableHeader.tsx
```diff
@@ -0,0 +1,108 @@
...
+  defaultSort?: [string, MuiTableSortLabelProps["direction"]];
+  onRequestSort?: (event: React.MouseEvent<unknown>, property: string) => void;
+  actionsColumn?: boolean;
+  compress?: boolean;
+}
+
+const TableHeader = ({
+  columns,
+  responsive,
+  defaultSort,
```
jdslaugh, line 38: nit but destructuring might make this a bit clearer vs referencing `[0]` / `[1]`

Pretty sure it would just be this but definitely double check :) 

```suggestion
  defaultSort: [sortKey, sortDir],
```

jdslaugh (approved):
Small nit but otherwise looks good!

PR #3060, opened 2024-07-09
frontend: add nonLinear feature to Stepper

scsjonatan:
## Description
* Add nonLinear feature to Stepper
* Change Wizard actions in reducer
* Handle onNext prop without submit

## Non-linear stepper
1- Done
2- Active
3- Inactive 
4- Inactive with hover

Screenshot 2024-07-30 at 1 59 31 p m

## Linear stepper
Screenshot 2024-07-30 at 2 02 18 p m


frontend/packages/wizard/src/wizard.tsx
```diff
@@ -135,11 +137,17 @@ const Wizard = ({
...
+    dispatch({ type: WizardActionType.NEXT });
+  };
+
+  const handleStepClick = (step: number) => {
+    dispatch({ type: WizardActionType.GO_TO_STEP, step });
   };
 
   const context = (child: JSX.Element) => {
     return {
+      getNextStepToComplete: () => state.nextStepToComplete,
```
jdslaugh, line 149: Im not seeing the updates to the WizardContext which this is based off of, so we'll need to update that to support the additional functionality.

frontend/packages/core/src/stepper.tsx
```diff
@@ -191,11 +192,22 @@ const Step: React.FC<StepProps> = ({ children }) => <>{children}</>;
 export interface StepperProps extends Pick<MuiStepperProps, "orientation"> {
   activeStep: number;
   children?: React.ReactElement<StepProps>[] | React.ReactElement<StepProps>;
+  nonLinear?: boolean;
```
jdslaugh, line 195: Let's remove this and just extend the `Pick` to take this field as well, I'd rather not recreate if we dont need to

frontend/packages/core/src/stepper.tsx
```diff
@@ -206,18 +218,31 @@ const Stepper = ({ activeStep, orientation = "horizontal", children }: StepperPr
...
+            stepProps.variant = step.props.error ? "failed" : "active";
+          }
+        } else if (idx === activeStep) {
           stepProps.variant = step.props.error ? "failed" : "active";
         } else if (idx < activeStep) {
           stepProps.variant = "success";
         }
 
+        const label = step.props.label ?? `Step ${idx + 1}`;
+        const icon = <StepIcon {...stepProps} />;
```
jdslaugh, line 235: Not sure if it's this change or not but when running locally it seems the number isn't centered on at least the horizontal view. So we'll need to verify that.

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -37,6 +37,7 @@ export interface WizardProps
     | React.ReactElement<WizardStepProps>[];
   dataLayout: ManagerLayout;
   heading?: string | React.ReactElement;
+  nonLinear?: boolean;
```
jdslaugh, line 40: Similar here about picking this versus recreating

frontend/packages/core/src/stepper.tsx
```diff
@@ -191,11 +192,22 @@ const Step: React.FC<StepProps> = ({ children }) => <>{children}</>;
 export interface StepperProps extends Pick<MuiStepperProps, "orientation"> {
   activeStep: number;
   children?: React.ReactElement<StepProps>[] | React.ReactElement<StepProps>;
+  nonLinear?: boolean;
+  completed?: { [key: number]: boolean };
```
jdslaugh, line 196: I'm not a huge fan of this property on the stepper. I think we should look at adjusting the `Step` component to take a property of `isCompleted` or something to that effect and let it manage whether it's been completed or not. Then we can have the step call into the wizard context which could keep the listing there.

Ultimately I want this to be more intuitive and allow the steps to define their own completion criteria that can then be propagated up versus having to maintain a list somewhere.

frontend/packages/core/src/stepper.tsx
```diff
@@ -206,18 +218,31 @@ const Stepper = ({ activeStep, orientation = "horizontal", children }: StepperPr
...
+        const StepProps = completed ? { completed: completed[idx] } : {};
         return (
-          <MuiStep key={step.props.label}>
-            {/* eslint-disable-next-line react/no-unstable-nested-components */}
-            <MuiStepLabel StepIconComponent={() => <StepIcon {...stepProps} />}>
-              {step.props.label ?? `Step ${idx + 1}`}
-            </MuiStepLabel>
+          <MuiStep key={step.props.label} {...StepProps}>
+            {nonLinear ? (
+              <MuiStepButton onClick={() => handleStepClick(idx)} icon={icon}>
```
jdslaugh, line 240: We'll probably need a setting or something to change the color of the steps because right now it just all looks disabled and I don't think many would realize the other steps are clickable.

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -161,7 +169,18 @@ const Wizard = ({
         if (params?.toOrigin && origin) {
           navigate(origin);
         } else {
-          dispatch(WizardAction.BACK);
+          dispatch({ type: WizardActionType.BACK });
+        }
+      },
+      onNext: (params: { toOrigin?: boolean; keepSearch?: boolean }) => {
```
jdslaugh, line 175: This and the onBack functions are pretty much identical, so we should probably pull it out into a helper function

jdslaugh (commented):
This is a great start! Some minor pieces but overall I'd like us to consider changing how we're approaching this by letting the children handle their own status versus the parent component needing to do so.

frontend/packages/core/src/stepper.tsx
scsjonatan: I just updated the PR description with the new stepper look, let me know what you think.

frontend/packages/core/src/stepper.tsx
jdslaugh: Might've been a cache problem since I'm not seeing it anymore. Easy enough :)

frontend/packages/core/src/stepper.tsx
jdslaugh: Hmm, what if we did something like this (as it currently is today).

So this would be the normal mode. In the non-linear mode it would have the background and number as the same color but the blue outline would not be there unless it's the selected step. What do you think?

Also, would you mind putting up some screenshots of normal / non-linear mode just so it's easy to verify the current state isn't changed.

Screenshot 2024-07-15 at 11 27 29 AM

scsjonatan: Updated.

I decided to keep a light gray in the border, removing the border completely gives a sensation of not clickable element or that the styles are broken.

frontend/packages/core/src/stepper.tsx
scsjonatan: Solved [here](https://github.com/lyft/clutch/pull/3071)

frontend/packages/core/src/Contexts/wizard-context.tsx
```diff
@@ -3,10 +3,13 @@ import React from "react";
 export interface ContextProps {
   displayWarnings: (warnings: string[]) => void;
   onBack: (params?: { toOrigin?: boolean; keepSearch?: boolean }) => void;
+  onNext: (params?: { toOrigin?: boolean; keepSearch?: boolean }) => void;
```
jdslaugh, line 6: Lets pull the params out into its own interface so we're not duplicating it.

frontend/packages/core/src/stepper.tsx
```diff
@@ -183,19 +195,30 @@ const StepIcon: React.FC<StepIconProps> = ({ index, variant }) => {
...
+  handleStepClick?: (index: number) => void;
 }
 
-const Stepper = ({ activeStep, orientation = "horizontal", children }: StepperProps) => (
-  <StepContainer $orientation={orientation}>
+const Stepper = ({
+  activeStep,
+  orientation = "horizontal",
+  children,
+  nonLinear,
```
jdslaugh, line 216: Similar comment as above about defaulting this to false

frontend/packages/core/src/stories/stepper.stories.tsx
```diff
@@ -36,20 +39,36 @@ const PrimaryTemplate = ({
...
     <>
       <Grid container direction={orientation === "horizontal" ? "column" : "row"}>
         <Grid item xs={1}>
-          <Stepper activeStep={curStep} orientation={orientation}>
+          <Stepper
+            activeStep={curStep}
+            orientation={orientation}
+            handleStepClick={handleStepClick}
+            nonLinear={nonLinear}
+            {...(nonLinear ? { completed: completedSteps } : {})}
```
jdslaugh, line 65: Looks like the storybook setup will need to be adjusted since the logic changed.

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -25,12 +25,13 @@ import type {
...
 
-import { useWizardState, WizardAction } from "./state";
+import { useWizardState, WizardActionType } from "./state";
 import type { WizardStepProps } from "./step";
 
 export interface WizardProps
   extends Pick<ContainerProps, "width" | "className">,
-    Pick<MuiStepperProps, "orientation"> {
+    Pick<MuiStepperProps, "orientation">,
+    Pick<MuiStepperProps, "nonLinear"> {
```
jdslaugh, line 34:
```suggestion
    Pick<MuiStepperProps, "orientation" | "nonLinear"> {
```

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -135,11 +137,32 @@ const Wizard = ({
...
-    dispatch(WizardAction.NEXT);
+    dispatch({ type: WizardActionType.NEXT });
+  };
+
+  const handleStepClick = (step: number) => {
+    dispatch({ type: WizardActionType.GO_TO_STEP, step });
+  };
+
+  const handleNavigation = (
+    params: { toOrigin?: boolean; keepSearch?: boolean },
```
jdslaugh, line 148: Once we create the interface in the wizard-context, we can export it for use here so we don't have to keep up with the different places its typed.

jdslaugh (commented):
Minor comments. Once we get it addressed lets do some verification in workflows utilizing it and make sure everything seems peachy.

frontend/packages/core/src/Contexts/wizard-context.tsx
```diff
@@ -1,8 +1,14 @@
 import React from "react";
 
+export interface NavigationProps {
```
jdslaugh, line 3: Lets rename this to `WizardNavigationProps` or adjust the export in the index to rename it, just so it doesn't get confused with anything (since we do have a navigation setup in Core already)

frontend/packages/core/src/stepper.tsx
```diff
@@ -183,19 +195,30 @@ const StepIcon: React.FC<StepIconProps> = ({ index, variant }) => {
...
+  isComplete?: boolean;
 }
 /* eslint-enable react/no-unused-prop-types */
 
 const Step: React.FC<StepProps> = ({ children }) => <>{children}</>;
 
-export interface StepperProps extends Pick<MuiStepperProps, "orientation"> {
+export interface StepperProps
+  extends Pick<MuiStepperProps, "orientation">,
+    Pick<MuiStepperProps, "nonLinear"> {
```
jdslaugh, line 206: You can combine these, ie:

```suggestion
  extends Pick<MuiStepperProps, "orientation" | "nonLinear"> {
```

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -4,6 +4,7 @@ import {
   ButtonGroup,
   FeatureOn,
   Grid,
+  NavigationProps,
```
jdslaugh, line 7: Lets adjust this to be an `import type`

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -156,15 +174,10 @@ const Wizard = ({
         setGlobalWarnings(warnings);
       },
       onBack: (params: { toOrigin?: boolean; keepSearch?: boolean }) => {
```
jdslaugh, line 176: Looks like we can update this to use the `NavigationProps`

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -156,15 +174,10 @@ const Wizard = ({
...
-          setSearchParams({});
-        }
-        if (params?.toOrigin && origin) {
-          navigate(origin);
-        } else {
-          dispatch(WizardAction.BACK);
-        }
+        handleNavigation(params, WizardActionType.BACK);
+      },
+      onNext: (params: { toOrigin?: boolean; keepSearch?: boolean }) => {
```
jdslaugh, line 179: Similar here as above

frontend/packages/core/src/stepper.tsx
```diff
@@ -205,19 +228,28 @@ const Stepper = ({ activeStep, orientation = "horizontal", children }: StepperPr
...
+        const icon = <StepIcon {...stepProps} />;
+
         return (
-          <MuiStep key={step.props.label}>
-            {/* eslint-disable-next-line react/no-unstable-nested-components */}
-            <MuiStepLabel StepIconComponent={() => <StepIcon {...stepProps} />}>
-              {step.props.label ?? `Step ${idx + 1}`}
-            </MuiStepLabel>
+          <MuiStep key={label} completed={isComplete}>
+            {nonLinear ? (
```
jdslaugh, line 246: Since `handleStepClick` is optional we should also verify that the function exists before trying to call it. Since the Stepper can be used outside of the Wizard we won't always have a guarantee of it being populated.

jdslaugh (commented):
Some minor comments

jdslaugh (approved):
Thanks for the updates!

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -255,11 +268,16 @@ const Wizard = ({
...
-            <Stepper activeStep={state.activeStep} orientation={orientation}>
+            <Stepper
+              activeStep={state.activeStep}
+              orientation={orientation}
+              nonLinear={nonLinear}
+              handleStepClick={handleStepClick}
+            >
               {filteredChildren.map((child: WizardChildren) => {
-                const { name } = child.props;
+                const { name, isComplete } = child.props;
```
jdslaugh, line 278: Sorry, didn't notice this until I manually went through the code.

We have the `isComplete` prop on the Step itself but we'll need to support the prop on the WizardStep as well so it's passing it down correctly. Right now the `child.props` might have it or it might not. Should be able to mimic the way we handle isLoading/error on the WizardStep.

jdslaugh (commented):
Sorry, had to revert the approval. 

Right now the `Step` itself handles the `isComplete` which is correct as it keeps the logic contained and this also handles utilizing the Stepper outside of the wizard.

But the WizardStep should also have this prop, similar to isLoading, which can be set and then passed through to the Stepper Step which then the stepper will understand. Then it can set its complete property in the context similar to the other properties.

jdslaugh (approved):

PR #3067, opened 2024-07-24
frontend: Add style prop to table component

lucechal14:
Scope: Add style prop to _TableRow_ so that styling can be dynamic for _TableCell_ components in body

Table with default column width
Screenshot 2024-07-24 at 3 55 29 p m

Table with set column width
Screenshot 2024-07-24 at 3 57 03 p m

### DIFF
--- a/frontend/packages/core/src/Table/table.tsx
+++ b/frontend/packages/core/src/Table/table.tsx
@@ -13,8 +13,9 @@ import {
 
 import { Popper, PopperItem } from "../popper";
 import styled from "../styled";
+import { getDisplayName } from "../utils";
 
-import TableCell, { TableCellType } from "./components/TableCell";
+import TableCell from "./components/TableCell";
 import TableHeader from "./components/TableHeader";
 import type { TableContainerProps, TableProps, TableRowProps } from "./types";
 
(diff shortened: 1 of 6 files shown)


frontend/packages/core/src/Table/table.tsx
```diff
@@ -132,12 +132,13 @@ const TableRow = ({
...
   responsive = false,
   colSpan,
+  style = [],
   ...props
 }: TableRowProps) => (
   <StyledTableRow onClick={onClick} $responsive={responsive} {...props}>
     {React.Children.map(children, (value, index) => (
       // eslint-disable-next-line react/no-array-index-key
-      <TableCell key={index} responsive={responsive} colSpan={colSpan}>
+      <TableCell key={index} responsive={responsive} colSpan={colSpan} style={style?.[index]}>
```
jdslaugh, line 141: Hmm, I'm not sure I'm a huge fan of this. It seems finicky if we only want to style one or two cells and not the entire thing.

I wonder if in the child map if we can just detect if we're already a `TableCell` and then just pass it through without nesting it like it currently does. That would allow us to make a styled cell in the parent component and then the row doesn't even have to know about it.

jdslaugh (approved):

PR #3071, opened 2024-07-29
Test stepper nonlinear

scsjonatan:
## Description

Includes changes to the `stepper` and `wizard` components to enhance their functionality and simplify the state management. The most important changes involve adding a new `isComplete` property to the `StepProps` interface, removing the `completed` property from `StepperProps`, and updating the wizard state management to use a new action type.

### Stepper Component Updates:
* Added `isComplete` property to `StepProps` and removed `completed` property from `StepperProps`.
* Updated the `Stepper` component to use the `isComplete` property for determining step completion status.

### Wizard Component Updates:
* Replaced the `ADD_COMPLETED_STEP` action type with `ON_COMPLETE` and removed the `completed` property from the state. 

* Updated the `Wizard` component to use the new `ON_COMPLETE` action type and pass the `isComplete` property to the `Step` components.

### DIFF
--- a/frontend/packages/core/src/stepper.tsx
+++ b/frontend/packages/core/src/stepper.tsx
@@ -234,6 +234,12 @@ const Stepper = ({
           stepProps.variant = step.props.error ? "failed" : "active";
         }
 
+        if (!nonLinear) {
+          if (idx < activeStep) {
+            stepProps.variant = "success";
+          }
+        }
+
         const label = step.props.label ?? `Step ${idx + 1}`;
         const icon = <StepIcon {...stepProps} />;
         const StepProps = isCompleted ? { completed: isCompleted } : {};
--- a/frontend/packages/core/src/stepper.tsx
+++ b/frontend/packages/core/src/stepper.tsx
@@ -228,16 +228,10 @@ const Stepper = ({
         };
         const { isComplete = false } = step.props;
 
-        if (isComplete) {
+        if (isComplete || (!nonLinear && idx < activeStep)) {
           stepProps.variant = "success";
         } else if (idx === activeStep) {
           stepProps.variant = step.props.error ? "failed" : "active";
-        }
-
-        if (!nonLinear) {
-          if (idx < activeStep) {
-            stepProps.variant = "success";
-          }
         }
 
         const label = step.props.label ?? `Step ${idx + 1}`;
--- a/frontend/packages/core/src/stepper.tsx
+++ b/frontend/packages/core/src/stepper.tsx
@@ -191,7 +191,7 @@ const StepIcon: React.FC<StepIconProps> = ({ index, variant, nonLinear }) => {
 export interface StepProps {
   label: string;
   error?: boolean;
-  isCompleted?: boolean;
+  isComplete?: boolean;
 }
 /* eslint-enable react/no-unused-prop-types */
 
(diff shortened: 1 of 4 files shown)


frontend/packages/core/src/stepper.tsx
```diff
@@ -191,6 +191,7 @@ const StepIcon: React.FC<StepIconProps> = ({ index, variant, nonLinear }) => {
 export interface StepProps {
   label: string;
   error?: boolean;
+  isCompleted?: boolean;
```
jdslaugh, line 194: nit on the naming

```suggestion
  isComplete?: boolean;
```

frontend/packages/core/src/stepper.tsx
```diff
@@ -227,22 +226,23 @@ const Stepper = ({
...
+        if (!nonLinear) {
+          if (idx < activeStep) {
+            stepProps.variant = "success";
+          }
         }
 
         const label = step.props.label ?? `Step ${idx + 1}`;
         const icon = <StepIcon {...stepProps} />;
-        const StepProps = completed ? { completed: completed[idx] } : {};
+        const StepProps = isCompleted ? { completed: isCompleted } : {};
```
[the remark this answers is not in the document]
jdslaugh: We probably dont need it anyways actually since we can just do:

```js
<MuiStep key={label} completed={isComplete}>
```

frontend/packages/core/src/stepper.tsx
```diff
@@ -227,22 +226,23 @@ const Stepper = ({
           variant: "pending" as StepIconVariant,
           nonLinear,
         };
+        const { isCompleted } = step.props;
```
jdslaugh, line 229:
```suggestion
        const { isComplete = false } = step.props;
```

frontend/packages/core/src/stepper.tsx
```diff
@@ -227,22 +226,23 @@ const Stepper = ({
...
 
-        if (completed) {
-          if (completed[idx]) {
-            stepProps.variant = "success";
-          } else if (idx === activeStep) {
-            stepProps.variant = step.props.error ? "failed" : "active";
-          }
+        if (isCompleted) {
```
jdslaugh, line 231: Looks like we can combine this a bit like:

```js
        if (idx === activeStep) {
          stepProps.variant = step.props.error ? "failed" : "active";
        } else if (isComplete || (!nonLinear && idx < activeStep)) {
          stepProps.variant = "success";
        }
```

frontend/packages/core/src/stepper.tsx
jdslaugh, line 194: nit on name:

```suggestion
  isComplete?: boolean;
```

jdslaugh (commented):
Few comments around naming etc but overall looks like this approach is what I was thinking!

Would you mind renaming this to a more descriptive title and description before we merge it into the other branch?

frontend/packages/core/src/stepper.tsx
scsjonatan: I'm thinking in the case when the user returns to a previous completed step, with your proposal if you select a completed step, the checkmark is replaced with the "active" state.

Is this how we want it to work?

scsjonatan: We could just invert the conditions to avoid the previous behavior:

```
if (isComplete || (!nonLinear && idx < activeStep)) {
  stepProps.variant = "success";
} else if (idx === activeStep) {
  stepProps.variant = step.props.error ? "failed" : "active";
}
```

PR #3084, opened 2024-08-07
frontend: Support for Workflow / Entire App Notification

lucechal14:
### Description
Support for Notifications across the app. We have 3 types of notifications:

1. **Header**: displayed in the header of the page with dynamic message, link and severity.
2. **Per Workflow:** displayed at the top of the main section of the page for each workflow specified in the configuration of the notifications with dynamic title, message, link and severity per workflow.
3. **Multi Workflow**: displayed at the top of the main section of the page for the workflows specified in the configuration of the notifications with dynamic title, message, link and severity.

### Examples
- Header and Per Workflow notification
Screenshot 2024-08-26 at 2 57 54 p m

- Header and Multi Workflow notification
Screenshot 2024-08-26 at 2 59 53 p m

### Testing
You can use an object similar to this to test:

`const banners = {
  header: {
    message: "header message",
    linkText: "link",
    link: "linkhere",
    severity: "info",
  },
  perWorkflow: {
    workflowName1: {
      title: "per workflow title for test 2",
      message: "per workflow message for test 2",
      linkText: "link",
      link: "linkhere",
      severity: "info",
    },
    workflowName2: {
      title: "per workflow title for test 3",
      message: "per workflow message for test 3",
      linkText: "link",
      link: "linkhere",
      severity: "info",
    },
  },
  multiWorkflow: {
    title: "multi workflow title",
    message: "multi workflow message",
    workflows: ["workflowName1", "workflowName3"],
    severity: "info",
    linkText: "link",
    link: "linkhere",
  },
};`

Send this const as _**banners**_ in the _**appConfiguration**_ prop of your _**ClutchApp**_


frontend/packages/core/src/AppNotifications/useCompareAppNotificationsData.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const bannersData = {
+    header: {},
+    multiWorkflow: {},
+    perWorkflow: {},
+  };
+  let shouldUpdate = false;
+
+  if (!isEmpty(banners?.header)) {
+    const headerPreferences = {
```
jecr, line 20: You think these could be handled with destructuring assignment?

lucechal14: I decided not to use destructuring for this one because none of the attributes are mandatory so if the user forgets to add them on the json or place where this data is taken from then it will throw an error

jecr:
LGTM, the only thing that feels odd is using the workflow's display name as key since it could be anything, didn't see an obvious alternative though 😅

jecr:
Kind of an edge case for the header alert, a long link text can displace the description message. Also, long description messages don't take the available height and overflow at 2 lines.
I don't think the component should hold huge chunks of text like that, but some handling could come handy just in case.
image

lucechal14:
> Kind of an edge case for the header alert, a long link text can displace the description message. Also, long description messages don't take the available height and overflow at 2 lines. I don't think the component should hold huge chunks of text like that, but some handling could come handy just in case. image

Yes the description has a set height and the it will overflow in case the description is too big, for the link the idea is that it is a small text, overall the idea of the header banner is just small text.

jecr (approved):

jdslaugh:
> > Kind of an edge case for the header alert, a long link text can displace the description message. Also, long description messages don't take the available height and overflow at 2 lines. I don't think the component should hold huge chunks of text like that, but some handling could come handy just in case. image
> 
> Yes the description has a set height and the it will overflow in case the description is too big, for the link the idea is that it is a small text, overall the idea of the header banner is just small text.

Good catch @jecr. @lucechal14 let's make sure it doesn't break free if it's too long. The idea is for it to not be very large but if it is I'd rather it be hidden or truncated versus breaking outside of the header. Can likely just `overflow: scroll` it and move the maxHeight to the alert itself.

lucechal14:
> > > Kind of an edge case for the header alert, a long link text can displace the description message. Also, long description messages don't take the available height and overflow at 2 lines. I don't think the component should hold huge chunks of text like that, but some handling could come handy just in case. image
> > 
> > 
> > Yes the description has a set height and the it will overflow in case the description is too big, for the link the idea is that it is a small text, overall the idea of the header banner is just small text.
> 
> Good catch @jecr. @lucechal14 let's make sure it doesn't break free if it's too long. The idea is for it to not be very large but if it is I'd rather it be hidden or truncated versus breaking outside of the header. Can likely just `overflow: scroll` it and move the maxHeight to the alert itself.

@jdslaugh @jecr done, here is how it looks like
Screen Recording 2024-09-05 at 12 39 47 p m

jdslaugh (approved):

PR #3086, opened 2024-08-09
Set is completed

scsjonatan:
## Description
* add `isComplete` to `WizardStep ` props
* Disable Step button on nonLinear state


frontend/packages/core/src/Contexts/wizard-context.tsx
```diff
@@ -13,6 +13,7 @@ export interface ContextProps {
   setOnSubmit: (f: (...args: any[]) => void) => void;
   setIsLoading: (isLoading: boolean) => void;
   setHasError: (hasError: boolean) => void;
+  setIsComplete: (isComplete: boolean) => void;
```
jdslaugh, line 16: Lets keep this optional similar to the `next` property.

jdslaugh (approved):

PR #3104, opened 2024-08-28
backend: Support file deletion in GitHub CreateBranch

septum:
### Description

- Enabled functionality to delete a file while creating a branch in GitHub service

### Testing Performed

Manual

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -374,15 +374,19 @@ func (s *svc) CreateBranch(ctx context.Context, req *CreateBranchRequest) error 
 	}
 
 	for filename, contents := range req.Files {
-		fh, err := wt.Filesystem.Create(filename)
-		if err != nil {
-			return err
-		}
-		if _, err := io.Copy(fh, contents); err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


backend/service/github/github.go
```diff
@@ -374,15 +374,19 @@ func (s *svc) CreateBranch(ctx context.Context, req *CreateBranchRequest) error
...
-		fh, err := wt.Filesystem.Create(filename)
-		if err != nil {
-			return err
-		}
-		if _, err := io.Copy(fh, contents); err != nil {
-			return err
-		}
-		if err := wt.AddWithOptions(&git.AddOptions{Path: filename}); err != nil {
-			return err
+		if contents == nil {
```
septum, line 377: I feel that this check could potentially be a problem.

I'm relying on the assumption that `io.ReadCloser` will **never** be nil if there is content specified for a given file.

I explored the alternative of modifying `github.FileMap` to be a struct with a Reader and Delete properties, but it would required changes in several places.

jdslaugh (approved):

PR #3105, opened 2024-08-29
backend: Add create commit method for GitHub

septum:
### Description

- Added CreateCommit method in GitHub service

### Testing Performed

Manual


backend/service/github/github.go
```diff
@@ -621,3 +622,71 @@ func (s *svc) DeleteFile(ctx context.Context, ref *RemoteRef, path, sha, message
 	}
 	return contentRes, nil
 }
+
+func (s *svc) CreateCommit(ctx context.Context, ref *RemoteRef, message string, files FileMap) (*Commit, error) {
```
jdslaugh, line 626: Most of this functionality is identical to the CreateBranch function, so I think we can pull out the similar pieces into a helper function and utilize it in both places so we don't have duplicate logic.

jdslaugh (changes requested):

backend/service/github/github.go
septum: Sounds good! I'll wait until we merge the changes in #3106 to encapsulate this logic.

jdslaugh (approved):

PR #3106, opened 2024-08-29
backend: Add support for deleting dirs in GitHub CreateBranch

septum:
### Description

- Added support to delete directories while creating a branch, retrying the delete operation if the [error message](https://github.com/go-git/go-billy/blob/master/memfs/storage.go#L161) states that the filePath provided is a directory.

### Testing Performed

Manual

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -346,28 +346,6 @@ func commitOptionsFromClaims(ctx context.Context, commitTime time.Time) *git.Com
 	return ret
 }
 
-func deleteWorktreeDirFiles(wt *git.Worktree, dirPath string) error {
-	files, err := wt.Filesystem.ReadDir(dirPath)
-	if err != nil {
-		return err
-	}
-
-	for _, file := range files {
-		filePath := wt.Filesystem.Join(dirPath, file.Name())
-		if file.IsDir() {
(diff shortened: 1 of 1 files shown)


backend/service/github/github.go
```diff
@@ -374,20 +396,26 @@ func (s *svc) CreateBranch(ctx context.Context, req *CreateBranchRequest) error
 		return err
 	}
 
-	for filename, contents := range req.Files {
+	for filePath, contents := range req.Files {
 		if contents == nil {
-			if err := wt.Filesystem.Remove(filename); err != nil {
-				return err
+			if err := wt.Filesystem.Remove(filePath); err != nil {
```
jdslaugh, line 401: So I don't think all this new logic is necessary. I should've seen it in the other PR but we can just rotate from `wt.Filesystem.Remove` to `wt.Remove` (https://github.com/go-git/go-git/blob/master/worktree_status.go#L573)

Which will delete it if it's a file or a directory

smonero (approved):
maybe not needed, but are there plans to add unit tests later?

septum:
> maybe not needed, but are there plans to add unit tests later?


Yes, we'll encapsulate the logic of CreateBranch and CreateCommit in a later commit. That will be the piece we'll mock in the tests.

PR #3109, opened 2024-09-06
Add paths attribute to per workflow banner settings

lucechal14:
### Description

For the App Notification Banner, user can now specify the paths of a workflow that need to show the banner configured in the appConfiguration banners settings.


frontend/packages/core/src/utils/pathMatching.tsx
```diff
@@ -0,0 +1,17 @@
+import { matchPath } from "react-router";
+
+const findPathMatchList = (locationPathname: string, pathstoMatch: string[]) => {
```
jecr, line 3: <sup><sub>nit</sub></sup>
```suggestion
const findPathMatchList = (locationPathName: string, pathsToMatch: string[]) => {
```

jdslaugh (approved):

jecr (approved):

PR #3137, opened 2024-10-08
housekeeping: Fixing frontend scaffolding

jdslaugh:
### Description
Frontend scaffolding has been broken for a bit but it also wasn't smart enough to handle cases where it was extending functionality for the internal application versus an external.

- Broke apart the logic to make it easier to parse and update
- Added flags to overwrite templates as well as an internal flag

### Testing Performed
manual


jecr (approved):
Gave the update a few test runs, worked great for me 🙌

PR #3138, opened 2024-10-09
frontend: Set theme spacing

septum:
### Description

- Set the theme spacing
- Create and export a `ThemeSpacing` enum

### Testing Performed

Manual testing

### DIFF
--- a/frontend/packages/core/src/Theme/theme.tsx
+++ b/frontend/packages/core/src/Theme/theme.tsx
@@ -15,6 +15,35 @@ declare module "@emotion/react" {
   export interface Theme extends MuiTheme {}
 }
 
+declare module "@mui/material/styles" {
+  interface Theme {
+    clutch: {
+      spacing: {
+        none: number;
+        xs: number;
+        sm: number;
+        base: number;
+        md: number;
+        lg: number;
+        xl: number;
+      };
+    };
+  }
+  interface ThemeOptions {
+    clutch: {
+      spacing: {
+        none: number;
+        xs: number;
+        sm: number;
+        base: number;
+        md: number;
+        lg: number;
+        xl: number;
+      };
+    };
+  }
+}
+
 // Create a Material UI theme is propagated to all children.
 const createTheme = (variant: ThemeVariant): MuiTheme => {
   return createMuiTheme({
@@ -23,6 +52,17 @@ const createTheme = (variant: ThemeVariant): MuiTheme => {
     // `8` is the default scaling factor in MUI, we are setting it again to make it explicit
     // https://v5.mui.com/material-ui/customization/spacing/
     spacing: 8,
+    clutch: {
+      spacing: {
+        none: 0,
+        xs: 0.5,
+        sm: 1,
+        base: 2,
+        md: 3,
+        lg: 4,
+        xl: 5,
+      },
+    },
     transitions: {
       // https://material-ui.com/getting-started/faq/#how-can-i-disable-transitions-globally
       create: () => "none",
--- a/frontend/packages/core/src/Theme/theme.tsx
+++ b/frontend/packages/core/src/Theme/theme.tsx
@@ -20,6 +20,9 @@ const createTheme = (variant: ThemeVariant): MuiTheme => {
   return createMuiTheme({
     colors: clutchColors(variant),
     palette: palette(variant),
+    // `8` is the default scaling factor in MUI, we are setting it again to make it explicit
+    // https://v5.mui.com/material-ui/customization/spacing/
+    spacing: 8,
     transitions: {
       // https://material-ui.com/getting-started/faq/#how-can-i-disable-transitions-globally
       create: () => "none",
--- a/frontend/packages/core/src/Theme/types.tsx
+++ b/frontend/packages/core/src/Theme/types.tsx
@@ -41,3 +41,12 @@ export interface ClutchColors {
 }
 
 export type ComponentState = "hover" | "focused" | "pressed" | "selected" | "disabled";
+
+export enum ThemeSpacing {
+  None = 0,
+  ExtraSmall = 0.5,
+  Small = 1,
+  Base,
+  Medium,
+  Large,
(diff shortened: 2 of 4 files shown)


frontend/packages/core/src/Theme/types.tsx
```diff
@@ -41,3 +41,12 @@ export interface ClutchColors {
...
 
 export type ComponentState = "hover" | "focused" | "pressed" | "selected" | "disabled";
+
+export enum ThemeSpacing {
+  None = 0,
+  ExtraSmall = 0.5,
+  Small = 1,
+  Base,
+  Medium,
+  Large,
```
jdslaugh, line 51: What do these represent?

frontend/packages/core/src/Theme/types.tsx
```diff
@@ -41,3 +41,12 @@ export interface ClutchColors {
 }
 
 export type ComponentState = "hover" | "focused" | "pressed" | "selected" | "disabled";
+
+export enum ThemeSpacing {
+  None = 0,
+  ExtraSmall = 0.5,
+  Small = 1,
```
jdslaugh, line 48: I assume the numbers would be like `0.5 * 8`?

frontend/packages/core/src/Theme/types.tsx
septum: I was going for generic names for the spacing variants as using something like `PageGutter` could be self-descriptive, but the value could be repeated somewhere else.

frontend/packages/core/src/Theme/types.tsx
septum: That's right. The `spacing` value set in the theme is the scaling factor. 

At first I was trying to document each member with the resulting value, like `4px` , but it would be outdated if the spacing value changes in the future.

frontend/packages/core/src/Theme/types.tsx
jdslaugh: Do these map to something already? Curious why some are given a value while the others aren't. I haven't looked at the spacing variants for mui so maybe it's explained there.

jecr:
Just to get a clearer idea, how would this be used in the field?, does this replace the values the current `theme.spacing` outputs?

frontend/packages/core/src/Theme/types.tsx
septum: Not really a map, these values, which are in order 0, 0.5, 1, 2, 3 and 4, are used as a multiplier of the scaling factor to produce a value, so this enum is just a giving them a name.

It's expected to apply the enum members as following examples:

```diff
<Grid
- spacing={2}
+ spacing={ThemeSpacing.Base}
>
  <Component />
 </Grid>
 ```
 
 ```diff
const StyledComponent = styled(Component)(({ theme }: { theme: Theme }) => ({
- margin: "24px",
+ margin: theme.spacing(ThemeSpacing.Medium),
}));
 ```
 
 There's an example in [MUI's docs](https://v5.mui.com/material-ui/customization/spacing/#custom-spacing) talking about the custom spacing.

**Comment by septum** [comment]
> Just to get a clearer idea, how would this be used in the field?, does this replace the values the current `theme.spacing` outputs?

Copy-pasting the example from above, it's expected to be used as the following examples:

```diff
<Grid
- spacing={2}
+ spacing={ThemeSpacing.Base}
>
  <Component />
 </Grid>
 ```
 
 ```diff
const StyledComponent = styled(Component)(({ theme }: { theme: Theme }) => ({
- margin: "24px",
+ margin: theme.spacing(ThemeSpacing.Medium),
}));
 ```

File: frontend/packages/core/src/Theme/theme.tsx [hunk 3]
```diff
@@ -20,6 +20,9 @@ const createTheme = (variant: ThemeVariant): MuiTheme => {
   return createMuiTheme({
     colors: clutchColors(variant),
     palette: palette(variant),
+    // `8` is the default scaling factor in MUI, we are setting it again to make it explicit
+    // https://v5.mui.com/material-ui/customization/spacing/
+    spacing: 8,
```
jecr, line 25: Thanks for the usage example 🙌
So, just a thought, would it be a good idea to include the spacing options as [properties](https://mui.com/material-ui/customization/theming/#custom-variables) of the theme object?

septum: I was going with that approach at first, but after discovering the `theme.spacing` method, I wanted to take advantage of what it could be done with it, like the [multiple arity](https://v5.mui.com/material-ui/customization/spacing/#multiple-arity).

jecr: Gotcha, I think they can coexist, something like `theme.spacing(theme.standardSpacing.medium)` and in that `standardSpacing` you pretty much do the same you already have
I can see the naming and having to access the properties in `theme` twice a bit cumbersome 😅 though
What I do like about this is centralising everything in `theme`

septum: Agreed, I think centralizing it would improve its DevEx. 

Maybe we could add a new function to be used instead of `spacing()`, or come up with a different naming strategy for the spacing to make it concise?

jecr: Something like `theme.clutchSpacing.medium`?, it is still long 😆 but yeah, that sounds easy to use

septum: I think the `clutch` prefix makes sense, so I turned that into an object in b40fc87. It ended up like:

```
theme.spacing(theme.clutch.spacing.md);
```

I was aiming to create something that only relies on the variant being the functions themselves, but that would require ugly workarounds, like creating the theme twice. Also, using `this` was unstable, even with arrow functions.

Should we settle for this?

cc @jdslaugh

jdslaugh: Yeah I think this is a good way to go for the time being.

PR #3143, opened 2024-10-16
frontend: Add workflow layout component

septum:
### Description

- Added a new WorkflowLayout component

##### Before

localhost_3000_dynamodb_capacity (1)

##### After

localhost_3000_dynamodb_capacity (3)

##### Before

localhost_3000_k8s

##### After

localhost_3000_k8s_dashboard

### Testing Performed

Manual


frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const WorkflowLayout = ({
+  variant,
+  showHeader,
+  heading,
+  children,
+}: React.PropsWithChildren<LayoutProps>) => {
+  return (
+    <LayoutContainer $variant={variant}>
+      {showHeader !== false && (
```
jecr, line 70: Wondering about this condition, seems funny checking for` !== false`, wouldn't it be the same as just checking for `showHeader = true`?, or maybe a prop name change to `hideHeader` if the default is to show it?

septum: My reasoning is that it won't be used unless setting it to `false`. 

I could default it as `showHeader = true` in the props destructuring and doing the check to `=== true`, though.

jecr: 🤔 my concern is the prop name `showHeader`, and that it will only take effect if set to `false`, kinda goes against what `show` conveys. Or maybe I'm getting it wrong, double negatives confuse me 😆

septum: Ahh, I see now. I think that makes sense, I'll submit a change for it.

septum: Done in [275ea93](https://github.com/lyft/clutch/pull/3143/commits/275ea93fb34028b3978fd5888e2b308de56e1011)

frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+    },
+    // No styles
+    custom: {},
+  };
+  return layoutVariantStylesMap[variant];
+};
+
+const LayoutContainer = styled("div")(
+  ({ $variant, theme }: StyledVariantComponentProps) =>
+    getContainerVariantStyles($variant, theme) as any
```
jecr, line 47: Can this have a type other than `any`?

jecr (approved):

frontend/packages/core/src/WorkflowLayout/index.tsx
septum: Yes, done in 076904d

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -214,21 +215,32 @@ const ClutchApp = ({
...
-                                    type="layout"
-                                    workflow={workflow?.displayName}
-                                    banners={appConfiguration?.banners}
-                                  >
-                                    {React.cloneElement(<route.component />, {
-                                      ...route.componentProps,
-                                      heading,
-                                    })}
-                                  </AppNotification>
+                                  <WorkflowLayout {...workflowLayoutProps}>
```
jdslaugh, line 231: Sorry, had one thought about this. Let's add an appConfig variable that basically is `useWorkflowSpacing` or something, disabled by default (this way it doesn't negatively affect others) and then we enable it manually.

jdslaugh (approved):

PR #3144, opened 2024-10-16
[WIP]native sidecar pod status

murongl-lyft:
How to get the correct Pod status with native sidecars. 

### GitHub Issue

Fixes #

### DIFF
--- a/backend/service/k8s/pods.go
+++ b/backend/service/k8s/pods.go
@@ -327,9 +327,17 @@ func getPodStatus(pod *corev1.Pod) string {
 	}
 
 	initializing := false
+	totalInitContainers := len(pod.Spec.InitContainers)
 	for i := range pod.Status.InitContainerStatuses {
 		container := pod.Status.InitContainerStatuses[i]
 		restarts += int(container.RestartCount)
+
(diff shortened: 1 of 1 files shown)


backend/service/k8s/pods.go
```diff
@@ -326,10 +326,30 @@ func getPodStatus(pod *corev1.Pod) string {
...
+			// if the init container has a restart policy, it is native sidecar and should not be counted
+			// https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/753-sidecar-containers/README.md
+			totalInitContainers--
+			nativeSidecarContainerId[i] = struct{}{}
+		}
+	}
+
 	initializing := false
+	initContainerCount := 0
 	for i := range pod.Status.InitContainerStatuses {
```
abhinavdahiya, line 344: One goal of showing initializing containers can be - show customers that their application container run is waiting on initializing of platform containers.

So i would think it would be really nice to include the sidecar containers in this information, since now they are all in initContainers.

* normal init container -- make sure it's terminated and exit code is 0
* sidecar init container -- make sure it's running and ready

PR #3145, opened 2024-10-16
frontend: Add breadcrumbs component

septum:
### Description

- Added Breadcrumbs component

#### PRs

- Relies on #3143 

#### Before

localhost_3000_catalog_clutch

#### After

localhost_3000_catalog

### Testing Performed

Manual


jdslaugh (commented):
Looks good! From the image it looks like the last entry is getting the link applied, is that happening? I assume it's from test data but want to make sure.

Also, is this actually reliant on `workflow-layout`? Don't see anything related to it. So it could go in separately.

septum:
> Looks good! From the image it looks like the last entry is getting the link applied, is that happening? I assume it's from test data but want to make sure.
> 
> Also, is this actually reliant on `workflow-layout`? Don't see anything related to it. So it could go in separately.

It was used for testing only, I'll make sure to retest after the merge of #3143

frontend/packages/core/src/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
+import { Breadcrumbs as MuiBreadcrumbs, Theme } from "@mui/material";
+import { alpha } from "@mui/system";
+
+import styled from "../styled";
+
+import Breadcrumb from "./breadcrumb";
+import { BreadcrumbEntry } from "./types";
```
jecr, line 8:
```suggestion
import type { BreadcrumbEntry } from "./types";
```

frontend/packages/core/src/Breadcrumbs/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const Breadcrumbs = ({ entries }: BreadcrumbsProps) => {
+  return (
+    <StyledBreadcrumbs>
+      {entries.map((entry: BreadcrumbEntry) => (
+        <Breadcrumb key={entry.label} {...entry} />
+      ))}
+    </StyledBreadcrumbs>
+  );
+};
```
jecr, line 29: nit
```suggestion
const Breadcrumbs = ({ entries }: BreadcrumbsProps) => (
    <StyledBreadcrumbs>
      {entries.map((entry: BreadcrumbEntry) => (
        <Breadcrumb key={entry.label} {...entry} />
      ))}
    </StyledBreadcrumbs>
  );
```

jecr (approved):
Left a few minor comments

jdslaugh (approved):

PR #3155, opened 2024-10-22
frontend: Adjust workflow layout whitespace

septum:
### Description

- Modified spacing from Breadcrumbs and WorkflowLayout to provide better whitespace defaults

#### Before

Screenshot 2024-10-22 at 1 29 56 p m

#### After

Screenshot 2024-10-22 at 1 29 18 p m

### Testing Performed

Manual

### DIFF
--- a/frontend/packages/core/src/WorkflowLayout/index.tsx
+++ b/frontend/packages/core/src/WorkflowLayout/index.tsx
@@ -13,7 +13,7 @@ import { generateBreadcrumbsEntries } from "../utils";
 export type LayoutVariant = "standard" | "wizard" | "custom";
 
 export type LayoutProps = {
-  workflow?: Workflow;
+  workflow: Workflow;
   variant?: LayoutVariant;
   heading?: string | React.ReactElement;
   hideHeader?: boolean;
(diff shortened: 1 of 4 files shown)


frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -74,11 +81,11 @@ const WorkflowLayout = ({
   children,
 }: React.PropsWithChildren<LayoutProps>) => {
   const location = useLocation();
-  const workflowPaths = workflow.routes.map(({ path }) => `/${workflow.path}/${path}`);
+  const workflowPaths = workflow?.routes.map(({ path }) => `/${workflow?.path}/${path}`);
```
jecr, line 84: It there is no `workflow.path` what would be rendered?, could it do with a fallback or other handling?

jecr, line 84: Maybe this in case of no `workflow?.routes` so `workflowPaths` is not left `undefined`
```suggestion
  const workflowPaths = workflow?.routes.map(({ path }) => `/${workflow?.path}/${path}`) || [];
```

frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -56,13 +56,20 @@ const LayoutContainer = styled("div")(
...
   ),
+  paddingBottom: theme.spacing(theme.clutch.spacing.base),
   width: "100%",
 }));
 
-const HeaderTitle = styled(Typography)({
+const PageHeaderMainContainer = styled("div")({
+  display: "flex",
+  alignItems: "center",
+  height: "70px",
```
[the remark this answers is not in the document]
septum: Similarly to ShipIt, the layout header may contain QuickLinks and 70px is the height of that component. The `height` is set to that to handle that case in the near future.

frontend/packages/core/src/WorkflowLayout/index.tsx
septum: I see, because of `workflow` being undefined. 

I think I'm going to change the type signature to reflect that workflow (and the path property) are required.

jecr (approved):

PR #3156, opened 2024-10-23
frontend: Refactor workflows to use new layout and spacing units

septum:
### Description

- Modified the configuration and component of all routes of every workflow to use new layout and spacing units

##### Before

clutch lyft net_catalog_ (1)

##### After

localhost_3000_ec2_asg_resize (2)

> [!NOTE]
> New props were added to the layout to display only the breadcrumbs and preserve previous route component header

##### Before

clutch lyft net_dash_ (1)

##### After

379093432-227ce42a-f890-4889-8860-e7c6c27c929f

##### Before

clutch lyft net_dash_

##### After

localhost_3000_dash

### Testing Performed

Manual

### TODO

- [x] Change the application of spacing units to rely on the Clutch spacing

### DIFF
--- a/frontend/packages/core/src/WorkflowLayout/index.tsx
+++ b/frontend/packages/core/src/WorkflowLayout/index.tsx
@@ -112,7 +112,7 @@ const WorkflowLayout = ({
   );
 
   if (variant === "custom") {
-    return children;
+    return <>{children}</>;
   }
 
   return (
--- a/frontend/packages/core/src/WorkflowLayout/index.tsx
+++ b/frontend/packages/core/src/WorkflowLayout/index.tsx
@@ -19,7 +19,7 @@ export type LayoutProps = {
   variant?: LayoutVariant;
   title?: string | ((params: Params) => string);
   subtitle?: string;
-  onlyBreadcrumbs?: boolean;
+  breadcrumbsOnly?: boolean;
   hideHeader?: boolean;
 };
 
@@ -97,7 +97,7 @@ const WorkflowLayout = ({
   variant = "standard",
   title = null,
   subtitle = null,
-  onlyBreadcrumbs = false,
+  breadcrumbsOnly = false,
   hideHeader = false,
   children,
 }: React.PropsWithChildren<LayoutProps>) => {
@@ -120,10 +120,10 @@ const WorkflowLayout = ({
     <LayoutContainer $variant={variant}>
       {!hideHeader && (
         <PageHeader $variant={variant}>
-          <PageHeaderBreadcrumbsWrapper $onlyBreadcrumbs={onlyBreadcrumbs}>
+          <PageHeaderBreadcrumbsWrapper>
             <Breadcrumbs entries={breadcrumbsEntries} />
           </PageHeaderBreadcrumbsWrapper>
-          {!onlyBreadcrumbs && (
+          {!breadcrumbsOnly && (title || subtitle) && (
             <PageHeaderMainContainer>
               <PageHeaderInformation>
                 {title && (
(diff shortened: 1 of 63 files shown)


frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -90,22 +112,29 @@ const WorkflowLayout = ({
...
         <PageHeader $variant={variant}>
-          <Breadcrumbs entries={breadcrumbsEntries} />
-          <PageHeaderMainContainer>
-            {heading && (
-              <>
-                {React.isValidElement(heading) ? (
-                  heading
-                ) : (
-                  <Heading variant="h2">{heading}</Heading>
+          <PageHeaderBreadcrumbsWrapper $onlyBreadcrumbs={onlyBreadcrumbs}>
```
jecr, line 123: 🙋 what is `$onlyBreadcrumbs` used for?, in `PageHeaderBreadcrumbsWrapper`

frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -90,22 +112,29 @@ const WorkflowLayout = ({
...
+            <Breadcrumbs entries={breadcrumbsEntries} />
+          </PageHeaderBreadcrumbsWrapper>
+          {!onlyBreadcrumbs && (
+            <PageHeaderMainContainer>
```
jecr, line 127: If no `title` and `subtitle` should the containers be rendered?

frontend/packages/core/src/WorkflowLayout/index.tsx
septum: Good catch! It's a residue from a previous change, will remove it.

frontend/packages/core/src/WorkflowLayout/index.tsx
septum: Done in [84aacb4](https://github.com/lyft/clutch/pull/3156/commits/84aacb421d6b0d9ebdabae52825012747ef654e2).

jecr (approved):

jdslaugh (changes requested):
Overall looking good, lets see if we can break up this PR into 2 so that we can reduce the scope that this one impacts.

frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -63,23 +66,42 @@ const PageHeader = styled("div")(({ $variant, theme }: StyledVariantComponentPro
...
+  color: alpha(theme.colors.neutral[900], 0.45),
+}));
+
 const WorkflowLayout = ({
   workflow,
   variant = "standard",
-  heading = null,
+  title = null,
+  subtitle = null,
+  breadcrumbsOnly = false,
```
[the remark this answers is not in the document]
septum: So should these be defined inside the [AppProvider](https://github.com/lyft/clutch/pull/3156/files#diff-4327e2aff12665e532737317ff7aeb684fe579812690ed68f83a175fb989e768L219) or somewhere else?

frontend/workflows/projectCatalog/src/index.tsx
```diff
@@ -20,24 +20,41 @@ const register = (): WorkflowConfiguration => {
         description: "A searchable catalog of services",
         component: Catalog,
         featureFlag: "projectCatalog",
+        layoutProps: {
+          variant: "standard",
+          title: "Project Catalog",
```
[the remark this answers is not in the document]
septum: For the title and the subtitle, I agree that it is redundant with the other config props. I initially defined it as so, because  some of the workflows define their own in a component, ignoring the `heading` prop. 

We could definitely just use the other attributes and leaving the `breadcrumbsOnly` prop, because the title and subtitle will be always defined (from what I've seen in the workflows config).

PR #3159, opened 2024-10-29
frontend: Refactor workflows to use new layout

septum:
### Description

- Modified the configuration and component of all routes of every workflow to use the new layout

##### Before

clutch lyft net_catalog_ (1)

##### After

localhost_3000_catalog_testbuildnotify

> [!NOTE]
> New props were added to the layout to display only the breadcrumbs and preserve previous route component header

##### Before

clutch lyft net_dash_ (1)

##### After

localhost_3000_catalog

##### Before

clutch lyft net_dash_

##### After

localhost_3000_dash

### Testing Performed

Manual


frontend/packages/core/src/WorkflowLayout/index.tsx
```diff
@@ -63,49 +64,73 @@ const PageHeader = styled("div")(({ $variant, theme }: StyledVariantComponentPro
...
+  }
+
   const workflowPaths = workflow.routes.map(({ path }) => `/${workflow.path}/${path}`);
   const breadcrumbsEntries = generateBreadcrumbsEntries(
     location,
-    (url: string) =>
-      `/${workflow.path}` !== url &&
-      !workflowPaths.includes(url) &&
-      !workflowPaths.find(path => !!matchPath({ path }, url))
+    url => !!workflowPaths.find(path => !!matchPath({ path }, url))
```
septum, line 112: There's an issue with the list of workflow paths available:  

- When rendering the workflow paths not all possible routes are not available to this check, because some workflow s are fragmented between instances but not reconciled into one single workflow.

frontend/packages/core/src/AppProvider/index.tsx
```diff
@@ -217,9 +217,12 @@ const ClutchApp = ({
...
 
                             const workflowLayoutProps: LayoutProps = {
-                              ...route.layoutProps,
-                              heading: route.layoutProps?.heading || heading,
                               workflow,
+                              title: heading,
+                              subtitle: route.description,
+                              variant: route.layoutProps ? route.layoutProps.variant : null,
+                              breadcrumbsOnly: route.layoutProps?.breadcrumbsOnly || false,
+                              hideHeader: route.layoutProps?.hideHeader || false,
```
jecr, line 225:
```suggestion
                              variant: route?.layoutProps?.variant || null,
                              breadcrumbsOnly: route?.layoutProps?.breadcrumbsOnly ?? false,
                              hideHeader: route?.layoutProps?.hideHeader ?? false,
```
not sure about  the `route?.layoutProps` optional chaining, can a route be missing the `layoutProps` prop?

frontend/workflows/projectSelector/src/index.tsx
```diff
@@ -29,6 +29,9 @@ const register = (): WorkflowConfiguration => {
         displayName: "Dash",
         description: "Display helpful information for multiple services.",
         component: Dash,
+        layoutProps: {
+          variant: null,
```
jecr, line 33: Wouldn't this default to `null` anyway?

frontend/workflows/ec2/src/reboot-instance.tsx
```diff
@@ -89,7 +89,7 @@ const RebootInstance: React.FC<WorkflowProps> = ({ heading, resolverType, notes
   };
 
   return (
-    <Wizard dataLayout={dataLayout} heading={heading}>
```
jecr, line 92: The heading is no longer used elsewhere it seems, so should be safe to remove from the destructuring at the top. I think the case applies for several other files.

frontend/packages/core/src/AppProvider/workflow.tsx
jecr, line 1: There's a BaseWorkflowProps interface in this file, with `heading` as its only property, is it still necessary to keep it around?

frontend/packages/core/src/AppProvider/index.tsx
septum: I agree on this change, but not with the route chaining, I think a `route` will always be defined.

jdslaugh:
@septum looking at this:

localhost_3000_catalog_testbuildnotify

Curious why `Catalog` is black and the project name is blue. Would think it was the other way around since the route looks to be on the project versus the main catalog

frontend/packages/core/src/AppProvider/workflow.tsx
septum: Yes, in some workflows we only display the breadcrumbs or hide the header entirely, and those make use of the `heading` prop.

frontend/packages/wizard/src/wizard.tsx
```diff
@@ -72,7 +72,6 @@ const Header = styled(Grid)<{ $orientation: MuiStepperProps["orientation"] }>(
 
 const Container = styled(MuiContainer)<{ $width: ContainerProps["width"] }>(
   {
-    paddingBlock: "24px 32px",
     height: "calc(100% - 56px)",
```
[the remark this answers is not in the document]
septum: Looks like it fills out the available space, even overflowing from this component. Was this defined to prevent scrolling?

frontend/workflows/audit/src/index.tsx
```diff
@@ -25,13 +25,19 @@ const register = (): WorkflowConfiguration => {
         displayName: "Logs",
         description: "View audit log",
         component: AuditLog,
+        layoutProps: {
+          variant: "standard",
+        },
```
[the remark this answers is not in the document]
septum: This would be great! The only issue is leaving the layout variant undefined, and doing something like this:

```
layoutProps: {}
```

I think for that case setting it as `null` could work.

septum:
> @septum looking at this:
> 
> localhost_3000_catalog_testbuildnotify
> 
> Curious why `Catalog` is black and the project name is blue. Would think it was the other way around since the route looks to be on the project versus the main catalog


This is because the comment above:

> There's an issue with the list of workflow paths available:
> 
> * When rendering the workflow paths not all possible routes are not available to this check, because some workflows are fragmented between instances but not reconciled into one single workflow.

Basically, when rendering this path, the `/catalog/` root path is not available, because it's in another instance of clutch. I added another check to manually add the `/${route.path}`, but it was providing some root paths that did not have a component.

jecr (approved):
It'd be nice having JSDocs or something akin to give users some guidance on using the new feature

PR #3160, opened 2024-10-29
frontend: Refactor workflows to use new spacing units

septum:
### Description

- Modified the styling props and config for every workflow to use the new spacing units

###### Before

localhost_3000_catalog_ (1)

###### After

localhost_3000_catalog_

###### Before

localhost_3000_ec2_instance_terminate

###### After

localhost_3000_catalog_ (2)

###### Before

localhost_3000_k8s_job-logs

###### After

localhost_3000_k8s_job-logs

### Testing Performed

Manual


frontend/workflows/k8s/src/k8s-dash-search.tsx
```diff
@@ -14,20 +14,21 @@ import { useDataLayout } from "@clutch-sh/data-layout";
...
 import { yupResolver } from "@hookform/resolvers/yup";
 import SearchIcon from "@mui/icons-material/Search";
+import type { Theme } from "@mui/material";
 import * as yup from "yup";
 
-const Container = styled.div({
-  margin: "32px 0",
-});
+const Container = styled.div(({ theme }: { theme: Theme }) => ({
+  margin: theme.spacing(theme.clutch.spacing.lg, theme.clutch.spacing.none),
```
jdslaugh, line 21: No idea but I wonder if we can override how theme.spacing works and then just be able to provide the levels, like:

```
theme.spacing('lg', 'none');
...
theme.spacing('lg', '16px');
```

Not anything that needs to happen here and might be more trouble than it's worth but could be nice for the future

jdslaugh (commented):
Looks good, can you provide a few before/after screenshots (doesn't have to be everything) just so there's a reference.

frontend/workflows/k8s/src/k8s-dash-search.tsx
septum: The example provided in the documentation seems to indicate that this is possible when using string values:
- https://mui.com/material-ui/customization/spacing/#multiple-arity

septum: It seems that we could add our own spacing function:

- https://github.com/mui/material-ui/blob/429b48a408c8388e2a5cf629874eb467abe4af35/packages/mui-system/src/createTheme/createSpacing.ts#L28 

Seemingly, there's no docs describing this functionality but this may enable the wanted functionality.

jecr: I believe there's an [option](https://mui.com/material-ui/customization/spacing/#custom-spacing) for intercepting string values passed to `theme.spacing` and using the values you defined previously to make a substitution
```
const theme = createTheme({
  spacing: (...args) => // Parse args, substitute strings with clutch.spacing values
});
```

frontend/workflows/projectCatalog/src/details/components/card.tsx
```diff
@@ -50,16 +50,16 @@ interface CardProps extends CatalogDetailsCard, BaseCardProps {}
...
 
 const StyledGrid = styled(Grid)({
   height: "fit-content",
 });
 
 const StyledProgressContainer = styled("div")(({ theme }: { theme: Theme }) => ({
-  marginBottom: theme.spacing(theme.clutch.spacing.sm),
-  marginTop: theme.spacing(-theme.clutch.spacing.base),
+  marginBottom: theme.spacing("sm"),
+  marginTop: theme.spacing("-16px"),
```
jecr, line 62: This could be done, right?
```suggestion
marginTop: theme.spacing(-2),
```

jecr (approved):
Neat!

PR #3174, opened 2024-11-13
frontend: Set workflow layout title as either route or workflow display names

septum:
### Description

-  Sets workflow layout title as either route or workflow display names

### Testing Performed

Manual


frontend/packages/core/src/AppProvider/index.tsx
jecr, line 1: Is the block above required?, or may it need adjustment?
```
const heading = route.displayName
  ? `${workflow.displayName}: ${route.displayName}`
  : workflow.displayName;
```

septum: It is needed as the fallback is `useWorkflowLayout` is not enabled.

jdslaugh (approved):

PR #3224, opened 2025-01-31
backend: Create GetFileContents method in GitHub service

septum:
### Description

- Created GetFileContents method in GitHub service

This method behaves similarly to `getFile` but is able to retry the request if the file is larger than 1MB. See [Get repository content](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content).

### Testing Performed

Manual and mocks

### DIFF
--- a/backend/service/github/github.go
+++ b/backend/service/github/github.go
@@ -5,12 +5,15 @@ package github
 // <!-- END clutchdoc -->
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


septum:
Not needed right now.

tlin4194 (approved):

backend/service/github/github.go
```diff
@@ -5,12 +5,15 @@ package github
...
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
```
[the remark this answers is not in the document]
septum: /other will update in it's own PR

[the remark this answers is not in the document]
septum: /other will update in it's own PR

jecr (approved):

backend/service/github/github.go
```diff
@@ -673,3 +734,31 @@ func (s *svc) SearchCode(ctx context.Context, query string, opts *githubv3.Searc
 
 	return results, nil
 }
+
+func (s *svc) GetFileContents(ctx context.Context, ref *RemoteRef, filePath string) (*githubv3.RepositoryContent, error) {
```
jdslaugh, line 738: Curious why this was necessary instead of adjusting GetFile to be able to support larger files / raw.

backend/service/github/github.go
```diff
@@ -42,17 +46,69 @@ const (
...
+				})
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			// Recreate the response body
+			resp.Body = io.NopCloser(bytes.NewReader(fileContentBytes))
+			resp.ContentLength = int64(len(fileContentBytes))
+		}
```
jdslaugh, line 110: Lets break this out into a utility function to keep it contained and easier to parse.

septum: Done in bfb3bf3.

backend/service/github/github.go
septum: Overall it would have required more work, and messing with an endpoint that is used by many other features apart from Project Deprecation.

backend/service/github/github.go
jdslaugh: Thanks!

jdslaugh (approved):

